PR #55, opened 2022-10-10
CSVParser MVP

MarshallVielmetti:


emilbon99:
Is this complete? Are we pull requesting to main or a new branch?

MarshallVielmetti:
I would suggest main because none of the files really conflict with anything else except for the one change to fix the bug we were dealing with because of the type issue. 
From there, I will just make a new branch to continue working on it but might as well have it in main to start in my opinion.

emilbon99:
Let's put it on another branch with the ticket number in it. I'd like all code on main to be production ready, which means docs and test cases.

Sorry for the delay, I've been working on dealing with the data irregularity and think I'm almost at a spot that should work.

PR #348, opened 2023-09-15
Ddded GetVersion and moved VERSION to /pkg/version

LeonLiur:
Also changed synnax.buid.yaml


synnax/pkg/version/GetVersion.go
emilbon99, line 1: Go naming conventions are snake case. I think this file should be named `version.go` or `get.go`

synnax/pkg/version/GetVersion.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"embed"
+	"fmt"
+	"io/ioutil"
+)
+
+//go:embed VERSION
+var fs embed.FS
+
+func GetVersion() string{
```
emilbon99, line 12: When you import this, you'd call `version.GetVersion()`, which is redundant. Instead, rename to `version.Get()` or (and this is technically considered OK by the go style standards) `version.Version()`

emilbon99 (changes requested):
Have you tested this out? We use [ginkgo](https://onsi.github.io/ginkgo/) to write our tests. After installing, make a new test suite in a package use `ginkgo bootstrap` and then to generate a new test file run `ginkgo generate FILE_NAME` . You can refer to the test files in some of the other directories as well. 

Other than those small things, looks super good!

emilbon99 (approved):
LGTM

PR #362, opened 2023-09-24
modify createLocal to check for duplicate names

t1an-xyz:
Modify createLocal method to check for channels with duplicate names before creating a new channel in the storage.

Added a new test case testing duplicate channel names. Modified existing test cases to use UUIDs as channel names.


emilbon99:
I'm going to go through and do test case checks for client libraries today.

PR #369, opened 2023-09-25
Create test-freighter-cpp.yml

sverlon:
Needed to make sure further changes to prod doesn't break cpp freighter

### DIFF
--- a/.github/workflows/test-freighter-cpp.yml
+++ b/.github/workflows/test-freighter-cpp.yml
@@ -1,0 +1,34 @@
+name: Test Freighter CPP
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'synnax/freighter/cpp/**'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


.github/workflows/test-freighter-cpp.yml
```diff
@@ -0,0 +1,34 @@
+name: Test Freighter CPP
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'synnax/freighter/cpp/**'
```
emilbon99, line 8: Could be a good idea to add the CI yaml file itself to this.

.github/workflows/test-freighter-cpp.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
```
emilbon99, line 15: One of the  useful tools I use in a monorepo is the `dorny/paths-vilter@v2`. It actually does a diff check on the git history as well.

emilbon99 (changes requested):
Can you just add this as another job to `freighter.test.yaml`? I like to keep tests for related services together

PR #371, opened 2023-09-28
Sy 182 streaming interface cpp

sverlon:
- Implemented streaming gRPC in freighter
- Tested using multiple instances of streamer, sending multiple messages


emilbon99 (approved):
LGTM

Issue #11, opened 2022-09-04
Client - TDMS ColumnReader Implementation


emilbon99:
I'd like an easy way to import TDMS files for use within Synnax. The ingestion CLI uses protocol classes to ingest multiple file types under the same interactive workflow. There are two main protocol classes involved here, both of them are in `synnax/client/py/cli/io/protocol.py`. The first is `RowFileReader` which is a protocol for reading row based files (CSV, TSV, etc.). The second is `ColumnFileReader` which is for reading columnar files like HDF5 and TDMS.  

This ticket involves implementing the `ColumnFileReader` protocol for reading TDMS files using the [npTDMS](https://nptdms.readthedocs.io/en/stable/reading.html#reading-large-files) python library. 

At a high level, the steps for working this issue are as follows:

1. Add nptdms as a dependency to the pyproject TOML.
2. Create a new class called `TDMSReader` in `synnax/client/py/cli/io/tdms.py`.
3. Implement the matcher protocol in the same way it was done in `synnax/client/py/cli/io/csv.py`. 
4. Implement the `File` protocol by defining a `path` method which returns the path to the file.
5. Implement the `BaseReader` protocol.
6. Implement the `ColumnReader` protocol.  It's important that we support setting an appropriate chunk size for reading large files.

ALL PROTOCOLS ARE IN `synnax/client/py/cli/io/protocol.py`.

PR #375, opened 2023-10-01
#11 client tdms columnreader implementation

Orbital-Web:
Implemented TDMS reader
Supports:
- `File`, `BaseReader`, and `ColumnReader` Protocols
- Has several tests to ensure reader works as expected

Notes:
- `nsamples` will sometimes overestimate (but never underestimate), i.e., when chunk_size is greater than the actual size of the last chunk
- defaults to reading every chunk if `keys` is not specified


client/py/pyproject.toml
```diff
@@ -18,6 +18,7 @@ python = "^3.11"
 websockets = "^11.0.3"
 mypy = "^1.3.0"
 pandas = "^2.0.2"
+pyTDMS = "^0.0.3"
```
emilbon99, line 21: Do we need this dep?

client/py/synnax/io/tdms.py
```diff
@@ -0,0 +1,144 @@
...
+        """Seeks the reader to the first sample in the file."""
+        self._current_chunk = 0
+    
+    def read(self, *keys: str) -> pd.DataFrame:
+        """Reads a dataframe with chunk size * number of columns samples. The returned
+        dataframe contains columns for each key in keys.
+        """ 
+        # if we already read everything, return an empty chunk
+        if self._current_chunk >= self.n_chunks:
+            return pd.DataFrame()
```
emilbon99, line 109: We need some way to indicate when the read method is exhausted. In the csv reader implementation I raise a `StopIteration` error. This is a bit clunky as `read` might be used outside of an iterator. It could be reasonable to change the protocol to add a boolean indicator that says whether there is more data or not. Thoughts?

client/py/synnax/io/tdms.py
```diff
@@ -0,0 +1,144 @@
...
+        return (self.read() for _ in range(self.n_chunks))
+    
+    @property
+    def n_chunks(self) -> int:
+        """Returns number of chunks in a channel.
+        Assumes every channel has the same number of chunks
+        """
+        if self._n_chunks is None:
+            # assume every channel has the same number of data
+            with TdmsFile.open(self._path) as tdms_file:
```
emilbon99, line 138: Do we need to open this handle multiple times? I guess the overhead isn't huge, but I do like the thought of opening the file once and keeping a handle as a class property. Reduces the number of file descriptors.

client/py/tests/test_io_reader.py
```diff
@@ -59,3 +62,62 @@ def test_read(self, valid_file: RowFileReader):
             assert len(d) == 1
             count += 1
         assert count == 4
+
+
+@pytest.mark.tdms
```
emilbon99, line 67: We might eventually want to generalize this test suite to work with multiple column reader types (i.e. parameterize), but this LGTM for now.

client/py/synnax/io/tdms.py
```diff
@@ -0,0 +1,144 @@
...
+                group0 = tdms_file.groups()[0]
```
emilbon99, line 139: This might raise an obscure error if the TDMS file is empty.

emilbon99 (changes requested):
See my comments. Can you also format with black and bump the minor version in `pyproject.toml`?

PR #402, opened 2023-10-08
314 math parser

t1an-xyz:
Implements a library to parse and evaluate math expressions
Also implements basic queue and stack data structures using slices


x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+	tokens := []string{}
+	currToken := ""
+	for i, c := range s {
+		if c == ' ' {
+			if currToken != "" {
+				tokens = append(tokens, currToken)
+				currToken = ""
+			}
+			continue
+		}
```
emilbon99, line 40: Can we compress this into an `&&` statement and then use an `if else` block in the next clause instead of the `continue`?

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+		//	Check if token is a negative number
+		if c == '-' && currToken == "" && (i == 0 || tokens[len(tokens)-1] == "(" || tokens[len(tokens)-1] == "+" || tokens[len(tokens)-1] == "-" || tokens[len(tokens)-1] == "*" || tokens[len(tokens)-1] == "/" || tokens[len(tokens)-1] == "^") {
```
emilbon99, line 42: No need to repeatedly access `tokens[len(tokens-1)]`. Put it into a temporary variable instead. Also, instead of explicitly doing if checks on each valid token, you can create a private, global map and check if the token is in the map. Not only will this be cleaner, it'll also be faster.

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+			currToken += string(c)
+		} else if c == '+' || c == '-' || c == '*' || c == '/' || c == '(' || c == ')' || c == '^' {
```
emilbon99, line 44: See optimization above.

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+
+type Expression struct {
+	exp ast.Expr
+}
+
+type Resolver interface {
+	Resolve(string) (float64, error)
+}
+
+func findTokens(s string) ([]string, error) {
```
emilbon99, line 30: There's no case where `findTokens` encounters an error, so why do we return one?

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+	tokens := []string{}
```
emilbon99, line 31: Convert this to a named return value.

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+
+// Given a string representing a valid mathematical expression, sets exp to the
+// parsed expression
+func (e *Expression) Build(s string) error {
+	tokens, err := findTokens(s)
+	if err != nil {
+		return err
+	}
+	//	Uses the shunting yard algorithm to create an ast
+	precedence := map[string]int{
```
emilbon99, line 68: Declare these maps globally and avoid additional allocations. You could also use them for the purpose I described on L42,L44

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+		"+": 1,
+		"-": 1,
+		"*": 2,
+		"/": 2,
+		"(": 0,
+		"^": 3,
+	}
+	tokensDict := map[string]token.Token{
```
emilbon99, line 76: See prev comment

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+		case "+", "-", "*", "/":
+			for operators.Len() > 0 && precedence[*operators.Peek()] >= precedence[t] {
+				op, err := operators.Pop()
+				if err != nil {
+					return err
+				}
+				Y, err1 := output.Pop()
+				X, err2 := output.Pop()
+				if err1 != nil || err2 != nil {
+					return errors.New("Invalid expression")
```
emilbon99, line 97: If you're going to return an error that has the same message in multiple places, it's best to extract it into a public global var. This lets callers handle it with `errors.Is()`. Even better, create a base error called `InvalidExpression`, then use `errors.Wrapf` to wrap it with a detailed message.

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	exp, err := output.Pop()
+	if err != nil {
+		return err
+	}
+	e.exp = exp.(ast.Expr)
+	return nil
+}
+
+func (e Expression) GetTree() ast.Expr {
```
emilbon99, line 165: Go discourages prefixing methods with `Get`. Just use `Tree` or `AST`

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+			_, err := strconv.ParseFloat(t, 64)
+			if err != nil {
+				output.Push(&ast.Ident{Name: t})
+			} else {
+				output.Push(&ast.BasicLit{Kind: token.FLOAT, Value: t})
+			}
+
+		}
+	}
+	for operators.Len() > 0 {
```
emilbon99, line 145: It seems like this is a close math with the code from L119 to L134.  Would it make sense to extract into a separate method?

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package calc
+
+import (
+	"errors"
```
emilbon99, line 13: Don't use internal errors package. Use `github.com/cockroachdb/errors` instead, it adds nice features and network portability.

x/go/queue/queue.go
```diff
@@ -0,0 +1,46 @@
...
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package queue
+
+import (
+	"errors"
```
emilbon99, line 13: Don't use internal errors package. Use `github.com/cockroachdb/errors` instead.

x/go/stack/stack.go
```diff
@@ -0,0 +1,46 @@
...
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package stack
+
+import (
+	"errors"
```
emilbon99, line 13: Don't use internal errors package. Use `github.com/cockroachdb/errors` instead

x/go/queue/queue.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+type Queue[T any] struct {
```
emilbon99, line 16: For generic, STL like containers, it's a good idea to add tests and to document.

x/go/stack/stack.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+type Stack[T any] struct {
```
emilbon99, line 16: For generic, STL like containers, it's a good idea to add tests and to document.

x/go/calc/calc.go
```diff
@@ -0,0 +1,209 @@
...
+			currToken += string(c)
+		}
+	}
+	if currToken != "" {
+		tokens = append(tokens, currToken)
+	}
+	return tokens, nil
+}
+
+// Given a string representing a valid mathematical expression, sets exp to the
```
emilbon99, line 60: This is an invalidate docstring format. See https://tip.golang.org/doc/comment for a guide on documenting go code.

x/go/calc/calc_test.go
```diff
@@ -0,0 +1,230 @@
...
+				expectedTree := &ast.BinaryExpr{X: &ast.BasicLit{Kind: token.FLOAT, Value: "1"}, Op: token.ADD, Y: &ast.BasicLit{Kind: token.FLOAT, Value: "1"}}
+				actualTree := e.GetTree().(*ast.BinaryExpr)
+				Expect(treesEqual(expectedTree, actualTree)).To(BeTrue())
+			})
+			It("Should build 1+2*4", func() {
+				e := calc.Expression{}
+				err := e.Build("1 + 2 * 4")
+				Expect(err).ToNot(HaveOccurred())
+				expectedTree := &ast.BinaryExpr{X: &ast.BasicLit{Kind: token.FLOAT, Value: "1"}, Op: token.ADD, Y: &ast.BinaryExpr{X: &ast.BasicLit{Kind: token.FLOAT, Value: "2"}, Op: token.MUL, Y: &ast.BasicLit{Kind: token.FLOAT, Value: "4"}}}
+				actualTree := e.GetTree().(*ast.BinaryExpr)
```
emilbon99, line 90: Will `GetTree` always return `*ast.BinaryExpr`. If so, why don't we return the concrete type instead of an interface.

x/go/calc/calc_test.go
```diff
@@ -0,0 +1,230 @@
...
+				Expect(err).ToNot(HaveOccurred())
+				expectedTree := &ast.BinaryExpr{X: &ast.BasicLit{Kind: token.FLOAT, Value: "1"}, Op: token.ADD, Y: &ast.BasicLit{Kind: token.FLOAT, Value: "1"}}
+				actualTree := e.GetTree().(*ast.BinaryExpr)
+				Expect(treesEqual(expectedTree, actualTree)).To(BeTrue())
+			})
+			It("Should build 1+2*4", func() {
+				e := calc.Expression{}
+				err := e.Build("1 + 2 * 4")
+				Expect(err).ToNot(HaveOccurred())
+				expectedTree := &ast.BinaryExpr{X: &ast.BasicLit{Kind: token.FLOAT, Value: "1"}, Op: token.ADD, Y: &ast.BinaryExpr{X: &ast.BasicLit{Kind: token.FLOAT, Value: "2"}, Op: token.MUL, Y: &ast.BasicLit{Kind: token.FLOAT, Value: "4"}}}
```
emilbon99, line 89: For long expressions like this, expand onto multiple lines for readability.

emilbon99 (changes requested):
Great work! I can't comment too much on the algorithmic integrity, but if your tests pass I'm happy! The changes I requested are focused on improving readability.

x/go/calc/calc_test.go
t1an-xyz: The `GetTree` method was created mostly so I could make sure that the trees were correct when writing tests. In an actual application of this library, I feel like the tree should be abstracted away, so the user only needs to interact with an Expression object.

emilbon99: Ah understood, then this is OK for now. When things like this happen, however, it sometimes points to a type that has too much responsibility. Perhaps split `Expression` into a `Parser`, and an `Evaluator` and then compose them back into the original `Expression` struct?

emilbon99 (approved):
LGTM

PR #404, opened 2023-10-10
0016 frame flight protocol

Fazanza:
Please leave any comments about the code that u find bad. I will get to as soon as I can. Also will work on the document in the upcoming days


synnax/pkg/distribution/framer/encoder/encoder.go
```diff
@@ -0,0 +1,183 @@
...
+		index += 16
+	}
+
+	if channelFlag == 1 {
+		returnStruct.Keys = m.keys
+	}
+
+	for i := 0; i < len(m.keys); i++ {
+		if channelFlag == 0 {
+			if index < len(src) && channel.Key(binary.LittleEndian.Uint32(src[index:])) == m.keys[i] {
```
emilbon99, line 147: If statement could be better structured. Just check for the converse and use continue? (or is it inverse). Anyway, you don't need an else statement here.

synnax/pkg/distribution/framer/encoder/encoder_test.go
```diff
@@ -0,0 +1,113 @@
...
+				Keys:   keySlice[0:2],
+				Series: []telem.Series{series1, series5},
+			}
+			byteArray, err := cd.Encode(testStruct)
+			returnStruct, err := cd.Decode(byteArray)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(testStruct).To(Equal(returnStruct))
+		})
+	})
+})
```
emilbon99, line 113: I want performance tests!

emilbon99 (changes requested):
Some changes to make. I'd also like to see performance profiling in terms of heap allocation, especially when compared to JSON.

synnax/pkg/distribution/framer/encoder/encoder.go
```diff
@@ -0,0 +1,183 @@
...
+		curDataSize, index               int             = -1, 0
+		startTime, endTime               telem.TimeStamp = 0, 0
+		returnArray                      []byte
+		sizeFlag, alignFlag, channelFlag int = 1, 1, 0
+	)
+
+	channelFlag = map[bool]int{false: 0, true: 1}[len(m.keys) == len(src.Keys)]
+
+	for i := 0; i < len(m.keys); i++ {
+		if index >= len(src.Keys) || src.Keys[index] != m.keys[i] {
```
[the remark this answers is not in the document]
Fazanza: The way this is set up, the index will go over the len of the src keys, and was causing an error. Had to implement this condition to stop that.

emilbon99:
LGTM

PR #407, opened 2023-10-12
401 cpp channel svc

sverlon:
Not merge ready, opening for contributing comments


client/cpp/synnax/channel/channel.cpp
```diff
@@ -0,0 +1,123 @@
...
+using namespace Synnax;
+
+std::string RETRIEVE_ENDPOINT = "/channel/retrieve";
+std::string CREATE_ENDPOINT = "/channel/create";
+
+
+Channel::Channel(const api::v1::Channel &ch) {
+    name = ch.name();
+    key = ch.key();
+    index = ch.index();
```
sverlon, line 20: would probably just use a member initializer list here. I.e.
```
Channel::Channel(const api::v1::Channel &ch) 
: name(ch.name()), key(ch.key()) ... dataType(Telem::DataType(ch.data_type())) {}
```

client/cpp/synnax/channel/channel.cpp
```diff
@@ -0,0 +1,123 @@
...
+    dataType = Telem::DataType(ch.data_type());
+}
+
+Channel::Channel(
+        std::string name,
+        Telem::DataType dataType,
+        Telem::Rate rate,
+        bool is_index,
+        std::uint32_t leaseholder,
+        ChannelKey index,
```
sverlon, line 33: Continuing from above, makes this redundant. Unless you want to have this additional ctor.

client/cpp/synnax/channel/channel.cpp
```diff
@@ -0,0 +1,123 @@
...
+    a->set_data_type(data_type.value);
+    a->set_rate(rate.value);
+    auto response = create_client->send(CREATE_ENDPOINT, req);
+    return {response.first.channels(0)};
+}
+
+void ChannelClient::create(std::vector<Channel> &channels) {
+    auto req = api::v1::ChannelCreateRequest();
+    for (const auto &ch: channels) {
+        auto a = req.add_channels();
```
sverlon, line 76: The type of a is hard to infer here, and usually you only use auto when the type is easy to deduce, like

`auto a = make_unique<int>(42);`

Had to look up that this was a pointer and whatnot. Otherwise this is clean code, like the flow of passing the element pointer to `to_proto` and setting the required values, seems efficient.

client/cpp/synnax/channel/channel.cpp
```diff
@@ -0,0 +1,123 @@
...
+    auto response = create_client->send(CREATE_ENDPOINT, req);
+    for (auto i = 0; i < response.first.channels_size(); i++)
+        channels[i] = Channel(response.first.channels(i));
+}
+
+
+Channel ChannelClient::retrieve(ChannelKey key) {
+    auto req = api::v1::ChannelRetrieveRequest();
+    req.add_keys(key);
+    std::pair<api::v1::ChannelRetrieveResponse, grpc::Status> response =
```
sverlon, line 88: way cleaner to just write `auto [response, status] = retrieve_client->send(...)`.

C++ allows for unpacking pairs in a similar way to other languages, think this was as of c++11 or 14.

client/cpp/synnax/channel/channel.cpp
```diff
@@ -0,0 +1,123 @@
...
+    auto payload = api::v1::ChannelRetrieveRequest();
+    payload.add_names(name);
+    retrieve_client->send(RETRIEVE_ENDPOINT, payload);
+}
+
+std::vector<Channel> ChannelClient::retrieve(const std::vector<ChannelKey> &keys) {
+    auto req = api::v1::ChannelRetrieveRequest();
+    for (auto key: keys)
+        req.add_keys(key);
+    std::pair<api::v1::ChannelRetrieveResponse, grpc::Status> response =
```
sverlon, line 105: same as above

client/cpp/synnax/channel/channel.cpp
```diff
@@ -0,0 +1,123 @@
...
+            retrieve_client->send(RETRIEVE_ENDPOINT, req);
+    std::vector<Channel> channels;
+    for (auto i = 0; i < response.first.channels_size(); i++)
+        channels.emplace_back(response.first.channels(i));
```
sverlon, line 109: was this also an IDE suggestion? I also had people telling me `emplace_back` was better than `push_back`, still don't know why...

client/cpp/synnax/channel/channel.h
```diff
@@ -0,0 +1,72 @@
...
+#include "freighter/freighter.h"
+
+#pragma once
+
+using namespace Synnax;
+
+using ChannelKey = std::uint32_t;
+
+namespace Channel {
+    typedef Freighter::UnaryClient<
```
sverlon, line 15: From a style point of view, I wouldn't indent all of this inside the `namespace Channel` block, especially since *all* the code in this file is in this namespace.

sverlon (commented):
Didn't get through all of this, gonna continue later.

client/cpp/synnax/auth/auth.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+namespace Auth {
+    class Auth {
+    public:
+        std::string username;
+        std::string password;
+
+        Auth(std::string username, std::string password) :
+                username(std::move(username)), password(std::move(password)) {}
```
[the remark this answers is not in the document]
emilbon99: Removed

client/cpp/synnax/channel/channel.cpp
emilbon99: Fixed.

client/cpp/synnax/channel/channel.cpp
emilbon99: Fixed.

client/cpp/synnax/channel/channel.cpp
emilbon99: The code actually didn't compile when using `push_back`, still don't completely understand why. I figured out a much bettter way to do this translation so, so resolved.

client/cpp/synnax/channel/channel.h
emilbon99: Fixed the formatter in my IDE to not indent.

client/cpp/synnax/channel/channel.cpp
emilbon99: Yep, done.

client/cpp/synnax/channel/channel.cpp
emilbon99: Cleaned up these constructors quite a bit.

sverlon:
All TODOs are cleaned up in Freighter. Need to update release folder with freighter-1.2.tar for external projects.

PR #422, opened 2023-10-22
[pluto/vis] - relief valve initial commit

am3005:
Created the inline and angled relief valves in Figma and exported svg files to synnax.


emilbon99:
Rejected until we get all symbols added and tested.

PR #441, opened 2023-11-08
Flight Protocol for C++

brshin:
Flight Protocol for C++. 

The current frame and series implementation serves as a placeholder. Once the official C++ implementation is clarified, adapting the protocol to the new framework will be a straightforward process. Minimal adjustments in a few key areas will be all that's required.


emilbon99:
@sverlon can you review this code.

emilbon99:
@brshin This needs test cases using GTest.

PR #444, opened 2023-11-08
361 supporting db deletion operations

LeonLiur:
Allows deletion of cesium channel through DeleteChannel(), which checks for whether the channel is written to, then deletes the channel


cesium/internal/unary/db.go
```diff
@@ -28,9 +30,10 @@
 
 type DB struct {
 	Config
-	Domain     *domain.DB
-	Controller *controller.Controller[controlledWriter]
-	_idx       index.Index
+	Domain      *domain.DB
+	Controller  *controller.Controller[controlledWriter]
+	_idx        index.Index
+	openWriters *atomic.Int32Counter
```
emilbon99: We're also tracking the number of open iterators here, so we should change the name to something more semantic.

cesium/internal/unary/iterator.go
```diff
@@ -24,6 +24,7 @@ type Iterator struct {
 	alamos.Instrumentation
 	IteratorConfig
 	Channel  core.Channel
+	db       *DB
```
emilbon99, line 27: I think it might be a good idea not to pass in the entire DB here since we're only modifying the counter. What about passing in an anonymous function that we just call on release?

cesium/internal/unary/writer.go
```diff
@@ -40,6 +40,7 @@ func (c WriterConfig) controlTimeRange() telem.TimeRange {
 type Writer struct {
 	WriterConfig
 	Channel core.Channel
+	db      *DB
```
emilbon99, line 43: See my note inside `Iterator`. Probably want to avoid passing in the entire DB.

cesium/control.go
```diff
@@ -25,6 +25,9 @@ type ControlUpdate struct {
 }
 
 func (db *DB) ConfigureControlUpdateChannel(ctx context.Context, key ChannelKey) error {
+	if key == 0 {
+		return errors.New("[cesium] update channel key cannot be 0")
```
emilbon99, line 29: Missing a dash here -> should be "[cesium] - update channel key cannot be 0"

cesium/internal/unary/db.go
```diff
@@ -87,9 +90,19 @@ func (db *DB) OpenIterator(cfg IteratorConfig) *Iterator {
...
+		db:             db,
 	}
 	i.SetBounds(cfg.Bounds)
+	db.openIteratorWriters.Add(1)
 	return i
 }
 
+func (db *DB) TryClose() error {
+	if db.openIteratorWriters.Value() > 0 {
+		return errors.New("[unary] channel being written to")
```
emilbon99, line 102: This error might confuse people because there may be open iterators and no open writers. Should probably say `channel has N iterators/writers accessing it.`

cesium/control_test.go
```diff
@@ -29,6 +29,7 @@ var _ = Describe("Control", Ordered, func() {
 	var db *cesium.DB
 	BeforeAll(func() {
 		db = openMemDB()
+		Expect(db.ConfigureControlUpdateChannel(ctx, 0).Error()).To(ContainSubstring("cannot be 0"))
```
emilbon99, line 32: Move this into its own test case

cesium/delete.go
```diff
@@ -0,0 +1,43 @@
...
+			delete(db.unaryDBs, ch)
+			return db.fs.Remove(strconv.Itoa(int(ch)))
+		} else {
+			return err
+		}
+	}
+	vdb, vok := db.virtualDBs[ch]
+	if vok {
+		if db.GetDigestKey() == ch {
+			return errors.New("[cesium] cannot delete update digest channel")
```
emilbon99, line 32: Same formatting note. Add a `d` after bracket

cesium/delete.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+		if err := vdb.TryClose(); err == nil {
+			delete(db.virtualDBs, ch)
+			return db.fs.Remove(strconv.Itoa(int(ch)))
+		} else {
+			return errors.New("[cesium] channel being written to")
```
emilbon99, line 38: See above

cesium/delete.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/cockroachdb/errors"
+	"strconv"
+)
+
+func (db *DB) DeleteChannel(ch ChannelKey) error {
+	db.mu.Lock()
+	defer db.mu.Unlock()
+	udb, uok := db.unaryDBs[ch]
+	if uok {
+		if err := udb.TryClose(); err == nil {
```
emilbon99, line 22: You dont need an else block here because you're doing an early return. The way I typically do it is try to keep the normal or 'happy' path as un-nested as possible. I would swap this case around, check for a non-nil error and return early if you find one.

cesium/delete.go
```diff
@@ -0,0 +1,43 @@
...
+		} else {
+			return err
+		}
+	}
+	vdb, vok := db.virtualDBs[ch]
+	if vok {
+		if db.GetDigestKey() == ch {
+			return errors.New("[cesium] cannot delete update digest channel")
+		}
+		if err := vdb.TryClose(); err == nil {
```
emilbon99, line 34: Same notes as above.

cesium/control.go
```diff
@@ -99,6 +103,10 @@ func (db *DB) ControlUpdateToFrame(ctx context.Context, u ControlUpdate) Frame {
 	}
 }
 
+func (db *DB) GetDigestKey() ChannelKey {
```
emilbon99, line 106: Get rid of this method

cesium/channel_test.go
```diff
@@ -31,6 +32,11 @@ var _ = Describe("Channel", Ordered, func() {
 				Expect(db.CreateChannel(ctx, ch)).To(Succeed())
 				Expect(ch.Key).To(Equal(core.ChannelKey(1)))
 			})
+			It("Should delete a channel when we delete it", func() {
+				Expect(db.DeleteChannel(1)).To(Succeed())
```
emilbon99, line 36: Move this test case to a delete specific block

cesium/writer_behavior_test.go
```diff
@@ -83,6 +94,11 @@ var _ = Describe("Writer Behavior", Ordered, func() {
 					Start:    10 * telem.SecondTS,
 				}))
 
+				Expect(db.DeleteChannel(basic1)).ToNot(Succeed())
```
emilbon99, line 97: I think we should move this into its own test

cesium/streamer_behavior_test.go
```diff
@@ -136,6 +140,9 @@ var _ = Describe("Streamer Behavior", Ordered, func() {
 			Expect(r.Frame.Keys).To(HaveLen(1))
 			i.Close()
 			Expect(sCtx.Wait()).To(Succeed())
+
+			Expect(db.DeleteChannel(basic3)).To(Succeed())
+			Expect(db.DeleteChannel(controlKey).Error()).To(ContainSubstring("digest"))
```
emilbon99, line 145: Shouldn't assert on this here

emilbon99 (changes requested):

cesium/internal/virtual/db.go
```diff
@@ -27,24 +30,37 @@ func (e *controlEntity) ChannelKey() core.ChannelKey { return e.ck }
...
 	Config
-	controller *controller.Controller[*controlEntity]
+	controller  *controller.Controller[*controlEntity]
+	openWriters *atomic.Int32Counter
 }
 
 type Config struct {
 	alamos.Instrumentation
 	Channel core.Channel
+	FS      fs.FS
```
emilbon99, line 40: Remove this

emilbon99 (changes requested):

cesium/delete.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+		}
+	}
+
+	return udb.Delete(ctx, tr)
+}
+
+func (db *DB) GarbageCollect(ctx context.Context, maxsizeRead uint32, maxGoRoutine int64) (collected bool, err error) {
+	db.mu.RLock()
+	defer db.mu.RUnlock()
```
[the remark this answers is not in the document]
LeonLiur: Even if it is a Rlock? Rlock would not block other read and write operations right?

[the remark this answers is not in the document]
emilbon99: RLock will not block concurrent reads, but it will block concurrent writes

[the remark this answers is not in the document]
emilbon99: So for the entire time you're garbage collecting, deleting or creating channels cannot happen

cesium/internal/domain/file_controller.go
```diff
@@ -200,19 +207,56 @@ func (fc *fileController) gcReaders() (bool, error) {
...
-			fc.writers.RUnlock()
-			fc.writers.Lock()
-			err := w.TrackedWriteCloser.Close()
-			fc.writers.open = append(fc.writers.open[:i], fc.writers.open[i+1:]...)
-			fc.writers.Unlock()
-			return true, err
+	fc.readers.RLock()
+	defer func() {
+		fc.writers.RUnlock()
+		fc.readers.RUnlock()
```
Lham42, line 215: I would generally release locks in the reverse order you acquired them because it lets you better ensure a deadlock won't occur (I can explain more since it might not be immediately clear why)

cesium/internal/domain/db.go
```diff
@@ -138,6 +140,183 @@ func (db *DB) NewIterator(cfg IteratorConfig) *Iterator {
...
+		db.idx.mu.RUnlock()
+		return errors.New("Invalid starting position")
+	}
+	end, ok := db.idx.get(endPosition)
+	if !ok {
+		db.idx.mu.RUnlock()
+		return errors.New("Invalid ending position")
+	}
+	db.idx.mu.RUnlock()
+
```
Lham42, line 168: So this is what I think the process youre doing in this function:
1. you acquire a read lock to get the start and end positions
2. you release the read lock and are holding no locks 
3. you acquire a write lock and then do more work involving the start and end positions

Can you guarantee that in step 2, another thread (goroutine?) doesn't come in and alter the start and end positions? I am not convinced that you can.

Lham42 (changes requested):

emilbon99 (changes requested):

cesium/internal/domain/index.go
```diff
@@ -21,7 +21,8 @@ type index struct {
 	alamos.Instrumentation
 	mu struct {
 		sync.RWMutex
-		pointers []pointer
+		pointers   []pointer
+		tombstones map[uint16][]pointer
```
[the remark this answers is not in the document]
LeonLiur: It's easier to garbage collect all the tombstones in a file all together

x/go/io/fs/fs.go
```diff
@@ -98,8 +114,19 @@ func (d *defaultFS) List(pth string) ([]os.FileInfo, error) {
 	return infos, nil
 }
 
-func OSDirFS(dir string) (FS, error) {
-	return (&defaultFS{perm: defaultPerm}).Sub(dir)
+func (d *defaultFS) Remove(pth string) error {
+	if e, err := d.Exists(pth); err != nil || !e {
```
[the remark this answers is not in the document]
LeonLiur: I'm a bit confused on this one.

I think I understand the idea: is that if Remove runs without error, we can be 100% certain that that path no longer exists, correct?

cesium/internal/domain/db.go
```diff
@@ -138,10 +140,187 @@ func (db *DB) NewIterator(cfg IteratorConfig) *Iterator {
...
+	p := db.idx.mu.pointers[0]
+	tr.Start = p.Start
+	p = db.idx.mu.pointers[len(db.idx.mu.pointers)-1]
+	tr.End = p.End
+
+	return tr
+}
+
+// Delete tombstones all pointers ranging from [db.get(startPosition).start + startOffset, db.get(endPosition).end - endOffset)
+func (db *DB) Delete(ctx context.Context, startPosition int, endPosition int, startOffset int64, endOffset int64, tr telem.TimeRange) error {
```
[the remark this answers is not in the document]
LeonLiur: The reason we need no lock here is because in its caller, unary delete, a lock is already acquired through a `LockedIterator`, so this thread already has a lock. This is unintuitive behaviour though and even took me a while to remember. I will pass in a flag like `withLock bool`

emilbon99 (changes requested):

cesium/internal/controller/controller.go
```diff
@@ -273,27 +285,103 @@ func (c *Controller[E]) LeadingState() *State {
...
+		// Check if there is an existing region that overlaps with that time range,
+		// if there is, then that means a gate is already in control of it,
+		// therefore this method should not create an absolute gate.
+		if r.timeRange.OverlapsWith(tr) {
+			if exists {
+				return nil, t, errors.Newf("[controller] - encountered multiple control regions for time range %s", tr)
+			}
+
+			r.Lock()
+			if r.curr != nil {
```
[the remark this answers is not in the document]
LeonLiur: it's possible for r.curr to be nil, when we register a region using `register`, r.curr will be nil

cesium/internal/domain/db.go
```diff
@@ -138,10 +140,187 @@ func (db *DB) NewIterator(cfg IteratorConfig) *Iterator {
...
+
+	newPointers := []pointer{
+		{
+			TimeRange: telem.TimeRange{
+				Start: start.Start,
+				End:   tr.Start,
+			},
+			fileKey: start.fileKey,
+			offset:  start.offset,
+			length:  uint32(startOffset), // length from start.Start to tr.Start
```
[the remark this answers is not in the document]
LeonLiur: Yes, I could use a if statement to check it – would this not result in an extra hotpath comparison? Or is this case not a "hotpath"?

cesium/options.go
```diff
@@ -51,6 +53,12 @@ func WithFS(fs xfs.FS) Option {
 	}
 }
 
+func WithGC(config *GCConfig) Option {
```
emilbon99, line 56: Why do you need to pass this as a pointer? Also, where do you merge in the default options that you defined?

x/go/io/counter.go
```diff
@@ -11,21 +11,22 @@ package io
...
+	mu      sync.RWMutex
 	buf     []byte
 }
 
 // NewInt32Counter opens a new, atomic counter backed by the given file. The counter
 // must have exclusive write access to the file.
 func NewInt32Counter(f ReaderAtWriterAtCloser) *Int32Counter {
 	i := &Int32Counter{
-		wrapped: &atomic.Int32Counter{},
+		wrapped: int32(0),
```
emilbon99, line 29: No need to initialize this field. Go will already do it for you.

x/go/io/fs/fs.go
```diff
@@ -43,19 +46,31 @@ type subFS struct {
...
-	return s.FS.Sub(path.Join(s.dir, name))
+	return s.FS.Sub(goPath.Join(s.dir, name))
 }
 
 func (s *subFS) Exists(name string) (bool, error) {
-	return s.FS.Exists(path.Join(s.dir, name))
+	return s.FS.Exists(goPath.Join(s.dir, name))
 }
 
 func (s *subFS) List(name string) ([]os.FileInfo, error) {
```
emilbon99, line 60: Be consistent as to whether you're using `name` or `path`

emilbon99 (changes requested):

cesium/internal/domain/iterator.go
```diff
@@ -52,8 +52,26 @@ type Iterator struct {
...
+	locked bool
+	// onClose stores a function that runs upon closing or unlocking the iterator, e.g. unlocking a Mutex
+	onUnlock func()
 	// readerFactory gets a new reader for the given domain pointer.
 	readerFactory func(ctx context.Context, ptr pointer) (*Reader, error)
+	closed        bool
+}
+
+// Lock sets an iterator to be locked, and takes in a function that gets executed directly, and another function to be stored in onClose
+func (i *Iterator) Lock(onLock func(), onUnlock func()) {
```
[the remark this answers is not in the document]
LeonLiur: I agree this is confusing, but I tried other ways to implement this and this is the one that I think works best (can remove Unlock).

It is not true that every time we create a locked iterator, we want to lock the mutex, but if we leaving the duty of locking the iterator to the caller, out-of-package callers won't be able to lock the mutex.

Therefore I think this is the cleanest way to allow the user to customize whether they want to lock the mutex. I could do a boolean flag, but I think another `withLock` just makes the matter worse.

Let's discuss this more.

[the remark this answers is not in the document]
emilbon99: "It is not true that every time we create a locked iterator, we want to lock the mutex, but if we leaving the duty of locking the iterator to the caller, out-of-package callers won't be able to lock the mutex." -> Isn't this the definition of a Locked Iterator? What would be 'locked' about it if it doesn't lock anything? Why not use a regular iterator in this case?

cesium/internal/domain/iterator.go
```diff
@@ -113,21 +118,33 @@ func (i *Iterator) Prev() bool {
...
 	return i.readerFactory(ctx, i.value)
 }
 
 // Len returns the number of bytes occupied by the telemetry in the current domain.
 func (i *Iterator) Len() int64 { return int64(i.value.length) }
 
 // Close closes the iterator.
-func (i *Iterator) Close() error { return nil }
+func (i *Iterator) Close() error {
+	if i.closed {
```
emilbon99, line 141: I don't think this clause is necessary. I generally prefer it when methods like `Close` are idempotent. Unlike other iterator methods, there's no hard in letting someone call `Close` as much as they want.

cesium/internal/unary/db.go
```diff
@@ -83,13 +88,63 @@ func (db *DB) OpenIterator(cfg IteratorConfig) *Iterator {
...
 	i := &Iterator{
-		idx:            db.index(),
-		Channel:        db.Channel,
-		internal:       iter,
-		IteratorConfig: cfg,
+		idx:              db.index(),
+		Channel:          db.Channel,
+		internal:         iter,
+		IteratorConfig:   cfg,
+		decrementCounter: func() { db.openIteratorWriters.Add(-1) },
```
emilbon99, line 95: This is my stylistic preference, but I would generally name this function `onClose` because that makes it clearer that it should only be called when the iterator is closed, and it could be expanded to do other things in the future (such as end a trace) instead of just decrementing a counter.

emilbon99 (approved):
LGTM. I left a few small, stylistic notes.

PR #460, opened 2024-01-21
459 queue

t1an-xyz:
Implement a deque

Modify stack and queue to use a deque as the underlying container

Modify `Peek()` in stack and queue to return a copy rather than a pointer

Modify calc to accommodate for the previous change


emilbon99:
If stack and queue are just using a deque under the hood, then why are we defining them as concrete types? Can't we just (1) use a deque for everything instead or (2) have the stack and queue defined as interfaces that the deque implements?

emilbon99 (commented):
See my comment

t1an-xyz:
> If stack and queue are just using a deque under the hood, then why are we defining them as concrete types? Can't we just (1) use a deque for everything instead or (2) have the stack and queue defined as interfaces that the deque implements?

I mostly did that for backwards compatibility since I thought it was bad practice to remove code in case some older code relies on it. I have replaced the stack used in the calc library with a deque.

emilbon99:
Fantastic! This looks great

PR #464, opened 2024-02-10
463 search quality

t1an-xyz:
This PR improves the search used in the ontology library. The search now ignores case and is also more restrictive. Before, the search algorithm would return a result if any word in the search phrase matches the index. Now, the search only returns if every word matches. This should, overall, improve the quality of the search.

**Important Limitations**
Due to poor documentation, I was unable to get the indexing algorithm the ignore underscores. In the future, if someone can figure out the Bleve library, it would be good to create a custom tokenizer that splits strings by spaces, underscores, dashes, and periods. Instead, I used a regex workaround in the Disjunction Search.


t1an-xyz:
Create a function in search.go that takes a string and escapes regex symbols. For example, given the string `"A* algorithm"`, the function should return `"A\* algorithm`. `"array[i-1]"` should return `"array\[i\-1\]"`. `"hello\s"` should return `"hello\\s"`. And `"hello\\s"` should return `"hello\\\\s`. `"question?"` should return `"question\?"`. Etc.

emilbon99:
Merged this PR in the 0.16.0 release.

PR #523, opened 2024-04-23
522 sy 550 cesium fix mathmaxuint32 digest check in virtual writer

LeonLiur:


cesium/writer_stream.go
```diff
@@ -493,7 +500,7 @@ func (w virtualWriter) Close() (ControlUpdate, error) {
 		Transfers: make([]controller.Transfer, 0, len(w.internal)),
 	}
 	for _, chW := range w.internal {
-		if chW.Channel.Key == math.MaxUint32 {
+		if chW.Channel.Key == w.digestKey {
```
emilbon99, line 503: Please leave a comment as to why we do this!! It's very hard to parse if you don't already know

emilbon99 (changes requested):

PR #527, opened 2024-04-24
524 sy 578 make channel deletion work by renaming folder

LeonLiur:
Note: this does not prevent the database from entering undefined state if the user repeatedly deletes a channel and then creating one with the same key in a very short duration. Possible workthrough is generating a random sequence trailing the directory name to delete.


cesium/delete.go
```diff
@@ -56,20 +57,31 @@ func (db *DB) DeleteChannel(ch ChannelKey) error {
 		return err
 	}
 
-	db.mu.Unlock()
+	// Rename the file first, so we can avoid hogging the mutex while deleting the directory
+	// may take a longer time.
+	// Rename the file to have a random suffix in case the channel is repeatedly created
+	// and deleted.
+	var newName = channelDirName(ch) + "-DELETE-" + strconv.Itoa(rand.Int())
```
emilbon99, line 64: Don't use var here. Use a `:=` initializer.

cesium/delete.go
```diff
@@ -56,20 +57,31 @@ func (db *DB) DeleteChannel(ch ChannelKey) error {
...
+	err = db.fs.Rename(channelDirName(ch), newName)
```
emilbon99, line 65: Minor note, but you call `channelDirName` twice. It has no impact on performance, but from a style perspective i'd make a new variable called `oldName` and call the function once.

cesium/delete.go
```diff
@@ -56,20 +57,31 @@ func (db *DB) DeleteChannel(ch ChannelKey) error {
...
 	var (
 		indexChannels       = make([]ChannelKey, 0)
 		directoriesToRemove = make([]string, 0)
 	)
 
 	defer func() {
-		// Pool all directories to remove
+		db.mu.Unlock()
+
 		for _, name := range directoriesToRemove {
```
emilbon99, line 85: Why is this in a defer statement?

cesium/internal/unary/unary_suite_test.go
```diff
@@ -20,15 +19,16 @@ import (
...
 )
 
 func TestUnary(t *testing.T) {
 	RegisterFailHandler(Fail)
-	fileSystems, cleanUp = testutil.FileSystems()
 	RunSpecs(t, "Unary Suite")
-	Expect(cleanUp()).To(Succeed())
 }
+
+var _ = BeforeSuite(func() { Expect(fileSystems).To(HaveLen(2)) })
```
emilbon99, line 32: Is this assertion useful?

emilbon99 (changes requested):

cesium/internal/unary/unary_suite_test.go
LeonLiur: Yes, if there is an error in creating the osFS `tempDir`, it would not get added to the map. we would know if that went wrong this way.

PR #532, opened 2024-05-05
516 sy 380 writer auto commit

LeonLiur:


cesium/db.go
```diff
@@ -67,7 +67,7 @@ func (db *DB) WriteArray(ctx context.Context, key core.ChannelKey, start telem.T
 }
 
 // Read implements DB.
-func (db *DB) Read(ctx context.Context, tr telem.TimeRange, keys ...core.ChannelKey) (frame Frame, err error) {
+func (db *DB) Read(_ context.Context, tr telem.TimeRange, keys ...core.ChannelKey) (frame Frame, err error) {
```
Lham42, line 70: is context used in this function? Otherwise why have it even as optional?

cesium/internal/domain/index.go
```diff
@@ -78,20 +101,50 @@ func (idx *index) overlap(tr telem.TimeRange) bool {
...
-	idx.mu.RLock()
+	idx.mu.Lock()
+
+	defer span.End()
+
 	if len(idx.mu.pointers) == 0 {
-		idx.mu.RUnlock()
-		return span.EndWith(RangeNotFound)
+		idx.mu.Unlock()
+		return span.Error(RangeNotFound)
```
Lham42, line 112: do you want a test that covers the case of the condition on 110 to be true?  maybe codecov is just not up to date and the tests you have above test it already but just wanted to make sure

cesium/internal/domain/index.go
```diff
@@ -78,20 +101,50 @@ func (idx *index) overlap(tr telem.TimeRange) bool {
...
 	}
-	idx.mu.RUnlock()
-	return span.EndWith(idx.updateAt(ctx, updateAt, p))
+
+	ptrs := idx.mu.pointers
+	oldP := ptrs[updateAt]
+	if oldP.Start != p.Start {
+		idx.mu.Unlock()
+		return span.Error(RangeNotFound)
+	}
```
Lham42, line 125: Want to test when the condition on 122 is true as well? maybe the codecov

x/go/testutil/cesium_test_utils.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/synnaxlabs/cesium"
+)
+
+var k cesium.ChannelKey = 1
+
+func GenerateCesiumChannelKey() (key cesium.ChannelKey) {
+	key = k
+	k++
+	return
```
Lham42, line 12: code cov

cesium/writer_open.go
```diff
@@ -148,11 +178,11 @@ func (db *DB) newStreamWriter(ctx context.Context, cfgs ...WriterConfig) (w *str
 			return
 		}
 		for _, idx := range domainWriters {
-			_, err_ := idx.Close()
+			_, err_ := idx.Close(ctx)
```
Lham42, line 181: codecov

cesium/writer_stream.go
```diff
@@ -233,15 +233,22 @@ func (w *streamWriter) sendRes(req WriterRequest, ack bool, err error, end telem
...
 			if errors.Is(err, control.Unauthorized) && !*w.ErrOnUnauthorized {
 				return nil
 			}
 			return err
 		}
+
+		if w.Mode.AutoCommit() {
+			_, err = idx.Commit(ctx, true)
+			if err != nil {
+				return
```
Lham42, line 249: is there no other action that needs to be taken on error?

Lham42 (changes requested):

cesium/writer_stream.go
LeonLiur: This is a behaviour I'm not sure about: ideally, we would want a failed commit (which is most likely due to the domain of the written telemetry being wrong) to also "revert" the write so the writer could re-try writing correct data. However, this is really hard and could be really slow since the data is already written to disk. So in this case we just return an error and not do anything. This is also partially because our driver when encountering a write error would just make a new writer instead of trying to write again.

Is this the desired behaviour? Does this make sense?

cesium/db.go
emilbon99: This implements an interface for forward compatibility

x/go/testutil/cesium_test_utils.go
emilbon99: This function will eventually fuck us if we try to run these tests concurrently. Just something to be aware of

cesium/internal/domain/index.go
LeonLiur: I think this case never happens: `update` is only called in `Commit`, and `Commit` would only call `update` if prevCommit is not 0, which means the `len(pointers) != 0`. This is just here as a safety, even though it never happens. Should this still be in the code?

cesium/internal/domain/index.go
LeonLiur: Same as above

cesium/internal/domain/writer.go
```diff
@@ -162,17 +186,27 @@ func (w *Writer) Commit(ctx context.Context, end telem.TimeStamp) error {
...
-func (w *Writer) Close() error {
+// uncommitted data will be discarded. Any committed, but unpersisted data will be persisted.
+// Close is idempotent, and is also not safe to call concurrently with any other writer methods.
+func (w *Writer) Close(ctx context.Context) error {
 	if w.closed {
 		return nil
 	}
+
+	if !w.AutoPersistTime.IsZero() {
+		w.idx.persist(ctx)
```
emilbon99, line 207: I'm a bit sussed out about this. It doesn't need a mutex lock on the index?

cesium/writer_open.go
```diff
@@ -27,18 +27,40 @@ import (
 
 // WriterMode sets the operating mode of the writer, optionally enabling or disabling
 // persistence and streaming.
+// WriterMode is an uint8, its right-most bit represents the value of Persist(),
+// second-to-right-most bit represents the value of Stream(), and the third-to-right-most
+// bit represents the value of AutoCommit().
 type WriterMode uint8
```
emilbon99, line 33: You need to make sure you let the rest of the team know when you make breaking changes to the public API of Cesium like this. This change to writer mode has pretty dramatic implications all the way upstream to the C++, TS, Python libraries as well as the Server.

emilbon99: If we're going to go the route of using bit masks here, why don't we just make them separate booleans in the config instead? I know you use bitmasks for file modes, but IMO it's confusing. Using booleans is way simpler.

cesium/writer_stream.go
```diff
@@ -336,65 +343,66 @@ type idxWriter struct {
...
 		}
 
-		if l == -1 {
-			l = fr.Series[i].Len()
+		if lengthOfFrame == -1 {
+			lengthOfFrame = fr.Series[i].Len()
 		}
 
-		if s.count == w.writeNum {
+		if uWriter.timesWritten == w.numWrites {
```
emilbon99, line 367: Why can't we track this info internally to the unary writer?

cesium/writer_stream.go
```diff
@@ -336,65 +343,66 @@ type idxWriter struct {
...
 	sampleCount int64
 }
 
-func (w *idxWriter) Write(fr Frame) (Frame, error) {
+func (w *idxWriter) Write(ctx context.Context, fr Frame) (Frame, error) {
 	var (
-		l int64 = -1
-		c       = 0
+		lengthOfFrame     int64 = -1
+		channelsWrittenTo       = 0
```
emilbon99, line 354: If you're going to rename this, at least make it related to a count i.e. `numChannelsWrittenTo` or `channelsWrittenToCount`

emilbon99: And be consistent across the naming

cesium/writer_stream.go
```diff
@@ -420,7 +428,7 @@ func (w *idxWriter) Write(fr Frame) (Frame, error) {
 	return fr, nil
 }
 
-func (w *idxWriter) Commit(ctx context.Context) (telem.TimeStamp, error) {
+func (w *idxWriter) Commit(ctx context.Context, autoPersist bool) (telem.TimeStamp, error) {
```
emilbon99, line 431: Why do we need to pass `autoPersist` around everywhere? Can't we pass this as part of the configuration?

cesium/writer_stream.go
emilbon99: @Lham42 Are you confused about why he's not returning the error here. If so, it's because go has implicit returns of named variables.

cesium/writer_stream.go
emilbon99: If you need to do this because we have a `writer.SetMode` function and we need to do this dynamically. If so, let's get rid of the ability to `SetMode` on the writer and just make it so you need to re-instantiate the writer if you want to turn auto commit on or off.

cesium/internal/domain/index.go
Lham42: If that is the correct way to handle events that should never happen in go (i just am not convinced it is yet), then it would be fine. In C++ I'd place an assertion for things I'm positive would not happen but not sure if the same paradigm is used here (stack suggests panic would be the equivalent way to handle). In any case, make sure its clear in the code at least with a comment that this case should never happen

cesium/writer_stream.go
Lham42: Ok that makes sense. It wasn't clear to me if errors were handled but if they can be and are handled by the caller then this is fine.

x/go/testutil/cesium_test_utils.go
LeonLiur: Yeah made this atomic

cesium/internal/domain/writer.go
LeonLiur: You're right it does

cesium/internal/domain/index.go
LeonLiur: @emilbon99 What should happen here? Should panic?

cesium/cesium_suite_test.go
```diff
@@ -34,7 +34,7 @@ func openDBOnFS(fs xfs.FS) *cesium.DB {
 	))
 }
 
-func pathInDBFromKey(key cesium.ChannelKey) string {
+func channelKeyToPath(key cesium.ChannelKey) string {
```
emilbon99, line 37: Much better name.

cesium/internal/domain/index.go
```diff
@@ -34,13 +35,17 @@ type indexUpdate struct {
...
+
 	defer span.End()
-	idx.mu.RLock()
+
 	insertAt := 0
+
 	if p.fileKey == 0 {
-		panic("fileKey must be set")
+		idx.mu.Unlock()
+		idx.L.DPanic("fileKey must be set")
```
emilbon99, line 48: Remember that this will only panic in development. In production this function will continue execution. Are you sure you don't want to return here?

cesium/internal/domain/index.go
```diff
@@ -49,14 +54,33 @@ func (idx *index) insert(ctx context.Context, p pointer) error {
...
+		idx.mu.pointers = append(idx.mu.pointers[:insertAt], append([]pointer{p}, idx.mu.pointers[insertAt:]...)...)
+	}
+
+	idx.persistHead = min(idx.persistHead, insertAt)
+
+	if persist {
+		update := indexUpdate{afterIndex: idx.persistHead}
+		idx.persistHead = len(idx.mu.pointers) - 1
+		idx.mu.Unlock()
+		idx.Observer.Notify(ctx, update)
```
emilbon99, line 78: Honestly we should probably just get rid of the observer here and call the persist function directly. Adding an observer just adds complexity and was kinda dumb on my part.

cesium/internal/domain/index.go
```diff
@@ -102,45 +163,11 @@ func (idx *index) beforeFirst(ts telem.TimeStamp) bool {
...
-
-func (idx *index) modifyAfter(ctx context.Context, i int, f func()) {
-	update := indexUpdate{afterIndex: i}
-	defer func() {
-		idx.mu.Unlock()
-		idx.Observer.Notify(ctx, update)
-	}()
-	idx.mu.Lock()
-	f()
+func (idx *index) persist(ctx context.Context) {
```
emilbon99, line 166: Why aren't you calling this function on line 78 instead of calling Notify explicitly?

cesium/internal/domain/index.go
```diff
@@ -78,20 +102,57 @@ func (idx *index) overlap(tr telem.TimeRange) bool {
...
+		idx.mu.pointers[updateAt] = p
+	}
+
+	idx.persistHead = min(idx.persistHead, updateAt)
+
+	if persist {
+		update := indexUpdate{afterIndex: idx.persistHead}
+		idx.persistHead = len(idx.mu.pointers) - 1
+		idx.mu.Unlock()
+		idx.Observer.Notify(ctx, update)
```
emilbon99, line 149: is section of code looks eerily similar to the code on line 74. Can we consolidate?

cesium/internal/domain/writer.go
```diff
@@ -80,12 +85,13 @@ func Write(ctx context.Context, db *DB, tr telem.TimeRange, data []byte) error {
...
-	fileKey    uint16
-	internal   xio.TrackedWriteCloser
-	presetEnd  bool
-	closed     bool
+	prevCommit  telem.TimeStamp
+	idx         *index
+	fileKey     uint16
+	internal    xio.TrackedWriteCloser
+	presetEnd   bool
+	lastPersist telem.TimeStamp
```
emilbon99, line 93: I would rename this variable to `lastIndexPersist` or something. That makes it clearer that we're tracking index persistence instead of some other form of general persistence.

cesium/internal/domain/writer.go
```diff
@@ -140,16 +147,35 @@ func (w *Writer) Write(p []byte) (n int, err error) {
...
 // return an error.
+// If WriterCommit.AutoPersistInterval is greater than 0, then the changes committed would only
+// be persisted to disk after the set interval.
 func (w *Writer) Commit(ctx context.Context, end telem.TimeStamp) error {
+	var (
+		now     = telem.Now()
+		persist = w.AutoPersistInterval <= 0 || w.lastPersist.Span(now) >= w.WriterConfig.AutoPersistInterval
+	)
+
+	if w.AutoPersistInterval > 0 && persist {
```
emilbon99, line 158: You just checked if the interval is less than or equal to 0. `w.AutoPersistInterval <= 0` will always return false if `w.AutoPersistInterval > 0`, which means your or check is redundant no?

cesium/internal/domain/writer_behavior_test.go
```diff
@@ -10,18 +10,39 @@
...
+	xfs "github.com/synnaxlabs/x/io/fs"
 	"github.com/synnaxlabs/x/telem"
 	. "github.com/synnaxlabs/x/testutil"
 	"github.com/synnaxlabs/x/validate"
+	"os"
+	"sync"
+	"time"
 )
 
+func extractPointer(f xfs.File) (p struct {
```
emilbon99, line 28: This is ok, and I get why you've created an anomymous struct instead of making pointer public, but ngl maybe we should make pointer public

cesium/internal/unary/delete_test.go
```diff
@@ -21,7 +21,7 @@ import (
 
 var _ = Describe("Delete", Ordered, func() {
 	for fsName, fs := range fileSystems {
```
emilbon99, line 23: Minor, but maybe you should rename the fs creator to `newFS` or something? Kinda confusing when you're assigning `fs := fs()`

cesium/internal/unary/writer.go
```diff
@@ -22,14 +22,16 @@ import (
...
-	End       telem.TimeStamp
-	Subject   control.Subject
-	Authority control.Authority
-	Persist   *bool
+	Start           telem.TimeStamp
+	End             telem.TimeStamp
+	Subject         control.Subject
+	Authority       control.Authority
+	Persist         *bool
+	AutoPersistTime telem.TimeSpan
```
emilbon99, line 34: Should be `AutoPersistInterval` no?

emilbon99: Also I just realized it but this variable is kind of confusing, and doesn't explain that this controls domain persistence and not persistence in general. Something to thing about. Also we should document all of the properties of this struct.

cesium/writer_stream.go
```diff
@@ -106,6 +103,8 @@ type streamWriter struct {
 	updateDBControl func(ctx context.Context, u ControlUpdate)
 }
 
+func (w *streamWriter) AutoCommitEnabled() bool { return w.AutoCommitInterval != NoAutoCommit }
```
emilbon99, line 106: Does this method need to be public?

emilbon99 (approved):
Almost there! I think it needs a few minor changes

cesium/internal/domain/index.go
LeonLiur: Agreed – also allows us to get rid of the unlocking - relocking semantic. Was there any advantage in having an observer? I guess the index persist would run concurrently with other db operations so faster?

cesium/internal/domain/writer_behavior_test.go
LeonLiur: I think unless we need to use something private in the actual code, there is no need to making that thing public for the sake of testing – I think a smaller interface is probably better

client/py/synnax/framer/writer.py
```diff
@@ -102,6 +104,9 @@ class Writer:
     error, close will raise the accumulated error.
     """
 
+    # constant
+    ALWAYS_INDEX_PERSIST_ON_AUTO_COMMIT: TimeSpan = -1
```
emilbon99, line 108: Why is this a static variable on the request class? Shouldn't you move it to the module scope?

client/py/tests/test_framer.py
```diff
@@ -92,6 +93,55 @@ def test_write_frame_idx_no_timestamps(
...
+                                enable_auto_commit=True,
+                                auto_index_persist_interval=50 * TimeSpan.MILLISECOND) as w:
+            data = np.random.rand(10).astype(np.float64)
+            w.write(pd.DataFrame({channel.key: data}))
+            w.write(pd.DataFrame({channel.key: data}))
+
+        f = client.read(TimeRange(0, TimeStamp(1 * TimeSpan.SECOND)), channel.key)
+        assert f.__len__() == 20
+
+    def test_writer_creation_with_bad_params(self, channel: sy.Channel, client: sy.Synnax):
```
emilbon99, line 134: I don't know if these should actually be error cases. Can't it just be that any negative number means always persist? And can't we just ignore the auto persist interval if auto commit is off?

client/ts/README.md
```diff
@@ -10,3 +10,16 @@ Detailed documentation is available [here](https://docs.synnaxlabs.com/typescrip
 npm install @synnaxlabs/client
 ```
 
+## Development
+
+Synnax's TypeScript client unit tests are written in [Vitest](https://vitest.dev/). To test the framework, run
+
+```bash
+npm test
```
emilbon99, line 18: We use pnpm. It should be `pnpm test`

client/ts/src/framer/writer.spec.ts
```diff
@@ -7,11 +7,11 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-import { DataType, Rate, TimeRange, TimeStamp } from "@synnaxlabs/x";
+import {DataType, Rate, TimeRange, TimeSpan, TimeStamp} from "@synnaxlabs/x";
```
emilbon99, line 10: There should be a space between the braces on imports

client/ts/src/framer/writer.ts
```diff
@@ -49,12 +49,16 @@ export enum WriterMode {
   StreamOnly = 3,
 }
 
+export const AlwaysIndexPersistOnAutoCommit : TimeSpan = new TimeSpan(-1)
```
emilbon99, line 52: This constant should be in all caps.

client/py/tests/test_framer.py
LeonLiur: Agreed

Lham42 (approved):

client/py/synnax/framer/writer.py
emilbon99: @LeonLiur You never addressed this

LeonLiur: Moved to module scope

PR #535, opened 2024-05-07
533 sy 554 writer auto file cutoff

LeonLiur:
Added file size check in domain/writer `Commit()`: if the current file size is too big, writer will commit the current domain, and start a new domain in a different file. The current domain's end timestamp is temporarily used as the domain writer's start timestamp. Upon next `unary/writer.Write()`call, a Stamp call is made to change the start timestamp of both unary and domain writers to the first timestamp written in the next domain.

Breaking / changed behaviour: When writing with a preset end, if the last commit of a writer is one where a file change was done, the End timestamp of that domain is the actual end, instead of the preset end (this is because if we set it to the preset end, upon opening the new domain, we would get an overlap error). Is this important?


cesium/internal/domain/file_controller.go
```diff
@@ -74,10 +74,10 @@ func (fc *fileController) acquireWriter(ctx context.Context) (uint16, xio.Tracke
 	for fileKey, w := range fc.writers.open {
 		s, err := fc.FS.Stat(fileKeyName(fileKey))
 		if err != nil {
-			return 0, nil, err
+			return 0, nil, span.Error(err)
```
Lham42, line 77: is there no test for error handling from fc.fs.stat()

cesium/internal/domain/file_controller.go
```diff
@@ -130,7 +130,11 @@ func (fc *fileController) newWriter(ctx context.Context) (*controlledWriter, err
 	fc.writers.Lock()
 	fc.writers.open[nextKey] = w
 	fc.writers.Unlock()
-	return &w, err
+	if err != nil {
+		return &w, span.Error(err)
```
Lham42, line 134: same as above

cesium/internal/domain/file_controller.go
Lham42: resolved

cesium/internal/domain/file_controller.go
Lham42: resolved

cesium/internal/unary/writer.go
```diff
@@ -195,8 +195,19 @@ func (w *Writer) commitWithEnd(ctx context.Context, end telem.TimeStamp) (telem.
...
 		return 0, controller.Unauthorized(w.control.Subject.String(), w.Channel.Key)
 	}
+
 	if end.IsZero() {
-		// we're using w.len - 1 here because we want the timestamp of the last
+		if w.lastCommitFileSwitch {
+			// Correct the start position.
+			approx, err := w.idx.Stamp(ctx, dw.Start, 1, true)
+			if err != nil {
+				return 0, err
```
Lham42, line 204: no panic?

cesium/internal/domain/writer.go
```diff
@@ -139,31 +141,66 @@ func (w *Writer) Write(p []byte) (n int, err error) {
...
+	} else {
+		w.prevCommit = commitEnd
+	}
+
+	return span.Error(f(ctx, ptr))
+}
+
+// resolveCommitEnd returns whether a file change is needed, the resolved commit end, and any errors.
+func (w *Writer) resolveCommitEnd(end telem.TimeStamp) (bool, telem.TimeStamp, error) {
+	s, err := w.fc.FS.Stat(fileKeyName(w.fileKey))
```
emilbon99, line 194: I don't like calling `fs.Stat` here. We have an exclusive write lock on the file and we know how many bytes are written to the domain, so we know the size of the file. Why make the extra sys call? Also I don't like accessing the FS directly inside of the writer.

emilbon99: Also why is this logic in `resolveCommitEnd`? It has nothing to do with resolving the commit end.

emilbon99: Also this only checks the size of the file after we commit. This means that if we just kept writing data without committing, it would never roll over the file?

emilbon99 (changes requested):

cesium/internal/domain/file_controller.go
```diff
@@ -126,7 +128,11 @@ func (fc *fileController) newWriter(ctx context.Context) (*controlledWriter, err
 	fc.writers.Lock()
 	fc.writers.open[nextKey] = w
 	fc.writers.Unlock()
-	return &w, err
+	if err != nil {
```
emilbon99, line 131: What's the point of this check?

cesium/internal/domain/writer.go
```diff
@@ -120,6 +120,12 @@ type Writer struct {
 	// fileKey represents the key of the file written to by the writer. One can convert it
 	// to a filename via the fileKeyToName function.
 	fileKey uint16
+	// fc is the file controller for the writer's FS.
+	fc *fileController
+	// fileSize is the writer's file's size
+	fileSize telem.Size
```
emilbon99, line 126: Just curious as to why were tracking both `fileSize` and `lenWritten`

cesium/internal/domain/writer.go
```diff
@@ -120,6 +120,12 @@ type Writer struct {
 	// fileKey represents the key of the file written to by the writer. One can convert it
 	// to a filename via the fileKeyToName function.
 	fileKey uint16
```
emilbon99, line 122: This means we can have a maximum of 65,000 files. Is that enough? Maybe. I don't know.

cesium/internal/domain/writer.go
emilbon99: Ah I see. We used to use the `Len()` method on `xio.TrackedWriteCloser`. Can we just get rid of this interface now? I think it's the only place we use it in the codebase. Can just use an `io.WriteCloser`.

cesium/internal/domain/writer.go
```diff
@@ -259,9 +301,9 @@ func (w *Writer) Close(ctx context.Context) error {
 	return w.internal.Close()
 }
 
-func (w *Writer) validateCommitRange(end telem.TimeStamp) error {
-	if !w.prevCommit.IsZero() && end.Before(w.prevCommit) {
-		return errors.Wrap(validate.Error, "commit timestamp must be strictly greater than the previous commit")
+func (w *Writer) validateCommitRange(end telem.TimeStamp, switchingFile bool) error {
+	if !w.prevCommit.IsZero() && !switchingFile && end.Before(w.prevCommit) {
```
emilbon99, line 305: Can you explain why we don't validate that the end was after the previous commit when switching files?

cesium/internal/unary/writer.go
```diff
@@ -233,10 +239,11 @@ func (w *Writer) commitWithEnd(ctx context.Context, end telem.TimeStamp) (telem.
 	if !ok {
 		return 0, controller.Unauthorized(w.control.Subject.String(), w.Channel.Key)
 	}
+
 	if end.IsZero() {
 		// We're using w.len - 1 here because we want the timestamp of the last
 		// written frame.
-		approx, err := w.idx.Stamp(ctx, w.Start, w.len(dw.Writer)-1, true)
+		approx, err := w.idx.Stamp(ctx, w.Start, w.len(dw.Writer)-1, false)
```
emilbon99, line 246: Why did we change this to false?

cesium/internal/unary/writer.go
```diff
@@ -247,7 +254,9 @@ func (w *Writer) commitWithEnd(ctx context.Context, end telem.TimeStamp) (telem.
 		end = approx.Lower + 1
 	}
 
-	return end, dw.Commit(ctx, end)
+	err := dw.Commit(ctx, end)
```
emilbon99, line 257: What is the point of this modification and space?

emilbon99 (changes requested):

cesium/internal/domain/writer.go
LeonLiur: If there is a preset end, the previous commit will take that as prevCommit, but we must terminate the commit prematurely in order to open up a new one. therefore.

cesium/internal/unary/writer.go
LeonLiur: now we must stamp across domains if the underlying "domain" was artificially split in two by file cutoff

cesium/internal/domain/file_controller.go
LeonLiur: thought it wasn't good to have `span.Error(nil)`. turns out that's fine.

cesium/internal/domain/writer.go
LeonLiur: Note for future: 65000 Files of 1gb each ~= 65TB / channel. this should be fine.

cesium/internal/domain/writer.go
LeonLiur: We still need the `Offset()` method – while we could bake that into the `domain/Writer` definition, I still like the writer having its own interface better, so I might instead just remove `Len()` and `Reset()` but keep offset. Which one would be the better approach?

cesium/internal/domain/writer.go
LeonLiur: Changed to `&& !(switchingFile && w.presetEnd)`

cesium/internal/domain/writer.go
LeonLiur: NVM – we still need both of those things: one to track how many bytes the domain writer has written, and the other to track how many bytes the individual trackedWriter has written – that needs to go into the pointer

cesium/internal/index/domain.go
```diff
@@ -155,18 +182,58 @@ func (i *Domain) Stamp(
...
+
+	lowerTs, err := readStamp(r, iter.Len()+(endOffset-(startApprox.Upper-startApprox.Lower))*8, make([]byte, 8))
+	return Between(lowerTs, upperTs), err
+}
+
+// resolveEffectiveDomain returns the TimeRange and length of the underlying domain(s).
+// The effective domain can be many continuous domains as long as they're immediately
+// continuous, i.e. the end of one domain is the start of the other.
+func resolveEffectiveDomain(i *domain.Iterator) (telem.TimeRange, int64) {
+	var (
```
emilbon99, line 218: This is vary minor, but you could turn both `effectiveDomainBounds` and `effectiveDomainLen` into named return variables, then instead of calling `breaker` you could do an early return and clean your cod eup a bit.

cesium/internal/index/domain.go
```diff
@@ -133,11 +144,27 @@ func (i *Domain) Stamp(
...
+		}
+	}
+
 	gap := iter.Len() / 8
 	if endOffset >= iter.Len()/8 {
 		for {
 			if !iter.Next() {
+				// exhausted
 				if continuous {
 					err = ErrDiscontinuous
```
emilbon99, line 169: Just want to make sure there's no way we could trigger this error with effectively continuous domains.

emilbon99 (commented):
This looks good to me once you resolve noted issues.

PR #540, opened 2024-05-11
539 sy 601 cesium channel rekey

LeonLiur:
- Channel rekey
- Fixed behaviour of memFS with O_CREATE flags: no longer creates a new file when a file already exists
- File-based Int32 counter now attempts to read the given file and store as value on creation – if found a non-EOF value, it will store it as the counter

After rekey, we will lose writers to all previous files ie they will not be filled up, if they are not full – and the last written file will be not full. Considering the frequency of rekey, this is fine for now. Eventually this will be fixed by rewriting file controller, possibly during implementation of timerange deletion & tombstoning.


cesium/channel.go
```diff
@@ -111,3 +111,67 @@
 	}
 	return v.Error()
 }
+
+// RenameChannel changes the key of channel oldKey into newKey. This operation is
+// idempotent and does not return an error if the channel does not exist.
+// RenameChannel returns an error if there are open iterators/writers on the given channel.
+func (db *DB) RenameChannel(oldKey ChannelKey, newKey core.ChannelKey) error {
```
emilbon99, line 118: Should we call this `RekeyChannel`?

cesium/channel.go
```diff
@@ -111,3 +111,67 @@ func (db *DB) validateNewChannel(ch Channel) error {
...
+	db.mu.Lock()
+	defer db.mu.Unlock()
+	udb, uok := db.unaryDBs[oldKey]
+	if uok {
+		if err := udb.TryRekey(); err != nil {
```
emilbon99, line 123: This method name is a bit strange no? Because its not actually doing the rekey operation. Shouldn't we name it something like `CanRekey`, or even simpler `NoOpenReadersOrWriters`

cesium/channel.go
```diff
@@ -111,3 +111,67 @@ func (db *DB) validateNewChannel(ch Channel) error {
...
+			return err
+		}
+
+		err := db.fs.Rename(keyToDirName(oldKey), keyToDirName(newKey))
+		if err != nil {
+			return err
+		}
+
+		udb.Config.Channel.Key = newKey
```
emilbon99, line 132: Not a tremendous fan of editing the internal properties of a class, but this seems ok.

cesium/channel.go
```diff
@@ -111,3 +111,67 @@ func (db *DB) validateNewChannel(ch Channel) error {
...
+		newFS, err := db.fs.Sub(keyToDirName(newKey))
+		if err != nil {
+			return err
+		}
+
+		err = udb.ReconfigureFS(newFS)
```
emilbon99, line 138: Hmm, can we move more of the responsibility of re-keying to the unary DB itself? Feels like we're putting a lot of logic inside of cesium that should be inside of `unary`

cesium/channel.go
```diff
@@ -111,3 +111,67 @@ func (db *DB) validateNewChannel(ch Channel) error {
...
+		}
+
+		vdb.Config.Channel.Key = newKey
+		delete(db.virtualDBs, oldKey)
+		db.virtualDBs[newKey] = vdb
+
+		return nil
+	}
+
+	return nil
```
emilbon99, line 176: Big note. Won't the meta file still have the same key in both unary and virtual? This just renames the folder.

cesium/channel_test.go
```diff
@@ -74,6 +75,162 @@ var _ = Describe("Channel", func() {
 					),
 				)
 			})
+
+			Describe("Rekey", func() {
```
emilbon99, line 79: These cases look good, but I think they miss the edge case mentioned above, where we don't actually change the key of the channel in the meta file.

cesium/channel.go
emilbon99, line 123: Why don't we just use `TryClose` here, and then instead of having this complicated `ReconfigureFS` function, we just re-open the DB on the new directory. Seems simpler. Less methods, fewer lines of code. Still need to resolve the meta file issue I mention later though.

cesium/internal/domain/file_controller.go
```diff
@@ -279,6 +283,8 @@ func (fc *fileController) removeReadersWriters(ctx context.Context, key uint16)
 
 func (fc *fileController) close() error {
 	fc.writers.Lock()
+	fc.readers.Lock()
+	defer fc.readers.Unlock()
```
emilbon99, line 287: Could probably stick these two into a single defer call. Minor.

x/go/io/counter.go
```diff
@@ -24,12 +25,21 @@ type Int32Counter struct {
...
-func NewInt32Counter(f ReaderAtWriterAtCloser) *Int32Counter {
+func NewInt32Counter(f ReaderAtWriterAtCloser) (*Int32Counter, error) {
 	i := &Int32Counter{
 		f:   f,
 		buf: make([]byte, 4),
 	}
-	return i
+	val, err := i.load()
+	if err != nil {
+		if err == io.EOF {
```
emilbon99, line 35: Dont use `err ==`, use `errors.Is`

emilbon99 (changes requested):

emilbon99:

PR #542, opened 2024-05-14
[client/ts & client/py] - implemented auto commit on write methods

emilbon99:


client/py/synnax/framer/client.py
```diff
@@ -147,11 +147,14 @@ def write(
         :returns: None.
         """
         with self.open_writer(
-            start=start, channels=to, strict=strict, mode=WriterMode.PERSIST_ONLY
+            start=start, 
+            channels=to, 
+            strict=strict, 
+            mode=WriterMode.PERSIST_ONLY,
+            enable_auto_commit=True,
+            auto_index_persist_interval=TimeSpan.MAX,
```
LeonLiur, line 155: this makes a lot more sense as a "never auto persist", maybe we should remove the -1 option. thoughts?

LeonLiur: why never persist and not the default 1 second persist?

LeonLiur (approved):

PR #544, opened 2024-05-14
543 sy 617 cesium panics when unable to parse channel key

LeonLiur:


cesium/channel_test.go
```diff
@@ -74,6 +75,31 @@ var _ = Describe("Channel", func() {
...
+					Expect(f.Close()).To(Succeed())
+
+					db, err := cesium.Open("", cesium.WithFS(s))
+					Expect(err).ToNot(HaveOccurred())
+					Expect(db.Close()).To(Succeed())
+				})
+
+				It("Should turn parsable numeric folders into channels", func() {
+					s := MustSucceed(fs.Sub("sub"))
+					_, err := s.Sub("1")
```
emilbon99, line 91: This is kind of concerning. This DB has no meta or index file, so we don't know the data type of the channel (along with other critical info). What will happen if we try to open a reader on this directory? I feel like this case should fail

emilbon99 (changes requested):

cesium/channel_test.go
LeonLiur: The DB will create the meta file during `openUnaryOrVirtual` and will create `index.domain` and `counter.domain` files while opening the unary and domain DBs. So I think this is fine. If those files are corrupted, then I feel like DB creation should still succeed and internal errors should reflect those errors

LeonLiur: NVM. On a second thought it doesn't make sense to create a meta file if we don't know the meta properties of that channel. I will make a check in `ReadOrCreate` meta that makes it fail if channel's properties are not set.

emilbon99: This has been resolved?

cesium/internal/meta/meta.go
```diff
@@ -29,16 +30,19 @@ func ReadOrCreate(fs xfs.FS, ch core.Channel, ecd binary.EncoderDecoder) (core.C
...
-				"[meta] - a channel is required when creating a new database",
-			)
-		}
-		return ch, Create(fs, ecd, ch)
+	if exists {
+		return Read(fs, ecd)
 	}
-	return Read(fs, ecd)
+
+	if ch.DataType == telem.UnknownT {
```
emilbon99, line 37: We should validate other properties of the channel here no? Like checking that the key is not zero. Make sure we also check other properties that need to be well defined that I did not mention. Also, I feel like we should move this check into the `Create` function no?

cesium/open.go
```diff
@@ -42,12 +42,14 @@ func Open(dirname string, opts ...Option) (*DB, error) {
...
 	for _, i := range info {
-		key := core.ChannelKey(lo.Must(strconv.Atoi(i.Name())))
-		if err != nil {
-			return nil, err
-		}
 		if i.IsDir() {
-			if err = _db.openVirtualOrUnary(Channel{Key: key}); err != nil {
+			key, err := strconv.Atoi(i.Name())
+			if err != nil {
+				_db.options.L.Error("failed parsing existing folder to channel key", zap.Error(err))
```
emilbon99, line 48: I don't think there's any point to call this variable `_db` anymore. Let's just rename to `db`

cesium/open.go
```diff
@@ -42,12 +42,14 @@ func Open(dirname string, opts ...Option) (*DB, error) {
...
+				continue
+			}
+
+			if err = _db.openVirtualOrUnary(Channel{Key: ChannelKey(key)}); err != nil {
 				return nil, err
 			}
 		}
```
emilbon99, line 55: Maybe we should warn that there is an unknown file in the directory?

emilbon99 (changes requested):

cesium/channel_test.go
```diff
@@ -118,6 +119,108 @@ var _ = Describe("Channel", func() {
...
+					Expect(err).ToNot(HaveOccurred())
+					_, err = f.Write([]byte("heheheha"))
+					Expect(err).ToNot(HaveOccurred())
+					Expect(f.Close()).To(Succeed())
+
+					db, err = cesium.Open("", cesium.WithFS(s))
+					Expect(err).To(MatchError(ContainSubstring("error decoding meta file")))
+				})
+
+				It("Should error if the meta.json file got changed with impossible configurations", func() {
```
emilbon99, line 141: This should be a table spec no?

cesium/channel_test.go
```diff
@@ -118,6 +119,108 @@ var _ = Describe("Channel", func() {
...
+					Expect(ch.Key).To(Equal(key))
+					Expect(ch.Index).To(Equal(indexKey))
+					Expect(ch.DataType).To(Equal(telem.Int64T))
+
+					ch = MustSucceed(db.RetrieveChannel(ctx, indexKey))
+					Expect(ch).ToNot(BeNil())
+					Expect(ch.Key).To(Equal(indexKey))
+					Expect(ch.IsIndex).To(BeTrue())
+					Expect(ch.DataType).To(Equal(telem.TimeStampT))
+
```
emilbon99, line 211: I would add some `By` clauses here just so its easier to identify where things go wrong.

cesium/internal/domain/index.go
```diff
@@ -237,7 +237,7 @@ func (idx *index) read(f func()) {
 func (idx *index) persist(ctx context.Context, persistAtIndex int) error {
 	ctx, span := idx.T.Bench(ctx, "domain/index.persist")
 	defer span.End()
-	encoded := idx.indexPersist.encode(idx.persistHead, idx.mu.pointers)
```
emilbon99, line 240: Where do we update the persist head?

cesium/channel_test.go
emilbon99: Also, wouldn't it be possible to move these specs into the `meta` package instead?

emilbon99 (commented):
Good to merge once you address my comments

PR #545, opened 2024-05-16
Sy 573 fix channel retrieval

pjdotson:


synnax/pkg/distribution/channel/verification/service.go
```diff
@@ -0,0 +1,215 @@
...
+	ticker := time.NewTicker(24 * time.Hour)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			if expireTime.Before(time.Now()) {
+				s.Ins.L.Error("License key expired. Access has been limited", zap.String("expiredAt", expireTime.String()))
```
emilbon99, line 178: All of these logs should also be base64 encoded

synnax/pkg/distribution/channel/verification/service.go
```diff
@@ -0,0 +1,215 @@
...
+			} else if timeLeft := time.Until(expireTime); timeLeft <= s.ExpireWarning {
+				s.Ins.L.Warn("License key will expire soon. Access will be limited.", zap.String("expiresIn", timeLeft.String()))
+			} else {
+				s.Ins.L.Info("License key is not expired.", zap.String("expiresIn", timeLeft.String()))
+			}
+		}
+	}
+}
+
+func errExpireKey(channelsAllowed int) error {
```
emilbon99, line 188: Functions like `errExpireKey` and `channelsAllowed` are very easy to search for.

synnax/pkg/distribution/channel/verification/service.go
```diff
@@ -0,0 +1,215 @@
...
+	encoded1 := "dXNpbmcgYW4gZXhwaXJlZCBwcm9kdWN0IGxpY2Vuc2Uga2V5LCB1c2UgaXMgbGltaXRlZCB0byB0aGUgZmlyc3Qg"
```
emilbon99, line 189: These encoded keys don't need to be defined within the function. they can be constants. In fact, the entire error can be extracted outside of this function.

synnax/pkg/distribution/channel/verification/service.go
emilbon99: My recommendation would be to get rid of these sorts of functions entirely

synnax/pkg/distribution/channel/verification/service.go
```diff
@@ -0,0 +1,215 @@
...
+	"go.uber.org/zap"
+	"golang.org/x/net/context"
+)
+
+type Config struct {
+	// DB is the database used to store information.
+	DB kv.DB
+	// Ins is the instrumentation used for logging
+	Ins alamos.Instrumentation
+	// ExpireWarning is the warning time until the key expires
```
emilbon99, line 34: Easily searchable

synnax/pkg/distribution/channel/verification/service.go
```diff
@@ -0,0 +1,215 @@
...
+	DefaultConfig                       = Config{
+		ExpireWarning: 7 * 24 * time.Hour,
+	}
+)
+
+// Validate implements config.Properties.
+func (c Config) Validate() error {
+	v := validate.New("key")
+	validate.NotNil(v, "DB", c.DB)
+	validate.NonZero(v, "ExpireWarning", c.ExpireWarning)
```
emilbon99, line 49: Base64 encode and decode the name of this field

synnax/pkg/distribution/channel/verification/service.go
```diff
@@ -0,0 +1,215 @@
...
+// key stored in DB.
+func OpenService(key string, cfgs ...Config) (*Service, error) {
+	cfg, err := config.New(DefaultConfig, cfgs...)
+	if err != nil {
+		return nil, err
+	}
+	service := &Service{Config: cfg}
+
+	// This stuff needs to be updated when I decide what to do with the
+	// background key validation process
```
emilbon99, line 77: easily searchable

synnax/pkg/distribution/channel/verification/service.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// Close will shutdown the service
+func (s *Service) Close() error {
+	return s.shutdown.Close()
+}
+
+// ValidateChannelCount returns an error if inUse is greater than the number of
+// channels the user is allowed to use.
+func (s *Service) ValidateChannelCount(ctx context.Context, inUse int64) error {
```
emilbon99, line 103: Documenting this function is also dangerous. Remember that a user could just edit this function to `return nil` and they would have an infinite no. of channels for free

synnax/pkg/distribution/channel/verification/service.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// create sets  key in s.DB. Returns an error if the key is invalid - you can
+// only set a valid key in the database
+func (s *Service) create(ctx context.Context, key string) error {
+	err := validateKey(key)
+	if err != nil {
+		return err
+	}
+	return s.DB.Set(ctx, []byte("productKey"), []byte(key))
```
emilbon99, line 151: All calls of `productKey` should be base64 encoded. Not to mention, they can be extracted outside of this function because they are constants.

synnax/pkg/distribution/channel/verification/service.go
```diff
@@ -0,0 +1,215 @@
...
+// updateKeyValidation will run a background process that continuously checks if
+// the key is expired.
+func (s *Service) updateKeyValidation(ctx context.Context) error {
+
+	key, err := s.retrieve(ctx)
+	if err != nil {
+		return err
+	}
+	expireTime := getExpirationDate(key)
+	ticker := time.NewTicker(24 * time.Hour)
```
emilbon99, line 169: The check interval should be configurable

synnax/cmd/start.go
```diff
@@ -79,10 +79,11 @@ var (
...
 	var (
-		ins      = configureInstrumentation(v)
-		insecure = viper.GetBool("insecure")
-		verbose  = viper.GetBool("verbose")
-		autoCert = viper.GetBool("auto-cert")
+		ins        = configureInstrumentation(v)
+		insecure   = viper.GetBool("insecure")
+		verbose    = viper.GetBool("verbose")
+		autoCert   = viper.GetBool("auto-cert")
+		productKey = viper.GetString("product-key")
```
emilbon99, line 86: Should be base64 encoded, and variable should not be named `productKey`

synnax/cmd/start.go
```diff
@@ -290,6 +295,7 @@ func buildDistributionConfig(
 	ins alamos.Instrumentation,
 	storage storage.Config,
 	transports *[]fgrpc.BindableTransport,
+	productKey string,
```
emilbon99, line 298: Shouldn't be named productKey

synnax/cmd/start_flags.go
```diff
@@ -63,4 +63,10 @@ func configureStartFlags() {
 		false,
 		"Automatically generate self-signed certificates.",
 	)
+
+	startCmd.Flags().String(
+		"product-key",
```
emilbon99, line 68: Base64 encode all of these arguments

synnax/pkg/api/channel.go
```diff
@@ -37,6 +38,7 @@ type Channel struct {
 	IsIndex     bool                 `json:"is_index" msgpack:"is_index"`
 	Index       channel.Key          `json:"index" msgpack:"index"`
 	Alias       string               `json:"alias" msgpack:"alias"`
+	Internal    bool                 `json:"internal" msgpack:"alias"`
```
emilbon99, line 41: The struct tags here are incorrect. Should be `json:"internal" msgpack:"internal"`

synnax/pkg/distribution/channel/channel_group.go
emilbon99, line 1: This file is already inside of the package channel, why does it need to be called `channel_group` instead of just `group`?

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+			r.start = localKeys[i]
+			r.size = 1
+		}
+	}
+	c.remove(r)
+}
+
+// GetChannelsBeforeKey returns the number of channels in c before key. If key
+// is in c, that does not count towards the number of channels returned.
+func (c *ChannelGroup) GetChannelsBeforeKey(key Key) uint16 {
```
emilbon99, line 131: This would be better named `GetNumChannelsBeforeKey` or `GetNumBeforeKey`. Remember when you're writing go code, you should ready all package internal methods in the context of their package name. If the method is called `GetNumBeforeKey` and the package name is `channel`, you should assume that it is the number of channels before a particular key.

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+			counter += localKey - (*c)[i].start
+		} else {
+			counter += (*c)[i].size
+		}
+	}
+	return counter
+}
+
+// IsKeyIn return true if the key is in c and false if key is not in c
+func (c ChannelGroup) IsKeyIn(key Key) bool {
```
emilbon99, line 145: Just call this method `Contains` no?

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// Less returns true if the key of (*c)[i] is before the key of (*c)[j].
+// Implements sort.interface.
+func (c *ChannelGroup) Less(i, j int) bool {
+	return (*c)[i].start < (*c)[j].start
+}
+
+// InsertKey allows you to insert one single key into the channelGroup c.
+func (c *ChannelGroup) InsertKey(key Key) {
```
emilbon99, line 58: Call this `Insert`. If the first argument is already a channel key, it's simpler to just call the method `Insert`

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+package channel
+
+import (
+	"slices"
+	"sort"
+)
+
+// channelRange is the basic struct for ChannelGroup, and is defined as a set of
+// consecutive channel keys.
+type channelRange struct {
```
emilbon99, line 24: There's no point in starting the name of a struct with the package name unless the struct has the name of the package itself. This should just be called `Range`, or, even better, `KeyRange`

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+	size uint16
+}
+
+// end returns the key that is at the end of channelRange r
+func (r channelRange) end() uint16 {
+	return r.start + r.size - 1
+}
+
+// ChannelGroup extends []channelRange with a few convenience methods.
+type ChannelGroup []channelRange
```
emilbon99, line 37: This should be called `Group` and not `ChannelGroup`. Also, I feel like group is an ambiguous name. Isn't this more of a set?

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+
+// Len returns the length of c. Implements sort.interface.
```
emilbon99, line 39: What is `sort.interface`?

synnax/pkg/distribution/channel/channel_group.go
emilbon99: Also does this struct need to be public to the package? Shouldn't it be internal to the channel package.

emilbon99: If you want to be able to test this, move it to an internal package inside the `channel` package

synnax/pkg/distribution/channel/lease_proxy.go
```diff
@@ -1,301 +1,352 @@
...
+)
+
+type leaseProxy struct {
+	ServiceConfig
+	createRouter     proxy.BatchFactory[Channel]
+	deleteRouter     proxy.BatchFactory[Key]
+	leasedCounter    *kv.AtomicInt64Counter
+	freeCounter      *kv.AtomicInt64Counter
+	externalCounter  *kv.AtomicInt64Counter
+	deletedChannels  *ChannelGroup
```
emilbon99, line 32: These variables should just be called `deleted` and `internal`. Again, package contextual

synnax/pkg/distribution/channel/service.go
```diff
@@ -1,123 +1,154 @@
...
+
+func (s *service) NewWriter(tx gorp.Tx) Writer {
+	return writer{proxy: s.proxy, tx: s.DB.OverrideTx(tx)}
+}
+
+func (s *service) NewRetrieve() Retrieve {
+	return newRetrieve(s.DB, s.otg, func(channels []Channel) ([]Channel, error) {
+		maxAllowed, err := s.proxy.ChannelsNeedValidation()
+
+		if err != nil {
```
emilbon99, line 134: This is a great example where an early return could solve deep nesting. Instead of doing

```
if err != nil {
/// A bunch of logic
}
```

Do
```
if err == nil {
   return
}
/// a bunch of logic
```

synnax/pkg/distribution/channel/service.go
```diff
@@ -1,123 +1,154 @@
...
+			var fErr error
+			keys := KeysFromChannels(channels)
+
+			var returnedChannels []Channel
+			var err error
```
emilbon99, line 139: Why does this `err` variable need to be declared? Also, you can combine multiple declarations into a single block

```
var (
   varOne = 1  
   varTwo = 2
)
```

synnax/pkg/distribution/channel/service.go
```diff
@@ -1,123 +1,154 @@
...
+	}
+	return s, nil
+}
+
+func (s *service) NewWriter(tx gorp.Tx) Writer {
+	return writer{proxy: s.proxy, tx: s.DB.OverrideTx(tx)}
+}
+
+func (s *service) NewRetrieve() Retrieve {
+	return newRetrieve(s.DB, s.otg, func(channels []Channel) ([]Channel, error) {
```
emilbon99, line 131: I feel like this function is a little bit heavy to pass in anonymously at the service level. Can we move this logic into a named function?

emilbon99 (changes requested):

synnax/pkg/distribution/channel/channel_group.go
LeonLiur: I would change the comment to be something without 'you', more like `InsertKey inserts one single key into the group c`.

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+func (c *ChannelGroup) Less(i, j int) bool {
+	return (*c)[i].start < (*c)[j].start
+}
+
+// InsertKey allows you to insert one single key into the channelGroup c.
+func (c *ChannelGroup) InsertKey(key Key) {
+	c.InsertKeys(Keys{key})
+}
+
+// InsertKeys allows you to insert several keys into the channelGroup c.
```
LeonLiur, line 62: see above

LeonLiur: do you really need two methods where one is for inserting one key and the other is for inserting many keys? IMO could you make the method signature `Insert(k ...Key)`

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+	c.insert(r)
+}
+
+// RemoveKey removes key from c.
+func (c *ChannelGroup) RemoveKey(key Key) {
+	c.RemoveKeys(Keys{key})
+}
+
+// RemoveKeys removes all keys in keys from c.
+func (c *ChannelGroup) RemoveKeys(keys Keys) {
```
LeonLiur, line 95: Same as above, maybe use unpacking?

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+
+	localKeys := make([]uint16, len(keys))
+	i := 0
+	for _, key := range keys {
+		if c.IsKeyIn(key) {
+			localKeys[i] = key.LocalKey()
+			i++
+		}
+	}
+	localKeys = localKeys[0:i]
```
LeonLiur, line 108: If every time you add an element to `localKeys`, you increment to `i`, what's the point of doing `[0:i]`? isn't i always going to be equal to the length of `localKeys`?

synnax/pkg/distribution/channel/channel_group.go
```diff
@@ -0,0 +1,273 @@
...
+			r.size -= overlap
+			c.insert(r)
+			return
+		}
+	}
+
+	// At this point, r does not overlap at all with any of the
+	// channelRanges in c.
+
+	// This checks if we need to compress c
```
LeonLiur, line 208: Unify comment syntax: either they are all capitalized and have punctuation, or they are not. I think the standard is that if the comment is more than a few words, it should be a full sentence with capitalization and punctuation. @emilbon99

synnax/pkg/distribution/channel/retrieve.go
```diff
@@ -1,137 +1,149 @@
...
+
+// Offset offsets the results returned by the query. This is an identical interface to
+// gorp.Retrieve.
+func (r Retrieve) Offset(offset int) Retrieve {
+	r.gorp.Offset(offset)
+	return r
+}
+
+// Exec executes the query, binding
+func (r Retrieve) Exec(ctx context.Context, tx gorp.Tx) (err error) {
```
LeonLiur, line 97: Why changed to named return? IMO if a function is named return then it's return should not have any arguments, where possible

synnax/pkg/distribution/channel/service.go
LeonLiur: you can also combine multiple delcarations of the same type into a same line
```var (
fErr, err error
)```

File: synnax/pkg/distribution/channel/verification/verification.go [hunk 29]
```diff
@@ -0,0 +1,176 @@
...
+	"errors"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/synnaxlabs/x/crypto"
+	"github.com/synnaxlabs/x/date"
+)
+
+const maxFreeChannels = 50
```
LeonLiur, line 23: encode

synnax/pkg/distribution/channel/verification/verification.go
LeonLiur, line 1: This whole file is very searchable no? I can just search 'channel key' in the repo and find this. Is this file hidden in some way?

synnax/pkg/distribution/distribution.go
```diff
@@ -1,155 +1,169 @@
...
+
+	d.Channel, err = channel.New(ctx, channel.ServiceConfig{
+		HostResolver:           d.Cluster,
+		ClusterDB:              gorpDB,
+		TSChannel:              d.Storage.TS,
+		Transport:              channelTransport,
+		Ontology:               d.Ontology,
+		Group:                  d.Group,
+		ValidateChannelCount:   func(count int64) error { return ca.ValidateChannelCount(ctx, count) },
+		ChannelsNeedValidation: func() (int, error) { return ca.GetMaxFreeChannels(), ca.IsExpired(ctx) },
```
LeonLiur, line 123: Same with these, these variabls should not be searchable

x/go/math/math.go
```diff
@@ -0,0 +1,30 @@
...
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+// math is a package that extends math functionality.
+
+package math
+
+// IntPow returns the integer n to the m power. IntPow requires m to be a
```
LeonLiur, line 14: Not sure how applicable this is, but there's a nice way to do power in log(m) time https://en.wikipedia.org/wiki/Exponentiation_by_squaring – might be useful if m can be large

LeonLiur (changes requested):

emilbon99:
A note on this PR that just came to my mind. Now that we've added the `Internal` field we need to make sure we update all internal channels to actually set them as internal, and make sure users can't create internal channels.

PR #552, opened 2024-05-17
551 sy 627 fix memfs implementation of open

LeonLiur:
Fixed 2 things:
1. Calling open with `os.O_WRONLY` or `os.O_RDWR` **without** `os.O_CREATE` no longer creates a new file – this was a bug.
2. Opening a file with `os.O_APPEND` in memFS now properly opens the file in append mode, so writes would go to the end of the file.


emilbon99 (approved):
LGTM

PR #557, opened 2024-05-21
514 synnax tombstone deletion and garbage collection

LeonLiur:
Timerange deletion changed behaviours:
- File controller will prioritize opening a handle on "old" files that have space before writing to a "new" file


cesium/internal/domain/delete.go
```diff
@@ -130,107 +62,224 @@ func (db *DB) Delete(ctx context.Context, startPosition int, endPosition int, st
...
+		db.idx.mu.Unlock()
+		return err
+	}
+	db.idx.mu.Unlock()
+
+	if err = db.files.rejuvenate(key); err != nil {
+		return err
+	}
+
+	return db.FS.Remove(name + "_temp")
```
emilbon99, line 228: Let's say the garbage collection fails halfway through. We probably want to try to recover the DB to a certain state no? This probably means renaming files again and removing certain things. Stuff to think about

cesium/internal/unary/delete.go
```diff
@@ -23,45 +26,55 @@ import (
...
-	g, _, err := db.Controller.OpenAbsoluteGateIfUncontrolled(tr, control.Subject{Key: "Delete Writer"}, func() (controlledWriter, error) {
-		return controlledWriter{
-			Writer:     nil,
-			channelKey: db.Channel.Key,
-		}, nil
-	})
-
+	g, _, err := db.Controller.OpenAbsoluteGateIfUncontrolled(
+		tr,
+		control.Subject{Key: "delete_writer_" + uuid.New().String()},
```
emilbon99, line 60: My take would be to leave the key as a UUID and then set the name to `delete_writer`.

cesium/internal/unary/delete.go
```diff
@@ -23,45 +26,55 @@ import (
...
+		func() (controlledWriter, error) {
+			return controlledWriter{Writer: nil, channelKey: db.Channel.Key}, nil
+		})
 	if err != nil {
 		return err
 	}
 
 	_, ok := g.Authorize()
```
emilbon99, line 68: If we're ignoring ok, why don't we just set it to an underscore?

cesium/internal/unary/delete.go
```diff
@@ -70,54 +83,43 @@ func (db *DB) Delete(ctx context.Context, tr telem.TimeRange) error {
...
+	err = db.Domain.Delete(
+		ctx,
+		int64(density.Size(startOffset)),
+		int64(density.Size(endOffset)),
+		tr,
+	)
 
-	g.Release()
-	return i.Close()
+	return errors.CombineErrors(err, i.Close())
```
emilbon99, line 120: I'm ok with this, but honestly I would still defer closing the iterator

cesium/internal/unary/delete.go
```diff
@@ -70,54 +83,43 @@ func (db *DB) Delete(ctx context.Context, tr telem.TimeRange) error {
...
-func (db *DB) GarbageCollect(ctx context.Context, maxSizeRead uint32) error {
-	db.mu.RLock()
-	defer db.mu.RUnlock()
-	if db.mu.openIteratorWriters > 0 {
-		return nil
-	}
-	err := db.Domain.CollectTombstones(ctx, maxSizeRead)
-	return err
+func (db *DB) GarbageCollect(ctx context.Context) error {
+	return db.wrapError(db.Domain.GarbageCollect(ctx))
```
emilbon99, line 124: Nice! No garbage collection logic here

emilbon99 (changes requested):
This PR looks so good! I only reviewed the core deletion related files (Delete methods and GC). I left a few minor and stylistic comments. Well done!

cesium/internal/unary/race_test.go
```diff
@@ -0,0 +1,191 @@
...
+						wg          = sync.WaitGroup{}
+						receivedErr error
+					)
+					wg.Add(10)
+
+					for i := 0; i < 10; i++ {
+						go func() {
+							defer GinkgoRecover()
+							defer wg.Done()
+							err := rateDB.Delete(ctx, (11 * telem.SecondTS).Range(13*telem.SecondTS))
```
LeonLiur, line 174: df

cesium/internal/unary/delete.go
```diff
@@ -11,122 +11,85 @@ package unary
...
+}
 
+func (db *DB) delete(ctx context.Context, tr telem.TimeRange) error {
+	if !tr.Valid() {
+		return errors.Newf("delete start %d cannot be after delete end %d", tr.Start, tr.End)
+	}
+
+	g, _, err := db.Controller.OpenAbsoluteGateIfUncontrolled(
```
emilbon99, line 37: Do we still need to open this?

cesium/internal/unary/delete.go
```diff
@@ -11,122 +11,85 @@ package unary
...
+	if !approxDist.Exact() {
+		approxStamp, err := db.index().Stamp(ctx, domainStart, stampEnd, true)
 		if err != nil {
-			return errors2.CombineErrors(i.Close(), err)
+			return offset, err
+		}
+		if !approxStamp.Exact() {
+			panic("cannot find exact timestamp")
+		}
+		*ts = approxStamp.Upper
```
emilbon99, line 87: This is weird. Don't pass as a pointer. Just take it as an argument and return by value.

cesium/internal/unary/delete.go
```diff
@@ -11,122 +11,85 @@ package unary
...
-			Start: i.TimeRange().Start,
-			End:   tr.End,
-		}, false)
+	if !approxDist.Exact() {
+		approxStamp, err := db.index().Stamp(ctx, domainStart, stampEnd, true)
 		if err != nil {
-			return errors2.CombineErrors(i.Close(), err)
+			return offset, err
+		}
+		if !approxStamp.Exact() {
```
emilbon99, line 84: It seems like all of this logic is just useful when `isStart` is true. Why don't we pass two different callbacks to the domain `delete` method? One for resolving the start and one for the end. It would be more expressive.

cesium/internal/domain/delete.go
```diff
@@ -11,235 +11,355 @@ package domain
...
+	}
 
-			for tombstonePtr < len(tombstones) && currentPointer.offset > tombstones[tombstonePtr].offset {
-				priorTombstoneLength += db.idx.mu.tombstones[fileKey][tombstonePtr].length
-				tombstonePtr++
-			}
+	db.idx.mu.RLock()
+	defer db.idx.mu.RUnlock()
+	persist := db.idx.indexPersist.prepare(0)
+	// We choose to keep the mutex locked while persisting pointers: the time sacrifice
```
emilbon99, line 197: Be careful, an RLock does not guarantee exclusivity. You can still race two persists here.

cesium/internal/domain/delete.go
```diff
@@ -11,235 +11,355 @@ package domain
...
+		if db.files.hasWriter(fileKey) {
+			continue
 		}
-		f, err := db.FS.Open(strconv.Itoa(int(fileKey))+"_temp.domain", os.O_CREATE|os.O_RDWR)
+		s, err := db.FS.Stat(fileKeyToName(fileKey))
 		if err != nil {
-			return err
+			return span.Error(err)
+		}
+		if s.Size() < int64(db.FileSize) {
```
emilbon99, line 185: This is a bad check. What if the file is 4 bytes underneath the max. Will we still open a new writer on it? I guess so, but we probably shouldn't because we'll end up with a tiny domain. Probably a good idea to add a better check than this.

emilbon99 (changes requested):

PR #564, opened 2024-05-22
SY-660 add driver to docker build in CI

Lham42:
Update CI to build the server with the embedded driver.

- updated earthfile to run with driver
- updated github workflow so the docker job runs after the build job completes and gets the driver binary from the build job

NOTE:

https://github.com/actions/upload-artifact#maintaining-file-permissions-and-case-sensitive-files

"File permissions are not maintained during artifact upload. All directories will have 755 and all files will have 644. For example, if you make a file executable using chmod and then upload that file, post-download the file is no longer guaranteed to be set as an executable.

If you must preserve permissions, you can tar all of your files together before artifact upload. Post download, the tar file will maintain file permissions and case sensitivity."

open pr on github actions team since 2019 and so this is the workaround they suggest ^


.github/workflows/synnax.build.yaml
```diff
@@ -110,24 +35,13 @@ jobs:
       - name: Create Release
         if: steps.filter.outputs.changed == 'true' && github.ref == 'refs/heads/main'
         id: create-release
-        uses: actions/github-script@v6
+        uses: softprops/action-gh-release@v1
```
Lham42, line 38: I dont recall making this change but I don't know whether it makes a difference?

PR #570, opened 2024-05-23
Eislam/sy 670 opc ua driver improvements

Lham42:
Changes to opc scanner:
- made max depth as optional config passed in with cmd.args (default to 10)

Changes to acquisition pipeline:
- changes to start to handle restarting run thread once it exits on a caught error

Changes to OPC reader:
- fixed null access being done in read function
- refactored some code so that the acquisiton pipeline is constructed upon reader task construction rather than using copy semantics

Changes to NI 
- refactored some code so that the acquisiton pipeline is constructed upon reader & writer task construction rather than using copy semantics

Changes to breaker:
- implement a condition variable for breaker that will trigger break to shutdown earlier than exceeding its max retry count
- impl a shutdown function that is now used in the classes that need the breaker (heartbeat, task maanager, ctrl, acquisition)

changes to loop:
- implemented a more precise timer using welfords algorithm for calculating variance in the call to system sleep (at a configurable interval, right now itll have resolution of 0.1ms


console/src/layout/Window.tsx
emilbon99, line 1: Note for @emilbon99 - This was the wrong merge resolution. Resolve

driver/heartbeat/heartbeat.cpp
```diff
@@ -46,6 +46,7 @@ freighter::Error heartbeat::Heartbeat::stop() {
     running.wait(false);
     running = false;
     run_thread.join();
+    breaker.close();
```
emilbon99, line 49: Why do we put this after joining the run thread? Is there a way we can do all of this without needing to manage multiple things i.e. can we get the breaker to exit when running is set to false

driver/loop/loop.h
```diff
@@ -32,20 +32,52 @@ class Timer {
...
         const auto elapsed = now - last;
         const auto interval_nanos = interval.nanoseconds();
         if (elapsed < interval_nanos) {
             auto remaining = interval_nanos - elapsed;
-            std::this_thread::sleep_for(std::chrono::nanoseconds(remaining));
+            // std::this_thread::sleep_for(std::chrono::nanoseconds(remaining));
+            this->exactSleep(std::chrono::nanoseconds(remaining));
             last = std::chrono::high_resolution_clock::now();
             return {elapsed, true};
+        } else {
```
emilbon99, line 72: Since we're returning early we don't need this else statement. Also, I think we should get rid of the warning logs as it could happen a lot in prod and obscure other, more useful logs

driver/meminfo/meminfo.h
```diff
@@ -71,13 +71,15 @@ class MemInfo final : public task::Task {
...
-        pipe = pipeline::Acquisition(
-            ctx,
-            writer_cfg,
-            std::move(source),
-            breaker_config
-        );
-        pipe.start();
+
+        // TODO: change this so that you are not using the acquisition pipeline copy constructor (follow ni pattern w/ configure)
+        // pipe = pipeline::Acquisition(
```
emilbon99, line 76: Why is this commented out. Did we just turn off the meminfo task?

driver/opc/dev/.server-extended.py.un~
emilbon99, line 1: Remove all of these files. Might be worth updating the gitignore for this directory to be the same as that for the python client

driver/opc/reader.cpp
```diff
@@ -543,7 +542,7 @@ std::unique_ptr<task::Task> Reader::configure(
         return nullptr;
     }
 
-    auto source = std::make_unique<ReaderSource>(
+    auto source = std::make_shared<ReaderSource>(
```
emilbon99, line 545: Why does this need to be a shared pointer?

driver/opc/scanner.cpp
```diff
@@ -89,15 +93,25 @@ static UA_StatusCode nodeIter(
                                     browseName.name.length);
             auto node_id = nodeIdToString(child_id);
             auto [dt, is_array] = variant_data_type(value);
+            auto [dt, is_array] = variant_data_type(value);
+            std::cout << "Node id: " << node_id << " Name: " << name << " Is array: " << is_array << " Data type: " << dt.value << std::endl;
```
emilbon99, line 97: Get rid of this

driver/opc/scanner.cpp
```diff
@@ -75,6 +76,9 @@ static UA_StatusCode nodeIter(
     // LOG(INFO) << "Node class: " << nodeClass;
     // LOG(INFO) << "Node id: " << nodeIdToString(child_id);
     // LOG(INFO) << "Node depth: " << ctx->depth;
+    // LOG(INFO) << "Node class: " << nodeClass;
```
emilbon99, line 79: Get rid of these

driver/opc/scanner.cpp
```diff
@@ -89,15 +93,25 @@ static UA_StatusCode nodeIter(
...
                 });
+                ctx->channels->push_back({
+                    dt,
+                    name,
+                    node_id,
+                    is_array
+                });
         }
     }
+    if (ctx->depth >= ctx->max_depth || child_id.namespaceIndex == 0) return
```
emilbon99, line 113: Duplicated

driver/opc/scanner.cpp
emilbon99, line 1: A bunch of duplicated stuff in this file

driver/pipeline/acquisition.cpp
```diff
@@ -53,14 +49,14 @@ void Acquisition::stop() {
 synnax::TimeStamp resolve_start(const synnax::Frame &frame) {
     for (size_t i = 0; i < frame.size(); i++) {
         if (frame.series->at(i).data_type == synnax::TIMESTAMP) {
-            std::int64_t ts = frame.series->at(i).at<int64_t>(0);
+            std::int64_t ts = frame.series->at(i).at<int64_t>(-1);
```
emilbon99, line 52: This should be 0, not negative 1

driver/pipeline/acquisition.cpp
```diff
@@ -21,30 +23,24 @@ Acquisition::Acquisition(
     WriterConfig writer_config,
     std::shared_ptr<Source> source,
```
emilbon99, line 24: Why does this need to be a shared pointer?

driver/task/manager.cpp
```diff
@@ -94,6 +94,7 @@ freighter::Error task::Manager::stop() {
     streamer->closeSend();
     run_thread.join();
     for (auto &[key, task]: tasks) task->stop();
+    breaker.close();
```
emilbon99, line 97: I feel like we're using several different mechanisms to shut things down. This is confusing, and will make it hard to reliably implement startup and shutdown logic for new tasks. We need to simplify

emilbon99 (changes requested):

driver/pipeline/acquisition.cpp
Lham42: Because in the case of NI for example, we have the DigitalWriteSink which has a member StateSource which it needs to update the state of the digital outputs, but it also needs to be accessed through the acquisition pipe to read out those states.

driver/loop/loop.h
Lham42: wdym returning early?

driver/meminfo/meminfo.h
Lham42: i did temporary cus i needed to make some changes. ive just made those changes and included it in main again.

driver/opc/dev/.server-extended.py.un~
Lham42: added gitignore

driver/heartbeat/heartbeat.cpp
Lham42: refactored to close internally on destruction

driver/opc/scanner.cpp
```diff
@@ -89,13 +86,20 @@ static UA_StatusCode nodeIter(
...
             auto [dt, is_array] = variant_data_type(value);
+            // std::cout << "Node id: " << node_id << " Name: " << name << " Is array: " << is_array << " Data type: " << dt.value << std::endl;
             if (dt != synnax::DATA_TYPE_UNKNOWN && !dt.is_variable())
                 ctx->channels->push_back({
                     dt,
                     name,
                     node_id,
                     is_array
                 });
+                ctx->channels->push_back({
```
emilbon99, line 97: Duplicated code

emilbon99 (approved):
Make sure you've addressed all duplicated code issues and then we're good to go.

PR #571, opened 2024-05-23
566 sy 634 version migration

LeonLiur:
This PR is NOT going to merge into main and is going to merge into timerange deletion instead, for the purpose that migration must be in place before tombstone deletion gets released.
This PR is also tricky to test: we have to write data in one version and assert that it is the same in another version: when we write data using the dev branch's code, we will write it in V2 – we could manually manipulate the data to make it become V1 (i.e. removing the first few bytes of a file / deleting some file), then read it into the DB to force a migration – however, there is no programatic way to guarantee this data is actually V1.

Another thing is that once merged, this will fail right off the bat because none of the existing json files have a version field: something worth thinking about: maybe we don't store the version in the `meta.json` file and instead store it in a separate file.


cesium/internal/version/version.go
```diff
@@ -0,0 +1,14 @@
...
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package version
+
+type Version = uint8
+
+const CurrentVersion = 2
```
emilbon99, line 14: Just call this `Current` and make sure its of type `Version` because I think it will be an int right now

cesium/internal/testdata/v1/db-data/2/1.domain
```diff
@@ -0,0 +1,2 @@
+
```
emilbon99, line 1: Interesting. Why does this file have a blank line at the start?

cesium/internal/testdata/v1/generate.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/synnaxlabs/x/telem"
+)
+
+var (
+	ctx      = context.Background()
+	channels = testdata.Channels
+	frames   = testdata.Frames
+)
+
+func GenerateV1() error {
```
emilbon99, line 16: Just keep in mind that this function will actually generate database files for the **current** version, not v1.

cesium/internal/testutil/fs.go
```diff
@@ -17,3 +19,61 @@ var FileSystems = map[string]FSFactory{
 		return MustSucceed(xfs.Default.Sub(dirName)), func() error { return xfs.Default.Remove(dirName) }
 	},
 }
+
+func ReplicateFS(srcFS, destFS xfs.FS) error {
```
emilbon99, line 23: Do we need this function so we can test with memfs? If not why can't we just use a regular copy command? Also, name this function `CopyFS` instead of replicate.

cesium/internal/version/migrate.go
emilbon99, line 1: Now that we have access to `Truncate`, do we actually need to store the first 4 bytes as the counter?

cesium/open.go
```diff
@@ -97,6 +105,18 @@ func (db *DB) openVirtualOrUnary(ch Channel) error {
 		if isOpen {
 			return nil
 		}
+		if ch.Version != version.CurrentVersion {
+			err = version.Migrate(fs, ch.Version, version.CurrentVersion)
```
emilbon99, line 109: My gut is that this logic should be inside of `unary` no? I don't see why it needs to be at the cesium level

emilbon99 (approved):
Looks good once you address my comments

cesium/internal/testdata/v1/db-data/2/1.domain
LeonLiur: Seems like all binary files have a blank line at the start, will investigate

cesium/internal/version/migrate.go
LeonLiur: No, in theory we don't. I'll make a decision when timerange deletion is done whether to keep it or not

cesium/open.go
LeonLiur: This is kind of weird: the reason it's here is because we must encode a new meta after the migration is done, which requires the `metaECD`: we currently don't store the `metaECD` in any lower level DBs.

cesium/internal/testutil/fs.go
LeonLiur: Yes, we need to be able to test with memfs

cesium/open.go
LeonLiur: Added a method in unary that takes `ecd` as an argument to rewrite meta

cesium/internal/testutil/fs.go
LeonLiur: Yes, we need to use it to test memfs

PR #574, opened 2024-05-24
574 sy 373 better error messages, sy 623 store channel names in meta, sy 662 better handling closing

### LINKED ISSUE #561: sy-662-allow-close-on-domain-readers

### LINKED ISSUE #563: sy-373-better-error-messages

### LINKED ISSUE #572: sy-623-store-channel-name-in-metajson

LeonLiur:
closes #563, closes #561 , closes #572
Changed:
- Renamed `cesium.ChannelNotFound` to `cesium.ErrChannelNotFound`.
- Added error in cesium: `ErrDomainClosed`.
- Removed public variable `IteratorClosedError` from cesium.
- `TryClose` is now the default behaviour of `Close`: a `unary` or `virutal` DB cannot be closed if there are writers or iterators on it.
- made variable `errDBClosed` private


cesium/channel.go
```diff
@@ -100,10 +148,10 @@ func (db *DB) validateNewChannel(ch Channel) error {
...
 		if ch.IsIndex {
-			v.Ternary("index", ch.DataType != telem.TimeStampT, "index channel must be of type timestamp")
+			v.Ternary("data type", ch.DataType != telem.TimeStampT, "index channel must be of type timestamp")
 			v.Ternaryf("index", ch.Index != 0 && ch.Index != ch.Key, "index channel cannot be indexed by another channel")
 		} else if ch.Index != 0 {
-			validate.MapContainsf(v, ch.Index, db.unaryDBs, "index %v does not exist", ch.Index)
+			validate.MapContainsf(v, ch.Index, db.unaryDBs, "index channel %v does not exist", ch.Index)
 			v.Funcf(func() bool {
 				return !db.unaryDBs[ch.Index].Channel.IsIndex
 			}, "channel %v is not an index", ch.Index)
```
emilbon99, line 157: Can we add the name of the channel to this error as well?

cesium/channel.go
```diff
@@ -54,7 +67,42 @@ func (db *DB) RetrieveChannel(_ context.Context, key ChannelKey) (Channel, error
...
+		}
+		db.unaryDBs[key] = udb
+		return nil
+	}
+	vdb, vok := db.virtualDBs[key]
+	if vok {
+		if vdb.Channel.Name == newName {
+			return nil
+		}
+		vdb.Channel.Name = newName
```
emilbon99, line 96: This is a race condition. What if somebody is reading from the channel name while it is being renamed? Same case with `unary`. This doesn't mean we should resolve. But we should make a note that we're ok with a race in this scenario because we can guarantee it wont cause bugs

cesium/db.go
```diff
@@ -28,12 +29,12 @@ type (
 	Frame      = core.Frame
 )
 
-var ChannelNotFound = core.ChannelNotFound
```
emilbon99, line 31: Did you redefine this somewhere else. If not, we need this to be public so we can access in layers above. Note that renaming this is a breaking change.

cesium/db.go
```diff
@@ -45,11 +46,15 @@ type DB struct {
 		inlet  confluence.Inlet[WriterRequest]
 		outlet confluence.Outlet[WriterResponse]
 	}
+	closed   bool
 	shutdown io.Closer
 }
 
 // Write implements DB.
 func (db *DB) Write(ctx context.Context, start telem.TimeStamp, frame Frame) error {
```
emilbon99, line 54: Can we get better documentation on these methods. These no longer 'implement DB' because DB is the concrete type itself. Same with WriteArray and Read

cesium/delete.go
```diff
@@ -151,7 +153,7 @@ func (db *DB) removeChannel(ch ChannelKey) error {
 				}
 				otherDB := db.unaryDBs[otherDBKey]
 				if otherDB.Channel.Index == udb.Config.Channel.Key {
-					return errors.New("[cesium] - could not delete index channel with other channels depending on it")
+					return errors.Newf("cannot delete channel %d because it indexes data in channel %d", ch, otherDBKey)
```
emilbon99, line 156: Channel name in message!!! Please do this for all errors

cesium/internal/controller/controller.go
```diff
@@ -203,7 +205,7 @@ func (r *region[E]) unprotectedOpen(g *Gate[E]) (t Transfer, err error) {
 	// Check if any gates have the same subject key.
 	for og := range r.gates {
 		if og.Subject.Key == g.Subject.Key {
-			err = errors.Wrapf(validate.Error, "[controller] - gate with subject key %s already exists", g.Subject.Key)
+			err = errors.Wrapf(validate.Error, "control subject %s is already registered in the region", g.Subject.Key)
```
emilbon99, line 208: Subject name in error message!

cesium/internal/controller/controller.go
```diff
@@ -220,17 +222,45 @@ func (r *region[E]) unprotectedOpen(g *Gate[E]) (t Transfer, err error) {
...
+}
+
+var (
+	_ config.Config[Config] = Config{}
+	// DefaultConfig is the default configuration for opening a Controller.
+	DefaultConfig = Config{Concurrency: control.Exclusive}
+)
+
+func (c Config) Validate() error {
+	v := validate.New("controller config")
```
emilbon99, line 238: Weird. Just return nil

emilbon99: Although I would honestly validate that the key is not empty. Warning: this might break a few things upstream that we need to check.

cesium/internal/controller/controller.go
```diff
@@ -306,20 +336,21 @@ func (c *Controller[E]) OpenAbsoluteGateIfUncontrolled(tr telem.TimeRange, s con
 		// therefore this method should not create an absolute gate.
 		if r.timeRange.OverlapsWith(tr) {
 			if exists {
-				return nil, t, errors.Newf("[controller] - encountered multiple control regions for time range %s", tr)
+				c.L.DPanic(fmt.Sprintf("encountered multiple control regions for time range %s", tr))
```
emilbon99, line 339: Tidiness, just put the message once into a variable here.

cesium/internal/controller/controller.go
```diff
@@ -306,20 +336,21 @@ func (c *Controller[E]) OpenAbsoluteGateIfUncontrolled(tr telem.TimeRange, s con
...
+				return nil, t, errors.Newf("encountered multiple control regions for time range %s", tr)
 			}
 
 			r.Lock()
 			if r.curr != nil {
 				r.Unlock()
-				return nil, t, errors.Newf("[controller] - region already being controlled")
+				return nil, t, errors.Newf("cannot create a gate on the region %s because it is already controlled", r.timeRange)
```
emilbon99, line 346: Would be good to add the gates subject name and key

cesium/internal/controller/controller.go
```diff
@@ -360,10 +391,11 @@ func (c *Controller[E]) OpenGateAndMaybeRegister(cfg GateConfig, callback func()
 		if r.timeRange.OverlapsWith(cfg.TimeRange) {
 			// v1 optimization: one writer can only overlap with one region at any given time.
 			if exists {
-				return nil, t, errors.Newf("[controller] - encountered multiple control regions for time range %s", cfg.TimeRange)
+				c.L.DPanic(fmt.Sprintf("encountered multiple control regions for time range %s", cfg.TimeRange))
```
emilbon99, line 394: Same error as a few lines above. Can we condense somehow?

cesium/internal/controller/controller.go
```diff
@@ -400,7 +432,7 @@ func (c *Controller[E]) register(
 ) error {
 	for _, r := range c.regions {
 		if r.timeRange.OverlapsWith(t) {
-			return errors.Newf("entity already registered for time range %s", t)
+			return errors.Newf("time range %s collides with the time range for another entity", t)
```
emilbon99, line 435: Would be good to include relevant subject names and keys as well

cesium/internal/controller/controller.go
```diff
@@ -436,5 +468,5 @@ func (c *Controller[E]) remove(r *region[E]) {
 }
 
 func Unauthorized(name string, ch core.ChannelKey) error {
-	return errors.Wrapf(control.Unauthorized, "writer %s does not have control authority over channel %s", name, ch)
+	return errors.Wrapf(control.Unauthorized, "writer %s does not have control authority over channel %d", name, ch)
```
emilbon99, line 471: Don't know if you do this at the unary level, but would be good to add channel name here as well

cesium/internal/core/channel.go
```diff
@@ -33,9 +33,13 @@ const (
...
-	// Key is a unique identifier to the channel within the cesium.
+	// Key is a unique identifier to the channel within the cesium data engine.
 	// [REQUIRED]
-	Key     ChannelKey `json:"key" msgpack:"key"`
+	Key ChannelKey `json:"key" msgpack:"key"`
+	// Name is a non-unique, human-readable identifier to the channel within the data
+	// engine. Note that it is never used to index or retrieve a channel.
+	// [OPTIONAL]
+	Name    string `json:"name" msgpack:"name"`
 	IsIndex bool
```
emilbon99, line 43: Document this field

cesium/internal/domain/file_controller.go
```diff
@@ -332,14 +330,30 @@ func (fc *fileController) removeReadersWriters(ctx context.Context, key uint16)
...
+	fc.readers.RLock()
+	defer func() {
+		fc.readers.RUnlock()
+		fc.writers.RUnlock()
+	}()
 	c := errors.NewCatcher(errors.WithAggregation())
 	for _, w := range fc.writers.open {
+		c.Exec(func() error {
+			if !w.tryAcquire() {
+				return errors.Newf("writer for file %d is in use and cannot be closed", w.fileKey)
```
emilbon99, line 343: Might be good to figure out a way to inject the channel key and name before we pass this up the callstack. Would be good to know exactly which db still has files open.

cesium/internal/index/domain.go
```diff
@@ -35,7 +35,7 @@ func (i *Domain) Distance(ctx context.Context, tr telem.TimeRange, continuous bo
 	defer func() { err = errors.CombineErrors(err, iter.Close()) }()
 
 	if !iter.SeekFirst(ctx) || (!iter.TimeRange().ContainsRange(tr) && continuous) {
-		err = ErrDiscontinuous
+		err = NewErrDiscontinuousTR(tr)
```
emilbon99, line 38: May just be a personal preference, but I typically don't put `New` in front of these error constructors. Regardless of which route you choose, please standardize across the cesium codebase

cesium/internal/index/domain.go
```diff
@@ -168,14 +174,18 @@ func (i *Domain) Stamp(
...
-					err = ErrDiscontinuous
+					err = errors.Wrapf(domain.ErrRangeNotFound, "cannot find stamp end with offset %d", offset)
 					return
 				}
 				approx = Between(iter.TimeRange().End, telem.TimeStampMax)
 				return
 			}
 			gap += iter.Len() / 8
 			if endOffset < gap {
+				err = r.Close()
```
emilbon99, line 185: Why did you add all of these `Close` lines?

cesium/internal/unary/db.go
```diff
@@ -47,39 +47,30 @@ type DB struct {
...
+	closed     bool
 }
 
+var ErrDBClosed = core.EntityClosed("unary.db")
+
 func (db *DB) Index() index.Index {
 	if !db.Channel.IsIndex {
-		panic(fmt.Sprintf("[control.unary] - database %v does not support indexing", db.Channel.Key))
+		// inconceivable state
+		panic(fmt.Sprintf("channel %v is not an index channel", db.Channel.Key))
```
emilbon99, line 59: Would be useful to include the names of these channels in the panic messages as well

cesium/internal/unary/db.go
```diff
@@ -157,10 +157,16 @@ func (db *DB) TryClose() error {
 	db.mu.RLock()
 	defer db.mu.RUnlock()
 	if db.mu.openIteratorWriters > 0 {
-		return errors.Newf("[cesium] - cannot close channel because there are currently %d unclosed writers/iterators accessing it", db.mu.openIteratorWriters)
+		return db.wrapError(errors.Newf("cannot close channel %d because there are currently %d unclosed writers/iterators accessing it", db.Channel.Key, db.mu.openIteratorWriters))
```
emilbon99, line 160: Channel name!

cesium/internal/core/errors.go
```diff
@@ -1,16 +1,25 @@
...
-	return fmt.Errorf("[cesium] - %s is already closed", entityName)
+	return errors.Newf("operation on %s is invalid because it is already closed", entityName)
+}
+
+func NewErrorWrapper(key ChannelKey, name string) func(error) error {
+	return func(err error) error {
+		if err == nil {
+			return nil
+		}
+		return errors.Wrapf(err, "error in channel [%s]<%d>", name, key)
```
emilbon99, line 23: Why add `error in`, the user already knows its an error

cesium/internal/unary/db.go
emilbon99: Ah I see what you did. Can you please get rid of the duplicate keys

cesium/internal/unary/writer.go
```diff
@@ -181,10 +194,12 @@ func (w *Writer) len(dw *domain.Writer) int64 {
 
 // Write validates and writes the given array.
 func (w *Writer) Write(series telem.Series) (a telem.AlignmentPair, err error) {
+	defer func() { err = w.wrapError(err) }()
```
emilbon99, line 197: Weird but I guess I'm ok with it

emilbon99: I don't know how much value this defer statements adds though. Looks ugly to me. If you want to keep I'm fine with it though

cesium/internal/unary/writer_behavior_test.go
```diff
@@ -123,6 +125,24 @@ var _ = Describe("Writer Behavior", Ordered, func() {
 					By("Releasing control of the DB")
 					Expect(db.LeadingControlState()).To(BeNil())
 				})
+				Specify("Sad path", func() {
```
emilbon99, line 128: Better test name. All happy paths are alike, all unhappy paths are unique

cesium/internal/virtual/db.go
```diff
@@ -65,12 +72,18 @@ func (db *DB) LeadingControlState() *controller.State {
...
 
 func (db *DB) TryClose() error {
+	if db.closed {
+		return nil
+	}
 	db.mu.RLock()
 	defer db.mu.RUnlock()
 	if db.mu.openWriters > 0 {
-		return errors.Newf(fmt.Sprintf("[cesium] - cannot close channel because there are currently %d unclosed writers accessing it", db.mu.openWriters))
+		return db.wrapError(errors.Newf(fmt.Sprintf("cannot close channel %d because there are currently %d unclosed writers accessing it", db.Channel.Key, db.mu.openWriters)))
```
emilbon99, line 81: Redundant restatement

cesium/iterator_open.go
```diff
@@ -38,9 +44,9 @@ func (db *DB) newStreamIterator(cfg IteratorConfig) (*streamIterator, error) {
 		if !ok {
 			_, vOk := db.virtualDBs[key]
 			if vOk {
-				return nil, errors.Wrapf(validate.Error, "cannot open iterator on virtual channel %d", key)
+				return nil, errors.Newf("cannot open iterator on virtual channel %d", key)
```
emilbon99, line 47: Channel name!

cesium/writer_behavior_test.go
```diff
@@ -42,14 +42,15 @@ var _ = Describe("Writer Behavior", func() {
...
+				_ = cleanUp
 				Expect(cleanUp()).To(Succeed())
 			})
 			Describe("Happy Path", func() {
 				Context("Indexed", func() {
 					Specify("Basic Write", func() {
 						var (
-							basic1      = GenerateChannelKey()
-							basic1Index = GenerateChannelKey()
+							basic1      = testutil.GenerateChannelKey()
```
emilbon99, line 52: Any reson why you removed the `.` here. I prefer importing testutils with `.` imports

emilbon99 (approved):

emilbon99 (changes requested):
Accidentally marked as approved. Please review my comments. If there are any place where I request a channel name and it is already added, feel free to ignore.

cesium/internal/domain/file_controller.go
LeonLiur: All failures at this level must go through the unary level, so is it fine if this error only passes the mesasge, and it will be wrapped later?

cesium/internal/index/domain.go
LeonLiur: IMO it makes more sense to have the actual error stored as `ErrDiscontinuous`, and make a constructor that constructs it. This way we can match the errors easier. How would you do this?

cesium/internal/index/domain.go
LeonLiur: need to close all domain/iterator level readers to avoid leaking them

cesium/internal/controller/controller.go
LeonLiur: Yes, let's check upstream

cesium/channel.go
LeonLiur: I don't think it's an issue – changing the name does not affect anything in cesium.

cesium/writer_behavior_test.go
LeonLiur: not really

cesium/db.go
LeonLiur: Re-added

cesium/channel.go
emilbon99: I know its not an issue. We should leave a comment that we know its a race condition and we're intentionally choosing to ignore it because its not an issue

cesium/internal/domain/db.go
```diff
@@ -12,23 +12,33 @@ package domain
...
 	// ErrDomainOverlap is returned when a domain overlaps with an existing domain in the DB.
 	ErrDomainOverlap = errors.Wrap(validate.Error, "domain overlaps with an existing domain")
-	// RangeNotFound is returned when a requested domain is not found in the DB.
-	RangeNotFound = errors.Wrap(query.NotFound, "domain not found")
+	// ErrRangeNotFound is returned when a requested domain is not found in the DB.
+	ErrRangeNotFound = errors.Wrap(query.NotFound, "domain not found")
+	errDBClosed      = core.EntityClosed("domain.db")
 )
 
+func NewErrDomainOverlap(tr1, tr2 telem.TimeRange) error {
```
emilbon99, line 34: Call this `ErrWriteConflict`

cesium/delete.go
```diff
@@ -151,20 +153,20 @@ func (db *DB) removeChannel(ch ChannelKey) error {
 				}
 				otherDB := db.unaryDBs[otherDBKey]
 				if otherDB.Channel.Index == udb.Config.Channel.Key {
-					return errors.New("[cesium] - could not delete index channel with other channels depending on it")
+					return errors.Newf("cannot delete channel [%s]<%d> because it indexes data in channel [%s]<%d>", udb.Channel.Name, udb.Channel.Key, otherDB.Channel.Name, otherDB.Channel.Key)
```
emilbon99, line 156: It's probably worth adding a stringer for the channel struct as opposed to repeating the `[%s]<%d>` format

cesium/internal/domain/db.go
```diff
@@ -12,23 +12,34 @@ package domain
...
-	RangeNotFound = errors.Wrap(query.NotFound, "domain not found")
+	// ErrWriteConflict is returned when a domain overlaps with an existing domain in the DB.
+	ErrWriteConflict = errors.Wrap(validate.Error, "write conflicts with existing timerange in database")
+	// ErrRangeNotFound is returned when a requested domain is not found in the DB.
+	ErrRangeNotFound = errors.Wrap(query.NotFound, "timerange not found")
+	errDBClosed      = core.EntityClosed("domain.db")
 )
 
+func NewErrWriteConflict(tr1, tr2 telem.TimeRange) error {
+	return errors.Wrapf(ErrWriteConflict, "write overlaps on timerange %v", tr1.Intersect(tr2))
```
emilbon99, line 36: I think this message could still be better -> "write overlaps with existing data occupying time range"

cesium/internal/unary/writer.go
```diff
@@ -63,7 +63,7 @@ var (
 		EnableAutoCommit:         config.False(),
 		AutoIndexPersistInterval: 1 * telem.Second,
 	}
-	WriterClosedError = core.EntityClosed("unary.writer")
+	writerClosedError = core.EntityClosed("unary.writer")
```
emilbon99, line 66: Small convention thing. It seems like you typically prefix your errors with `err` as opposed to a suffix. This is very small, but perhaps something to think about standardizing

cesium/internal/unary/writer.go
```diff
@@ -142,12 +154,12 @@ func (db *DB) OpenWriter(ctx context.Context, cfgs ...WriterConfig) (w *Writer,
...
 	})
 	if err != nil {
-		return nil, transfer, err
+		return nil, transfer, w.wrapError(err)
 	}
 
 	w.control = g
-	db.mu.Add(1)
-	return w, transfer, err
+	db.mu.add(1)
```
emilbon99, line 161: Hard to figure out why you're adding a value hear. Might be better to name this function something like `addOpenWriterIterator`

cesium/internal/unary/writer.go
```diff
@@ -249,7 +266,7 @@ func (w *Writer) commitWithEnd(ctx context.Context, end telem.TimeStamp) (telem.
 			return 0, err
 		}
 		if !approx.Exact() {
-			return 0, errors.New("could not get exact timestamp")
+			return 0, errors.Wrapf(validate.Error, "writer start %s cannot be resolved in the index", w.Start)
```
emilbon99, line 269: Would be good to include the name of the index channel here. I'm fine with adding a public method to the index to return information about its name/key for debugging.

x/go/control/transfer.go
```diff
@@ -9,14 +9,22 @@
...
 type Subject struct {
 	Key  string `json:"key" msgpack:"key"`
 	Name string `json:"name" msgpack:"name"`
 }
 
-func (s Subject) String() string { return lo.Ternary(s.Name != "", s.Name, s.Key) }
+func (s Subject) String() string {
+	if s.Name != "" {
+		return fmt.Sprintf("[%s]<%s>", s.Name, s.Key)
+	} else {
```
emilbon99, line 24: Useless else statement due to early return.

x/go/telem/telem.go
```diff
@@ -178,6 +178,38 @@ func (tr TimeRange) String() string {
 	return tr.Start.String() + " - " + tr.End.String()
 }
 
+func (tr TimeRange) Union(tr2 TimeRange) (ret TimeRange) {
```
emilbon99, line 181: I think it'd be a good idea to add some tests for these

emilbon99 (changes requested):
Note for all error messages. Time range is two words. Find and replace occurences of `timerange` and replace with `time range`

Almost there!

cesium/internal/unary/writer.go
LeonLiur: Since the mutex only controls the entity count, I think it's better to just rename the variable from mu to entityCount, so we do `db.entityCount.Add`, reminiscent of something like `wg.Add(1)` or `sem.Add(1)`.

cesium/internal/unary/writer.go
LeonLiur: This condition should be inconceivable, but I added a public method `Info` to index still – we will need to log similar errors in handling deletion.

cesium/internal/domain/writer.go
```diff
@@ -303,10 +311,10 @@ func (w *Writer) Close(ctx context.Context) error {
 
 func (w *Writer) validateCommitRange(end telem.TimeStamp, switchingFile bool) error {
 	if !w.prevCommit.IsZero() && !switchingFile && end.Before(w.prevCommit) {
-		return errors.Wrap(validate.Error, "commit timestamp must not be less than the previous commit")
+		return errors.Wrapf(validate.Error, "commit timestamp %s must not be less than the previous commit timestamp %s", end, w.prevCommit)
 	}
 	if !w.Start.Before(end) {
-		return errors.Wrap(validate.Error, "commit timestamp must be strictly greater than the starting timestamp")
+		return errors.Wrapf(validate.Error, "commit timestamp %s must be strictly greater than the starting timestamp %s", end, w.Start)
```
emilbon99, line 317: We get this error all the time when dealing with timing issues on the driver side. It'd be good to represent these as actual stringified timestamps. And then it would also be good to indicate what the delta between the timestamps are (as a nicely strinigified timespan). See how we do this in [typescript](https://github.com/synnaxlabs/synnax/blob/3aac1847a5d1411c4204e0a3166f3bee63aad733/x/ts/src/telem/telem.ts#L507). It has been very useful

cesium/internal/index/domain.go
emilbon99: Ok thats fine

cesium/internal/unary/iterator.go
```diff
@@ -227,7 +241,7 @@ func (i *Iterator) Len() (l int64) {
 	return
 }
 
-func (i *Iterator) Error() error { return i.err }
+func (i *Iterator) Error() error { return core.NewErrorWrapper(i.Channel)(i.err) }
```
emilbon99, line 244: Kinda weird but I guess its ok

cesium/internal/unary/writer.go
```diff
@@ -217,22 +230,26 @@ func (w *Writer) updateHwm(series telem.Series) {
 }
 
 // Commit commits the written series to the database.
-func (w *Writer) Commit(ctx context.Context) (telem.TimeStamp, error) {
+func (w *Writer) Commit(ctx context.Context) (ts telem.TimeStamp, err error) {
```
emilbon99, line 233: I don't understand what the point of this extraction into a named variable is

cesium/internal/unary/writer.go
```diff
@@ -217,22 +230,26 @@ func (w *Writer) updateHwm(series telem.Series) {
...
 	if w.closed {
-		return telem.TimeStampMax, WriterClosedError
+		err = w.wrapError(errWriterClosed)
+		return
 	}
+
 	if w.Channel.IsIndex {
-		return w.commitWithEnd(ctx, w.hwm+1)
+		ts, err = w.commitWithEnd(ctx, w.hwm+1)
```
emilbon99, line 240: Why aren't you wrapping this error?

cesium/internal/virtual/db.go
```diff
@@ -27,50 +28,61 @@ type controlEntity struct {
...
+	db.entityCount.RLock()
+	defer db.entityCount.RUnlock()
+	if db.entityCount.openWriters > 0 {
+		return db.wrapError(errors.Newf(fmt.Sprintf("cannot close channel because there are %d unclosed writers accessing it", db.entityCount.openWriters)))
 	}
-	return db.Close()
-}
 
-func (db *DB) Close() error { return nil }
+	db.closed.Swap(true)
```
emilbon99, line 86: Why are you using swap when you could use `Store`? You don't need the return value here

x/go/telem/telem.go
```diff
@@ -178,6 +178,38 @@ func (tr TimeRange) String() string {
...
+	if tr.End.After(tr2.End) {
+		ret.End = tr.End
+	} else {
+		ret.End = tr2.End
+	}
+
+	return
+}
+
+func (tr TimeRange) Intersect(tr2 TimeRange) (ret TimeRange) {
```
emilbon99, line 197: If you're going to call the other method `Union`, this should be called `Intersection`

emilbon99 (changes requested):

cesium/internal/unary/iterator.go
LeonLiur: I thought since the error wrapper was only used in two places, it did not make sense to store it as as field in the iterator struct

emilbon99 (approved):
Good to go!

PR #596, opened 2024-06-01
Pjdotson/sy 696 console deep linking: Workspaces, schematics, line plots, ranges

pjdotson:
Deep linking implemented for workspaces, schematics, line plots, and ranges.


console/src/schematic/link.tsx
```diff
@@ -0,0 +1,38 @@
...
+  if (resource != "schematic") return false;
+  client.workspaces.schematic
+    .retrieve(resourceKey)
+    .then((schematic) => {
+      if (schematic == null) return false;
+      const layoutCreator = create({
+        ...(schematic.data as unknown as State),
+        key: schematic.key,
+        name: schematic.name,
+        // snapshot: schematic.snapshot,
```
pjdotson, line 28: Want to discuss if we should use schematic.snapshot

console/src/hardware/task/ontology.tsx
```diff
@@ -77,6 +83,7 @@ const TreeContextMenu: Ontology.TreeContextMenu = (props) => {
       <Menu.Item itemKey="delete" startIcon={<Icon.Delete />}>
         Delete
       </Menu.Item>
+      <Ontology.LinkAddressMenuItem />
```
emilbon99, line 86: This seems like it'd be a better fit for the `Link` module instead of the ontology

console/src/layouts/LayoutMain/LayoutMain.tsx
```diff
@@ -28,13 +29,14 @@ import { Workspace } from "@/workspace";
...
-  const d = useDispatch();
+  const dispatch = useDispatch();
   useEffect(() => {
-    d(Layout.maybeCreateGetStartedTab());
+    dispatch(Layout.maybeCreateGetStartedTab());
   }, []);
   Version.useLoadTauri();
   Device.useListenForChanges();
   Cluster.useLocalServer();
+  Link.useDeepLink({ handlers: Link.HANDLERS });
```
emilbon99, line 39: Redundant name. Call it `Link.useDeep`

console/src/channel/ontology.tsx
```diff
@@ -209,6 +215,7 @@ const TreeContextMenu: Ontology.TreeContextMenu = (props) => {
       <Menu.Item itemKey="delete" startIcon={<Icon.Delete />}>
         Delete
       </Menu.Item>
+      <Ontology.LinkAddressMenuItem />
```
emilbon99, line 218: Would probably be better to make this part of the `Link` package no? Call it `Link.CopyMenuItem`.

console/src/cluster/link.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Link } from "@/link";
+import { setActive } from "@/cluster/slice";
+
+export const linkHandler: Link.Handler = ({
+  resource,
+  resourceKey,
+  clusters,
+  dispatch,
+}) => {
+  if (resource !== "cluster") {
```
emilbon99, line 19: I like one-lining these if statements. Feels cleaner -> `if (resource !== "cluster") return false;`

console/src/cluster/link.tsx
```diff
@@ -0,0 +1,32 @@
...
+    return false;
+  }
+  const clusterKey = resourceKey;
+  if (clusters.find((cluster) => cluster.key === clusterKey) == undefined) {
+    // Console does not have this cluster in store.
+    console.error(
```
emilbon99, line 25: Dispatch this as a notification instead, so we can give feedback to the user.

console/src/lineplot/link.tsx
```diff
@@ -0,0 +1,37 @@
...
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { create, State } from "@/lineplot/slice";
+import { Link } from "@/link";
+
+export const linkHandler: Link.Handler = ({
```
emilbon99, line 13: I would honestly make the link handler be an async function that way we can use the `await` syntax and not `.then`

console/src/lineplot/link.tsx
```diff
@@ -0,0 +1,37 @@
...
+      const layoutCreator = create({
+        ...(linePlot.data as unknown as State),
+        key: linePlot.key,
+        name: linePlot.name,
+      });
+      placer(layoutCreator);
+      return true;
+    })
+    .catch((error) => {
+      console.error("Error: ", error);
```
emilbon99, line 33: Dispatch a notification here. I know I don't adhere to this myself but we should be doing it a lot more.

console/src/link/link.tsx
```diff
@@ -0,0 +1,91 @@
...
+          dispatch,
+          placer,
+          clusters,
+        }),
+      );
+    });
+  }, [client]);
+};
+
+export const HANDLERS: Handler[] = [
```
emilbon99, line 85: By importing and specifying all of the handlers here, you're creating a circular dependency. I would think about defining `HANDLERS` elsewhere, such as in `LayoutMain`

console/src/ontology/ContextMenu.tsx
```diff
@@ -34,3 +34,9 @@ export const RenameMenuItem = (): ReactElement => (
     Rename
   </Menu.Item>
 );
+
+export const LinkAddressMenuItem = (): ReactElement => (
```
emilbon99, line 38: Move this to the `Link` module

emilbon99 (changes requested):

console/src/link/link.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+export const useDeepLink = ({ handlers }: UseDeepLinkProps): void => {
+  const client = PSynnax.use();
+  const dispatch = useDispatch();
+  const placer = Layout.usePlacer();
+  const clusters = Cluster.useSelectMany();
+
+  useEffect(() => {
+    onOpenUrl((urls) => {
```
emilbon99, line 48: Cleanup function here

emilbon99 (changes requested):

PR #601, opened 2024-06-02
Ebonilla/sy 672 drift allow persistence of multiple windows

emilbon99:


pjdotson:
We should talk in person about this but I can't get the console to open after building the app or by running it in dev mode.

emilbon99:
I fixed this issue

PR #604, opened 2024-06-02
Ebonilla/sy 718 fix certain streaming transports not disconnecting in

emilbon99:


freighter/ts/src/websocket.ts
```diff
@@ -116,6 +116,7 @@ class WebSocketStream<RQ extends z.ZodTypeAny, RS extends z.ZodTypeAny = RQ>
     };
 
     this.ws.onclose = (ev: CloseEvent) => {
+      console.log("CLOSED");
```
pjdotson, line 119: Do you want to keep line 119 here?

emilbon99: Removed

PR #611, opened 2024-06-04
Ebonilla/sy 233 fix deadlock in distribution mock tests

emilbon99:


LeonLiur (approved):
Did not review the pluto & console stuff

PR #613, opened 2024-06-04
Ebonilla/sy 736 server shutdown improvements

emilbon99:


synnax/cmd/start.go
```diff
@@ -80,11 +80,11 @@ func start(cmd *cobra.Command) {
 	v := version.Get()
 	decodedName, _ := base64.StdEncoding.DecodeString("bGljZW5zZS1rZXk=")
 	var (
-		ins      = configureInstrumentation(v)
-		insecure = viper.GetBool("insecure")
-		verbose  = viper.GetBool("verbose")
-		autoCert = viper.GetBool("auto-cert")
-		verifier = viper.GetString(string(decodedName))
+		ins, prettyLogger = configureInstrumentation(v)
```
LeonLiur, line 83: why do we need a separate logger and can't just use `ins` to log the start & shutdown messages?

LeonLiur (approved):
LGTM and left one question

synnax/cmd/start.go
emilbon99: Because the production logger sometimes won't let us print nice output

PR #619, opened 2024-06-06
585 sy 693 support for cesium writer err on unauthorized

emilbon99:


synnax/pkg/api/frame_writer.go
```diff
@@ -29,13 +29,39 @@ type FrameWriterConfig struct {
...
+	// [REQUIRED]
+	Keys channel.Keys `json:"keys" msgpack:"keys"`
+	// Mode sets the persistence and streaming mode for the writer. The default mode is
+	// WriterModePersistStream. See the ts.WriterMode documentation for more.
+	// [OPTIONAL]
+	Mode writer.Mode `json:"mode" msgpack:"mode"`
+	// ErrOnUnauthorized controls whether the writer will return an error when
+	// attempting to write to a channel that it does not have authority over.
+	// In non-control scenarios, this value should be set to true. In scenarios
+	// that require control handoff, this value should be set to false.
```
LeonLiur, line 53: There should be an [OPTIONAL] Here

synnax/pkg/distribution/framer/writer/service.go
```diff
@@ -61,6 +62,11 @@ type Config struct {
 	// absolute authority for all channels.
 	// [OPTIONAL]
 	Authorities []control.Authority `json:"authorities" msgpack:"authorities"`
+	// ErrOnUnauthorized controls whether the writer will return an error when
+	// attempting to write to a channel that it does not have authority over.
+	// In non-control scenarios, this value should be set to true. In scenarios
+	// that require control handoff, this value should be set to false.
```
LeonLiur, line 68: Here too

synnax/pkg/version/VERSION
```diff
@@ -1 +1 @@
-0.20.15
+0.20.6
```
LeonLiur, line 1: Should be 16

cesium/internal/controller/controller.go
```diff
@@ -67,9 +73,23 @@ func (g *Gate[E]) Authorize() (e E, ok bool) {
 	}
 	g.r.RUnlock()
 	if !ok {
-		return e, false
+		currState := g.r.curr.State()
+		return e, errors.Wrapf(
+			control.Unauthorized,
+			"%s has no control authority over channel %v - it is currently held by %s",
```
LeonLiur, line 79: should use gate subject stringer

cesium/writer_behavior_test.go
```diff
@@ -1233,7 +1231,7 @@ var _ = Describe("Writer Behavior", func() {
 					})
 				})
 			})
-			Describe("Data t Errors", func() {
+			Describe("Data & Errors", func() {
```
LeonLiur, line 1234: This should be data type error

cesium/internal/controller/controller.go
```diff
@@ -67,9 +73,23 @@ func (g *Gate[E]) Authorize() (e E, ok bool) {
...
+			currState.Resource,
+			currState.Subject.Name,
+		)
 	}
-	return g.r.entity, ok
+	return g.r.entity, nil
+}
+
+// Current returns the gate currently in control of the region's entity.
+func (g *Gate[E]) Current() State {
```
LeonLiur, line 89: Where does this get used?

cesium/internal/controller/controller.go
```diff
@@ -67,9 +73,23 @@ func (g *Gate[E]) Authorize() (e E, ok bool) {
 	}
 	g.r.RUnlock()
 	if !ok {
-		return e, false
+		currState := g.r.curr.State()
```
LeonLiur, line 76: Race here g.r.curr access is not safe

cesium/internal/virtual/writer.go
```diff
@@ -14,16 +14,65 @@ import (
...
+	closed    bool
+	WriterConfig
+}
+
+func (db *DB) OpenWriter(_ context.Context, cfgs ...WriterConfig) (w *Writer, transfer controller.Transfer, err error) {
 	if db.closed.Load() {
 		return nil, transfer, db.wrapError(dbClosed)
 	}
+	cfg, err := config.New(DefaultWriterConfig, cfgs...)
+	if err != nil {
```
LeonLiur, line 73: wrap error here

cesium/internal/virtual/writer.go
```diff
@@ -38,39 +87,28 @@ func (db *DB) OpenWriter(_ context.Context, cfg WriterConfig) (w *Writer, transf
 			align: a,
 		}, nil
 	})
+	if err != nil {
+		return nil, transfer, err
```
LeonLiur, line 91: wrap error here

cesium/internal/virtual/writer.go
```diff
@@ -38,39 +87,28 @@ func (db *DB) OpenWriter(_ context.Context, cfg WriterConfig) (w *Writer, transf
...
+	}
+	if *cfg.ErrOnUnauthorized {
+		if _, err = g.Authorize(); err != nil {
+			g.Release()
+			return nil, transfer, err
```
LeonLiur, line 96: and here

cesium/internal/virtual/writer.go
```diff
@@ -38,39 +87,28 @@ func (db *DB) OpenWriter(_ context.Context, cfg WriterConfig) (w *Writer, transf
...
 	if w.closed {
 		return 0, w.wrapError(errWriterClosed)
 	}
 	if err := w.Channel.ValidateSeries(series); err != nil {
 		return 0, w.wrapError(err)
 	}
-	e, ok := w.control.Authorize()
+	e, ok := w.control.Authorized()
 	if !ok {
 		return 0, nil
```
LeonLiur, line 113: virtual writer doesn't err on unauthoirzed?

LeonLiur (changes requested):
I will fix the minor stylistic stuff tomorrow morning. Do read over the comments I left though.

Are the OPC driver stuff supposed to be in this PR?

Also cesium test failing

cesium/delete.go
```diff
@@ -240,12 +237,12 @@ func (db *DB) garbageCollect(ctx context.Context, maxGoRoutine int64) error {
 	_, span := db.T.Debug(ctx, "garbage_collect")
 	defer span.End()
 	db.mu.RLock()
```
LeonLiur, line 240: can condense the two defer statements into one

cesium/delete.go
```diff
@@ -257,8 +254,6 @@ func (db *DB) garbageCollect(ctx context.Context, maxGoRoutine int64) error {
 			return udb.GarbageCollect(_ctx)
 		})
 	}
-
-	db.mu.RUnlock()
 	return sCtx.Wait()
```
LeonLiur, line 258: I feel like we shouldn't lock down channel creation during GC. Creating a channel that uses an index channel in GC as index is fine – we just won't be able to write data for it.

As for deletion, we should not allow that, but we should address that at the Unary level with `entityCount`

cesium/delete.go
```diff
@@ -55,41 +56,38 @@ func (db *DB) DeleteChannel(ch ChannelKey) error {
...
 	// may take a longer time.
 	// Rename the file to have a random suffix in case the channel is repeatedly created
 	// and deleted.
 	oldName := channelDirName(ch)
 	newName := oldName + "-DELETE-" + strconv.Itoa(rand.Int())
-	err = db.fs.Rename(oldName, newName)
-	if err != nil {
-		db.mu.Unlock()
-		return nil
+	if err := (func() error {
```
LeonLiur, line 65: wow getting fancy

cesium/internal/unary/writer.go
```diff
@@ -251,7 +257,7 @@ func (w *Writer) CommitWithEnd(ctx context.Context, end telem.TimeStamp) (err er
 }
 
 func (w *Writer) commitWithEnd(ctx context.Context, end telem.TimeStamp) (telem.TimeStamp, error) {
-	dw, ok := w.control.Authorize()
+	dw, ok := w.control.Authorized()
```
LeonLiur, line 260: Could we replace this syntax with `dw, err := w.control.Authorize()`

cesium/internal/virtual/writer.go
```diff
@@ -63,20 +73,56 @@ type Writer struct {
...
 	}
 	if err := w.Channel.ValidateSeries(series); err != nil {
 		return 0, w.wrapError(err)
 	}
-	e, ok := w.control.Authorize()
-	if !ok {
-		return 0, nil
+	e, err := w.control.Authorize()
+	if err != nil {
+		return 0, err
```
LeonLiur, line 121: wrap error

cesium/open.go
```diff
@@ -68,15 +68,7 @@ func Open(dirname string, opts ...Option) (*DB, error) {
 }
 
 func (db *DB) openVirtualOrUnary(ch Channel) error {
-	if db.closed.Load() {
-		return errDBClosed
-	}
-
-	db.mu.Lock()
-	defer db.mu.Unlock()
```
LeonLiur, line 76: We must not forget that this method is also called on db creation – but this is fine since we don't need a lock there as the db object is not exposed to the caller yet

client/py/synnax/exceptions.py
```diff
@@ -40,6 +40,27 @@ def __init__(self, field: str, message: str):
...
+
+    pass
+
+
+class UnauthorizedError(Exception):
+    """
+    Raised when an entity attempts to access or modify information it is not allowed.
+    """
+
+    TYPE = _FREIGHTER_EXCEPTION_PREFIX + "control.unauthorized"
```
LeonLiur, line 58: minor, but why not `ControlError.TYPE + ".unauthorized"`?

client/py/synnax/framer/client.py
```diff
@@ -107,6 +110,7 @@ def open_writer(
             authorities=authorities,
             name=name,
             mode=mode,
+            err_on_unauthorized=err_on_unauthorized,
```
LeonLiur, line 113: what about sendAuthError?

LeonLiur (approved):
What happened to `SendAuthErr`? it never got passed up the stack

driver/errors/errors.h
```diff
@@ -14,6 +14,6 @@
 namespace driver {
     const std::string ERROR_PREFIX = "sy.driver.";
     const std::string TYPE_CRITICAL_HARDWARE_ERROR = ERROR_PREFIX + "hardware.critical";
-    const std::string TYPE_TEMPORARY_HARDWARE_ERROR = ERROR_PREFIX + "hardware.temporary";
+    const std::string TEMPORARY_HARDWARE_ERROR = ERROR_PREFIX + "hardware.temporary";
     const std::string TYPE_CONFIGURATION_ERROR = ERROR_PREFIX + "configuration";
```
Lham42, line 18: why do the other errors need TYPE in front of them?

driver/opc/reader.cpp
```diff
@@ -569,21 +573,39 @@ std::unique_ptr<task::Task> Reader::configure(
...
+        ctx->setState({
+            .task = task.key,
+            .variant = "success",
+            .details = json{
+                {"running", true},
+                {"message", "Task started successfully"}
+            }
+        });
+    }
     else if (cmd.type == "stop") return stop();
```
Lham42, line 597: you got rid of the Reader::start() but kept Reader::stop() which are nearly identical. you should get rid of both start and stop, or keep start()

Lham42: I guess the reason for this is because task::Task has a pure virtual stop() because the task manager uses this when it needs to reconfigure a task so I would not remove stop()

driver/pipeline/control.h
```diff
@@ -16,55 +16,123 @@
...
-
-    /// @brief breaker
+    std::shared_ptr<StreamerFactory> factory;
+    synnax::StreamerConfig config;
+    std::shared_ptr<Sink> sink;
     breaker::Breaker breaker;
 
     void runInternal();
 
+    void maybeJoinThread() const;
```
Lham42, line 134: I don't like this name. The idea of this function should be that all joinable threads should be joined if this function returns (unless the run thread itself calls it which should not be done but is checked for sanity)

PR #620, opened 2024-06-06
Pjdotson/sy 747 console create new line plot from workspace under ontology

pjdotson:


console/src/workspace/ontology.tsx
```diff
@@ -6,118 +6,135 @@
...
+        tree: nodes,
+        keys: selection.resources.map((r) => r.id.toString()),
+      });
+      setNodes([...next]);
+    })();
+  };
+
+  const handleCreateNewSchematic = (): void => {
+    const workspace = selection.resources[0].id.key;
+    void (async () => {
```
emilbon99, line 56: We should improve this void async pattern. We can probably leverage something like a custom wrapper around a react query mutation to make things much cleaner and more reliable. We'll talk tomorrow.

emilbon99 (commented):
PR looks great - one comment we should address together tomorrow

PR #624, opened 2024-06-06
Ebonilla/sy 742 fix issues with recreating channels with the same name

emilbon99:


client/py/synnax/channel/retrieve.py
```diff
@@ -94,6 +94,32 @@ def __init__(
         self.instrumentation = instrumentation
         self.__retriever = retriever
 
+    def delete(self, keys: ChannelParams) -> None:
+        normal = normalize_channel_params(keys)
+        if normal.variant == "names":
+            for name in normal.params:
+                key = self.__names_to_keys.get(name)
```
LeonLiur, line 101: A name may correspond to multiple channels

emilbon99: Fixed

PR #629, opened 2024-06-12
618 sy 721 client interfaces for timerange deletions

LeonLiur:


synnax/pkg/distribution/channel/delete_test.go
```diff
@@ -29,72 +29,74 @@ var _ = Describe("Delete", Ordered, func() {
...
+					Expect(exists).To(BeFalse())
+				})
+				It("Should not be able to retrieve the channel from the storage DB", func() {
+					Expect(services[1].Delete(ctx, ch.Key(), true)).To(Succeed())
+					channels, err := builder.Cores[1].Storage.TS.RetrieveChannels(ctx, ch.Key().StorageKey())
+					Expect(err).To(MatchError(cesium.ErrChannelNotFound))
+					Expect(channels).To(BeEmpty())
+				})
 			})
+			//Context("Node is remote", func() {
```
emilbon99, line 71: Add a note at the top to explain why we commented this out

synnax/pkg/distribution/framer/deleter/lease_proxy.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (lp *leaseProxy) deleteTimeRangeByName(
+	ctx context.Context,
+	retriever channel.Readable,
+	names []string,
+	tr telem.TimeRange,
+) error {
+	var res []channel.Channel
+	retriever.NewRetrieve().WhereKeys()
```
emilbon99, line 70: Dead statement. Remove

synnax/pkg/distribution/framer/deleter/lease_proxy.go
```diff
@@ -0,0 +1,109 @@
...
+	return lp.deleteTimeRangeGateway(ctx, batch.Gateway, tr)
+}
+
+func (lp *leaseProxy) deleteTimeRangeByName(
+	ctx context.Context,
+	retriever channel.Readable,
+	names []string,
+	tr telem.TimeRange,
+) error {
+	var res []channel.Channel
```
emilbon99, line 69: I would pre-allocate this array with a specific capacity equal to the length of names. Deleting by name is also kind of sus because there could be many channels with the same name

emilbon99: But I guess that's just a caveat we need to note in the docs

synnax/pkg/distribution/transport/grpc/framer/transport.go
```diff
@@ -86,22 +101,22 @@ func New(pool *fgrpc.Pool) Transport {
 				ClientFunc: func(
 					ctx context.Context,
 					conn grpc.ClientConnInterface,
-				) (fgrpc.GRPCClientStream[*framerv1.WriterRequest, *framerv1.WriterResponse], error) {
-					return framerv1.NewWriterServiceClient(conn).Write(ctx)
+				) (fgrpc.GRPCClientStream[*tsv1.WriterRequest, *tsv1.WriterResponse], error) {
```
emilbon99, line 104: We should rename this back to framer. `tsv1` doesn't make a ton of sense in this context

emilbon99 (approved):
Phenomal! I told you it was going to start getting easier to build out big features without a ton of feedback comments. I left a few small notes - once you make those changes you're good to merge.

synnax/pkg/distribution/framer/deleter/lease_proxy.go
LeonLiur: problematic if a name matches multiple results and one name is not found: unique names would have same length as desired names. Is the main concern for not preallocating for memory?

synnax/pkg/distribution/transport/grpc/framer/transport.go
LeonLiur: I think the idea was that this package also imports the `framer` package. I will rename it to `framerv1`.

client/py/synnax/synnax.py
```diff
@@ -56,7 +56,7 @@ class Synnax(Client):
     signals: Registry
     hardware: HardwareClient
 
-    __client: Transport
+    __stream_client: Transport
```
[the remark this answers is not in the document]
LeonLiur: Not a typo, `_transport` collides with the name `_transport` in the framer client class. I will rename that to `stream_transport` or something

PR #632, opened 2024-06-12
SY-769: Console - ontology context menu did not appear correctly

pjdotson:
Changed the Pluto ContextMenu() React element to only up a context menu for multiple selected items if the context menu as was opened above one of the selected items.


pjdotson:
Adding a comment here because I wanted to talk about potential styling changes to make to CSS files to reflect the difference between "selected" (left-clicked) and right clicked (context-menu pops up for).

PR #633, opened 2024-06-12
[synnax] - search index improvements

emilbon99:


synnax/pkg/distribution/ontology/search/search_test.go
```diff
@@ -131,4 +137,20 @@ var _ = Describe("Search", func() {
...
+			tk := &search.SepTokenizer{}
+			tok := tk.Tokenize([]byte(input))
+			Expect(tok).To(HaveLen(len(expected)))
+			for i, term := range expected {
+				Expect(tok[i].Term).To(Equal([]byte(term)))
+			}
+		},
+		Entry("Single Word", "test", []string{"test"}),
+		Entry("Two Words", "test test", []string{"test", "test"}),
+		Entry("Two Words with Underscore", "test_test", []string{"test", "test"}),
```
LeonLiur, line 151: I would add a test for Two words with dot too

synnax/pkg/distribution/ontology/search/search_test.go
```diff
@@ -23,6 +23,7 @@ var _ = Describe("Search", func() {
 		var idx *search.Index
 		BeforeEach(func() {
 			idx = MustSucceed(search.New())
+			idx.Register(ctx, schema.Schema{Type: "test"})
 		})
 		DescribeTable("Term Searching",
```
LeonLiur, line 28: I recall the `[` and `]` characters were problematic, we could add a few tests on those while we're at it?

LeonLiur (approved):
LGTM. Added two comments for two additional test cases we should add

PR #638, opened 2024-06-13
[ops] - fix non-double-quoted json property

LeonLiur:


emilbon99:
This property should not be double quoted.

PR #646, opened 2024-06-15
Ebonilla/sy 731 schematic tab rename doesnt propagate to workspace

emilbon99:


pjdotson:
@emilbon99 I can't get this to work on my branch, it looks like the name changes still don't propagate to the workspace.

PR #648, opened 2024-06-15
Leo/sy 779 cesium codebase inspection

LeonLiur:
fixed bugs:
- Added a check to not configuring a control update channel on a non-virtual non-string channel
- Fixed behaviour where writing to a timestamp channel with int64 would not error: must write in timestamps.
- Fixed bug in unary iterator where after reading parts of a doamin, that domain is skipped directly
- Fixed bug in unary iterator where if view is entirely in a discontinuity, following reads are incorrect.
- Added lots of tests for cesium iterator


cesium/internal/unary/iterator.go
```diff
@@ -138,7 +138,7 @@ func (i *Iterator) Next(ctx context.Context, span telem.TimeSpan) (ok bool) {
 
 	i.reset(i.view.End.SpanRange(span).BoundBy(i.bounds))
 
-	if i.view.IsZero() {
+	if i.view.IsZero() || i.view.End.BeforeEq(i.internal.TimeRange().Start) {
```
emilbon99, line 141: I'd be interested in getting an explanation of why you added this

cesium/writer_stream.go
```diff
@@ -413,16 +412,14 @@ func (w *idxWriter) validateWrite(fr Frame) error {
...
 
+		if fr.Series[i].DataType != uWriter.Channel.DataType {
+			return errors.Wrapf(
+				validate.Error,
+				"invalid data type for channel %d, expected %s, got %s",
+				k, uWriter.Channel.DataType, fr.Series[i].DataType)
+		}
 		if lengthOfFrame == -1 {
-			s := fr.Series[i]
-			if s.DataType.Density() == 0 {
```
emilbon99, line 418: Can you explain this change?

emilbon99 (approved):
Approved once you address my questions

cesium/internal/index/domain.go
```diff
@@ -39,6 +39,12 @@ func (i *Domain) Distance(ctx context.Context, tr telem.TimeRange, continuous bo
 
 	if !iter.SeekFirst(ctx) || (!iter.TimeRange().ContainsRange(tr) && continuous) {
```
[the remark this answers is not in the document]
LeonLiur: Changed to a better error message

cesium/internal/unary/iterator.go
LeonLiur: If the entire view is before the start of the internal domain, that means there's no data to be read.

Added to address a bug where if the view is put between two domains via `Next`, it unnecessarily moves on to the next domain and skip the previous one. (domain 1: [10, 20), domain 2: [30, 40), domain 3: [50, 60), say view is at [23, 24), the iterator position is at domain 2 – instead of reading nothing and doing nothing, the iterator actually moved on to domain 3, skipping domain 2 entirely.

cesium/writer_stream.go
LeonLiur: We want to check every frame for whether it has the right data type, not just the first.

LeonLiur: We want to check whether every series has the right data type, not just the first

emilbon99: Shouldn't we be checking this in the unary layer? I added this so that getting the length of the series doesn't panic on a variable length series

LeonLiur: Yes, you're right, we're already checking at the unary level whether the data type is correct. But I still don't understand why we only chekc for the first channel if the data type is unknown. Could you explain what you mean by "variable length series". Isn't getting the length of the frame supposed to panic when the frame is not even?

LeonLiur: Also, just to confirm – at the unary level we check as if int64 and timestamp data type are interchangeable – should that be the case? Should we allow passing a int64 series into a timestamp channel?

PR #655, opened 2024-06-21
Leo/sy 822 client set chunk size for iterators

LeonLiur:

## Key Information

- [Linear Issue](https://linear.app/synnaxlabs/issue/SY-822/client-set-chunk-size-for-iterators)

## Description

Added option to specify chunk size used in autospan iteration from python and typescript client.

## Basic Readiness Checklist

## Manual QA Additions

## Breaking Changes


client/py/synnax/framer/client.py
```diff
@@ -127,11 +127,13 @@ def open_iterator(
         self,
         tr: TimeRange,
         params: ChannelParams,
+        chunk_size: int = 5e5,
```
emilbon99, line 130: We should reduce this. Could make it easy to overload the maximum message size for certain websockret connections

client/py/synnax/framer/iterator.py
```diff
@@ -72,18 +73,21 @@ class Iterator:
     tr: TimeRange
     instrumentation: Instrumentation
     value: Frame
+    _chunk_size: int
```
emilbon99, line 76: Why is this underscored?

emilbon99: Also why does this field need to be in the response?

client/ts/src/framer/iterator.ts
```diff
@@ -105,6 +107,7 @@ export class Iterator {
       command: Command.Open,
       keys: adapter.keys,
       bounds: new TimeRange(tr),
+      chunkSize: chunkSize || 5e5,
```
emilbon99, line 110: Use as a default argument in the function instead. Just like python. Also, when doing an op like this (if you do do it), you should use nullish coalescing instead (`chunkSize ?? 5e5`).  As with the note on python, let's make the auto chunk size significantly smaller.

docs/site/src/pages/reference/python-client/read-data.mdx
```diff
@@ -121,7 +121,9 @@ situations where it's necessary to process large volumes of data. Sometimes
 these reads may be too large to fit in memory.
 
 Synnax supports server side iterators that allow us to process large queries in
-consistently sized chunks:
+consistently sized chunks. To configure reading in chunks, pass in `chunk_size` to the
```
emilbon99, line 124: Fix this - it makes it seem like you need to specify the chunk size in order to read in chunks. By default, Synnax uses a chunk size of X. If you'd like to set a custom size, use the `chunk_size` argument.

client/ts/src/framer/client.ts
```diff
@@ -42,11 +42,12 @@ export class Client {
...
    *
    * @param tr - A time range to iterate over.
-   * @param keys - A list of channel keys to iterate over.
+   * @param channels - A list of channels (by name or key) to iterate over.
+   * @param chunkSize - Number of samples in each iteration under AutoSpan.
    * @returns a new {@link TypedIterator}.
    */
-  async openIterator(tr: CrudeTimeRange, channels: Params): Promise<Iterator> {
-    return await Iterator._open(tr, channels, this.retriever, this.streamClient);
+  async openIterator(tr: CrudeTimeRange, channels: Params, chunkSize?: number): Promise<Iterator> {
```
emilbon99, line 49: Since you can't have named arguments in TS, I would actually accept this as an extra `IteratorOptions` parameter that has `chunkSize` as a field. This makes it easy to extend options in the future without the caller needing to worry about the order of the arguments. I also makes it clearer what the number corresponds to when calling the function.

docs/site/src/pages/reference/typescript-client/read-data.mdx
```diff
@@ -68,15 +68,18 @@ For more information on the `Frame` class, see [Frames](series-and-frames#frames
...
 ```typescript
 const start = TimeStamp.now();
 const end = start.add(TimeStamp.seconds(10));
 
 const iterator = await client.openIterator(
-    { start, end }, 
-    ["temperature", "humidity"]
+    { start, end },
+    ["temperature", "humidity"],
+    100, // chunkSize
```
emilbon99, line 82: You wouldn't need the extra comment here if you use a TS options pattern instead. See my comment on the method signature

emilbon99 (changes requested):

emilbon99 (approved):

PR #670, opened 2024-07-01
Ebonilla/sy 842 fix cesium alignment regression

emilbon99:
# Fix Pull Request Template

## Key Information

- Linear Issue: [SY-842](https://linear.app/synnaxlabs/issue/SY-842/fix-cesium-alignment-regression)

## Description

Fixes an issue where a domain in a data channel and the corresponding domain in the index channel would not align correctly. 

## Basic Readiness Checklist

- [x] I have added sufficient regression tests to cover the changes.

## Manual QA Additions

### DIFF
--- a/cesium/internal/index/domain.go
+++ b/cesium/internal/index/domain.go
@@ -29,7 +29,11 @@ type Domain struct {
 var _ Index = (*Domain)(nil)
 
 // Distance implements Index.
-func (i *Domain) Distance(ctx context.Context, tr telem.TimeRange, continuous bool) (approx DistanceApproximation, err error) {
+func (i *Domain) Distance(
+	ctx context.Context,
+	tr telem.TimeRange,
+	continuous bool,
+) (approx DistanceApproximation, bounds DomainBounds, err error) {
 	var startApprox, endApprox DistanceApproximation
 	ctx, span := i.T.Bench(ctx, "distance")
 	defer func() { _ = span.EndWith(err, ErrDiscontinuous) }()
@@ -37,15 +41,23 @@ func (i *Domain) Distance(ctx context.Context, tr telem.TimeRange, continuous bo
 	iter := i.DB.NewIterator(domain.IteratorConfig{Bounds: tr})
 	defer func() { err = errors.CombineErrors(err, iter.Close()) }()
 
+	// Case 1 - If the domain with the given time range doesn't exist in the database, then
+	// there's nothing we can do.
+	// Case 2 - If the domain exists, but it doesn't contain the entire time range,
+	// then it's discontinuous, and we return early if the user doesn't want discontinuous
+	// results.
 	if !iter.SeekFirst(ctx) || (!iter.TimeRange().ContainsRange(tr) && continuous) {
 		err = NewErrDiscontinuousTR(tr)
 		return
 	}
 
+	// If the time range is zero, then the distance is zero.
 	if tr.IsZero() {
+		bounds = ExactDomainBounds(iter.Position())
 		return
 	}
 
+	// Open a new reader on the domain at the start of the range.
 	r, err := iter.NewReader(ctx)
 	if err != nil {
 		return
@@ -57,12 +69,15 @@ func (i *Domain) Distance(ctx context.Context, tr telem.TimeRange, continuous bo
 		return
 	}
 
+	// If the current domain contains the end of the time range, then everything
+	// is continuous and within the current domain.
 	if iter.TimeRange().ContainsStamp(tr.End) || tr.End == iter.TimeRange().End {
 		endApprox, err = i.search(tr.End, r)
 		approx = Between(
 			endApprox.Lower-startApprox.Upper,
 			endApprox.Upper-startApprox.Lower,
 		)
+		bounds = ExactDomainBounds(iter.Position())
 		return
 	} else if continuous {
 		err = NewErrDiscontinuousTR(tr)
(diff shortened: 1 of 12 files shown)


client/py/examples/dev/multi_stream.py
```diff
@@ -0,0 +1,67 @@
+import synnax as sy
```
LeonLiur, line 1: When I run this example I get a cesium error, could you double check you get that too?

cesium/internal/unary/iterator_behavior_test.go
```diff
@@ -643,6 +643,48 @@ var _ = Describe("Iterator Behavior", Ordered, func() {
...
+					It("Should correctly define the alignment of a series when a domain has already been written to the index channel", func() {
+						Expect(unary.Write(ctx, indexDB, 6*telem.SecondTS, telem.NewSecondsTSV(6, 7, 8, 9, 10, 11, 12, 13, 14, 15))).To(Succeed())
+						Expect(unary.Write(ctx, indexDB, 20*telem.SecondTS, telem.NewSecondsTSV(20, 21, 22, 23, 24, 25, 26))).To(Succeed())
+						Expect(unary.Write(ctx, db, 20*telem.SecondTS, telem.NewSeriesV[int64](8, 9, 10, 11, 12, 13, 14))).To(Succeed())
+						iter := db.OpenIterator(unary.IteratorConfig{
+							Bounds:        (20 * telem.SecondTS).SpanRange(15 * telem.Second),
+							AutoChunkSize: 3,
+						})
+						defer func() {
+							defer GinkgoRecover()
```
LeonLiur, line 677: Should we keep these debug helpers in the test files or remove them

cesium/internal/unary/iterator.go
```diff
@@ -318,33 +324,42 @@ func (i *Iterator) read(ctx context.Context, offset telem.Offset, size telem.Siz
...
+	err error,
+) {
+	target := i.internal.TimeRange().Start.SpanRange(0)
 	if i.internal.TimeRange().Start.Before(i.view.Start) {
-		target := i.internal.TimeRange().Start.Range(i.view.Start)
-		startApprox, err = i.idx.Distance(ctx, target, true)
+		target.End = i.view.Start
 	}
-	return
+	sa, da, err := i.idx.Distance(ctx, target, true)
```
LeonLiur, line 361: I understand `sa` and `da`, but I think it's better if we call it `startApprox` and `domainApprox` so we don't need ot think about what `sa` and `da` mean

cesium/internal/index/domain.go
```diff
@@ -29,23 +29,35 @@ type Domain struct {
 var _ Index = (*Domain)(nil)
 
 // Distance implements Index.
-func (i *Domain) Distance(ctx context.Context, tr telem.TimeRange, continuous bool) (approx DistanceApproximation, err error) {
+func (i *Domain) Distance(
+	ctx context.Context,
+	tr telem.TimeRange,
+	continuous bool,
+) (approx DistanceApproximation, bounds DomainBounds, err error) {
```
LeonLiur, line 36: very minor - ignore this if you don't think it's necessary - `domainBounds` may be a better name since there are too many things in this method that can be called `bounds`.

cesium/internal/index/rate.go
```diff
@@ -40,3 +40,7 @@ func (r Rate) Stamp(_ context.Context, ref telem.TimeStamp, distance int64, _ bo
 func (r Rate) Info() string {
 	return fmt.Sprintf("rate index: %v", r.Channel)
 }
+
+func (r Rate) Domain(_ context.Context, ref telem.TimeStamp) (DomainBounds, error) {
```
LeonLiur, line 44: Why does this method exist?

cesium/internal/index/domain.go
```diff
@@ -285,3 +309,11 @@ func readStamp(r io.ReaderAt, offset int64, buf []byte) (telem.TimeStamp, error)
 func (i *Domain) Info() string {
 	return fmt.Sprintf("domain index: %v", i.Channel)
 }
+
+func (i *Domain) Domain(ctx context.Context, ref telem.TimeStamp) (DomainBounds, error) {
```
LeonLiur, line 313: Is this method used anywhere? If so, since both rate and domain implement this method, it should go into the index interface.
Also, in the case where `ref` is not an exact timestamp in the continuous domain, the end result is discontinuous, but still 'exact' – as long as the caller is aware of this, it's fine.

LeonLiur (approved):
Could you look at the example file you added and check that it runs properly? Otherwise looks good once you address the minor changes I highlighted

cesium/internal/index/domain.go
emilbon99: Good idea

cesium/internal/index/domain.go
emilbon99: Oh you're right its not used. I meant to get rid of this - good catch

cesium/internal/unary/iterator.go
emilbon99: Agreed

cesium/internal/unary/iterator_behavior_test.go
emilbon99: No. I'll remove. I thought this was necessary to recover from an assertion failure in the `defer` statement

client/py/examples/dev/multi_stream.py
emilbon99: Works fine for me

LeonLiur: Perfect, works for me too now

PR #690, opened 2024-07-05
[ops/ts] - improved error matching semantics

emilbon99:
# Fix Pull Request Template

## Key Information

- Linear Issue: [SY-897](https://linear.app/synnaxlabs/issue/SY-897/ni-fix-channel-recreation-in-task-configuration-workflows)

## Description

Improves error matching by adding a new `matches` function instead of using `instanceof`. `instanceof` can cause problems when multiple projects that have different versions of dependencies are bundled together. In these cases, `instanceof` will not work. Instead, we match by the `type` filed on the error.

## Basic Readiness Checklist

- [x] I have added sufficient regression tests to cover the changes.

## Manual QA Additions

### DIFF
--- a/client/ts/src/auth/auth.ts
+++ b/client/ts/src/auth/auth.ts
@@ -73,7 +73,7 @@ export class Client {
       }
       reqCtx.params.Authorization = `Bearer ${this.token}`;
       const [resCtx, err] = await next(reqCtx);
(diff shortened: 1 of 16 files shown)


pjdotson:
@emilbon99 There is a failing TS freighter test in CI due to different error matching.

emilbon99:
> @emilbon99 There is a failing TS freighter test in CI due to different error matching.

Fixed

PR #705, opened 2024-07-09
Sy 874 server automatic panic recovery

LeonLiur:

## Key Information

- **Linear Issue**: [SY-874](https://linear.app/synnaxlabs/issue/SY-874/[server]-automatic-panic-recovery)

## Description

Panic recovery in X/Signal.

## Basic Readiness Checklist

## Migrations

## Manual QA Additions

## Breaking Changes


x/go/signal/routine.go
```diff
@@ -117,6 +128,21 @@ func CancelOnExitErr() RoutineOption {
...
 
+func WithPanicPolicy(p PanicPolicy) RoutineOption {
+	return func(r *routineOptions) {
+		r.panicPolicy = p
+	}
+}
+
+func WithMaxRestart(maxRestart int) RoutineOption {
+	return func(r *routineOptions) {
+		if r.panicPolicy != Restart {
```
emilbon99, line 139: I don't think this should panic. Just do an early return.

x/go/signal/routine.go
```diff
@@ -55,6 +58,14 @@ const (
 	Panicked
 )
 
+//go:generate stringer -type=PanicPolicy
+const (
```
emilbon99, line 62: If we're going to make this an enum, it might be worth to fix `cancelOnExit` and `cancelOnExitErr` to be an enum as well (albeit probably a private one so we can keep the same options pattern). This could get rid of the strange double boolean and extra `||` checks.

x/go/signal/routine.go
emilbon99: Now that I think about it, this option is somewhat confusing. It makes me think that `WithMaxRestart` sets the number of times to restart when the go-routine exits with an error, not necessarily when the routine panics.

I also don’t think that `Restart` should be a panic policy. The final policy after the "last straw" could still be one of the panic policies (RecoverNoErr, RecoverErr). Why can't you just make the routine restart if `maxRestarts` is greater than 0, and not make `Restart` a panic policy?

emilbon99 (changes requested):

x/go/signal/routine.go
LeonLiur: This is a great idea

LeonLiur: `CancelOnExitErr` and `CancelOnExit` also used to panic, I will also make that early return.

x/go/signal/routine.go
LeonLiur: Thoughts on renaming `cancelOnExitErr` to `cancelOnFail`?

emilbon99: Sure. But you should make it an enum.

x/go/signal/routine.go
emilbon99: If you make `CancelOnExitErr` and `CancelOnExit` and enum then you won't need to worry about that

LeonLiur: Right

PR #715, opened 2024-07-13
[cesium] added length check for series when updating high watermark

Lham42:
# Fix Pull Request Template

## Key Information

- **Linear Issue**: [SY-964]()

## Description

Maximum 2-3 sentence description describing the changes.

Server would panic when disconnecting an NI device which is in use by a read task in the server.  Added a check to `UpdateHighWater` to check for series length before using it to set the watermark.

## Basic Readiness Checklist

- [ ] I have added sufficient regression tests to cover the changes.

## Manual QA Additions

### DIFF
--- a/cesium/writer_stream.go
+++ b/cesium/writer_stream.go
@@ -472,7 +472,15 @@ func (w *idxWriter) updateHighWater(s telem.Series) error {
 			s.DataType,
 		)
 	}
+	if s.Len() == 0 {
+		return errors.Wrapf(
+			validate.Error,
+			"series for channel %d length is zero",
+			w.idx.key,
+		)
+	}
 	w.idx.highWaterMark = telem.ValueAt[telem.TimeStamp](s, s.Len()-1)
+
 	return nil
 }
 
--- a/x/go/telem/series.go
+++ b/x/go/telem/series.go
@@ -54,9 +54,9 @@ func (s Series) Split() [][]byte {
 	return o
 }
 
-func ValueAt[T types.Numeric](a Series, i int64) T {
-	start := i * int64(a.DataType.Density())
-	end := (i + 1) * int64(a.DataType.Density())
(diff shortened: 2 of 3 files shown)


cesium/writer_stream.go
```diff
@@ -472,7 +472,15 @@ func (w *idxWriter) updateHighWater(s telem.Series) error {
 			s.DataType,
 		)
 	}
+	if s.Len() == 0 {
```
LeonLiur, line 475: Why should this error rather than doing nothing and just return?

x/go/telem/series.go
```diff
@@ -55,6 +55,8 @@ func (s Series) Split() [][]byte {
 }
 
 func ValueAt[T types.Numeric](a Series, i int64) T {
```
LeonLiur, line 57: I would rename the `a` variable to `s` for Series

LeonLiur (changes requested):

cesium/writer_stream.go
emilbon99: Agreed, we definitely shouldn't error here. If this series has length 0, then all other series in the frame must have length 0 per `validateWrite` so no data gets written. It honestly might be clearer to do this check just after the call to `validateWrite` on line 340 though. It feels more explicit that silently writing no data to all unary writers.

emilbon99: If we do choose to keep it here, we should add a test case that asserts its ok to write an empty series to a unary DB

emilbon99 (changes requested):
Would be good to add a regression test that asserts we don't panic anymore

cesium/writer_stream.go
Lham42: So if the series is empty do we still want to be able to write? Then would it make more sense just to add to the condition on line 351 to not update the watermark if the series is empty?

PR #731, opened 2024-07-18
Sy 877 server api middleware

LeonLiur:

## Key Information

- **Linear Issue**: [SY-878](sy-878-server-logic-for-setting-user-permissions)

## Description

API middleware & client libraries supporting user registration and policy management.

Please review and consider these three questions I had while in development:
- In our typescript client, the `auth` field can be undefined if the user did not provide username & password upon starting the client. Should this still be the case? Shouldn't we allow `auth` to always be an available service so the user can `register` even if they did not log in?
- Should register be a public route? This is fine because even if a user registers itself, it does not have the privilege to do anything else. However this might leave vulnerabilities like DDoS? (which we can address in other ways than making it a private endpoint).
- Minor security issue: If a user has channel CREATE privileges but not RETRIEVE, they can still RETRIEVE channels they created because of the client's channel cache. Is this worth addressing?

## Basic Readiness Checklist

- [ ] I have updated user facing documentation accordingly.

## Migrations

## Manual QA Additions

## Breaking Changes

/api/v1/auth/protected/change-password -> /api/v1/user/protected/change-password
/api/v1/auth/protected/change-username -> /api/v1/user/protected/change-username
/api/v1/auth/register -> /api/v1/user/register


client/py/synnax/ontology/id.py
```diff
@@ -0,0 +1,20 @@
...
+
+from pydantic import BaseModel
+
+
+class OntologyID(BaseModel):
+    key: str | None = ""
+    type: str | None = ""
+
+
+user_ontology_type = OntologyID(type="user")
```
emilbon99, line 18: I think these should go inside the respective modules for the types.

client/ts/package.json
```diff
@@ -27,12 +27,15 @@
     "@synnaxlabs/freighter": "workspace:*",
     "@synnaxlabs/x": "workspace:*",
     "async-mutex": "^0.5.0",
+    "nanoid": "^5.0.7",
+    "uuid": "^10.0.0",
```
emilbon99, line 31: Why do we need this as a dependency? Also, the "@synnaxlabs/x" package has an `id` module that we can use to generate unique IDs.

client/py/synnax/auth.py
```diff
@@ -100,6 +101,17 @@ def authenticate(self) -> None:
         self.token = res.token
         self.user = res.user
 
+    def register(self, username: str, password: str) -> UserPayload:
```
emilbon99, line 104: This feels like it probably belongs inside of the `user` module instead, so we access it with `client.users.register` or something along those lines.

synnax/cmd/start.go
```diff
@@ -160,6 +161,12 @@ func start(cmd *cobra.Command) {
 		if err != nil {
 			return err
 		}
+		accessSvc, err := rbac.NewService(rbac.Config{
```
emilbon99, line 164: Make this inline

client/py/synnax/ontology/__init__.py
emilbon99, line 1: Add a license header to this file

client/ts/src/auth/auth.ts
```diff
@@ -84,4 +85,18 @@ export class Client {
...
+
+  async register(username: string, password: string): Promise<user.Payload> {
+    const [res, err] = await this.client.send(
+      REGISTER_ENDPOINT,
+      { username: username, password: password },
+      insecureCredentialsZ,
+      tokenResponseZ,
+    );
+
+    if (err != null) {
```
emilbon99, line 97: There is a utility function in freighter called `sendRequired` us you should use instead of using the `err != null` checks

synnax/pkg/api/access.go
```diff
@@ -0,0 +1,120 @@
...
+	AccessRetrievePolicyResponse struct {
+		Policies []rbac.Policy `json:"policies" msgpack:"policies"`
+	}
+)
+
+func (a *AccessService) RetrievePolicy(
+	ctx context.Context,
+	req AccessRetrievePolicyRequest,
+) (res AccessRetrievePolicyResponse, err error) {
+	res.Policies = make([]rbac.Policy, 0)
```
emilbon99, line 104: Useless allocation. Just use nil.

emilbon99 (changes requested):
See my comments. Also, please do a review of all endpoints to make sure we're not accidentally leaking information by using named return values after checking access.

emilbon99:
Note that Pluto tests do not pass because the client does not compile due to type errors

client/py/synnax/auth.py
LeonLiur: We don't have a user module – in both the server and the client the user module is the same as the auth module

LeonLiur: Should we maybe just rename the `auth` module to `user`?

emilbon99: Well the `auth` service and the `user` service are different in the server, which they should be. Maybe instead we should split the `auth` API in the server?

synnax/pkg/api/access.go
LeonLiur: If this is nil then upon encoding as json it won't get encoded into an empty list

LeonLiur:
> Note that Pluto tests do not pass because the client does not compile due to type errors

addressed

emilbon99:
@LeonLiur Test still fail

emilbon99 (changes requested):

emilbon99 (approved):
LGTM

synnax/pkg/api/workspace.go
Lham42, line 1: Should we add tests for the error checks that codecov says arent convered right now? Are they going to be covered by some higher level tests?

synnax/pkg/distribution/channel/access.go
Lham42, line 1: what's the point in this file now?

synnax/pkg/distribution/channel/channel.go
```diff
@@ -153,6 +153,11 @@ func (k Keys) Unique() Keys { return lo.Uniq(k) }
 // followed by the keys that are absent in k.
 func (k Keys) Difference(other Keys) (Keys, Keys) { return lo.Difference(k, other) }
 
+// OntologyIDs returns the ontology.ID for each key.
+func (k Keys) OntologyIDs() []ontology.ID {
+	return lo.Map(k, func(key Key, _ int) ontology.ID { return ontology.ID{Type: OntologyType, Key: key.String()} })
```
Lham42, line 158: Is there a test for this?

synnax/pkg/distribution/channel/ontology.go
```diff
@@ -20,16 +21,22 @@ import (
...
+const OntologyType ontology.Type = "channel"
 
 // OntologyID returns a unique identifier for a Channel for use within a resource
 // ontology.
 func OntologyID(k Key) ontology.ID {
-	return ontology.ID{Type: ontologyType, Key: k.String()}
+	return ontology.ID{Type: OntologyType, Key: k.String()}
+}
+
+func OntologyIDsFromChannels(chs []Channel) []ontology.ID {
```
Lham42, line 32: test/codecov

synnax/pkg/distribution/ontology/schema/resource.go
```diff
@@ -56,6 +56,8 @@ func (id ID) String() string { return string(id.Type) + ":" + id.Key }
 // IsZero true if the ID is the zero value for its type.
 func (id ID) IsZero() bool { return id.Key == "" && id.Type == "" }
 
+func (id ID) IsType() bool { return id.Type != "" && id.Key == "" }
```
Lham42, line 59: test/codecov

synnax/pkg/hardware/device/ontology.go
```diff
@@ -27,6 +28,18 @@ func OntologyID(k string) ontology.ID {
 	return ontology.ID{Type: OntologyType, Key: k}
 }
 
+func OntologyIDsFromDevices(ds []Device) []ontology.ID {
```
Lham42, line 31: test/codecov

synnax/pkg/hardware/device/ontology.go
Lham42, line 1: test/codecov

synnax/pkg/hardware/rack/ontology.go
```diff
@@ -28,6 +29,18 @@ func OntologyID(k Key) ontology.ID {
 	return ontology.ID{Type: OntologyType, Key: k.String()}
 }
 
+func OntologyIDs(keys []Key) []ontology.ID {
```
Lham42, line 32: there is alot of repeated declarations of this function. Is this necessary or is there a better way to abstract this out?

Lham42: same for the other functions below

synnax/pkg/ranger/access.go
Lham42, line 1: Same as file earlier. Is this file here just for the package inclusion?

synnax/pkg/distribution/channel/access.go
LeonLiur: removed

synnax/pkg/ranger/access.go
LeonLiur: Removed

LeonLiur: deleted

synnax/pkg/api/workspace.go
LeonLiur: Not sure how codecov's diffing works, but there's some lines that it marks as uncovered that is definitely hit as an endpoint by client test

synnax/pkg/hardware/rack/ontology.go
LeonLiur: The function itself simply casts a uuid or a int type to a ontologyID type while attaching the type of the ID ("channel", "label", etc). Do you think it's better to have the caller provide both the type and the ID instead of calling a specific type's `OntologyID` method while providing just the ID?

synnax/pkg/distribution/channel/channel.go
LeonLiur: This function should be hit as part of client tests. We decided that it's really hard to test at the API level what client tests can already test, so I think much of the server straight up has 0 coverage which is why some of these are being flagged by codecov. I think it's  not worth to add a test now only for a function with very predictable behaviours when in the future we're gonna hit it with more systematic tests. What do you think?

synnax/pkg/hardware/rack/ontology.go
LeonLiur: I think the latter is more succinct, but it is true that that involves defining the same method in every package

synnax/pkg/distribution/channel/channel.go
Lham42: Understood. I can agree with this as long we make sure the client tests we make are comprehensive with this in mind. Ill resolve this here then.

synnax/pkg/hardware/rack/ontology.go
Lham42: hm good point. I think the latter is clearer and the additional abstraction doesn't create much benefit here.

Lham42 (approved):
LGTM

synnax/pkg/access/rbac/policy.go
```diff
@@ -10,31 +10,81 @@
...
+
+		for _, o := range policy.Objects {
+			if o.Type == AllowAll.Type {
+				// If the subject has an AllowAll policy, allow all requests.
+				return true
+			}
+			if o.IsType() {
+				// If an object applies to an entire type, then all requested objects
+				// of that type may be satisfied.
+				for requestedO, _ := range requestedObjects {
```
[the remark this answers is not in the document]
LeonLiur: Good catch

synnax/pkg/api/api.go
```diff
@@ -193,14 +193,18 @@ type Transport struct {
...
+	AccessRetrievePolicy freighter.UnaryServer[AccessRetrievePolicyRequest, AccessRetrievePolicyResponse]
 }
 
 // API wraps all implemented API services into a single container. Protocol-specific
 // API implementations should use this struct during instantiation.
 type API struct {
 	provider     Provider
 	config       Config
-	Auth         *AuthService
+	Auth         *UserService
```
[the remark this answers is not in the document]
LeonLiur: @emilbon99 suggested we separate `User` and `Auth` into two separate APIs and now I think that makes a lot of sense, let's do that then. I'll make sure to change all the client dependencies as well. `Auth` is just responsible of login and user is responsible of `change_username`, `change_password`, and `register`.

synnax/pkg/api/channel_test.go
```diff
@@ -1,242 +0,0 @@
-// Copyright 2023 Synnax Labs, Inc.
```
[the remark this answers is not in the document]
LeonLiur: We decided that we're not going to test at the API level since it's hard to setup and most of the endpoints are hit by client tests anyways. On a second thought I think putting this back in the codebase can't hurt so I'm going to reinstate these test cases.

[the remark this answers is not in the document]
LeonLiur: OK I remember why we took it out now – it was because it's hard to incorporate token & user info into a request made from the unit test so the tests were broken after auth was added. Going to remove these tests.

client/ts/src/user/client.ts
```diff
@@ -0,0 +1,38 @@
...
+      typeof insecureCredentialsZ,
+      typeof tokenResponseZ
+    >(
+      this.client,
+      REGISTER_ENDPOINT,
+      { username: username, password: password },
+      insecureCredentialsZ,
+      tokenResponseZ,
+    );
+
```
emilbon99, line 35: Remove this space

client/ts/src/user/client.ts
```diff
@@ -0,0 +1,38 @@
...
+export class Client {
+  private readonly client: UnaryClient;
+
+  constructor(client: UnaryClient) {
+    this.client = client;
+  }
+
+  async register(username: string, password: string): Promise<Payload> {
+    const { user: usr } = await sendRequired<
```
emilbon99, line 25: Renaming this variable only removes a single letter. Just return the original.

synnax/pkg/access/rbac/policy.go
emilbon99, line 1: Are the functions in this file covered by tests?

synnax/pkg/api/http/http.go
```diff
@@ -20,9 +20,14 @@ import (
...
-	t.AuthRegistration = fhttp.UnaryServer[api.RegistrationRequest, api.TokenResponse](router, false, "/api/v1/auth/register")
-	t.AuthChangePassword = fhttp.UnaryServer[api.ChangePasswordRequest, types.Nil](router, false, "/api/v1/auth/protected/change-password")
-	t.AuthChangeUsername = fhttp.UnaryServer[api.ChangeUsernameRequest, types.Nil](router, false, "/api/v1/auth/protected/change-username")
+
+	// USER
+	t.UserRegistrationOld = fhttp.UnaryServer[api.RegistrationRequest, api.TokenResponse](router, false, "/api/v1/auth/register")
+	t.UserChangePasswordOld = fhttp.UnaryServer[api.ChangePasswordRequest, types.Nil](router, false, "/api/v1/auth/protected/change-password")
+	t.UserChangeUsernameOld = fhttp.UnaryServer[api.ChangeUsernameRequest, types.Nil](router, false, "/api/v1/auth/protected/change-username")
+	t.UserRegistration = fhttp.UnaryServer[api.RegistrationRequest, api.TokenResponse](router, false, "/api/v1/user/register")
+	t.UserChangePassword = fhttp.UnaryServer[api.ChangePasswordRequest, types.Nil](router, false, "/api/v1/user/protected/change-password")
```
emilbon99, line 29: Put the common endpoints next to each other and add a comment explaining why we made this change.

synnax/pkg/api/user.go
```diff
@@ -0,0 +1,126 @@
...
+// RegistrationRequest is an API request to register a new user.
+type RegistrationRequest struct {
+	auth.InsecureCredentials
+}
+
+// Register registers a new user with the provided credentials. If successful, returns a
+// response containing a valid JWT along with the user's details.
+func (s *UserService) Register(ctx context.Context, req RegistrationRequest) (tr TokenResponse, err error) {
+	return tr, s.WithTx(ctx, func(txn gorp.Tx) error {
+		if err := s.authenticator.NewWriter(txn).Register(ctx, req.InsecureCredentials); err != nil {
```
emilbon99, line 50: Where are the access control checks for this method? Please make sure all relevant endpoints are access controlled, or we're running a security risk

emilbon99 (changes requested):
Address my comments. Also freighter test is failing - flaky or did the formatting tests mess something up?

synnax/pkg/access/rbac/policy.go
LeonLiur: sí

LeonLiur:
> Address my comments. Also freighter test is failing - flaky or did the formatting tests mess something up?

Seemed flakey, passing now

pjdotson (approved):

Lham42 (changes requested):
Is there functionality here we should also test in QA?

LeonLiur:
> Is there functionality here we should also test in QA?

Yes, we should probably test the ability to register new users

emilbon99:
@LeonLiur @Lham42 We'll probably add QA changes with @pjdotson goes through and implements the console stuff. The changes in this PR can all be easily unit tested.

client/py/tests/conftest.py
```diff
@@ -15,12 +15,15 @@
...
 
 @pytest.fixture(scope="session")
 def client() -> synnax.Synnax:
     return synnax.Synnax(
-        host="localhost",
-        port=9090,
+        host=HOST,
+        port=PORT,
         username="synnax",
         password="seldon",
```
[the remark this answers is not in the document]
LeonLiur: Addressed

Lham42 (approved):
LGTM

emilbon99 (approved):

PR #750, opened 2024-07-29
SY-1016 Guarding write/iterate/stream endpoints with access control

LeonLiur:

## Key Information

- **Linear Issue**: [SY-1016](https://linear.app/synnaxlabs/issue/SY-1016/[server]-guarding-writeiteratestream-endpoints-with-access-control)

## Description

Introduced access checks on framer methods - iterate, write, stream, and delete.

## Basic Readiness Checklist

- [ ] I have updated user facing documentation accordingly.
- [ ] I have verified code coverage targets are met.

## Migrations

##  Additional Notes
- [ ] These changes deal with concurrency.
- [ ] These changes affect UI.

## Manual QA Additions

## Reviewer Checklist
- [ ] Sufficient test coverage of new additions.
- [ ] Verified all steps in the Readiness checklists.
- [ ] UI changes have been tested.
- [ ] Style and formatting is consistent.
- [ ] Reviewed any relevant changes to concurrent code for safety. 
- [ ] Sufficient comments and clarity of code.
- [ ] Any additional documentation necessary for feature is provided and has been reviewed for clarity.


synnax/pkg/api/api.go
```diff
@@ -210,7 +210,7 @@ type API struct {
 	config       Config
 	Auth         *AuthService
 	User         *UserService
-	Telem        *FrameService
+	Framer       *FrameService
```
pjdotson, line 213: @emilbon99 Will changing the name to `Framer` from `Telem` cause any problems with incompatibility?

synnax/pkg/api/api.go
```diff
@@ -355,10 +355,10 @@ func (a *API) BindTo(t Transport) {
...
 
 	// FRAME
-	t.FrameWriter.BindHandler(a.Telem.Write)
-	t.FrameIterator.BindHandler(a.Telem.Iterate)
-	t.FrameStreamer.BindHandler(a.Telem.Stream)
-	t.FrameDelete.BindHandler(a.Telem.FrameDelete)
+	t.FrameWriter.BindHandler(a.Framer.Write)
+	t.FrameIterator.BindHandler(a.Framer.Iterate)
+	t.FrameStreamer.BindHandler(a.Framer.Stream)
+	t.FrameDelete.BindHandler(a.Framer.FrameDelete)
```
pjdotson, line 361: You should change the handler to `Delete` instead of `FrameDelete`.

pjdotson: Also, there's a bunch of variables and functions in `framer.go` that have names that start with `Frame` or `Framer` - those names should be changed.

client/py/synnax/framer/streamer.py
```diff
@@ -56,6 +56,9 @@ def __init__(
 
     def __open(self):
         self.__stream.send(_Request(keys=self.__adapter.keys, start=self.from_))
+        r, exc = self.__stream.receive()
```
pjdotson, line 59: Replace with `_, exc`.

client/py/synnax/framer/streamer.py
```diff
@@ -119,6 +122,9 @@ def __init__(
     async def open(self):
         self.__stream = await self.__client.stream(_ENDPOINT, _Request, _Response)
         await self.__stream.send(_Request(keys=self.__adapter.keys, start=self.from_))
+        _, err = await self.__stream.receive()
```
pjdotson, line 125: `exc` not `err`.

client/py/tests/test_access.py
```diff
@@ -12,7 +12,7 @@
 
 import synnax as sy
 from synnax.channel.client import channel_ontology_type
-from synnax.ontology.id import OntologyID
+from synnax.ontology import OntologyID
```
pjdotson, line 15: Have you checked that this pattern is matched now in schematic, range, line plot, etc.?

client/py/synnax/framer/streamer.py
```diff
@@ -32,7 +32,7 @@ class _Request(Payload):
 
 class _Response(Payload):
     frame: FramePayload
-    error: ExceptionPayload | None
+    error: ExceptionPayload | str | None
```
pjdotson, line 35: Is there a reason that it could be a string here?

pjdotson (changes requested):

Lham42:
To address your comments

1. I think it makes sense to decouple the framer and channel client 
2. No action/return is sufficient ack for success so I'd make sure you return error that can be checked rather than an ack
3. Once this pr has been approved, documentation can be updated to have a section on this for py and ts clients. The rest of the docs can be held off til UI is done.

pjdotson:
On your comments:
1. @emilbon99 Do you have an idea on a better solution? This solves this problem but I think the problem is that channels are unique amongst all of our resources. I don't love the idea of making an ontology id for this but it seems preferable to adding separate actions for reading / updating data.
2. Shouldn't you send an error instead of an ack if a streamer failed to create? I feel like thats the more idiomatic solution in Go / matches the rest of the codebase better.
3. Yes lets add something on both Python / TypeScript client.

synnax/pkg/api/api.go
LeonLiur: This file is in `package api` and not `package framer`, so a lot of these variables do need to be prefixed with `frame` since otherwise they collide with other names in the `api` package, like channel delete.

client/py/synnax/framer/streamer.py
LeonLiur: gotcha, also changed other occurances

client/py/tests/test_access.py
LeonLiur: yep

client/py/synnax/framer/streamer.py
LeonLiur: no, you're right, removed.

LeonLiur:
> On your comments:
> 
> 1. @emilbon99 Do you have an idea on a better solution? This solves this problem but I think the problem is that channels are unique amongst all of our resources. I don't love the idea of making an ontology id for this but it seems preferable to adding separate actions for reading / updating data.
> 2. Shouldn't you send an error instead of an ack if a streamer failed to create? I feel like thats the more idiomatic solution in Go / matches the rest of the codebase better.
> 3. Yes lets add something on both Python / TypeScript client.

2. Yes, if there's an error, an empty frame + an error is sent; if there's no error, an empty frame + no error is sent. this way the `receive()` on the client side does not block.

Lham42 (changes requested):

client/py/synnax/access/payload.py
emilbon99, line 1: Do we still need the ALLOW_ALL here?

client/py/synnax/channel/client.py
emilbon99, line 1: The formatting on these elipses is very strange

client/py/synnax/framer/writer.py
emilbon99, line 1: Weird formatting on elipses

client/ts/src/access/policy/payload.ts
emilbon99, line 1: Do we still need the ALLOW_ALL type?

emilbon99 (approved):
LGTM once these comments are addressed.

PR #757, opened 2024-08-03
SY-1053 New API Reference Generator for TypeScript

pjdotson:
Added a workflow that allows for an automatically generated client.api.md file to build while running TypeScript tests. Viewing diffs to this file will let us know how the client API changes.

Questions I want addressed:
1. Right now this is all covered in the `test.client.yaml` GitHub action. Should we move this to a separate file?
2. It would be pretty trivial to add another set of changes + workflows for freighter/ts, x/ts, alamos/ts, and pluto - should we add them?
3. Right now a workflow won't fail if a the API changes - I can change one of the flags when the command is run and we will get a failure. I wanted to know if we are good with reviewing based off of the file diffs or would rather have an explicit action fail.


Lham42:
1.  I think its fine to keep it in the test client
2.  if it is pretty fast I'd add them to the other ts libs
3.  I think causing the checks to fail is probably ideal so we know immediately of we're making api breaking changes

Also can you add to the readme about how togen/view this?

emilbon99:
1. I would move this into a separate workflow that only runs when we're merging the RC into main. This workflow should fail if the API is incompatible.
2. No. Pluto is in beta and the other APIs are internal. I don't think it's worth it right now.
3. See #1

emilbon99 (changes requested):

pjdotson:
@Lham42 @emilbon99 I'll add a README on how to generate the API in the next commit, but I changed the CI action to only check if the API changes instead of generating it. So development steps would now be:

1. Developer makes a change to the API.
2. Developer runs `genApi:client` in root directory, the `client.api.md` file changes. Push changes to remote branch
3. Changes are viewable in PR, so now reviewers can look at the diffs in the `client.api.md` file.
4. CI runs `checkApi:client`: This only fails if dev didn't do step 2.

PR #781, opened 2024-08-19
SY-1096 Update Client API GitHub Action to Pass on Back Merges

pjdotson:
This PR changes the client API GitHub action to pass when there are API changes in the base branch.

### DIFF
--- a/.github/workflows/api.client.yaml
+++ b/.github/workflows/api.client.yaml
@@ -1,4 +1,4 @@
-name: "Check API"
+name: "Client API"
 on:
   pull_request:
     branches:
--- a/.github/workflows/api.client.yaml
+++ b/.github/workflows/api.client.yaml
@@ -17,6 +17,8 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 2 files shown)


Lham42:
So to clarify - will the check pass only once we backmerge the base branch which has API changes?

Lham42 (changes requested):
Can you give an example of the process that would happen now with the change to the workflow because it is not immediately clear to me now. Does this workflow now automatically backmerge the base branch  for an open pr too?

pjdotson:
Steps:
1. Checks for diffs in any package that could cause an API change (x, alamos, freighter, client)
2. Checks out the repo
3. Fetches the base branch (if PR is for merging into rc, it fetches rc).
4. On the local repo, merges the base branch (eg rc) into the head branch (the hash that the PR is opened on).
5. checks if the API has changed on the local repo, compared to the existing `client.api.md` document. If so, this workflow fails, if not, the workflow passes.

It does not affect the code on the remote branches because it never pushes code to the repo. It doesn't automatically backmerge the base branch for an open PR.

Lham42 (approved):
LGTM

PR #813, opened 2024-09-06
SY-1098 Add Ability to Resize Most Schematic Symbols

pjdotson:

## Key Information

- **Linear Issue**: [SY-1098](https://linear.app/synnax/issue/SY-1098/add-ability-to-resize-any-schematic-symbol)

## Description

Added ability to resize (almost) any schematic symbols - buttons, switches, values, and setpoints still cannot be resized. Also refactored some code in relevant files.

NOTE: Previous versions of the schematic will have a `scale` of `null`, so the options for scaling them will not show up. I don't know a great way to fix this problem, and don't think it matters too much - all it means is that the elements created before this version gets released cannot be scaled. If someone wants to scale that element, they could delete it and replace it with a new one.

## Basic Readiness Checklist

- [x] I have verified code coverage targets are met.

## Migrations

##  Additional Notes
- [ ] These changes deal with concurrency.
- [ ] These changes affect UI.

## Manual QA Additions

## Breaking Changes

## Reviwer Checklist
- [ ] Sufficient test coverage of new additions.
- [ ] Verified all steps in the Readiness checklists.
- [ ] UI changes have been tested.
- [ ] Style and formatting is consistent.
- [ ] Reviewed any relevant changes to concurrent code for safety. 
- [ ] Sufficient comments and clarity of code.
- [ ] Any additional documentation necessary for feature is provided and has been reviewed for clarity.

### DIFF
--- a/docs/site/src/pages/reference/console/schematics.mdx
+++ b/docs/site/src/pages/reference/console/schematics.mdx
@@ -4,6 +4,7 @@ title: "Schematics"
 heading: "Schematics"
 description: "Control hardware using schematics."
 ---
+
 import { Icon } from "@synnaxlabs/media";
 import { Divider } from "@synnaxlabs/pluto";
 import { Image, Video } from "@/components/Media";
@@ -13,7 +14,7 @@ A schematic is used to see a diagram of your physical or simulated system. It di
 live sensor data and can be used to exercise manual control by changing the values of
 channels tied to actuators. We've added a simple example of a schematic below:
 
-<Image client:only="react" id="console/schematics/example"  />
+<Image client:only="react" id="console/schematics/example" />
 
 <Divider.Divider direction="x" />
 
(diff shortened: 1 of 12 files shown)


pjdotson:
Leaving a comment here - I need to check how migrating old symbols into this works. I believe what will happen is that the scale will not show up on the form because it is null for old symbols.

emilbon99:
This generally looks ok. I think we need to make some other changes for this feature to be nice though. Label positions look strange when elements are larger. Take this rotary mixer for example. It's probably pretty common to want to make this thing pretty large. It looks strange to fill in the entire mixer when it is turned on. We should probably make it so you hook up a switch to turn the mixer on and off, then we can make it possible to hide and show the controls states on it, so then it becomes more of a 'static' element.

Also, if you can scale the element, you should also have the option to make the stroke width of the symbol larger or smaller.

Screenshot 2024-09-07 at 8 16 08 PM

emilbon99:
Screenshot 2024-09-07 at 8 17 52 PM

Just looks kinda strange to have this massive, white rotary mixer on the screen

emilbon99:
Even though you would probably want to have a rotary mixer that big on your P&ID

emilbon99 (approved):
Approved, although I think we need to spend more time thinking about the larger implications of making sure this feature has a nice UX

PR #888, opened 2024-11-04
SY-1374 Fix Windows Channel Deletion Bug

LeonLiur:

## Key Information

- **Linear Issue**: [SY-#](https://linear.app/synnax/issue/)

## Description

Fixed the following bugs that were failing the Windows CI since **deletion of a file to which a handle is open is allowed on Unix, but not Windows**.

1. index persist is not closed upon unary server close
2. after an unsuccessful unaryDB close, the db is marked as close even though the close failed, leading it be marked as closed although file handles are still open.
3. test cases asserting on opening a pre-existing DB in an FS does not close the DB.
4. windows implementation of dirInfo does not return size correctly
5. test in channel_test and testutil_test forgot to close file handle
6. test in unary_test and index_test forgot to close DB
7. when opening cesium DB fails, meta file handle was never closed.

## Basic Readiness

## Backwards Compatibility

### Data Structures

### API Changes

  - [ ] C++
  - [ ] TypeScript
  - [ ] Python

### Breaking Changes


cesium/db.go
```diff
@@ -103,7 +103,7 @@ func (db *DB) Read(ctx context.Context, tr telem.TimeRange, keys ...core.Channel
 // database, a deadlock is caused since the signal context is closed while the writers
 // attempt to send to relay.
 func (db *DB) Close() error {
-	if !db.closed.CompareAndSwap(false, true) {
+	if db.closed.Load() {
```
emilbon99, line 106: This is sussy. Just loading the closed state here could cause a race condition if close is called concurrently with another DB method

cesium/internal/domain/db.go
```diff
@@ -192,7 +192,7 @@ func (db *DB) HasDataFor(ctx context.Context, tr telem.TimeRange) (bool, error)
 
 // Close closes the DB. Close should not be called concurrently with any other DB methods.
 func (db *DB) Close() error {
-	if !db.closed.CompareAndSwap(false, true) {
+	if db.closed.Load() {
```
emilbon99, line 195: See my comment in the main cesium DB close method

emilbon99 (changes requested):

cesium/db.go
LeonLiur: resolved by first doing compare and swap then storing false if close failed.

.github/workflows/test.cesium.yaml
```diff
@@ -1,4 +1,5 @@
-name: "Test - Cesium"
+name: "Test: Cesium"
```
Lham42, line 1: hyphen

.github/workflows/test.x.yaml
```diff
@@ -1,4 +1,5 @@
-name: "Test - X"
+name: "Test: X"
```
Lham42, line 1: hyphen

Lham42 (changes requested):

pjdotson (approved):

cesium/db.go
Lham42: Isnt this still an issue? why is this not  a compare and swap rather than storing true at the end. I assume the intention of doing this is so that if a step later on fails, the db isnt still marked as closed. But doesn't this introduce the issue of another thread interpereting the db as open while its in the process of getting closed?

Lham42 (changes requested):

cesium/db.go
LeonLiur: No – during the process of the db being closed, the `CompareAndSwap` call makes it so that `closed` is `true` – meaning other operations are not allowed. If the close failed, then those operations are allowed again

cesium/internal/unary/db.go
```diff
@@ -151,11 +151,22 @@ func (db *DB) Read(ctx context.Context, tr telem.TimeRange) (frame core.Frame, e
...
+// still be marked closed and no read/write operations are allowed on it to protect
+// data integrity.
 func (db *DB) Close() error {
 	if !db.closed.CompareAndSwap(false, true) {
 		return nil
 	}
-	return db.wrapError(db.domain.Close())
+	err := db.domain.Close()
+	if err != nil {
+		if errors.Is(err, domain.ErrOpenEntity) {
```
emilbon99, line 163: It would be good to add a comment explaining why you do this here. It's not obvious

emilbon99 (approved):
Approved after my comments are addressed

cesium/db.go
```diff
@@ -122,5 +122,10 @@ func (db *DB) Close() error {
 	for _, u := range db.unaryDBs {
 		c.Exec(u.Close)
 	}
-	return c.Error()
+	if c.Error() != nil {
+		db.closed.Store(true)
```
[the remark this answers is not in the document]
LeonLiur: Outdated diff – resolved.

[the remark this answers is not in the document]
LeonLiur: For the cesium DB, closed is always swapped to `true` when `Close` is called – this is because even if one unaryDB failed because of open handles, we should still stop serving read/write requests since one of the channels in index/data pairs may be closed while the other is not.

cesium/internal/domain/db.go
```diff
@@ -204,5 +204,10 @@ func (db *DB) Close() error {
 	w := errors.NewCatcher(errors.WithAggregation())
 	w.Exec(db.fc.close)
 	w.Exec(db.idx.close)
-	return w.Error()
+	if w.Error() != nil {
+		db.closed.Store(false)
```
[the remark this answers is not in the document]
LeonLiur: This is outdated after our discussion – we only store `false` in `closed` if the close failed because of open entities

emilbon99 (approved):

Lham42 (approved):

PR #902, opened 2024-11-07
[pluto] - improved value font sizing and spacing

emilbon99:

## Key Information

- **Linear Issue**: [SY-1440](https://linear.app/synnax/issue/SY-1440/switch-schematic-values-to-monospaced-font)

## Description

Switches to use a monospaced font on schematic values, makes the "Small" font option slightly larger, and makes it so that values that switch between - and + frequently don't shift their digits.

## Basic Readiness

- [x] I have needed QA steps to the [release

## Backwards Compatibility

### Data Structures

### API Changes

  - [ ] C++
  - [ ] TypeScript
  - [ ] Python

### Breaking Changes

### DIFF
--- a/pluto/src/text/SelectLevel.tsx
+++ b/pluto/src/text/SelectLevel.tsx
@@ -35,7 +35,7 @@ const DATA: Entry[] = [
     label: "M",
   },
   {
-    key: "p",
+    key: "h5",
     label: "S",
   },
   {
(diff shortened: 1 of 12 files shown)


pjdotson (changes requested):
The value does not have enough padding on the right side when it is positive. If you look closely the second zero actually overflows into the units:
image

emilbon99:
Fixed

pjdotson (changes requested):
1. You should merge the value text size and units text size into one, otherwise we get monstrosities like this:
image
2. It looks weird that the size of the values box changes when you just remove a negative sign:
https://github.com/user-attachments/assets/0c7945af-3ebc-479c-9d5a-577782e27186

pjdotson (changes requested):
Theres still a problem with the width changing when going from plus to minus. To replicate it, put a default value of 0, and then put in 100, and then put in -100.

https://github.com/user-attachments/assets/65951ef8-ba74-478f-92bc-6459f57c4b05

pjdotson (approved):

PR #903, opened 2024-11-07
[client/py] - added type coersion for primitive non-string values in …

emilbon99:
…range KV

## Key Information

- **Linear Issue**: [SY-1452](https://linear.app/synnax/issue/SY-1452/add-auto-conversion-to-string-for-metadata-in-python)

## Description

Adds type coercion in the Python client for non-string values on the range key-value store. 

## Basic Readiness

## Backwards Compatibility

### Data Structures

### API Changes

  - [ ] C++
  - [ ] TypeScript
  - [ ] Python

### Breaking Changes

### DIFF
--- a/client/py/synnax/ranger/kv.py
+++ b/client/py/synnax/ranger/kv.py
@@ -12,13 +12,26 @@ import uuid
 from typing import overload
 from freighter import Payload, UnaryClient, send_required
 
+from synnax import ValidationError
 from synnax.util.normalize import normalize
+from synnax.util.primitive import Primitive, is_primitive
 
 
 class KVPair(Payload):
     range: uuid.UUID
     key: str
     value: str
+
+    def __init__(self, **kwargs):
+        value = kwargs.get("value")
+        if not isinstance(kwargs["value"], str):
+            if not is_primitive(value):
+                raise ValidationError(f"""
+                Synnax has no way of casting {value} to a string when setting meta-data
+                on a range. Please convert the value to a string before setting it.
+                """)
+        kwargs["value"] = str(value)
+        super().__init__(**kwargs)
 
 
 class _GetRequest(Payload):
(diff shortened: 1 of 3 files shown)


client/py/synnax/ranger/kv.py
```diff
@@ -12,14 +12,27 @@
...
+from synnax.util.primitive import Primitive, is_primitive
 
 
 class KVPair(Payload):
     range: uuid.UUID
     key: str
     value: str
 
+    def __init__(self, **kwargs):
+        value = kwargs.get("value")
```
pjdotson, line 26: Isn't the better solution here to check if an object has a `__str__()` method instead? No need to restrict ourselves to primitives.

pjdotson: Or call the `str` function and catch the exception and throw our own

emilbon99: The problem is that all python objects have a __str__ method. But classes have a representation like <__main__.MyClass object at 0x000001AB2C8C7A90>, which I don't think we should allow

pjdotson: You could check if they use the default object, ChatGPT gives something like this:

```python
def safe_str(obj):
    # Check if the object's __str__ method is the default one from `object`
    if type(obj).__str__ == object.__str__:
        raise TypeError(f"{type(obj).__name__} has not overridden the default __str__ method.")
    return str(obj)

# Example usage:
class MyClass1:
    pass

class MyClass2:
    def __str__(self):
        return "This is a custom string representation."

obj1 = MyClass1()
obj2 = MyClass2()

try:
    print(safe_str(obj1))
except TypeError as e:
    print(f"Error: {e}")

try:
    print(safe_str(obj2))
except TypeError as e:
    print(f"Error: {e}")
```

emilbon99: Yes we could do this, but what about custom stringers from imported libraries? Should we not error or warn on those?

Don't have a tremendously strong opinion here

emilbon99: For example, the `WebSocket` constructor has a custom stringer that overrides the default one - is that acceptable? I guess so

pjdotson: Yeah I think it is, that case seems relatively rarer and this is generally the better solution. This is kind of the "be liberal in what you accept, conservative in what you output" principle.

emilbon99: Ok, will change

pjdotson (approved):

PR #904, opened 2024-11-08
[HOTFIX]  SY 1460 labjack ux improvements

Lham42:

## Key Information

- **Linear Issue**: [SY-1441](https://linear.app/synnax/issue/SY-1441)

## Description

This PR makes several console and driver side improvements to setting up labjack tasks as well as some other minor console improvements. This has been labeled as a hotfix to make it clear it is being directly merged into main.

## Basic Readiness

## Backwards Compatibility

### Data Structures

### API Changes

  - [ ] C++
  - [ ] TypeScript
  - [ ] Python

### Breaking Changes


driver/labjack/reader_source.cpp
```diff
@@ -13,6 +13,37 @@
...
+            .base_interval = 1 * SECOND,
+            .max_retries = 20,
+            .scale = 1.2,
+    };
+    this->breaker = breaker::Breaker(breaker_config);
+    this->open_device();
+}
+
+void labjack::ReaderSource::open_device(){
+    std::lock_guard<std::mutex> lock(labjack::device_mutex);
```
emilbon99, line 34: This function is very similar to the one in the scanner. Why not consolidate a single function that accepts a serial number and adds it to the map, and then call that function from both the scanner and reader source?

Also, what about the writer? Does that need to be opened?

emilbon99: I also don't think we should use a global map here. Why not build a class called "DeviceManager" that is a field on the factory, and that you pass into the constructor of both the scanner and reader tasks. Dependency injection is much clearer than using globals

emilbon99 (changes requested):

driver/labjack/reader_source.cpp
Lham42: i was lazy

Lham42: i created device manager class

driver/main.cpp
```diff
@@ -7,6 +7,20 @@
...
+#ifdef _WIN32
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+
+#include <winsock2.h>
```
emilbon99, line 19: What are these for?

emilbon99 (approved):
Approved by me. Can you please format he C++ code?

driver/main.cpp
Lham42: just copying what i said in slack:

 the new stuff in main.cpp is because once i added a device manager to the other files, it for some reason links things in a way that causes an error because of symbols included in both windows and winsock2.h . its a known issue and the accepted solution is to add those ifdefs and explicitly define those libraries in that exact order
 
 
image

PR #989, opened 2024-12-16
Synnax v0.36.0 Release Notes

pjdotson:
Linear Issue: [SY-1711](https://linear.app/synnax/issue/SY-1711/synnax-v036-release-notes)

### DIFF
--- a/docs/site/src/pages/releases/0-36-0.mdx
+++ b/docs/site/src/pages/releases/0-36-0.mdx
@@ -1,0 +1,68 @@
+import Release from "@/components/releases/Release.astro";
+import { Image, Video } from "@/components/Media";
+import Diagram from "@/components/Diagram.astro";
+import LabJack from "@/pages/releases/LabJack.astro";
+
+<Release
+    version="0.36.0"
+    date="Dec 16, 2024"
+    title="Line Plot Annotation Tool"
+    breakingChanges
+>
+
+Synnax v0.36.0 introduces several new quality of life improvements and bug fixes,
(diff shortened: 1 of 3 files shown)


Lham42 (changes requested):
can this image be made smaller to focus on the annotations + graph

image

Lham42 (changes requested):
Can we also make 

"console versions 0.36.0 and newer will not work with server versions of 0.35.0 and older" be what's bolded in the breaking changes instead so it isn't missed.

Lham42 (changes requested):
image

can you just make this two sections like

Refactored:
  - ...
 
Bug fixes:
 - ...

Lham42 (changes requested):
incorrect title

image

Edit: i guess we did that in the last release too. But "Summary" seems like its make more sense

docs/site/src/pages/releases/0-36-0.mdx
```diff
@@ -0,0 +1,78 @@
...
+and console use to communicate. This change means that **console versions 0.36.0 and
+newer** will not work with **server versions of 0.35.0 and older**. To avoid issues,
+please ensure that your server and console are updated to the same version.
+
+### Other Improvements and Bug Fixes
+
+Improvements:
+
+- Sort child ranges of a range by start time.
+- Add copy link to task **toolbar** context menu.
```
Lham42, line 61: does this need to be bolded?

Lham42 (approved):
approved but one comment

PR #996, opened 2024-12-19
Sy 1724 fix fgrpc race condition/deadlock

Lham42:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1724](https://linear.app/synnax/issue/1724)

## Description

Originally an issue was found when the driver crashed if you very quickly started and stopped an OPC read task. I don't think this is the correct solution, but the crashing issue was resolved by removing the double inheritance happening with the stream class.  This, however, required me to call the middleware chain in this class no longer, which can pose other issues. This then led to a deadlock issue which was resolved by atomically setting the writes_done_called and the closed variables. 

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client

### DIFF
--- a/freighter/cpp/fgrpc/BUILD.bazel
+++ b/freighter/cpp/fgrpc/BUILD.bazel
@@ -10,7 +10,6 @@ cc_library(
     deps = [
         "//freighter/cpp:freighter",
         "@com_github_grpc_grpc//:grpc++",
-        "@com_github_google_glog//:glog",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


pjdotson (changes requested):
If merging into `main` can you bump the server version? Otherwise merge into `rc`.

emilbon99:
This PR should not be merged. I'll take a look at it

PR #1036, opened 2024-12-31
Sy 1800 lower contrast threshold to trigger schematic color change

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1767](https://linear.app/synnax/issue/SY-1800/lower-contrast-threshold-to-trigger-schematic-color-change)

## Description

When we change the theme from light to dark or vice versa, we need to change some of the low contrast schematic components to white or black. Previously we just used a contrast threshold to do this, but this changed elements that were yellow or pink in color. Now we do a 'grayness' check instead that ensures we only change grays.

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client

### DIFF
--- a/console/src/schematic/slice.ts
+++ b/console/src/schematic/slice.ts
@@ -401,23 +401,20 @@ export const { actions, reducer } = createSlice({
     fixThemeContrast: (state, { payload }: PayloadAction<FixThemeContrastPayload>) => {
       const { theme } = payload;
       const bgColor = new Color.Color(theme.colors.gray.l0);
+      const shouldChange = (crude: Color.Crude): boolean => {
+        const c = new Color.Color(crude);
+        return c.grayness > 0.85 && c.contrast(bgColor) < 1.3;
+      };
       Object.values(state.schematics).forEach((schematic) => {
         const { nodes, edges, props } = schematic;
         nodes.forEach((node) => {
           const nodeProps = props[node.key];
-          if ("color" in nodeProps) {
-            const c = new Color.Color(nodeProps.color as string);
-            // check the contrast of the color
-            if (c.contrast(bgColor) < 1.1)
-              // if the contrast is too low, change the color to the contrast color
+          if ("color" in nodeProps)
+            if (shouldChange(nodeProps.color as string))
               nodeProps.color = theme.colors.gray.l9;
-          }
         });
         edges.forEach((edge) => {
-          if (
-            edge.color != null &&
-            new Color.Color(edge.color as string).contrast(bgColor) < 1.1
-          )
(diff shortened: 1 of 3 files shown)


pluto/src/color/core/color.spec.ts
```diff
@@ -133,4 +133,21 @@ describe("color.Color", () => {
       expect(c.pickByContrast(c1, c2)).toEqual(c1);
     });
   });
+  describe("grayness", () => {
+    const tests: Array<[string, number]> = [
+      ["#000000", 0],
+      ["#ffffff", 1],
```
pjdotson, line 139: Does not make any sense to me why black is considered complete grayness and white is considered no grayness.

console/src/schematic/slice.ts
```diff
@@ -401,23 +401,20 @@ export const { actions, reducer } = createSlice({
     fixThemeContrast: (state, { payload }: PayloadAction<FixThemeContrastPayload>) => {
       const { theme } = payload;
       const bgColor = new Color.Color(theme.colors.gray.l0);
+      const shouldChange = (crude: Color.Crude): boolean => {
+        const c = new Color.Color(crude);
+        return c.grayness > 0.85 && c.contrast(bgColor) < 1.3;
```
pjdotson, line 406: Does this formula work well when going from dark to light instead of just light to dark?

pluto/src/color/core/color.ts
```diff
@@ -193,6 +193,17 @@ export class Color {
     return Number((0.2126 * r + 0.7152 * g + 0.0722 * b).toFixed(3));
   }
 
+  /**
+   * @returns the grayness of the color, between 0 and 1.
+   */
+  get grayness(): number {
```
pjdotson, line 199: I just am not sure that `grayness` makes sense here, since it takes into account the brightness as well. You could probably get a good measure of grayness with a formula like this:

avg = (r + g + b)/3
grayness = sqrt((r-avg)^2 + (g-avg)^2 + (b-avg)^2)

pjdotson (changes requested):

pluto/src/color/core/color.ts
emilbon99: Problem is that this makes yellow super gray

emilbon99: Updated

console/src/schematic/slice.ts
emilbon99: Updated grayness calc

pjdotson (approved):

PR #1041, opened 2025-01-05
SY-166 Key-Value State Forwarding

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-166](https://linear.app/synnax/issue/SY-166/key-value-state-forwarding)

## Description

Adds experimental support for multi-node clusters.

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client


x/go/kv/tx_test.go
```diff
@@ -0,0 +1,38 @@
+package kv_test
```
pjdotson, line 1: Add header

freighter/go/fgrpc/context.go
```diff
@@ -35,25 +32,25 @@ func parseContext(
...
 	grpcMD, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
-		zap.L().DPanic(
-			"failed to get metadata from context",
-			zap.String("service", serviceName),
-		)
+		//zap.L().DPanic(
+		//	"failed to get metadata from context",
+		//	zap.String("service", serviceName),
+		//)
```
pjdotson, line 53: If you are going to delete it delete it. Why did you comment all of this out?

pjdotson (changes requested):
Why are all of the protobuf generated files different now? And you also need to fix migration tests

emilbon99:
> Why are all of the protobuf generated files different now? And you also need to fix migration tests

Because I made changes to the distribution layer RPCs, and re-generating with a new GPRC version causes all the PBs to change.

client/py/synnax/control/client.py
```diff
@@ -7,7 +7,7 @@
 #  License, use of this software will be governed by the Apache License, Version 2.0,
 #  included in the file licenses/APL.txt.
 
-from synnax.channel.payload import ChannelParams
+from synnax.channel.payload import ChannelParams, ChannelKey, ChannelName
```
pjdotson, line 10: What was the point of this?

client/py/examples/control/scheduled/simulated_daq.py
```diff
@@ -0,0 +1,102 @@
...
+
+"""
+This example sets up a simulated data acquisition system (DAQ) that has a press valve
+(ox_mpv_cmd) and vent valve (ox_pre_vlv_cmd) connected to a pressure transducer
+(press_pt). When the press valve is open, the pressure increases, and when the vent
+valve is open, the pressure decreases. The pressure is written to the press_pt channel.
+
+This script can be run in conjuction with the `control_sequence.py` script to
+demonstrate how a control sequence can be written in Synnax.
+"""
```
pjdotson, line 18: This comment does not make sense here (what is `control_sequence.py`). And did someone ask for us to show scheduled
commands in autosequences? If not I think we should either delete these two files or
move them into the dev folder until we have the actual implementation.

.github/workflows/test.migration.yaml
```diff
@@ -78,7 +79,11 @@ jobs:
         working-directory: client/py
 
       - name: Stop Main Server
-        run: docker stop synnax-main && docker rm synnax-main
+        if: always() && steps.start_main.outcome == 'success'
```
pjdotson, line 82: Why do you need the `always()` here; isn't `steps.start_main.outcome == 'success'`
enough? And how could we get here if that `steps.start_main.outcome` failed, wouldn't
this test exit at that point?

.github/workflows/test.migration.yaml
```diff
@@ -122,4 +128,8 @@ jobs:
         working-directory: client/py
 
       - name: Stop Current Server
-        run: docker stop synnax-current && docker rm synnax-current
+        if: always() && steps.start_current.outcome == 'success'
+        run: |
+          docker logs synnax-current
+          docker stop synnax-current
+          docker rm synnax-current
```
pjdotson, line 135: Same comments as above

client/go/go.mod
pjdotson, line 1: How are there no dependencies to this file? And you need to figure out how we are
going to publish the go client, and what other go packages we will be publishing because
of that.

client/go/cmd/gen/main.go
pjdotson, line 1: Looks like this could be moved to the x module

pjdotson (changes requested):
1. Key-value state forwarding is good after comments are all addressed.
2. For the Go client, there are a few issues:
- Needs to be much much more documentation on public functions and data types. It is a public API after all.
- Need to figure out how to publish it and add a `deploy.go.yaml` file to CI.
- Need to add the go tests to the `test.client.yaml` CI file.
- Website documentation should also include pages on reading and streaming data.

client/py/synnax/control/client.py
emilbon99: There wasn't one. Removed

client/py/examples/control/scheduled/simulated_daq.py
emilbon99: Removed

.github/workflows/test.migration.yaml
emilbon99: In some actions, you might see some clean up jobs prefixed with `Post` that ensure that some resources are cleaned up before the action exists. `always()` ensures that a step runs even if a previous step fails, so that we can do cleanup. In this case, we should always do cleanup if the start command succeeded. This makes sure that containers are shut down properly, and we can view the logs (as was very useful in debugging the comments from your previous review).

.github/workflows/test.migration.yaml
emilbon99: See my reply aboce

client/go/go.mod
emilbon99: Deleted go client

client/go/cmd/gen/main.go
emilbon99: Deleted

aspen/internal/kv/kv.go
```diff
@@ -109,7 +109,7 @@ const (
 	operationReceiverAddr = "opReceiver"
 	feedbackSenderAddr    = "feedbackSender"
 	feedbackReceiverAddr  = "feedbackReceiver"
-	recoveryTransformAddr = "recoveryTransform"
+	recoveryTransformAddr = "gossipRecoveryTransform"
```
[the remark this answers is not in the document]
emilbon99: Renamed it back to what it was. Find and replace mistake

pjdotson (approved):
Approved after last open comments on PR are addressed.

freighter/go/fgrpc/stream.go
```diff
@@ -106,18 +106,21 @@ func (s *StreamClientCore[RQ, RQT, RS, RST]) Stream(
...
 			}
 			grpcClient, err := s.ClientFunc(ctx, conn.ClientConn)
 			stream = s.adaptStream(grpcClient)
-			return parseContext(
-				ctx,
-				s.ServiceDesc.ServiceName,
-				freighter.Client,
-				freighter.Stream,
-			), err
+			return freighter.Context{
```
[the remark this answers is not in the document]
emilbon99: You're not parsing anything. It's just contstructing the struct from scratch

PR #1044, opened 2025-01-08
SY-1193 Import / Export Workspaces

pjdotson:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1193](https://linear.app/synnax/issue/SY-1193/import-export-workspaces)

## Description

Added support for importing and exporting workspaces in the console. Also created

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client


console/src/export/extractor.ts
```diff
@@ -0,0 +1,26 @@
...
+  store: Store;
+  client: Synnax | null;
+}
+
+export interface ExtractorReturn {
+  data: string;
+  name: string;
+}
+
+export type Extractor = (
```
emilbon99, line 23: Would be good to add a quick command explaining what this does. Was able to back it out, but took me a sec

console/src/layout/selectors.ts
```diff
@@ -234,8 +234,10 @@ export const selectHauling = (state: StoreState): Haul.DraggingState =>
 export const useSelectHauling = (): Haul.DraggingState =>
   useMemoSelect(selectHauling, []);
 
-export const selectColorContext = (state: StoreState): Color.ContextState =>
-  selectSliceState(state).colorContext;
+export const selectColorContext = (state: StoreState): Color.ContextState => {
+  const rawContext = selectSliceState(state).colorContext;
+  return Color.contextStateZ.parse(rawContext);
```
emilbon99, line 239: Why do we need to parse this every time we select? Why don't we parse on import and then assume its valid?

console/src/layout/slice.ts
```diff
@@ -480,7 +461,7 @@ export const { actions, reducer } = createSlice({
       mosaic.focused = key;
     },
     setColorContext: (state, { payload }: PayloadAction<SetColorContextPayload>) => {
-      state.colorContext = payload.state;
+      state.colorContext = Color.transformColorsToHex(payload.state);
```
emilbon99, line 464: Kind of a weird utility

console/src/layout/selectors.ts
```diff
@@ -234,8 +234,10 @@ export const selectHauling = (state: StoreState): Haul.DraggingState =>
 export const useSelectHauling = (): Haul.DraggingState =>
   useMemoSelect(selectHauling, []);
 
-export const selectColorContext = (state: StoreState): Color.ContextState =>
-  selectSliceState(state).colorContext;
+export const selectColorContext = (state: StoreState): Color.ContextState => {
```
emilbon99, line 237: Seems pretty inefficient to parse this every time. Why don't we parse when setting it on the slice?

console/src/layouts/Mosaic.tsx
```diff
@@ -232,52 +229,61 @@ export const Mosaic = memo((): ReactElement => {
...
-            const fileAsJSON = JSON.parse(new TextDecoder().decode(buffer));
-
-            let handlerFound = false;
-            for (const fileHandler of FILE_HANDLERS)
-              if (
-                await fileHandler({
-                  file: fileAsJSON,
-                  placer,
+    async (nodeKey: number, loc: location.Location, event: React.DragEvent) => {
+      const items = Array.from(event.dataTransfer.items);
```
emilbon99, line 234: Might be a good idea to move this function into the import package. It seems like it would belong there mroe

console/src/layouts/Mosaic.tsx
```diff
@@ -232,52 +229,61 @@ export const Mosaic = memo((): ReactElement => {
...
+      for (const item of items)
+        try {
+          const entry = await handleDataTransferItem(item);
+          if (entry == null) throw new Error("path is null");
+          if (entry instanceof File) {
```
emilbon99, line 239: Can you add a comment saying that this is for a visualization, while the other case is for importing a workspace? Took me a while to figure out

console/src/layouts/Mosaic.tsx
```diff
@@ -232,52 +229,61 @@ export const Mosaic = memo((): ReactElement => {
...
           }
+
+          const parsedFiles = await Promise.all(
+            entry.files.map(async (file) => {
+              const buffer = await file.arrayBuffer();
+              const data = new TextDecoder().decode(buffer);
+              return { name: file.name, data };
+            }),
+          );
+          await ingest(entry.name, parsedFiles, {
```
emilbon99, line 273: Confusing that this function is called within the layout package but just called ingest, but it turns out its for ingesting a workspace. Can you rename to `ingestWorkspace`

emilbon99 (changes requested):

console/src/layout/selectors.ts
pjdotson: If we parse this on import, then we are storing a class `Color` in the Redux store. We don't want to do that so the colors are stored as `string`s in the store. `selectColorContext` returns a type that has `Color`s as some of the properties, so we need to reconstruct the colors when it is being used.

console/src/layout/slice.ts
pjdotson: Explained reason above but we can't store the `Color`s in the Redux store.

console/src/import/importDataTransferItem.ts
```diff
@@ -0,0 +1,124 @@
...
+
+import { type DirectoryIngestor, type FileIngestor } from "@/import/ingestor";
+import { type Layout } from "@/layout";
+
+interface DirectoryContent {
+  name: string;
+  files: File[];
+}
+
+const handleDataTransferItem = async (
```
emilbon99, line 22: I think this would be better named as `parseDataTransferItem`. Just not sure that this is a handler that handles any event or anything

console/src/import/importDataTransferItem.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (entries.length === 0) break;
+      await processEntries(entries);
+    }
+  };
+
+  await readAllEntries();
+  return { name: entry.name, files };
+};
+
+interface fromFileDropCtx {
```
emilbon99, line 63: Interfaces should be capitalized

console/src/import/importDataTransferItem.ts
```diff
@@ -0,0 +1,124 @@
...
+  client: Synnax | null;
+  fileIngestors: Record<string, FileIngestor>;
+  ingestDirectory: DirectoryIngestor;
+  layout: Partial<Layout.State>;
+  placeLayout: Layout.Placer;
+  store: Store;
+}
+
+export const importDataTransferItem = async (
```
emilbon99, line 72: Since the package is named import, this should just be called `Import.dataTransferItem`

emilbon99: File could be renamed as well

console/src/import/import.ts
```diff
@@ -0,0 +1,95 @@
...
+            message: `Failed to import ${type} at ${path}`,
+            description: e.message,
+            variant: "error",
+          });
+        }
+      }),
+    );
+  };
+
+export const useImport = (
```
emilbon99, line 86: Hook should just be called `use`

console/src/export/useExport.ts
```diff
@@ -0,0 +1,48 @@
...
+import { type DialogFilter, save } from "@tauri-apps/plugin-dialog";
+import { writeTextFile } from "@tauri-apps/plugin-fs";
+import { useStore } from "react-redux";
+
+import { type Extractor } from "@/export/extractor";
+import { type RootState } from "@/store";
+
+const FILTERS: DialogFilter[] = [{ name: "JSON", extensions: ["json"] }];
+
+export const useExport = (
```
emilbon99, line 20: Should just be called `use` as well as file name should be `use.ts`

console/src/import/import.ts
```diff
@@ -0,0 +1,95 @@
...
+import { open } from "@tauri-apps/plugin-dialog";
+import { readTextFile } from "@tauri-apps/plugin-fs";
+import { useStore } from "react-redux";
+
+import { type FileIngestor } from "@/import/ingestor";
+import { Layout } from "@/layout";
+import { type RootState } from "@/store";
+import { Workspace } from "@/workspace";
+
+interface ImportArgs {
```
emilbon99, line 23: Should be public

console/src/workspace/convertLayout.ts
```diff
@@ -0,0 +1,21 @@
...
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { deep } from "@synnaxlabs/x";
+
+import { type Layout } from "@/layout";
+
+export const convertLayout = (sliceState: Layout.SliceState): Layout.SliceState => {
```
emilbon99, line 14: I think a better name for this would be `purgeExcludedLayouts`

emilbon99: `convert` seems like it involves some change of type, this is just filtration

console/src/import/import.ts
```diff
@@ -0,0 +1,95 @@
...
+  addStatus: Status.AddStatusFn;
+  client: Synnax | null;
+  placeLayout: Layout.Placer;
+  store: Store;
+  workspaceKey?: string;
+}
+
+interface Importer {
```
emilbon99, line 31: Should be public

console/src/import/import.ts
```diff
@@ -0,0 +1,95 @@
...
+  (args: ImportArgs): Promise<void>;
+}
+
+interface ImporterCreator {
```
emilbon99, line 35: Should be public

emilbon99 (approved):
Approved after my comments are addressed

PR #1045, opened 2025-01-08
Create Common Exception Handling Function

pjdotson:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1821](https://linear.app/synnax/issue/SY-1821/create-error-handling-catch-function)

## Description

Refactor common exception handling mechanics in Pluto and Console to leverage a common callback

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client


emilbon99:
TO-DO:

- Add linting rule for passing promise callback into non-promise prop
- Adjust the `Status.handleException` function to be a constructor

emilbon99 (changes requested):

pluto/src/status/Aggregator.tsx
```diff
@@ -80,6 +84,18 @@ export const Aggregator = Aether.wrap<AggregatorProps>(
 
 export const useAggregator = (): AddStatusFn => useContext(Context).add;
 
+export const useHandleException = (): HandleExcFn => {
```
emilbon99, line 87: Might be better named `useExceptionHandler`. Also, exceptions are generally referred to as errors in JS, so it might be good to follow that convention and make it `useErrorHandler`

pluto/src/vis/diagram/aether/Diagram.ts
```diff
@@ -74,7 +74,12 @@ export class Diagram extends aether.Composite<
         this.children.map(async (child) => await child.render({ viewportScale })),
       );
     } catch (e) {
-      addStatus({ variant: "error", message: (e as Error).message });
+      if (!(e instanceof Error)) throw e;
+      addStatus({
```
emilbon99, line 78: Any reason why you're not using the new hook here?

emilbon99 (approved):

pluto/src/status/Aggregator.tsx
pjdotson: Changed to use `useExceptionHandler`. There is a slight difference between `Error`s and exceptions - anything can be thrown in JavaScript, but an `Error` is a specific class, although the vast majority of exceptions are `Error`s. See [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Control_flow_and_error_handling).

pluto/src/vis/diagram/aether/Diagram.ts
pjdotson: We are in aether so I just didn't feel like implementing an aether version of the hook.

PR #1060, opened 2025-01-21
SY-1890 - Improve Console Persistence Mechanism

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1890](https://linear.app/synnax/issue/SY-1890/fix-bad-cleanup-of-persisted-state)

## Description

Migrates the console persisted state to a more reasonable directory, fixes a number of issues causing the persistence to over-store values, and adds a bunch of test cases to ensure that behavior is as expected.

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client


x/ts/src/kv/types.ts
```diff
@@ -9,48 +9,45 @@
 
 import { z } from "zod";
```
pjdotson, line 10: Something in your typing here is not well thought out. It seems to me like the following code below should compile well:

```ts
export class NumberKV implements KV<number> {
  readonly store: Map<string, number> = new Map();

  get(key: string): number | null {
    return this.store.get(key) ?? null;
  }

  set(key: string, value: number): void {
    this.store.set(key, value);
  }

  delete(key: string): void {
    this.store.delete(key);
  }
}
```

However I get TS errors on the `get` and `set` methods about how type `IV` is not assignable to type `number`. It probably makes more sense for the interface `KV` to have two different type parameters, one that is the type of the input value to the writer and one that is the type of the output value of the reader, so something like this:

```ts
export interface KV<R = unknown, W = unknown> extends Reader<R>, Writer<W>, Deleter {}
export interface Reader<T = unknown> {
  get: (key: string) => T | null;
};
```

console/src/store.ts
```diff
@@ -126,25 +126,25 @@ export const migrateState = (prev: RootState): RootState => {
...
         ...LinePlot.MIDDLEWARE,
         ...Layout.MIDDLEWARE,
         ...Schematic.MIDDLEWARE,
-        persistMiddleware,
+        Persist.middleware(engine),
       ),
     reducer,
     enablePrerender: true,
-    debug: false,
+    debug: true,
```
pjdotson, line 147: Can you make this either `false` or `isDev()`?

console/src/persist/state.ts
```diff
@@ -7,115 +7,150 @@
...
-import { createTauriKV } from "@/persist/kv";
+import { openTauriKV, type SugaredKV } from "@/persist/kv";
 import { type Version } from "@/version";
 
-const PERSISTED_STATE_KEY = "console-persisted-state";
-const DB_VERSION_KEY = "console-version";
+export const PERSISTED_STATE_KEY = "console-persisted-state";
+export const DB_VERSION_KEY = "console-version";
+
+export const V1_STORE_PATH = "~/.synnax/console/persisted-state.dat";
```
pjdotson, line 21: Are we sure that this v1 store path loaded properly on Windows because of the different file separators?

pjdotson (changes requested):
Address comments and also you should do a sanity check on Windows

x/ts/src/kv/types.ts
emilbon99: I just removed all the type parameter from the class. This fits our current KV use case. No need to spend a lot of time building an extensive type system. Also doesn't really make sense to me that you would read a different type than you write in the majority of cases.

console/src/store.ts
emilbon99: Good catch

console/src/persist/state.ts
emilbon99: Yes. This was the path we were previously using for persistence. It's a bad path on windows. That's why it's v1 now, and I changed it to a better path.

console/src/persist/kv.ts
```diff
@@ -7,18 +7,28 @@
...
 // included in the file licenses/APL.txt.
 
-import { load, type Store } from "@tauri-apps/plugin-store";
+import { type kv } from "@synnaxlabs/x";
+import { LazyStore } from "@tauri-apps/plugin-store";
 
-export const multipleWindowsOpen = new Error("[persist] - windows open");
+/**
+ * A SugaredKV is a spiced up key-value store that provides a few extra goodies needed
+ * for efficient persistence.
```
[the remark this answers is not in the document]
emilbon99: It's a fun comment - sometimes its ok to have a bit of fun

pjdotson (approved):

PR #1064, opened 2025-01-24
SY-1898 - Calculations

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1898](https://linear.app/synnax/issue/SY-1898/calculations-v2)

## Description

Lua implementation of calculated channels.

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client


x/go/telem/series.go
```diff
@@ -65,8 +66,24 @@ func (s Series) Split() [][]byte {
...
+		i += s.Len()
+	}
 	b := s.Data[i*int64(s.DataType.Density()) : (i+1)*int64(s.DataType.Density())]
 	return UnmarshalF[T](s.DataType)(b)
 }
+
+// SetValueAt sets the value at the given index in the series. SetValueAt supports
+// negative indices, which will be wrapped around the end of the series. This function
+// cannot be used for variable density series.
+func SetValueAt[T types.Numeric](s Series, i int64, v T) {
```
Lham42, line 83: Is there a way we can see if a series is variable density and not modify the series in that case?

x/go/telem/series.go
```diff
@@ -65,8 +66,24 @@ func (s Series) Split() [][]byte {
 	return o
 }
 
-// ValueAt returns the value at the given index in the series.
+// ValueAt returns the numeric value at the given index in the series. ValueAt supports
+// negative indices, which will be wrapped around the end of the series. This function
+// cannot be used for variable density series.
 func ValueAt[T types.Numeric](s Series, i int64) T {
```
Lham42, line 72: Can we check if htere is a mismatch between the type T and the series passed in? What is the expected behaviour in that case/can there a be a test for that

synnax/pkg/service/framer/service.go
```diff
@@ -58,7 +57,7 @@ func (c Config) Override(other Config) Config {
 
 type Service struct {
 	Config
-	Calculated *calculated.Service
+	Calculated *calculation.Service
```
Lham42, line 60: can we rename this to Calculation?

synnax/pkg/service/framer/calculation/calculation.go
```diff
@@ -262,89 +330,80 @@ func (s *Service) startCalculation(
...
-		if len(s) == 0 {
-			continue
+func (c Calculator) close() { c.calculation.Close() }
+
+func (c Calculator) Calculate(fr framer.Frame) (of framer.Frame, err error) {
+	if len(fr.Series) == 0 {
+		return
+	}
+	os := telem.AllocSeries(c.ch.DataType, fr.Series[0].Len())
+	os.Alignment = fr.Series[0].Alignment
```
Lham42, line 384: So right now is this assuming that the rest of the series in the frame are the same size/alignment? Can we comment that to make it clear. What happens right now if that assumption is false?

console/src/channel/CreateCalculated.tsx
```diff
@@ -123,19 +158,19 @@ interface InternalProps extends Pick<RendererProps, "onClose"> {
...
+    values: initialValues,
+    sync: true,
+  });
 
   const addStatus = Status.useAggregator();
 
   const [createMore, setCreateMore] = useState(false);
   const { mutate, isPending } = useMutation({
     mutationFn: async (createMore: boolean) => {
       if (client == null) throw new Error("Client not available");
```
Lham42, line 172: we dont handle it like this in the original create channel window. we just return.

Lham42 (changes requested):

client/py/synnax/channel/client.py
```diff
@@ -64,7 +64,7 @@ def __init__(
         index: ChannelKey = 0,
         leaseholder: int = 0,
         key: ChannelKey = 0,
-        virtual: bool = False,
+        virtual: bool | None = None,
```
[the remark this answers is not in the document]
emilbon99: Yes. Resolved

synnax/pkg/distribution/channel/channel.go
```diff
@@ -212,6 +213,77 @@ func (c Channel) IsCalculated() bool {
 	return c.Virtual && c.Expression != ""
 }
 
+// Equals returns true if the two channels are meaningfully equal to each other. This
+// function should be used instead of a direct comparison, as it takes into account
+// the contents of the Requires field, ignoring the order of the keys.
+// If the exclude parameter is provided, the function will ignore the fields specified
+// in the exclude parameter.
+func (c Channel) Equals(other Channel, exclude ...string) bool {
```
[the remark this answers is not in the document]
emilbon99: AFAIK there are no other comparisons.

x/go/computron/computron.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	default:
+		return series
+	}
+	return series
+}
+
+var luaOptions = lua.Options{
+	// A callstack size of 20 is more than enough for all calculations.
+	CallStackSize: 5,
```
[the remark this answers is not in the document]
emilbon99: Updated call stack size to 10. Don't see us needing more than that.

x/ts/src/telem/series.ts
```diff
@@ -1089,6 +1089,12 @@ export class MultiSeries<T extends TelemValue = TelemValue> implements Iterable<
     return bounds.distance(b, start, end);
   }
 
+  parseJSON<Z extends z.ZodTypeAny>(schema: Z): Array<z.output<Z>> {
+    if (!this.dataType.equals(DataType.JSON))
+      throw new Error("cannot convert non-string series to strings");
```
pjdotson, line 1094: Update error message

synnax/pkg/service/framer/service.go
emilbon99: Renamed

console/src/channel/CreateCalculated.tsx
```diff
@@ -264,34 +309,38 @@ const Editor = (props: Code.EditorProps): ReactElement => {
...
-        if (channelKeys.length > 0)
-          requires.onChange(unique.unique([...valueRef.current, ...channelKeys]));
+
+        if (channelKeys.length > 0) {
+          const newRequires = unique.unique([...valueRef.current, ...channelKeys]);
+          requires.onChange(newRequires);
+        }
       } catch (error) {
-        console.error("Error initializing required channels:", error);
+        console.error("Error updating required channels:", error);
```
pjdotson, line 340: This should be an addStatus not a console error

pjdotson (commented):
In the resources toolbar, the extra icons on top of the channels are really hard to see - I think we should remove them, I couldn't even make out the ".00" without squinting:
image

In the calculated channel modal, the "Beta" looks like a button that we can click. It should probably get moved somewhere else, like right after the title in the top nav bar.

I don't like having to select the required channels in the console. I think that when we hit create we should be able determine what channels are required from the expression.

x/go/telem/series.go
emilbon99: It's impossible to check this without runtime reflection, which incures a huge overhead in a very hot path

x/go/telem/series.go
emilbon99: Yes

synnax/pkg/service/framer/calculation/calculation.go
emilbon99: Added a comment. Also added validation checks to calculated channels that prevent channels with different indexes from being used in the same calculation

console/src/channel/CreateCalculated.tsx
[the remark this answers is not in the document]
emilbon99: We should have a better error check here though, or disable the ability to hit the button. Let's discuss.

console/src/channel/CreateCalculated.tsx
emilbon99: @Lham42 I don't understand why this entire function needs to exists. Why do we need to generate the requires list? Shouldn't this just be a field on the channel?

Lham42: This is what I did to try to populate the required_channels list as you type it in full. Otherwise it only updates the list on autocomplete

emilbon99: So this really applies only when you copy and paste into a calculation, not when you're loading in an existing calculation? Just confused by the comment

x/go/computron/computron.go
```diff
@@ -0,0 +1,126 @@
+package computron
```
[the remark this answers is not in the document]
emilbon99: added

console/src/channel/CreateCalculated.tsx
emilbon99: I am refactoring FYI

Lham42: Or if you are typing it in the full channel name, not just copy pasting

Lham42: but ok

emilbon99: Ah gotcha

emilbon99: Refactored and added checks for hyphenated channels

console/src/channel/CreateCalculated.tsx
```diff
@@ -89,7 +90,40 @@ export const createCalculatedLayout = (base: Partial<Layout.State>): Layout.Stat
...
+      const s = await client.openStreamer({ channels: [CALCULATION_STATE_CHANNEL] });
+      return new framer.ObservableStreamer(s);
+    },
+    onChange: (frame) => {
+      const state = frame.get(CALCULATION_STATE_CHANNEL).parseJSON(calculationStateZ);
+      state.forEach(({ key, variant, message }) => {
+        client?.channels.retrieve(key).then((ch) => {
+          addStatus({
+            variant,
+            message: `Calculation for ${ch.name} failed`,
```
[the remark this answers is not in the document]
emilbon99: Resolved

emilbon99:
> In the calculated channel modal, the "Beta" looks like a button that we can click. It should probably get moved somewhere else, like right after the title in the top nav bar.

@Lham42 Wanted it in the bottom right

emilbon99:
> In the resources toolbar, the extra icons on top of the channels are really hard to see - I think we should remove them, I couldn't even make out the ".00" without squinting: image

They have been very useful to me. Is there an alternative? Should we just make the regular icon in the top right corner?

emilbon99:
> I don't like having to select the required channels in the console. I think that when we hit create we should be able determine what channels are required from the expression.

These get auto populated very nicely with tab completion, and in order to auto-select them we'd have to be able to parse lua Syntax to determine which variables are channels and which aren't. Not worth it for initial release IMO. The auto-completion works really well.

client/ts/src/channel/channel.spec.ts
```diff
@@ -306,40 +306,40 @@ describe("Channel", () => {
         name: "virtual-calc",
         dataType: DataType.FLOAT32,
         virtual: true,
-        expression: "result = np.array([])"
+        expression: "return np.array([])",
```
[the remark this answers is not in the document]
emilbon99: Removed

client/py/synnax/channel/client.py
```diff
@@ -91,6 +91,8 @@ def __init__(
         from the channel. This is provided by the Synnax py during calls to
         .channels.create() and .channels.retrieve() and should not be set by the caller.
```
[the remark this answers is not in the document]
emilbon99: They are in there

emilbon99:
> > In the resources toolbar, the extra icons on top of the channels are really hard to see - I think we should remove them, I couldn't even make out the ".00" without squinting: image
> 
> They have been very useful to me. Is there an alternative? Should we just make the regular icon in the top right corner?

Removed

pjdotson (approved):

x/go/telem/series.go
Lham42: Can do that then?

console/src/channel/CreateCalculated.tsx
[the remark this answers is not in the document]
Lham42: Has this been resolved?

x/go/telem/series.go
Lham42: got

client/py/synnax/channel/client.py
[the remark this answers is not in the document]
Lham42: I also don't see them? wouldnt there be a `:param` for them

Lham42 (approved):

PR #1065, opened 2025-01-25
SY-1899 - Pluto & Console - Quality of Life & Perf Improvements

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1899](https://linear.app/synnax/issue/SY-1899/various-console-perf-improvements)

## Description

Makes a few performance and quality of live improvements to pluto and the console

- Removes the `Aether.wrap` HOC that is no longer needed in React 19
- Removes all calls to `forwardRef` which are no longer needed in React 19
- Makes it so that `uint8` channels don't need to be converted to `float32` in order to be buffered and rendered on the gpu. Saves 4x memory in JS Aether, 4x memory in GPU, and 4x memory transfer between JS and GPU.

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client


pluto/src/vis/line/aether/frag_u8.glsl
```diff
@@ -0,0 +1,9 @@
+#version 300 es
+precision highp float;
```
pjdotson, line 2: Do we need high precision here?

pluto/src/vis/line/aether/vert.glsl
pjdotson, line 1: Can you delete some of these .glsl files that aren't being used?

pjdotson (changes requested):
Are we able to wrap the console app in `<StrictMode >` now?

pluto/src/telem/aether/convertSeries.ts
```diff
@@ -12,13 +12,18 @@ import { DataType, type math, type Series } from "@synnaxlabs/x";
...
-  return series.convert(DataType.FLOAT32, offset);
+export const convertSeries = (s: Series, offset?: math.Numeric): Series => {
+  if (s.dataType.isVariable || s.dataType.equals(DataType.UINT8)) return s;
+  if (offset == null && s.dataType.equals(DataType.TIMESTAMP))
+    offset = BigInt(s.data[0]);
+  return s.convert(DataType.FLOAT32, offset);
+};
+
+export const resolveDataType = (dt: DataType): DataType => {
+  if (dt.isVariable || dt.equals(DataType.UINT8)) return dt;
```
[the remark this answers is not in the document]
emilbon99: Main reason there's no support for UINT8 is that the vertex shader requires a uint, which will remove the sign from the integer. To support int8s we'd need to add another fragment + vertex shader. Since int8 is rarely used, while uint8 is very common, it's better to have a hot path optimization for uint8s than for int8s.

Boolean is just an alias for uint8

pluto/src/vis/line/aether/frag_u8.glsl
emilbon99: This file is dead code actually. Removed because we're justing using `frag.glsl` for both the float32s and uint8s

pluto/src/vis/line/aether/vert.glsl
emilbon99: Yep - solved

PR #1078, opened 2025-02-01
SY-1897 Fix Control State Issues

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1897](https://linear.app/synnax/issue/SY-1897/fix-schematic-controller-state-mangling)

## Description

This PR is a major refactor of the aether tree to correctly isolated context values that have been overriden by child components. It also improves typescript lint checks and addresses existing problems. 

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client


pluto/src/aether/aether/aether.ts
```diff
@@ -18,38 +18,17 @@ import { type MainMessage, type WorkerMessage } from "@/aether/message";
 import { state } from "@/state";
 import { prettyParse } from "@/util/zod";
 
-type UpdateVariant = "state" | "context";
+/**
+ * A selected subset of the Map interface used for efficiently propagating context
+ * updates through the tree.
+ */
+export interface ContextMap extends Pick<Map<string, any>, "get" | "forEach"> {}
```
pjdotson, line 25: Do you have to make the map a map of `string, any>`? Why can't you do `Map<string, unknown>`?

pluto/src/aether/aether/aether.ts
```diff
@@ -18,38 +18,17 @@ import { type MainMessage, type WorkerMessage } from "@/aether/message";
...
-  /**
-   * instrumentation is used for logging and tracing.
-   */
-  instrumentation: alamos.Instrumentation;
-}
+/**
+ * An internal function alias that creates a new component with the specified
+ * initial parent context
+ */
+type CreateComponent = (initialParentCtxValues: ContextMap) => Component;
```
pjdotson, line 31: Make this an interface instead of a type.

pluto/src/aether/aether/aether.ts
```diff
@@ -62,181 +41,148 @@ export interface Component {
...
-   * @param update - The update to propagate.
+   * @param path - The path of the component to update. The length of the path indicates
+   * the depth of the component in the aether tree. This method should propagate the
+   * state if the path has a length greater than 1.
+   * @param state - The state to update the component with.
+   * @param create - A function that creates a new component of the appropriate type if
+   * it doesn't exist in the tree.
    */
-  internalUpdate: (update: Update) => Promise<void>;
+  _updateState: (
```
pjdotson, line 55: Instead of naming these fields with an underscore, you should only export `Component`
with the fields `type` and `key`, and then make a new interface that does not get
exported that has these "private" fields. Then the classes could be `export class Class
implements Component, Internal`

pluto/src/aether/aether/aether.ts
```diff
@@ -62,181 +41,148 @@ export interface Component {
...
+  instrumentation: alamos.Instrumentation;
+  /** Initial parent context values for the component. These should be intentionally
+   * set to null if no values exists.
+   */
+  parentCtxValues: ContextMap | null;
 }
 
 /** A constructor type for an AetherComponent. */
-export type ComponentConstructor = new (update: Update) => Component;
+export type ComponentConstructor = new (props: ComponentConstructorProps) => Component;
```
pjdotson, line 91: Make this an interface instead of a type.

pluto/src/aether/aether/aether.ts
```diff
@@ -62,181 +41,148 @@ export interface Component {
...
-  set(key: string, value: any, trigger: boolean = true): void {
-    this.providers.set(key, value);
-    if (trigger) this.changedKeys.push(key);
-  }
-
-  setIfNotHas(key: string, value: any): void {
-    if (!this.providers.has(key)) this.set(key, value);
-  }
-
+  get<P>(key: string): P;
```
pjdotson, line 105: If it doesn't do any internal type checking there really isn't any point of the generic
type <P> here - just make the type of this method `get(key: string): unknown)` and then
the caller can say `const foo = ctx.get(key) as string`.

pluto/src/aether/aether/aether.ts
```diff
@@ -62,181 +41,148 @@ export interface Component {
...
-  async create<C extends Component>(update: Update): Promise<C> {
-    const Factory = this.registry[update.type];
-    if (Factory == null)
-      throw new Error(`[AetherRoot.create] - could not find component ${update.type}`);
-    const c = new Factory(update) as C;
-    await c.internalUpdate(update);
-    return c;
-  }
-
+  getOptional<P>(key: string): P | null;
```
pjdotson, line 112: Same here

pluto/src/aether/aether/aether.ts
```diff
@@ -62,181 +41,148 @@ export interface Component {
...
+    key,
+    type,
+    sender,
+    instrumentation,
+    parentCtxValues,
+  }: ComponentConstructorProps) {
+    this.type = type;
+    this.key = key;
+    this.sender = sender;
+    this._internalState = {} as InternalState;
```
pjdotson, line 173: Is there a reason to not just have them pass the initial internal state in the constructor?

pluto/src/aether/aether/aether.ts
```diff
@@ -62,181 +41,148 @@ export interface Component {
...
 }
 
 /**
- * Implements an AtherComponent that does not have any children, and servers as the base
+ * Implements an AetherComponent that does not have any children, and servers as the base
  * class for the AetherComposite type. The corresponding react component should NOT have
  * any children that use Aether functionality; for those cases, use AetherComposite instead.
  */
-export class Leaf<S extends z.ZodTypeAny, IS extends {} = {}> implements Component {
+export abstract class Leaf<State extends z.ZodTypeAny, InternalState extends {} = {}>
```
pjdotson, line 144: `StateSchema` would be a better name than `State`

pluto/src/aether/aether/aether.ts
pjdotson, line 1: Can you make `Composite` an abstract class too?

pluto/src/aether/aether/aether.ts
```diff
@@ -361,188 +323,231 @@ export class Composite<
...
-
-    const [key, subPath] = this.getRequiredKey(path);
-    // In this case, we can safely assume the context hasn't changed, so we can just use
-    // the internal, cached context.
-    const uCached = { ...u, ctx: this.ctx };
-    return subPath.length === 0
-      ? await this.updateThis(key, uCached)
-      : await this.updateChild(subPath, uCached);
+  getChild<T extends C = C>(key: string): T | null {
+    return (this._children.get(key) ?? null) as T | null;
```
pjdotson, line 340: I don't get the point of allow for typing this method, just return a type of `C | null` and let the caller narrow the method properly

pluto/src/aether/aether/aether.ts
```diff
@@ -361,188 +323,231 @@ export class Composite<
...
-    await super.internalUpdate(u);
-    for (const c of this.children)
-      await c.internalUpdate({ ...u, ctx: this.ctx, variant: "context" });
+  /**
+   * Finds all children of the component with the given type
+   *
+   * @param types - the type of the children to find
+   * @returns an array of all children of the component with the given type
+   */
+  childrenOfType<T extends C = C>(...types: Array<T["type"]>): readonly T[] {
```
pjdotson, line 349: Similar comment here about the type assertions doing more harm than good.

pluto/src/telem/aether/factory.ts
```diff
@@ -45,7 +45,7 @@ export class CompoundTelemFactory {
   }
 }
 
-export const factory = (client?: client.Client): CompoundTelemFactory => {
+export const newFactory = (client?: client.Client): CompoundTelemFactory => {
```
pjdotson, line 48: Call it `createFactory` instead of `newFactory` - we should really try to make all
function names verbs.

pjdotson (changes requested):

pluto/src/aether/aether/aether.ts
emilbon99: I disagree. For function types that are purely meant as aliases that should not be extended, this syntax is more condensed and easier to understand.

pluto/src/aether/aether/aether.ts
emilbon99: Doesn't this method assert that the components have the particular type? If you're worried about fixed type checking, then we should get rid of the entire 'C' type as well since it assumes children will be of a particular type. 

On the other hand, this means that we'll have to introduce about 40 `as` statements into the code. IMO there are places where typing is helpful and places where typing is harmful. This is one of the places where using `as` everywhere is harmful.

pluto/src/aether/aether/aether.ts
emilbon99: Not to mention this pattern is far more common in typescript than using interfaces to define functions

pluto/src/pluto/Pluto.tsx
```diff
@@ -64,13 +64,13 @@ export const Provider = ({
               <Status.Aggregator>
                 <Synnax.Provider connParams={connParams}>
                   <Channel.AliasProvider {...channelAlias}>
-                    <CanDisableTelem {...telem}>
-                      <Color.Provider {...color}>
-                        <Theming.Provider {...theming}>
+                    <Color.Provider {...color}>
+                      <Theming.Provider {...theming}>
+                        <CanDisableTelem {...telem}>
```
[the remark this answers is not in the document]
emilbon99: It reduces the number of context re-renders in the aether tree.

configs/eslint/index.js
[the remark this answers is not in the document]
emilbon99: Will correct

PR #1092, opened 2025-02-07
SY-1335 Console Hardware Refactor, Console Support for NI Analog Writer

pjdotson:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1335](https://linear.app/synnax/issue/SY-1335/console-task-refactor)

## Description

- Refactor of `console/src/hardware` directory
- NI Analog Write Task

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [x] Python Client
- [x] Server
- [ ] TypeScript Client


Lham42 (approved):
Didn't notice anything glaring but also review wasn't as thorough as emiliano will probably look at it. Mostly expect to find things through testing

client/ts/src/hardware/device/payload.ts
```diff
@@ -10,44 +10,42 @@
...
-  z.output<typeof deviceZ>,
-  "properties"
-> & { properties: P };
-
-export type DeviceKey = z.infer<typeof deviceKeyZ>;
-
-export const newDeviceZ = deviceZ.extend({
+export interface Device<
+  P extends UnknownRecord = UnknownRecord,
+  MK extends string = string,
```
emilbon99, line 34: These generic type names are hard to associate with properties. Just make them `Properties`, `Make` and `Model`

emilbon99: And make sure to do it for supertypes, consumers as well

client/ts/src/hardware/task/payload.ts
```diff
@@ -11,111 +11,84 @@ import { binary, type observe, type UnknownRecord } from "@synnaxlabs/x";
...
 });
-
-export const newTaskZ = taskZ.omit({ key: true }).extend({
-  key: taskKeyZ.transform((k) => k.toString()).optional(),
-  config: z.unknown().transform((c) => binary.JSON_CODEC.encodeString(c)),
-});
-
-export type NewTask<
+export interface Payload<
   C extends UnknownRecord = UnknownRecord,
```
emilbon99, line 51: Would be good to improve these generic type names as well

client/ts/src/ontology/payload.ts
```diff
@@ -95,36 +117,38 @@ export class ID {
   static readonly z = z.union([z.instanceof(ID), crudeIDZ.transform((v) => new ID(v))]);
 }
 
-export const Root = new ID({ type: BUILTIN_TYPE, key: "root" });
+export const RootID = new ID({ type: BUILTIN_TYPE, key: "root" });
```
emilbon99, line 120: `SCREAM_CASE` me baby! Kung fu panda!

client/ts/src/workspace/payload.ts
```diff
@@ -0,0 +1,41 @@
...
+export type Key = z.infer<typeof keyZ>;
+export type Params = Key | Key[];
+
+// --- VERY IMPORTANT ---
+// Synnax's encoders (in the binary package inside x) automatically convert the case
+// of keys in objects to snake_case and back to camelCase when encoding and decoding
+// respectively. This is done to ensure that the keys are consistent across all
+// languages and platforms. Sometimes workspaces have keys that are uuids, which have
+// dashes, and those get messed up. So we just use regular JSON for workspaces.
+const parse = (s: string): UnknownRecord => JSON.parse(s);
```
emilbon99, line 23: Should probably make this a common function that we apply to all the visualization types as well, as this warning is very relevant and goes unnotticed with visualization types.

console/src/channel/services/ontology.tsx
```diff
@@ -285,25 +282,20 @@ const TreeContextMenu: Ontology.TreeContextMenu = (props) => {
...
-      entry={{ ...props.entry, name: alias ?? props.entry.name }}
-    />
-  );
+export const Item: Tree.Item = ({ entry, ...rest }: Tree.ItemProps): ReactElement => {
+  const alias = Channel.useAlias(Number(new ontology.ID(entry.key).key));
+  return <Tree.DefaultItem {...rest} entry={{ ...entry, name: alias ?? entry.name }} />;
 };
 
 export const ONTOLOGY_SERVICE: Ontology.Service = {
+  ...Ontology.BASE_SERVICE,
```
emilbon99, line 291: Does `BASE_SERVICE` have noop implementations for everything? If so, might be better to call it `NOOP_SERVICE`

console/src/cluster/Dropdown.tsx
```diff
@@ -182,31 +182,31 @@ interface ListItemProps extends CoreList.ItemProps<string, Cluster> {
   validateName: (name: string) => boolean;
 }
 
-const ListItem = ({ validateName, ...props }: ListItemProps): ReactElement => {
+const ListItem = ({ validateName, ...rest }: ListItemProps): ReactElement => {
```
emilbon99, line 185: If we're going to use the `...rest` naming convention, we should apply it to call components. I think we should stick with `props`

console/src/hardware/common/device/Configure.tsx
```diff
@@ -151,19 +146,57 @@ export const Configure = <P extends UnknownRecord = UnknownRecord>({
...
+  const { data, error, isError, isPending } = useQuery({
+    queryKey: [layoutKey, client?.key],
+    queryFn: async () => {
+      if (client == null) throw new Error("Cannot reach server");
+      return await client.hardware.devices.retrieve<P>(layoutKey);
+    },
+  });
+  if (isPending)
+    return (
+      <Status.Text.Centered level="h2" variant="loading">
```
emilbon99, line 184: This is a pretty ugly loading indicator. I think we can make prettier

console/src/hardware/common/device/Configure.tsx
```diff
@@ -151,19 +146,57 @@ export const Configure = <P extends UnknownRecord = UnknownRecord>({
...
+        Fetching device from server
+      </Status.Text.Centered>
+    );
+  if (isError) {
+    const color = Status.variantColors.error;
+    return (
+      <Align.Space align="center" grow justify="center">
+        <Text.Text color={color} level="h2">
```
emilbon99, line 192: h2 seems to big for this. h4 would be better

console/src/hardware/common/device/Provider.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export const Provider = <
+  P extends UnknownRecord = UnknownRecord,
+  MK extends string = string,
+  MO extends string = string,
+>({
+  canConfigure,
+  children,
+  configureLayout,
+  noneSelectedElement = DEFAULT_NONE_SELECTED_ELEMENT,
```
emilbon99, line 51: Rename to `noneSelectedContent` to match convention of other props such as `emptyContent`

console/src/hardware/common/device/Select.tsx
```diff
@@ -0,0 +1,77 @@
...
+      path={path}
+      style={{ width: "100%" }}
+    >
+      {(p) => (
+        <Device.SelectSingle
+          {...p}
+          allowNone={false}
+          autoSelectOnNone={false}
+          emptyContent={
+            typeof emptyContent === "string" ? (
```
emilbon99, line 61: Could be good to drill this conditional logic into the select component itself, as this seems like it could be useful in a bunch of places.

emilbon99 (changes requested):
Looking very good so far. A few comments to look over before I go on a run.

console/src/hardware/common/task/Form.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const [state, setState] = useState(tsk.key, tsk.state ?? undefined);
+    const configureMutation = useMutation({
+      mutationFn: async () => {
+        if (client == null) throw new Error("Client not found");
+        if (!(await methods.validateAsync())) return;
+        const { config, name } = methods.value();
+        if (config == null) throw new Error("Config is required");
+        const [newConfig, rackKey] = await onConfigure(client, config, tsk.key, name);
+        methods.set("config", newConfig);
+        // current work around for Pluto form issues
```
emilbon99, line 97: Could you explain what the issue is in a bit more detail?

console/src/hardware/common/task/ParentRangeButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+      if (client == null) return;
+      const parent = await task.snapshottedTo();
+      if (parent != null) setParent(parent);
+      const tracker = await client.ontology.openDependentTracker({
+        target: task.ontologyID,
+        dependents: parent == null ? [] : [parent],
+        relationshipDirection: ontology.TO_RELATIONSHIP_DIRECTION,
+      });
+      tracker.onChange((parents) => {
+        const rng = parents.find((p) => p.id.matchesType(ranger.ONTOLOGY_TYPE));
```
emilbon99, line 38: Could this value ever be undefined?

console/src/hardware/common/task/useCreate.ts
```diff
@@ -0,0 +1,39 @@
...
+    async (task: task.New<C, T>, rackKey: rack.Key) => {
+      if (client == null) throw new Error("Client not found");
+      const rck = await client.hardware.racks.retrieve(
+        rackKey ?? rack.DEFAULT_CHANNEL_NAME,
+      );
+      const createdTask = await rck.createTask<C, D, T>(task);
+      dispatch(Layout.setArgs({ key: layoutKey, args: { taskKey: createdTask.key } }));
+      return createdTask;
+    },
+    [client, dispatch, layoutKey],
```
emilbon99, line 37: client?.key would be better

console/src/hardware/common/task/useCreate.ts
```diff
@@ -0,0 +1,39 @@
...
+  D extends {} = UnknownRecord,
+  T extends string = string,
+>(
+  layoutKey: string,
+) => {
+  const client = Synnax.use();
+  const dispatch = useDispatch();
+  return useCallback(
+    async (task: task.New<C, T>, rackKey: rack.Key) => {
+      if (client == null) throw new Error("Client not found");
```
emilbon99, line 29: I think this error message could be improved. We do this check in a bunch of other places too right? Could be good to extract into a common error/assertion

console/src/hardware/common/task/useState.ts
```diff
@@ -0,0 +1,80 @@
...
+    state: isRunning ? "running" : "paused",
+    message: initialState?.details?.message,
+  });
+  const client = Synnax.use();
+  Observe.useListener({
+    key: [client?.key, key, setState],
+    open: async () => client?.hardware.tasks.openCommandObserver(),
+    onChange: (command) => {
+      if (command.task !== key) return;
+      const type = command.type;
```
emilbon99, line 51: Small note, but it seems like a pretty useless extraction of a temporary variable. Also overloads a built-in keyword.

console/src/hardware/common/task/useState.ts
```diff
@@ -0,0 +1,80 @@
...
+    initialState?.details?.running ?? false,
+  );
+  const [state, setState] = useReactState<ReturnState>({
+    state: isRunning ? "running" : "paused",
+    message: initialState?.details?.message,
+  });
+  const client = Synnax.use();
+  Observe.useListener({
+    key: [client?.key, key, setState],
+    open: async () => client?.hardware.tasks.openCommandObserver(),
```
emilbon99, line 48: Could be a lot more efficient to create a `client.hardware.tasks.openActivityObserver` that has a streamer on both `sy_task_cmd` and `sy_task_state`, that way we can avoid opening two streamers where it could easily just be one.

console/src/hardware/common/task/useTare.ts
```diff
@@ -0,0 +1,62 @@
...
+  task,
+  isChannelTareable,
+  isRunning,
+}: UseTareProps<C>): UseTareReturn<C> => {
+  const client = Synnax.use();
+  const handleException = Status.useExceptionHandler();
+  const tare = useMutation({
+    onError: (e) => handleException(e, "Failed to tare channels"),
+    mutationFn: async (key: channel.Key[]) => {
+      if (client == null) throw new Error("Client not connected");
```
emilbon99, line 33: Another place we could consolidate

console/src/hardware/common/task/useTare.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (!(task instanceof clientTask.Task))
```
emilbon99, line 34: Why can't we do a null check? Using `instanceof` can be very error prone when checking across modules. Can explain verbally why.

console/src/hardware/common/task/useTare.ts
```diff
@@ -0,0 +1,62 @@
...
+        (channel) => keySet.has(channel.key) && isChannelTareable?.(channel) !== false,
+      );
+    },
+    [isChannelTareable],
+  );
+  const tareSingle = useCallback((key: channel.Key) => tare([key]), [tare]);
+  const allowTare = useCallback(
+    (keys: string[], channels: C[]) =>
+      getTareableChannels(keys, channels).length > 0 && isRunning,
+    [getTareableChannels, isRunning],
```
emilbon99, line 52: If you look at the code in `labjack/task/Read.tsx`, `getTareableChannels` is instantiated as a new variable on every render, which means all of these callbacks will be updated on every render, which means every item in the list will be rendered on every render.

You have to options here, either memoize every time you take in `getTareableChannels` or use the `useSyncedRef` pattern I showed you a few months ago. Second alternative is more bulletproof if you want to tolerate the caller passing in non-callbacked props.

emilbon99: Also, would be good to look at how tasks re-render with the react profiler.

console/src/hardware/labjack/device/Configure.tsx
```diff
@@ -7,48 +7,19 @@
...
-  };
+export const CONFIGURE_LAYOUT: Omit<Layout.BaseState, "key"> = {
+  ...Common.Device.CONFIGURE_LAYOUT,
+  name: "Device.Configure",
+  icon: "Logo.LabJack",
+  type: CONFIGURE_LAYOUT_TYPE,
+};
+
+export const Configure: Layout.Renderer = (props) => (
+  <Common.Device.Configure {...props} zeroProperties={ZERO_PROPERTIES} />
```
emilbon99, line 24: I think we typically use `initialProperties` or `initial*` as a convention in general. Not a huge deal - up to you.

console/src/hardware/labjack/device/SelectPort.tsx
```diff
@@ -0,0 +1,47 @@
...
+  > {
+  model: Model;
+  portType: PortType;
+}
+
+export const SelectPort = ({ model, portType, ...rest }: SelectPortProps) => (
+  <Select.Single<string, Port>
+    allowNone={false}
+    {...rest}
+    columns={[
```
emilbon99, line 32: This is 100% be memoized. Will cause a ton of re-renders.

emilbon99: Actually, can't this be moved to a SCREAM_CASE constant? kung fu panda!

console/src/hardware/labjack/task/getOpenPort.ts
```diff
@@ -0,0 +1,36 @@
...
+// included in the file licenses/APL.txt.
+
+import { Device } from "@/hardware/labjack/device";
+import { type Channel } from "@/hardware/labjack/task/types";
+
+export const getOpenPort = <T extends Device.PortType>(
+  channels: Channel[],
+  model: Device.Model,
+  types: T[],
+): T extends "AI"
```
emilbon99, line 17: Wild type. Could you extract into a named type so it's easier to understand what this is?

console/src/hardware/labjack/task/getOpenPort.ts
```diff
@@ -0,0 +1,36 @@
...
+  ? Device.AIPort | null
+  : T extends "DI"
+    ? Device.DIPort | null
+    : T extends "DO"
+      ? Device.DOPort | null
+      : T extends "AO"
+        ? Device.AOPort | null
+        : null => {
+  const portsInUse = new Set(channels.map((c) => c.port));
+  for (const type of types) {
```
emilbon99, line 27: Use `find` and not a for loop.

console/src/hardware/labjack/task/types.ts
```diff
@@ -10,17 +10,53 @@
...
+export const DI_CHANNEL_TYPE = "DI";
+type DIChannelType = typeof DI_CHANNEL_TYPE;
+
+export const TC_CHANNEL_TYPE = "TC";
+export type TCChannelType = typeof TC_CHANNEL_TYPE;
+
+export const AO_CHANNEL_TYPE = "AO";
+export type AOChannelType = typeof AO_CHANNEL_TYPE;
+
+export const AI_CHANNEL_TYPE = "AI";
```
emilbon99, line 28: Why is `AI_PORT_TYPE` in `labjack/device` the same as `AI_CHANNEL_TYPE` in `labjack/task`. Can we consolidate?

console/src/hardware/ni/task/types/v0.ts
```diff
@@ -383,14 +387,14 @@ export const ZERO_AI_FORCE_BRIDGE_TABLE_CHAN: AIForceBridgeTableChan = {
   physicalUnits: NEWTONS,
 };
 
-const twoPointLinZ = z.object({
+const twoPointLinShape = {
```
emilbon99, line 390: Nice approach. Hadn't thought about doing extension like this before.

console/src/hardware/ni/task/types/v0.ts
```diff
@@ -848,10 +846,10 @@ const aiChanZ = z.union([
   aiVoltageChanZ,
 ]);
 
-type AIChan = z.infer<typeof aiChanZ>;
-export type AIChanType = AIChan["type"];
+type AnalogInputChannel = z.infer<typeof analogInputChannelZ>;
```
emilbon99, line 849: Is there a reason why we renamed to this more verbose name? Afaik we use `AI` everywhere else, so why not here?

console/src/hardware/ni/task/types/v1.ts
```diff
@@ -241,11 +235,17 @@ export const ZERO_AI_CHANNELS: Record<v0.AIChanType, AIChan> = {
   [v0.AI_VOLTAGE_CHAN_TYPE]: ZERO_AI_VOLTAGE_CHAN,
 };
 
-export type Chan = v0.DIChan | AIChan | v0.DOChan;
+export const ZERO_AI_CHANNEL: AIChannel = ZERO_AI_CHANNELS[v0.AI_VOLTAGE_CHAN_TYPE];
+
+export type Channel =
+  | AIChannel
```
emilbon99, line 241: Mixing the conventions here doesn't make much sense. I like the condensed version.

console/src/hardware/common/device/Configure.tsx
pjdotson: There's an issue somewhere to improve loading indicators in general. I would like to tackle that + using query a lot more intelligentally somewhere the same time. There is a lot of components in our code base that follow the logic of 1) open query, 2) check if query is pending or is an error, 3) return actual content you want to see, and the basic query logic is always like ~80 lines of boilerplate.

emilbon99 (changes requested):
Left some more comments. One major note is the lack of consistency between using the shorthand channel convention `AI` and longhand `AnalogInput`. This is true in both variable names and in file names (ex. `SelectAIChannelTypeField.tsx`, and `SelectAnalogOutputChannelTypeField.tsx`.

Personally, I like the shorthand. It's the way we do it in Python and C++, and the way it's done in the NI DAQmx documentation. I know you'll probably disagree with me, so go with what you want. Regardless, be consistent.

console/src/hardware/common/device/Select.tsx
pjdotson: Added this to the List.tsx component, although I am not sure if that is the best place to put it.

console/src/hardware/common/task/layouts/ListAndDetails.tsx
```diff
@@ -0,0 +1,115 @@
...
+    ChannelListProps<C>,
+    "onTare" | "allowTare" | "isSnapshot" | "ListItem"
+  > {
+  Details: ComponentType<DetailsProps>;
+  generateChannel: GenerateChannel<C>;
+  initalChannels: C[];
+}
+
+export const ListAndDetails = <C extends Channel>({
+  ListItem: listItem,
```
emilbon99, line 42: If you're going to keep this as an `FC`, it should be `PascalCase`

console/src/hardware/common/task/layouts/ListAndDetails.tsx
```diff
@@ -0,0 +1,115 @@
...
+  Details,
+  initalChannels,
```
emilbon99, line 44: Typo

console/src/hardware/common/task/ChannelList.tsx
```diff
@@ -0,0 +1,169 @@
...
+    Pick<Align.SpaceProps, "onDragOver" | "onDrop"> {
+  emptyContent: ReactElement;
+  header: ReactNode;
+  isDragging?: boolean;
+  ListItem: FC<ChannelListItemProps<C>>;
+  selected: string[];
+}
+
+export const ChannelList = <C extends Channel>({
+  ListItem,
```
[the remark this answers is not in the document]
emilbon99: Not to mention, there's only one place in the codebase where you don't pass in a prop that would be equally compatible as a render prop.

console/src/hardware/labjack/task/Read.tsx
```diff
@@ -7,630 +7,308 @@
...
-          label="CJC Source"
-        >
-          {(p) => <SelectCJCSourceField {...p} model={model} />}
-        </Form.Field>
-        <Form.NumericField path={`${prefix}.cjcSlope`} label="CJC Slope" grow />
-        <Form.NumericField path={`${prefix}.cjcOffset`} label="CJC Offset" grow />
-      </Align.Space>
-    </Align.Space>
+    <Common.Task.Layouts.ListAndDetails<ReadChannel>
+      ListItem={(p) => <ChannelListItem {...p} onTare={tare} isRunning={isRunning} />}
```
emilbon99, line 208: Passing in props without memoizing them like this will cause the entire list to rerender every time the parent rerenders

console/src/hardware/ni/task/AnalogRead.tsx
```diff
@@ -7,533 +7,219 @@
...
-                  onTare={(key) => onTare([key])}
-                />
-              )}
-            </List.Core>
-          </List.Selector>
-        </List.List>
-      </Menu.ContextMenu>
-    </Align.Space>
+    <Common.Task.Layouts.ListAndDetails<AnalogInputChannel>
+      ListItem={(p) => <ChannelListItem {...p} onTare={tare} isRunning={isRunning} />}
```
emilbon99, line 120: Will cause the entire list to rerender every time the parent rerenders

console/src/log/services/ontology.tsx
```diff
@@ -101,11 +101,20 @@ const loadLog = async (client: Synnax, id: ontology.ID, placeLayout: Layout.Plac
...
 
-const handleSelect: Ontology.HandleSelect = async ({
+const handleSelect: Ontology.HandleSelect = ({
   client,
   selection,
   placeLayout,
-}) => await loadLog(client, selection[0].id, placeLayout);
+  handleException,
+}) => {
+  loadLog(client, selection[0].id, placeLayout).catch((e) => {
```
emilbon99, line 110: Kind of interesting to move towards promise callbacks and away from `async/await`. Async/await is more modern and user friendly IMO.

console/src/range/overview/Snapshots.tsx
```diff
@@ -33,16 +33,21 @@ const SNAPSHOTS: Record<"schematic" | "task", SnapshotService> = {
...
-      void (async () => {
-        const s = await client.workspaces.schematic.retrieve(res.id.key);
-        place(create({ ...s.data, key: s.key, name: s.name, snapshot: s.snapshot }));
-      })();
+      client.workspaces.schematic
+        .retrieve(res.id.key)
+        .then((s) =>
+          place(create({ ...s.data, key: s.key, name: s.name, snapshot: s.snapshot })),
+        )
+        .catch(console.error);
```
emilbon99, line 41: We can't improve to use `handleException`?

console/src/range/Toolbar.tsx
```diff
@@ -367,7 +367,7 @@ const List = (): ReactElement => {
   };
 
   return (
-    <PMenu.ContextMenu menu={(p) => <ContextMenu {...p} />} {...menuProps}>
+    <PMenu.ContextMenu menu={ContextMenu} {...menuProps}>
```
emilbon99, line 370: Don't pass a component as a render prop.

console/src/services.tsx
```diff
@@ -7,53 +7,63 @@
...
+  [schematic.ONTOLOGY_TYPE]: SchematicServices.ONTOLOGY_SERVICE,
+  [ontology.CLUSTER_TYPE]: ClusterServices.ONTOLOGY_SERVICE,
+  [user.ONTOLOGY_TYPE]: UserServices.ONTOLOGY_SERVICE,
+  [ontology.BUILTIN_TYPE]: generateEmpty(ontology.BUILTIN_TYPE),
+  [ontology.NODE_TYPE]: Node.ONTOLOGY_SERVICE,
+  [group.ONTOLOGY_TYPE]: Group.ONTOLOGY_SERVICE,
+  [ranger.ONTOLOGY_TYPE]: RangeServices.ONTOLOGY_SERVICE,
+  [workspace.ONTOLOGY_TYPE]: WorkspaceServices.ONTOLOGY_SERVICE,
+  [linePlot.ONTOLOGY_TYPE]: LinePlotServices.ONTOLOGY_SERVICE,
+  [ranger.ALIAS_ONTOLOGY_TYPE]: generateEmpty(ranger.ALIAS_ONTOLOGY_TYPE),
```
emilbon99, line 58: Rename to `createEmptyService` or just `createEmpty`. We almost always use `create` and should stick to that convention.l

emilbon99 (changes requested):

console/src/hardware/common/device/Configure.tsx
emilbon99: Sounds like a plan

console/src/log/services/ontology.tsx
pjdotson: I agree, but the problem of not handling errors in promises is pretty prevalant throughout our code base. IMO what we should move towards is something like a lot of asynchonous helper functions written with async / await, but instead of giving these functions a `handleException` arg or having them `console.error`, they will instead throw Errors. Then, when we call these functions in something synchronous like event handlers, the event handlers calls the async function and then catches it to handle errors. This works nicely because the main message of the error should probably be defined in the event handler ("failed to place log"), and the description can be from whatever error was thrown in the async function ("client not found").

console/src/hardware/common/task/ParentRangeButton.tsx
pjdotson: I don't think so, this will always be an ontology ID unless there is screwed up code in
the ontology client.

console/src/hardware/common/task/useCreate.ts
pjdotson: I made an `errors.ts` file with a `NullClientError` that gets thrown whenever we run
into something like this, feel free to change that message as you see fit.

console/src/hardware/common/task/Form.tsx
pjdotson: Added relevant issue number

console/src/log/services/ontology.tsx
emilbon99: Sounds good. Let's keep it this way.

PR #1110, opened 2025-02-20
SY-2047 - Remove Circular Dependencies from TypeScript codebase

pjdotson:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2047](https://linear.app/synnax/issue/SY-2047/remove-circular-dependencies-from-typescript-codebase)

## Description

- Removes circular dependencies from TypeScript codebase
- adds a dev tool for checking circular depencies

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client


client/ts/src/ontology/payload.ts
emilbon99, line 1: Why do we need a bunch of wilidcard imports when we only access one field?

client/ts/src/hardware/device/payload.ts
emilbon99, line 1: I don't understand the point of these wildcard imports

console/src/channel/services/palette.tsx
```diff
@@ -10,8 +10,8 @@
 import { Icon } from "@synnaxlabs/media";
 
 import { Channel } from "@/channel";
+import { Layout } from "@/layout";
```
emilbon99, line 13: Why was this move done? It doesn't make sense in layout. If it needs to be outside of version for some reason, then just move it into general purpose components.

console/src/cluster/Dropdown.tsx
```diff
@@ -20,23 +20,18 @@ import {
...
 import { useSelect, useSelectMany } from "@/cluster/selectors";
 import { type Cluster, remove, rename, setActive } from "@/cluster/slice";
-import { Menu } from "@/components/menu";
+import { Menu } from "@/components";
 import { CSS } from "@/css";
 import { Layout } from "@/layout";
 import { Link } from "@/link";
 
-export const List = (): ReactElement => {
+export const List = () => {
```
emilbon99, line 34: I'm a fan of forcing the user to specify a return type. It makes sure the caller doesn't accidentally return two mismatching types. If you want ot remove, that's ok though

console/src/hardware/opc/task/external.ts
```diff
@@ -22,4 +22,4 @@ export const LAYOUTS: Record<string, Layout.Renderer> = {
   [WRITE_TYPE]: Write,
 };
 
-export const SELECTABLES: Layout.Selectable[] = [READ_SELECTABLE, WRITE_SELECTABLE];
+export const SELECTABLES = [READ_SELECTABLE, WRITE_SELECTABLE];
```
emilbon99, line 25: Why did we get rid of these type annotations? I feel like they are helpful in understanding/enforcing what's supposed to be in there. If we get rid of this annotation, why don't we get rid of the one for the record above?

console/src/hardware/external.ts
```diff
@@ -11,29 +11,22 @@ import { task } from "@synnaxlabs/client";
...
 import { type Link } from "@/link";
-import { type Notifications } from "@/notifications";
-import { type Palette } from "@/palette";
 
 export * from "@/hardware/device";
 export * from "@/hardware/rack";
 export * from "@/hardware/task";
 
-export const COMMANDS: Palette.Command[] = [...Device.COMMANDS, ...Task.COMMANDS];
+export const COMMANDS = [...Device.COMMANDS, ...Task.COMMANDS];
```
emilbon99, line 20: Like my comment above, I don't understand getting rid of these type annotations. They are very helpful

emilbon99 (approved):
Overall looks pretty good. Only thing I'm not a fan of is the extensive removal of return types and type annotations. 

I used to have a lint rule that forced the caller to define the return type of a function. I think this rule was very useful, as it makes sure that you're returning a value that matches the expected function signature and doesn't just infer it. This is particularly relevant for functions that have multiple return paths and you need to validate that each return path is confmant.

PR #1116, opened 2025-02-22
SY-2097 - Warn on Significant Clock Skew Between Client and Server

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2097](https://linear.app/synnax/issue/SY-2097)

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client


pjdotson (changes requested):
The code here looks fine, but I think we should reconsider the logic of just comparing the time midpoints of send the request and getting the response - there is an implicit assumption that the time delays in both sending and receiving a message are equivalent. Why not look at an implementation where you create the client timerange with a start and end time, the server time range with a start and end time, and make sure that the server time range is completely encapsulated by the client time range?

Should add this for TS + PY clients as well.

PR #1117, opened 2025-02-22
SY-2099 Fix Dead Rack Warning

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2099](https://linear.app/synnax/issue/SY-2099)

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client

### DIFF
--- a/driver/cmd/exec.cpp
+++ b/driver/cmd/exec.cpp
@@ -12,24 +12,31 @@
 void print_usage() {
     std::cout << "Usage: synnax-driver <command> [options]\n"
             << "Commands:\n"
-            << "  start           Start the Synnax driver service\n"
-            << "    --standalone  Run in standalone mode (not as a service)\n"
-            << "    -s           Short form for --standalone\n"
-            << "  stop            Stop the Synnax driver service\n"
-            << "  restart         Restart the Synnax driver service\n"
-            << "  login           Log in to Synnax\n"
-            << "  install         Install the Synnax driver as a system service\n"
-            << "  uninstall       Uninstall the Synnax driver service\n"
-            << "  logs            View the driver logs\n"
-            << "  version         Display the driver version\n"
-            << "  clear           Clear the persisted state\n";
+            << "  start                     Start the Synnax driver service\n"
+            << "    --standalone/-s         Run in standalone mode (not as a service)\n"
+            << "    --debug                 Enable debug logging\n"
+            << "    --no-color              Disable color output in logs\n"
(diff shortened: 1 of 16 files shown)


driver/task/manager.cpp
```diff
@@ -80,7 +80,7 @@ void task::Manager::stop() {
 
 bool task::Manager::skip_foreign_rack(const TaskKey &task_key) const {
     if (synnax::task_key_rack(task_key) != this->rack.key) {
-        LOG(INFO) << "[driver] received task for foreign rack: " << task_key << ", skipping";
+        VLOG(1) << "[driver] received task for foreign rack: " << task_key << ", skipping";
```
pjdotson, line 83: What does this `VLOG(1)` do?

driver/cmd/exec.cpp
```diff
@@ -12,24 +12,31 @@
...
-            << "  uninstall       Uninstall the Synnax driver service\n"
-            << "  logs            View the driver logs\n"
-            << "  version         Display the driver version\n"
-            << "  clear           Clear the persisted state\n";
+            << "  start                     Start the Synnax driver service\n"
+            << "    --standalone/-s         Run in standalone mode (not as a service)\n"
+            << "    --debug                 Enable debug logging\n"
+            << "    --no-color              Disable color output in logs\n"
+            << "    --disable-sig-stop      Prevent SIGINT and SIGTERM from stopping the driver\n"
+            << "    --disable-stdin-stop    Prevent typing 'STOP' into stdin from stopping the driver\n"
```
pjdotson, line 20: Do we still need to use "STOP" as a command for stopping the driver? and what is the use case for disabling SIGINT from stopping the driver?

pjdotson (changes requested):

driver/task/manager.cpp
emilbon99: It turns this into a debug level log. Basically only gets turned on when `--debug` is true

driver/cmd/exec.cpp
emilbon99: Yes, it's the only way we can correctly manage the embedded driver lifecycle on Windows. If you enable sigint stops on Windows, then doing ctrl+c on the server will also send ctrl+c to the driver at the same time, which means that we don't gracefully shut down the driver in the order that its supposed to be. So we need to manually pipe a "STOP" command instead. Unix has a way to manage with this within go, but Windows does not. So it goes...

pjdotson (approved):

PR #1127, opened 2025-03-03
SY-2115: Fix Virtual Channel Bug

pjdotson:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2115](https://linear.app/synnax/issue/SY-2115)

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client


synnax/pkg/distribution/channel/lease_proxy.go
```diff
@@ -58,15 +58,33 @@ func newLeaseProxy(
 	if err != nil {
 		return nil, err
 	}
+	keyRouter := proxy.BatchFactory[Key]{Host: cfg.HostResolver.HostKey()}
+	externalNonVirtualChannels := make([]Channel, 0)
+	tx := cfg.ClusterDB.OpenTx()
```
emilbon99, line 63: No need to open a tx. `cfg.ClusterDB` already implements tx.

synnax/pkg/distribution/channel/lease_proxy.go
```diff
@@ -58,15 +58,33 @@ func newLeaseProxy(
...
+	if err := gorp.
+		NewRetrieve[Key, Channel]().
+		Where(func(c *Channel) bool { return !c.Internal && !c.Virtual }).
+		Entries(&externalNonVirtualChannels).
+		Exec(context.TODO(), tx); err != nil {
```
emilbon99, line 68: Don't use context.TODO. Add an argument to newLeaseProxy

synnax/pkg/distribution/channel/lease_proxy.go
```diff
@@ -58,15 +58,33 @@ func newLeaseProxy(
...
+
 	p := &leaseProxy{
 		ServiceConfig:   cfg,
 		createRouter:    proxy.BatchFactory[Channel]{Host: cfg.HostResolver.HostKey()},
-		keyRouter:       proxy.BatchFactory[Key]{Host: cfg.HostResolver.HostKey()},
+		keyRouter:       keyRouter,
 		renameRouter:    proxy.BatchFactory[renameBatchEntry]{Host: cfg.HostResolver.HostKey()},
 		leasedCounter:   c,
 		group:           group,
 		externalCounter: extCtr,
```
emilbon99, line 86: Why do we still have the external counter. Do we still use it?

emilbon99 (changes requested):
We should add test cases for this code.

synnax/pkg/distribution/channel/lease_proxy.go
pjdotson: Yes - the external counter still needs to be in aspen so all nodes have access to the
counter. The set.Integer only tracks keys that are leased to that particular node.
Whenever a node joins the cluster, it will reset externalCounter, but other than that
the same functionality applies.

emilbon99 (approved):
LGTM as long as you've verified functionality

PR #1145, opened 2025-03-13
SY-2156 - Properly Handle Disabled Channels in Scale Transform

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2156](https://linear.app/synnax/issue/SY-2156)

## Description

Properly handle disabled channels in scale transform.

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client

### DIFF
--- a/driver/transform/transform.h
+++ b/driver/transform/transform.h
@@ -173,22 +173,23 @@ class Scale final : public Transform {
 
 public:
     explicit Scale(const xjson::Parser &parser, const std::unordered_map<synnax::ChannelKey, synnax::Channel> &channels) {
-        parser.iter("channels", [this, &channels](xjson::Parser &channel_parser) {
-            const auto key = channel_parser.required<synnax::ChannelKey>("channel");
-            auto scale = channel_parser.optional_child("scale");
-            if (!channel_parser.ok()) return;
-            auto ch_t = channels.find(key);
+        parser.iter("channels", [this, &channels](xjson::Parser &channel_p) {
+            const auto key = channel_p.required<synnax::ChannelKey>("channel");
+            const auto enabled = channel_p.optional<bool>("enabled", true);
+            auto scale_p = channel_p.optional_child("scale");
+            if (!channel_p.ok() || !enabled) return;
+            const auto ch_t = channels.find(key);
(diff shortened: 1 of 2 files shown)


driver/transform/transform.h
```diff
@@ -173,22 +173,23 @@ class Scale final : public Transform {
 
 public:
     explicit Scale(const xjson::Parser &parser, const std::unordered_map<synnax::ChannelKey, synnax::Channel> &channels) {
-        parser.iter("channels", [this, &channels](xjson::Parser &channel_parser) {
-            const auto key = channel_parser.required<synnax::ChannelKey>("channel");
-            auto scale = channel_parser.optional_child("scale");
-            if (!channel_parser.ok()) return;
-            auto ch_t = channels.find(key);
+        parser.iter("channels", [this, &channels](xjson::Parser &channel_p) {
```
pjdotson, line 176: Naming something `channel_p` is kind of confusing. When I ask ChatGPT about this it made a comment about suffixing p being common for indicating pointers back when Hungarian notation was a thing. What's the harm with calling it `channel_parser` and below `scale_parser`? The code is much easier to read that way.

driver/transform/transform.h
```diff
@@ -173,22 +173,23 @@ class Scale final : public Transform {
...
+            const auto key = channel_p.required<synnax::ChannelKey>("channel");
+            const auto enabled = channel_p.optional<bool>("enabled", true);
+            auto scale_p = channel_p.optional_child("scale");
```
pjdotson, line 179: It makes sense to me to this variable assignment and initialization down to where it is first used.

pjdotson (changes requested):

driver/transform/transform.h
emilbon99: Ok.

driver/transform/transform.h
emilbon99: This needs to be here because it will accumulate an error internally if it fails to find a "scale" key with a valid format i.e. we need to place it before we check for `channel_p.ok()` which is on the next line.

PR #1146, opened 2025-03-14
SY-2179 - Fix Shutdown Panic: Missing Cancel Error

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2179](https://linear.app/synnax/issue/SY-2179)

## Description

There are some strange implementation details related to fiber websockets. The fiber request context provided is only valid for the lifetime of the http request before we upgrade it to a websocket. We were mistakenly using the context past it's available lifetime. This fix makes it so that we use the underlying server context that is valid throughout the lifetime of the socket. Also makes changes to prevent these sorts of lifecycle mistakes in the future.

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client

### DIFF
--- a/freighter/go/fhttp/stream.go
+++ b/freighter/go/fhttp/stream.go
@@ -293,6 +293,10 @@ func (s *streamServer[RQ, RS]) BindHandler(
 
 const closeReadWriteDeadline = 500 * time.Millisecond
 
+// fiberHandler handles the incoming websocket connection and upgrades the connection
+// to a websocket connection.
+//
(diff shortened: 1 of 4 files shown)


pjdotson (changes requested):
Can you add a test case for this?

emilbon99:
Not possible to add a reliable test case AFAIK. Fiber doesn't provide a guarantee on when the context will be invalidated, as it gets reused across requests.

pjdotson (approved):

PR #1149, opened 2025-03-14
SY-2181 - Python Client Better Communication of Writer Errors

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2181](https://linear.app/synnax/issue/SY-2181)

## Description

Fixes an issue where the python client writer would fail to accumulate errors correctly, and cause difficult to debug issues. Also updates Python client docs to add docs for error handling.

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client

### DIFF
--- a/client/py/synnax/framer/writer.py
+++ b/client/py/synnax/framer/writer.py
@@ -191,21 +191,18 @@ class Writer:
             raise exc
 
     @overload
-    def write(self, channels_or_data: ChannelName, series: CrudeSeries):
-        ...
+    def write(self, channels_or_data: ChannelName, series: CrudeSeries): ...
 
     @overload
     def write(
         self, channels_or_data: ChannelKeys | ChannelNames, series: list[CrudeSeries]
-    ):
-        ...
+    ): ...
 
     @overload
     def write(
         self,
         channels_or_data: CrudeFrame,
-    ):
-        ...
+    ): ...
 
     def write(
         self,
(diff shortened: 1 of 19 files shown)


docs/site/src/pages/reference/python-client/write-data.mdx
```diff
@@ -170,10 +170,12 @@ with client.open_writer(
     channels=["timestamps", "my_precise_tc"],
 ) as writer:
     for i in range(100):
-        writer.write({
+        if not writer.write({
             "timestamps": sy.TimeStamp.now(),
             "my_precise_tc": i,
-        })
+        }):
+            print("Failed to write")
```
pjdotson, line 177: If you stick with code like this you should make sure to actually show printing / throwing the error instead of just printing "Failed to write".

client/py/synnax/framer/writer.py
```diff
@@ -319,8 +329,8 @@ def commit(self) -> tuple[TimeStamp, bool]:
         should acknowledge the error by calling the error method or closing the writer.
         After the error is acknowledged, the caller can attempt to commit again.
         """
-        if self.__stream.received():
-            return TimeStamp.ZERO, False
+        if self._check_for_bad_ack():
+            return TimeStamp.now(), False
```
pjdotson, line 333: Why not throw the exception from `self.error` here?

pjdotson (changes requested):
Update the python client examples as well with the `if not writer.write` pattern

docs/site/src/pages/reference/python-client/write-data.mdx
emilbon99: This is not how the writer protocol currently works. The `write` method returns false when there's an error, and then the error can be discovered on writer close. We we want to change the protocol, we should have a discussion about it and turn it into a larger refactor.

client/py/synnax/framer/writer.py
emilbon99: Because it doesn't adhere to the way we structure the writer protocol in Synnax. We can talk about a long term change to the protocol that changes the error handling semantics. The intention with this PR is just to make sure that the error are communicated correctly.

emilbon99:
Updated python client examples as well

pjdotson (approved):

PR #1154, opened 2025-03-15
SY-1116 - Fix Bleve Race Condition on Server Startup

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-1116](https://linear.app/synnax/issue/SY-1116)

## Description

Fixes an issue where bleve would panic due to a concurrent map read and write on startup. This fixes the issue by indexing all resources after services have been indexed in the ontology. See issue for more details. 

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client

### DIFF
--- a/synnax/pkg/distribution/ontology/ontology.go
+++ b/synnax/pkg/distribution/ontology/ontology.go
@@ -26,8 +26,6 @@ package ontology
 
 import (
 	"context"
-	"io"
-
 	"github.com/synnaxlabs/alamos"
 	"github.com/synnaxlabs/synnax/pkg/distribution/ontology/schema"
 	"github.com/synnaxlabs/synnax/pkg/distribution/ontology/search"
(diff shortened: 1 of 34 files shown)


pjdotson:
@emilbon99 Failing tests

emilbon99:
Fixed race condition

synnax/pkg/distribution/ontology/ontology.go
```diff
@@ -116,9 +115,6 @@ func Open(ctx context.Context, configs ...Config) (*Ontology, error) {
```
pjdotson, line 115: In the above set of lines, if the `err` from `o.NewWriter(cfg.DB).DefineResource(ctx, RootID)` is not `nil`, then it won't get immediately returned. That should be fixed:

```go
	err = o.NewRetrieve().WhereIDs(RootID).Exec(ctx, cfg.DB)
	if errors.Is(err, query.NotFound) {
		err = o.NewWriter(cfg.DB).DefineResource(ctx, RootID)
	}
	if err != nil {
	    return nil, err
	}
	```

pjdotson (changes requested):

synnax/pkg/distribution/ontology/ontology.go
emilbon99: Fixed

pjdotson (approved):

PR #1165, opened 2025-03-16
SY-2104 - Improved Driver & Server Early Shutdown Mechanics

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2104](https://linear.app/synnax/issue/SY-2104)

## Description

Improves the early shutdown mechanics for both the driver and server. Also when through and verified that the TLS certificate setup works correctly. 

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client

### DIFF
--- a/driver/task/manager.cpp
+++ b/driver/task/manager.cpp
@@ -43,6 +43,7 @@ xerrors::Error task::Manager::open_streamer() {
             this->channels.task_delete = channel;
         else if (channel.name == TASK_CMD_CHANNEL) this->channels.task_cmd = channel;
 
+    if (this->exit_early) return xerrors::NIL;
     std::lock_guard lock{this->mu};
     auto [s, open_err] = this->ctx->client->telem.open_streamer(StreamerConfig{
         .channels = {
@@ -72,6 +73,7 @@ xerrors::Error task::Manager::configure_initial_tasks() {
 }
 
 void task::Manager::stop() {
+    this->exit_early = true;
     std::lock_guard lock{this->mu};
     // Very important that we do NOT set the streamer to a nullptr here, as the run()
     // method still needs access before shutting down.
@@ -88,7 +90,12 @@ bool task::Manager::skip_foreign_rack(const TaskKey &task_key) const {
 }
 
 xerrors::Error task::Manager::run(std::promise<void> *started_promise) {
+    if (this->exit_early) return xerrors::NIL;
     if (const auto err = this->configure_initial_tasks()) return err;
+    if (this->exit_early) {
+        this->stop_all_tasks();
+        return xerrors::NIL;
+    }
     if (const auto err = this->open_streamer()) return err;
     LOG(INFO) << xlog::GREEN() << "[driver] started successfully" << xlog::RESET();
     if (started_promise != nullptr) started_promise->set_value();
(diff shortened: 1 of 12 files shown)


synnax/cmd/start.go
```diff
@@ -251,14 +272,17 @@ func start(cmd *cobra.Command) {
...
 
 		// Provision the root user.
-		if err = maybeProvisionRootUser(ctx, gorpDB, authenticator, userSvc, rbacSvc); err != nil {
+		if err := maybeProvisionRootUser(ctx, gorpDB, authenticator, userSvc, rbacSvc); err != nil || ctx.Err() != nil {
 			return err
 		}
 
 		// Set the base permissions for all users.
-		if err = maybeSetBasePermission(ctx, gorpDB, rbacSvc); err != nil {
+		if err := maybeSetBasePermission(ctx, gorpDB, rbacSvc); err != nil || ctx.Err() != nil {
```
pjdotson, line 285: Just need the assignment, no need to redeclare `err`

synnax/cmd/start.go
```diff
@@ -129,9 +130,8 @@ func start(cmd *cobra.Command) {
 	// Perform the rest of the startup within a separate goroutine, so we can properly
 	// handle signal interrupts.
 	sCtx.Go(func(ctx context.Context) (err error) {
-
 		secProvider, err := configureSecurity(ins, insecure)
-		if err != nil {
+		if err != nil || ctx.Err() != nil {
 			return err
```
pjdotson, line 135: I wonder if `ctx` should have an `isError` method that just returns `ctx.Err() != nil`. Obviously the `err != nil` pattern is very idiomatic, but if `ctx.Error()` is mostly used to determine if you should return early (and the result isn't actually returned), it might make sense.

pjdotson: Also might be worth extending the above comment to explain why we are checking for `ctx.Err()` so often

driver/task/manager.cpp
```diff
@@ -88,7 +90,12 @@ bool task::Manager::skip_foreign_rack(const TaskKey &task_key) const {
 }
 
 xerrors::Error task::Manager::run(std::promise<void> *started_promise) {
+    if (this->exit_early) return xerrors::NIL;
```
pjdotson, line 93: It's interesting that you decided to use this Go-like pattern of returning an error instead of throwing it in all of the C++ code. Obviously there is no reason to change it now because all of the code works and we are the only ones using the C++ client library, but I wonder if it makes sense to keep this pattern when we make the library public and others start using it.

pjdotson (changes requested):
A few minor requests

driver/task/manager.cpp
emilbon99: From my reading, very few companies actually use C++'s standard exception mechanisms for error handling. It's more common to return errors as values or using status codes. For example, NI DAQmx and LabJack return status codes and not errors.

emilbon99: Many companies have their own internal error handling libraries like ours - it's a mess.

synnax/cmd/start.go
emilbon99: context.Context is part of the go STL. Nothing we can do to add methods or change signature. I will add a doc.

pjdotson (approved):

PR #1166, opened 2025-03-16
SY-2166 - Task Rename Synchronization in Console

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2166](https://linear.app/synnax/issue/SY-2166)

## Description

Made it so that renaming a task synchronizes properly across the layout mosaic, task form, and resources toolbar.

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client

### DIFF
--- a/console/src/layout/selectors.ts
+++ b/console/src/layout/selectors.ts
@@ -64,6 +64,18 @@ export const selectAltKey = (state: StoreState, key: string): string | undefined
 export const useSelectAltKey = (key: string): string | undefined =>
   useMemoSelect((state: StoreState) => selectAltKey(state, key), [key]);
 
+const selectName = (state: StoreState, key: string): string | undefined =>
+  select(state, key)?.name;
+
+export const useSelectName = (key: string): string | undefined =>
+  useMemoSelect((state: StoreState) => selectName(state, key), [key]);
+
+export const selectByFilter = (
+  state: StoreState,
+  filter: (layout: State) => boolean,
+): State | undefined =>
+  Object.values(selectSliceState(state).layouts).find((l) => filter(l));
+
 /**
  * Selects a layout from the store by key.
  *
--- a/console/src/hardware/common/task/Form.tsx
+++ b/console/src/hardware/common/task/Form.tsx
@@ -19,7 +19,8 @@ import {
 } from "@synnaxlabs/pluto";
 import { TimeSpan, type UnknownRecord } from "@synnaxlabs/x";
 import { useMutation } from "@tanstack/react-query";
-import { type FC } from "react";
+import { type FC, useEffect } from "react";
+import { useDispatch } from "react-redux";
 import { z } from "zod";
 
 import { CSS } from "@/css";
@@ -37,7 +38,7 @@ import {
 import { RUNNING_STATUS, type StartOrStopCommand } from "@/hardware/common/task/types";
 import { useCreate } from "@/hardware/common/task/useCreate";
 import { type StateDetails, useState } from "@/hardware/common/task/useState";
-import { type Layout } from "@/layout";
+import { Layout } from "@/layout";
 
 export type Schema<Config extends UnknownRecord = UnknownRecord> = z.ZodObject<{
   name: z.ZodString;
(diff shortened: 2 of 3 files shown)


console/src/layout/selectors.ts
```diff
@@ -64,6 +64,18 @@ export const selectAltKey = (state: StoreState, key: string): string | undefined
...
+  select(state, key)?.name;
+
+export const useSelectName = (key: string): string | undefined =>
+  useMemoSelect((state: StoreState) => selectName(state, key), [key]);
+
+export const selectByFilter = (
+  state: StoreState,
+  filter: (layout: State) => boolean,
+): State | undefined =>
+  Object.values(selectSliceState(state).layouts).find((l) => filter(l));
```
pjdotson, line 77:
```suggestion
  Object.values(selectSliceState(state).layouts).find(filter);
```

pjdotson (changes requested):
1. Add synchronization when renaming from task toolbar
2. Should we prompt a message about the task restarting if it is currently running?
3. For task snapshots, what should renaming the tab do? Should you be able to rename the tab?

emilbon99:
1. Done.
2. If you rename from the tab it won't actually re-configure the task. You need to hit the configure button to save. I think that's how it should be.
3. I disabled renaming the tab for snapshot.s

synnax/pkg/api/grpc/v1/hardware.pb.go
```diff
@@ -309,6 +309,7 @@ type Task struct {
 	Type          string                 `protobuf:"bytes,3,opt,name=type,proto3" json:"type,omitempty"`
 	Config        string                 `protobuf:"bytes,4,opt,name=config,proto3" json:"config,omitempty"`
 	Internal      bool                   `protobuf:"varint,5,opt,name=internal,proto3" json:"internal,omitempty"`
+	Snapshot      bool                   `protobuf:"varint,6,opt,name=snapshot,proto3" json:"snapshot,omitempty"`
```
pjdotson, line 312: Should you also include the task state here?

driver/task/manager.cpp
```diff
@@ -124,6 +125,10 @@ void task::Manager::process_task_set(const telem::Series &series) {
         }
 
         auto [sy_task, err] = this->rack.tasks.retrieve(task_key);
+        if (sy_task.snapshot) {
+            VLOG(1) << "[driver] ignoring snapshot task";
```
pjdotson, line 129: Should you add the task name here?

driver/task/manager.cpp
```diff
@@ -58,8 +58,9 @@ xerrors::Error task::Manager::open_streamer() {
 
 xerrors::Error task::Manager::configure_initial_tasks() {
     auto [tasks, tasks_err] = this->rack.tasks.list();
-    if (tasks_err)return tasks_err;
+    if (tasks_err) return tasks_err;
     for (const auto &task: tasks) {
+        if (task.snapshot) continue;
```
pjdotson, line 63: Add a test for this as well

console/src/hardware/task/layouts.ts
```diff
@@ -43,5 +43,10 @@ export const retrieveAndPlaceLayout = async (
   if (client == null) throw NULL_CLIENT_ERROR;
   const t = await client.hardware.tasks.retrieve(key);
   const layout = createLayout(t);
+  if (t.snapshot)
+    layout.tab = {
+      ...layout.tab,
```
pjdotson, line 48: Are there weird mechanics here if `layout.tab` was previously undefined? I'm not sure if
that would be trouble or not.

pjdotson (changes requested):

synnax/pkg/api/grpc/v1/hardware.pb.go
emilbon99: I think that's an eventually added. It's not built into client side logic rn so I don't think it's worth the add in this branch. Can add when needed.

driver/task/manager.cpp
emilbon99: Added

driver/task/manager.cpp
emilbon99: added

synnax/pkg/api/grpc/v1/hardware.pb.go
pjdotson: Fine with me

PR #1189, opened 2025-03-28
SY-2245: Fix Analog Sample Acquisition

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2245](https://linear.app/synnax/issue/SY-2245)

## Description

Another fix for analog sample acquisition clock skew. Gives up on using a stable waveform from the device and instead calculates timestamps using the system clock.

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client


driver/task/common/sample_clock.h
```diff
@@ -70,20 +70,17 @@ class HardwareTimedSampleClock final : public SampleClock {
     }
 
     void reset() override {
-        this->high_water = telem::TimeStamp::now();
+        this->high_water = telem::TimeStamp(0);
```
pjdotson, line 73: why not call `telem::TimeStamp::now()` here?

pjdotson (changes requested):

pjdotson:
@emilbon99 Fix failing test as well.

driver/task/common/sample_clock.h
emilbon99: We use a timestamp of 0 to check whether we're on the first read. If we are we skip. I know it's weird, but it lets the DAQmx task start regulating the timing.

x/cpp/telem/series.h
```diff
@@ -731,7 +731,7 @@ class Series {
         if (count == 1) return Series(start);
         Series s(TIMESTAMP_T, count);
         if (count == 0) return s;
-        const auto step = (end - start) / (static_cast<int64_t>(count) - 1);
+        const auto step = (end - start) / (static_cast<int64_t>(count));
```
pjdotson, line 734: If you think this is what the behavior should be than you need to make sure the documentation says it is end exclusive, and probably show an example to people know what you are talking about. Personally I think you should add a boolean option to include the endpoint or not like numpy: https://numpy.org/doc/2.1/reference/generated/numpy.linspace.html

x/cpp/loop/loop.h
```diff
@@ -123,16 +123,17 @@ class Gauge {
     telem::TimeSpan total_duration{0};
     telem::TimeSpan min_duration{std::numeric_limits<int64_t>::max()};
     telem::TimeSpan max_duration{0};
-    telem::TimeStamp curr_start = telem::TimeStamp(0);
+    std::chrono::time_point<std::chrono::high_resolution_clock> curr_start;
 
 public:
     void start() {
-        curr_start = telem::TimeStamp::now();
+        curr_start = std::chrono::high_resolution_clock::now();
```
pjdotson, line 130: What do you think about instead being able to call `telem::TimeStamp::now(true)` with an argument for high precision that wraps `std::chrone::high_resolution_clock`?

driver/opc/read_task.h
```diff
@@ -230,7 +230,7 @@ class ArrayReadTaskSource final : public BaseReadTaskSource {
             fr.emplace(ch.synnax_key, std::move(s));
         }
         auto start = telem::TimeStamp::now();
-        auto end = start + this->cfg.array_size * this->cfg.sample_rate.period();
+        auto end = start + (this->cfg.array_size + 1) * this->cfg.sample_rate.period();
```
pjdotson, line 233: See comment in `series.h` to allow `linspace` to accept an option and then change `generate_index_data` and this section of code.

driver/task/common/sample_clock.h
pjdotson: Okay, could you add a comment explaining this?

pjdotson (changes requested):

x/cpp/telem/series.h
emilbon99: Done

x/cpp/telem/series.h
```diff
@@ -722,16 +722,20 @@ class Series {
     /// @param start The starting timestamp
     /// @param end The ending timestamp
     /// @param count The number of points to generate
+    /// @param inclusive Whether or not to include the end timestamp as the last value
+    /// in th series.
```
pjdotson, line 726: typo

x/cpp/telem/series.h
```diff
@@ -722,16 +722,20 @@ class Series {
...
+        const bool inclusive = false
     ) {
         if (count == 1) return Series(start);
         Series s(TIMESTAMP_T, count);
         if (count == 0) return s;
-        const auto step = (end - start) / (static_cast<int64_t>(count) - 1);
+        const auto adjusted_count = inclusive ? count - 1 : count;
+        const auto step = (end - start) / adjusted_count;
         for (size_t i = 0; i < count; i++) s.write(start + step * i);
         s.size_ = count;
```
pjdotson, line 740: size should be `adjusted_count`

pjdotson (changes requested):
A couple minor fixes then you are good to merge

x/cpp/telem/series.h
emilbon99: Size should not be adjusted count. inclusive just means whether the linspace calculation should have the end as its last value or the sample just before the end as its last value.

PR #1197, opened 2025-03-31
SY-2300: Fix Sequencing Shutdown Mechanics

emilbon99:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2300](https://linear.app/synnax/issue/SY-2300)

## Description

Makes it so that sequences shutdown gracefully even after failed startup.

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client

### DIFF
--- a/driver/sequence/task.cpp
+++ b/driver/sequence/task.cpp
@@ -33,6 +33,8 @@ sequence::Task::Task(
 
 void sequence::Task::run() {
     if (const auto err = this->seq->begin(); err) {
+        if (const auto end_err = this->seq->end())
+            LOG(ERROR) << "[sequence] failed to end after failed start:" << end_err;
         this->state.variant = "error";
         this->state.details["running"] = false;
         this->state.details["message"] = err.message();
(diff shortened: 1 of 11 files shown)


driver/sequence/task.cpp
```diff
@@ -33,6 +33,8 @@ sequence::Task::Task(
 
 void sequence::Task::run() {
     if (const auto err = this->seq->begin(); err) {
+        if (const auto end_err = this->seq->end())
+            LOG(ERROR) << "[sequence] failed to end after failed start:" << end_err;
```
pjdotson, line 37: Should we also be writing `end_err` to the state as well?

pjdotson: And you should add a test for this

pjdotson (changes requested):

driver/sequence/task.cpp
emilbon99: I don't think so. I think the log here is the most useful, as the start error is what we really want to communicate.

emilbon99: https://linear.app/synnax/issue/SY-2309/improve-sequence-task-tests

PR #1219, opened 2025-04-07
SY-2350: Additional Schematic Symbols

evanhekman:
# Pull Request

Go to the `Preview` tab and select the appropriate template for your pull request:

- [Issue](?expand=1&template=issue.md)
- [RC](?expand=1&template=rc.md)


emilbon99 (approved):
Notes:

- Corner rounding on polygon does not work.
- Nozzle preview should be smaller.
- Side length and corner rounding should be on the same line as rotation and Number of Sides.
- Make '# of Sides' label 'Number of Sides'
- Use the unicode ° symbol instead of 'degrees'

emilbon99 (changes requested):
Notes:

Corner rounding on polygon does not work.
Nozzle preview should be smaller.
Side length and corner rounding should be on the same line as rotation and Number of Sides.
Make '# of Sides' label 'Number of Sides'
Use the unicode ° symbol instead of 'degrees'

evanhekman:
- fixed corner rounding for Polygon
- fixed formatting for Polygon Form
- fixed size of Nozzle preview
Also added in memoization for Polygon path/rounding calculations - was noticing a minor slowdown in console if I tried to drag around multiple shapes at the same time that was no longer noticeable once I added the memoization.

pjdotson (changes requested):
"Number of Sides" box on the Polygon shouldn't show "px"
image

Otherwise it looks excellent

evanhekman:
Sweet, thanks for the review! Would say double checking I didn't break the rules of hooks under Polygon would be worth doing, I think the majority of changes are easy to visually confirm by looking at the console. Would you want any reordering or organization inside registry.ts?

pluto/src/vis/schematic/primitives/Primitives.tsx
```diff
@@ -1315,6 +1384,186 @@ export const Tank = ({
...
+}: PolygonProps): ReactElement => {
+  const theme = Theming.use();
+  const padding = useMemo(() => 2 * ((strokeWidth ?? 2) + 1), [strokeWidth]);
+  const vertices = useMemo(
+    () => calculatePolygonVertices(numSides, sideLength, rotation, padding),
+    [numSides, sideLength, rotation, padding],
+  );
+  const path = useMemo(
+    () => generateRoundedPolygonPath(vertices, sideLength, cornerRounding ?? 0),
+    [vertices, cornerRounding],
```
pjdotson, line 1481: Forgot to include `sideLength` in the dependencies array here.

PR #1235, opened 2025-04-17
SY-2359 - Make `x/status` package in all languages

pjdotson:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-2359](https://linear.app/synnax/issue/SY-2359/make-statusvariant-x-package-in-all-languages)

## Description

Abstracts out all status packages to use the same four common statuses (info, success, warning, error) in all languages.

## Basic Readiness

### Data Structures

- [x] Server
- [x] Console

### API Changes

- [x] Python Client
- [x] Server
- [x] TypeScript Client


x/ts/src/status/types.ts
```diff
@@ -0,0 +1,23 @@
...
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { z } from "zod";
+
+export const SUCCESS_VARIANT = "success";
```
emilbon99, line 12: Assigning union value types to generic utilities like this in typescript isn't useful IMO. It's kind if like having direction="x" and direction={direction.X} inside of an Align.Space. The typescript compiler will already restrict the union types to a point where "x" is as type safe as {direction.X} and we don't introduce a whole new variable. Obv this is not the case in C++ or Go.

emilbon99 (changes requested):

emilbon99 (changes requested):

synnax/pkg/service/hardware/tracker/tracker.go
```diff
@@ -43,10 +43,10 @@ import (
 	"go.uber.org/zap"
 )
 
-// RackState is the state of a hardware rack. Unfortunately, we can't put this into
-// the rack package because it would create a circular dependency.
+// RackState is the state of a hardware rack. Unfortunately, we can't put this into the
+// rack package because it would create a circular dependency.
 type RackState struct {
-	rack.State
+	*rack.State
```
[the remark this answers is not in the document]
pjdotson: Yeah, if this is a struct instead of a pointer, then retrieving the state from a client library won't give you null / undefined / something similar, but instead give you a state where all of the fields are 0.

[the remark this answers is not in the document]
pjdotson: Actually, the problem is more in the API layer. If someone calls retrieve rack /
hardware / device / etc. and includeState is false, then the returned value will have
the 0 initialized state. This causes problems with the parsing because the status field
will be an empty string.

client/py/synnax/hardware/task/payload.py
```diff
@@ -22,12 +24,16 @@ class TaskPayload(Payload):
 
 class TaskState(Payload):
     task: int = 0
-    variant: str = ""
+    variant: Variant = ""
```
emilbon99, line 27: Wouldn't this violate `mypy` typing? As this is not a valid variant.

client/py/synnax/hardware/task/payload.py
```diff
@@ -22,12 +24,16 @@ class TaskPayload(Payload):
...
     key: str = ""
     details: dict
 
     def __init__(
-        self, task: int = 0, variant: str = "", key: str = "", details: dict | str = ""
+        self,
+        task: int = 0,
+        variant: Variant = "",
```
emilbon99, line 34: Same here

client/py/synnax/status/variant.py
```diff
@@ -0,0 +1,18 @@
...
+#  included in the file licenses/APL.txt.
+
+from typing import Literal
+
+SUCCESS_VARIANT = "success"
+INFO_VARIANT = "info"
+WARNING_VARIANT = "warning"
+ERROR_VARIANT = "error"
+
+Variant = Literal["success", "info", "warning", "error"]
```
emilbon99, line 17: Since we're using a literal everywhere, do we need the constants? Don't have a super strong opinion in the python case, but it seems that if we're restricting the arguments to functions via literals, then we don't need the constants ya know.

client/ts/src/hardware/device/payload.ts
```diff
@@ -36,7 +36,10 @@ export const deviceZ = z.object({
   location: z.string(),
   configured: z.boolean().optional(),
   properties: z.record(z.unknown()).or(z.string().transform(decodeJSONString)),
-  state: stateZ.optional(),
+  state: stateZ
```
emilbon99, line 39: Assuming this is both optional and nullable because sometimes you get null and sometimes undefined from the server? How lovely...

client/ts/src/hardware/rack/payload.ts
```diff
@@ -16,17 +16,20 @@ export type Key = z.infer<typeof keyZ>;
...
+  lastReceived: TimeStamp.z,
 });
 
 export interface State extends z.infer<typeof stateZ> {}
 
 export const rackZ = z.object({
   key: keyZ,
   name: z.string(),
-  state: stateZ.optional(),
+  state: stateZ
```
emilbon99, line 29: Since this pattern is used in several places it might be worth putting it in a common function kind of like we do with `nullableArrayZ`

emilbon99 (commented):
Refactor status to use disabled and loading

pluto/src/form/Form.tsx
```diff
@@ -50,8 +50,8 @@ export interface UseNullableFieldProps<I, O = I>
 export interface UseFieldReturn<I extends Input.Value, O extends Input.Value = I>
   extends FieldState<I> {
   onChange: (value: O) => void;
-  setStatus: (status: status.CrudeSpec) => void;
-  status: status.CrudeSpec;
+  setStatus: (status: Status.CrudeSpec) => void;
```
emilbon99, line 53: Why is `CrudeSpec` in pluto>?

pluto/src/status/Circle.tsx
```diff
@@ -8,15 +8,18 @@
...
 import { VARIANT_COLORS } from "@/status/colors";
 
 export interface CircleProps extends IconProps {
   variant?: status.Variant;
 }
 
-export const Circle = ({ variant = "info", ...rest }: CircleProps): ReactElement => (
-  <Icon.Circle color={VARIANT_COLORS[variant]} {...rest} />
-);
+export const Circle = ({ variant = "info", ...rest }: CircleProps): ReactElement =>
```
emilbon99, line 20: This component is probably better called `Indicator` right. `Circle` is just weird

pluto/src/table/aether/table.ts
emilbon99, line 1: Not sure if making all these lower case is the right move, as we usually have only one aether component in there that is the parallel of `Table.tsx`. That way there's an easy correlation between `Table.tsx` and `aether/Table.ts`. 

Opinion is not super strong here

synnax/pkg/service/hardware/tracker/tracker.go
```diff
@@ -158,51 +165,67 @@ func (c Config) Validate() error {
...
 	t.mu.Devices = make(map[string]device.State)
 
 	for _, r := range racks {
-		// Initialize rack state with empty maps
 		rck := &RackState{
+			State: rack.State{
+				Key:          r.Key,
+				LastReceived: telem.Now(),
+				Variant:      status.WarningVariant,
+				Message:      "Unknown rack state",
```
emilbon99, line 199: These would ready better as `x state unknown` as a user.

synnax/pkg/service/hardware/tracker/tracker.go
```diff
@@ -458,13 +500,18 @@ func (t *Tracker) handleRackChanges(ctx context.Context, r gorp.TxReader[rack.Ke
...
-				nState.LastReceived = telem.Now()
-				nState.Key = c.Key
-				t.mu.Racks[c.Key] = nState
-			}
+			continue
+		}
+		if _, rackExists := t.mu.Racks[c.Key]; !rackExists {
+			t.mu.Racks[c.Key] = &RackState{
+				Tasks: make(map[task.Key]task.State),
+				State: rack.State{
```
emilbon99, line 508: These default states should be consolidated into standard functions.

emilbon99: That way we don't need to modify every time we want to change messaging.

emilbon99 (changes requested):
LGTM once my comments are addressed

pluto/src/form/Form.tsx
pjdotson: Why shouldn't it be?

PR #1245, opened 2025-05-07
Sy 2225 improve console data sharing mechanism

pjdotson:
# Issue Pull Request

## Key Information

- **Linear Issue**: [SY-####](https://linear.app/synnax/issue/)

## Basic Readiness

### Data Structures

- [ ] Server
- [ ] Console

### API Changes

- [ ] Python Client
- [ ] Server
- [ ] TypeScript Client


client/ts/src/framer/streamer.ts
```diff
@@ -70,6 +70,14 @@ export class Streamer implements AsyncIterator<Frame>, AsyncIterable<Frame> {
     return this.adapter.adapt(new Frame((await this.stream.receive()).frame));
   }
 
+  updateKeys(keys: channel.Key | channel.Keys): void {
```
emilbon99, line 73: See above comment. Why does this function need to exist? Increases API surface area on a critical component of the public client. Is the point that you don't want an async function? I don't see this used anywhere else in the codebase.

client/ts/src/hardware/task/client.ts
```diff
@@ -411,6 +398,34 @@ export class Client implements AsyncTermSearcher<string, Key, Payload> {
...
+    timeout: CrudeTimeSpan,
+    args?: Args,
+    name?: string,
+  ): Promise<State<Details>> {
+    return await executeCommandSync(
+      this.frameClient,
+      task,
+      type,
+      timeout,
+      name ?? (await this.retrieve(task)).name,
```
emilbon99, line 424: This function only ever uses `name` if it encounters an error, yet we're doing a `retrieve` without caching on every single call to execute command? Needs to be a better way to handle this without needing to fetch every time.

client/ts/src/hardware/task/client.ts
```diff
@@ -460,3 +475,73 @@ export class Client implements AsyncTermSearcher<string, Key, Payload> {
...
+
+const executeCommand = async <Args extends {} = UnknownRecord>(
+  frameClient: framer.Client | null,
+  task: Key,
+  type: string,
+  args?: Args,
+): Promise<string> => {
+  if (frameClient == null) throw NOT_CREATED_ERROR;
+  const key = id.create();
+  await frameClient.write(TimeStamp.now(), COMMAND_CHANNEL_NAME, [
```
emilbon99, line 487: Did you test if this actually works? `frameClient.write` uses writer mode persist only by default, so unless there's a bug I don't think this will actually get propagated to devices.

client/ts/src/hardware/task/client.ts
```diff
@@ -460,3 +475,73 @@ export class Client implements AsyncTermSearcher<string, Key, Payload> {
...
+  task: Key,
+  type: string,
+  timeout: CrudeTimeSpan,
+  tskName: string | Promise<string>,
+  args?: Args,
+): Promise<State<Details>> => {
+  if (frameClient == null) throw NOT_CREATED_ERROR;
+  const streamer = await frameClient.openStreamer(STATE_CHANNEL_NAME);
+  const cmdKey = await executeCommand(frameClient, task, type, args);
+  const timeOut = new TimeSpan(timeout);
```
emilbon99, line 507: Wild naming convention. Just name it `parsedTimeout` or something

client/ts/src/hardware/task/client.ts
```diff
@@ -460,3 +475,73 @@ export class Client implements AsyncTermSearcher<string, Key, Payload> {
...
+    }
+  } finally {
+    clearTimeout(timeoutID);
+    streamer.close();
+  }
+};
+
+const formatTimeoutError = async (
+  type: string,
+  name: string | Promise<string>,
```
emilbon99, line 530: I don't think there's anywhere where you actually return a promise here.

emilbon99: So making this function async is not very useful

client/ts/src/ranger/client.ts
```diff
@@ -183,10 +186,19 @@ export class Range {
     wrapper.setCloser(async () => await base.close());
     return wrapper;
   }
-}
 
-export const sort = (a: Range, b: Range): -1 | 0 | 1 =>
-  sortTimeRange(a.timeRange, b.timeRange);
+  static readonly convertOntologyResourceToPayload = (
```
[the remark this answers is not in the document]
emilbon99: I don't know I'm kind of skewed on whether it should be a standalone fucntion or not. NOt a big deal, but they should at least be methods and not arrow functions

console/src/channel/useListenForCalculationState.ts
```diff
@@ -0,0 +1,45 @@
...
+      handleError(async () => {
+        if (client == null) throw NULL_CLIENT_ERROR;
+        try {
+          const { name } = await client.channels.retrieve(key);
+          addStatus({
+            ...baseStatus,
+            description: message,
+            message: `Calculation for ${name} failed`,
+          });
+        } catch {
```
emilbon99, line 37: If we're already using `handleError` what's the point of this try catch?

emilbon99: Or just get rid of the handleError if you want to use the base status.

console/src/range/Explorer.tsx
```diff
@@ -190,7 +190,7 @@ const ChangeLoader = () => {
   const { setSourceData } = List.useDataUtils<string>();
   const client = Synnax.use();
   useAsyncEffect(async () => {
-    const obs = await client?.ranges.openTracker();
+    const obs = await client?.ranges.openTracker(); //todo
```
emilbon99, line 193: todo?

pluto/src/hardware/task/useCommandExecutor.ts
```diff
@@ -0,0 +1,99 @@
...
+        details: {
+          message: `${caseconv.capitalize(type)} command to ${
+            taskName ?? `task with key ${taskKey}`
+          } timed out after ${timeout.toString()}`,
+        },
+      });
+      setKey(null);
+    }, timeout.milliseconds);
+  };
+  const addListener = Synch.useAddListener();
```
emilbon99, line 72: Any reason why you're not using the `useParsedListener` hook here? Seems like the perfect use case for it

pluto/src/channel/AliasContext.ts
```diff
@@ -0,0 +1,79 @@
...
+): [string, (newName: string) => Promise<void>] => {
+  const currentAlias = useAlias(key);
+  const { getName, setAlias } = useContext();
+  const [name, setName] = useState<string | undefined>(defaultName);
+  const client = Synnax.use();
+  useAsyncEffect(async () => {
+    const n = await getName(key);
+    setName(n);
+  }, [key, getName]);
+  const rename = useCallback(
```
emilbon99, line 61: This seems like the ideal place to use a mutation right? Instead of implementing our own undo logic.

emilbon99 (changes requested):
This is a good step forward and I think it's close to being able to be merged. That said, there's some things we should still think about:

1. What if I just want a hook called `Channel.use` that always gives me the most up to date version of a channel? I think this is where the power really comes in, because it guarantees that every place we display a channel's properties (such as the vis selector, or edit calculation dialog), we can make sure we always have the most up to date value without needing to keep track of making sure we've set up all of `Synch` listeners.

2. What's the decision on `useMutation` and `useQuery`. There is no consistency between using `useQuery` and then just raw dogging fetches inside of an async effect in other places. There's also some spots where we make modifications and don't use a mutation? Seems like there's a lot of simplification we can do in these places, especially surrounding forms, which is a major pain in the ass rn.

3. As far as a bunch of tiny files goes, I will lodge a small complain against it. There are a bunch of places where a file has 6 lines and is next to 5 or so other logically related files that basically all import from the same place and do very similar things. I don't think that these should be broken up. For example, just put all of the 5 line synchronization hooks into a file called `useSycnchronizers` or `synchronizers`. All for splitting up files when hooks are large and not logically related, but seems excessive in many cases here.

x/ts/src/telem/telem.ts
```diff
@@ -1170,15 +1176,10 @@ export class TimeRange implements Stringer {
...
-}
 
-export const sortTimeRange = (a: TimeRange, b: TimeRange): -1 | 0 | 1 => {
-  if (a.start.before(b.start)) return -1;
-  if (a.start.after(b.start)) return 1;
-  if (a.end.before(b.end)) return -1;
-  if (a.end.after(b.end)) return 1;
-  return 0;
-};
+  static readonly sort = (a: TimeRange, b: TimeRange): number =>
```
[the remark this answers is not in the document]
pjdotson: Sort by start first and then end. Added docs + spec

client/ts/src/framer/adapter.ts
```diff
@@ -33,11 +34,15 @@ export class ReadAdapter {
     return adapter;
   }
 
+  updateKeys(keys: channel.Key | channel.Keys): void {
```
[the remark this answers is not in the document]
pjdotson: Removed, was for an earlier implementation that I wanted to be synchronous but didn't
end up needing.

client/ts/src/framer/streamer.ts
pjdotson: Removed

client/ts/src/hardware/task/client.ts
pjdotson: lol, didn't even realize that

client/ts/src/hardware/task/client.ts
pjdotson: It is async to use the functionality for waiting for the name. Should I just remove this
functionality - it adds a lot of complexity to the code just to make an error message
easier to read in one case that we don't actually use?

client/ts/src/hardware/task/client.ts
pjdotson: Fixed but see below comment

client/ts/src/hardware/task/client.ts
pjdotson: Changed and added tests

pluto/src/hardware/task/useCommandExecutor.ts
pjdotson: This was dead code that I ended up not using

console/src/channel/useListenForCalculationState.ts
pjdotson: To add the separate message to get the name of the channel

console/src/hardware/common/task/ParentRangeButton.tsx
```diff
@@ -19,36 +18,21 @@ export interface ParentRangeButtonProps {
...
 }
 
+const filter = (resource: ontology.Resource): boolean =>
+  resource.id.type === ranger.ONTOLOGY_TYPE;
+
 export const ParentRangeButton = ({ taskKey }: ParentRangeButtonProps) => {
-  const client = Synnax.use();
-  const handleError = Status.useErrorHandler();
-  const [parent, setParent] = useState<ontology.Resource>();
+  const parentRange = Ontology.useParents(task.ontologyID(taskKey), filter)?.find(
```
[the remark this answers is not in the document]
pjdotson: To me it makes sense to keep this in the console because it relies on the fact that if a
task has a range as a parent, then it is snapshotted to it. Which is a constraint we
have made on the console, not on the server + client libraries.

pjdotson:
> This is a good step forward and I think it's close to being able to be merged. That said, there's some things we should still think about:
> 
> 1. What if I just want a hook called `Channel.use` that always gives me the most up to date version of a channel? I think this is where the power really comes in, because it guarantees that every place we display a channel's properties (such as the vis selector, or edit calculation dialog), we can make sure we always have the most up to date value without needing to keep track of making sure we've set up all of `Synch` listeners.
> 2. What's the decision on `useMutation` and `useQuery`. There is no consistency between using `useQuery` and then just raw dogging fetches inside of an async effect in other places. There's also some spots where we make modifications and don't use a mutation? Seems like there's a lot of simplification we can do in these places, especially surrounding forms, which is a major pain in the ass rn.
> 3. As far as a bunch of tiny files goes, I will lodge a small complain against it. There are a bunch of places where a file has 6 lines and is next to 5 or so other logically related files that basically all import from the same place and do very similar things. I don't think that these should be broken up. For example, just put all of the 5 line synchronization hooks into a file called `useSycnchronizers` or `synchronizers`. All for splitting up files when hooks are large and not logically related, but seems excessive in many cases here.

1. I added a hook in pluto/ontology called `useResource` that gives an ontology resource for any ontology ID, which should support most of those problems. For specific data types those should be implemented as needed.
2. Decided to mostly not use useQuery, as it isn't super helpful to us unless we seriously wanted to start cacheing server requests. I tried to move most modifications into useMutation hooks, but that might not be 100% yet.
3. changed file sizes in a lot of places

pluto/src/ranger/childRangeHooks.ts
emilbon99, line 1: This file is badly named. What does `childRangeHooks` mean? It's not clear here that this is for _retrieving_ children. Also, there is a hook to retrieve the parent? Just put these in synchronizers. Or you could call these hooks `useRetrieveChildRanges` or `useRetrieveParentRange` and then call this `useRetrieve.ts`.

pluto/src/sync/Context.ts
```diff
@@ -0,0 +1,18 @@
...
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { createContext, use } from "react";
+
+import { type ListenerAdder } from "@/sync/types";
+
+export const Context = createContext<ListenerAdder>(() => {
```
emilbon99: We have a hook called `useRequiredContext` that implements the functionality you manually implemented here.

pluto/src/sync/Provider.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export const Provider = (props: PropsWithChildren): ReactElement => {
+  const client = Synnax.use();
+  const handlersRef = useRef(new Map<FrameHandler, Set<channel.Name>>());
+  const streamerRef = useRef<framer.Streamer>(null);
+  const handleError = Status.useErrorHandler();
+
+  useAsyncEffect(async () => {
+    if (client == null) {
+      streamerRef.current?.close();
```
emilbon99: This cleanup code should happen inside of the effect cleanup clause. Also, check whether you need to call both streamerRef.current?.close() and observableStreamer.close()

pluto/src/sync/Provider.tsx
emilbon99, line 1: None of this module is tested. Should it be tested now or later?

pjdotson: I spent a good while trying to test but it turns out that testing components that update
from subscriptions is pretty hard. This is because you can't properly use `act`
normally, where you press a button and then the component updates, due to you not
knowing exactly when these things happen. There is also race conditions with
waiting for the streamer to open in the `Sync.Provider` as well. I can try to get some
type of test working, but my sense is that it won't be high value add and will probably
be pretty flaky.

emilbon99: I don't disagree that testing this won't be high value add. This is core code that interacts with pretty much every part of the Console. This needs to be tested. If there's a race condition then you should use a mutex. 

I'm fine with testing this later for the purposes of getting it merged but it does need tests.

client/ts/src/testutil/newTestClient.ts
```diff
@@ -9,19 +9,12 @@
...
-  let _props = {};
-  if (props.length > 0) _props = props[0];
-  return new Synnax({
-    host: HOST,
-    port: PORT,
-    username: USERNAME,
-    password: PASSWORD,
-    ..._props,
-  });
+export const TEST_CLIENT_CONNECTION_PROPS: SynnaxProps = {
```
emilbon99: Minor, but why is this called `CONNECTION_PROPS`. Should just be `TEST_CLIENT_PROPS` since its the props for the test client.

console/src/channel/useListenForCalculationState.ts
emilbon99: Then why are we using `handleError`? It doesn't do anything. I guess because of null client error? I'm fine with having this code here, but there's a bunch of extra exception handling logic that is not nice

console/src/hardware/common/task/Rack.tsx
```diff
@@ -23,8 +23,10 @@ export const Rack = ({ taskKey }: RackProps) => {
   const rackKey = task.getRackKey(taskKey);
   const rack = useQuery({
```
emilbon99: Shouldn't this be a `Rack.use` hook from pluto?

console/src/hardware/device/Toolbar.tsx
```diff
@@ -7,92 +7,44 @@
...
-
-  return <StateContext.Provider value={states}>{children}</StateContext.Provider>;
-};
-
-interface RackStateContextValue extends Record<string, rack.State> {}
-
-const RackStateContext = createContext<RackStateContextValue>({});
-
-const RackStateProvider = ({ children }: { children: ReactElement }) => {
+const StateProvider = (props: PropsWithChildren) => {
```
emilbon99: Why do we need a `Rack.StateProvider` and a `Device.StateProvider` anymore? Shouldn't I just be able to use a hook from pluto called `Device.useState` and `Rack.useState`?

pluto/src/channel/Select.tsx
```diff
@@ -105,7 +105,7 @@ export const SelectMultiple = ({
   const searcher = useMemo(
     () =>
       client?.channels.newSearcherWithOptions({
-        rangeKey: activeRange,
+        rangeKey: activeRange ?? undefined,
```
emilbon99, line 101: This is kinda ridiculous? Is it so we coalesce null into `undefined`?

pluto/src/hooks/useRequiredContext.ts
```diff
@@ -9,8 +9,10 @@
 
 import { use } from "react";
 
-export const useRequiredContext = <T>(context: React.Context<T | null>): T => {
+export const useRequiredContext = <T>(
+  context: React.Context<T | null>,
+): NonNullable<T> => {
   const value = use(context);
-  if (value === null) throw new Error(`useRequiredContext: context value is null`);
+  if (value == null) throw new Error(`useRequiredContext: context value is null`);
```
emilbon99, line 16: This change is fine, but we should default to instantiating required contexts to null and not undefined.

pluto/src/label/use.ts
```diff
@@ -0,0 +1,86 @@
...
+import { useCallback, useState } from "react";
+
+import { NULL_CLIENT_ERROR } from "@/errors";
+import { useAsyncEffect } from "@/hooks";
+import { useDeleteSynchronizer, useSetSynchronizer } from "@/label/synchronizers";
+import { Ontology } from "@/ontology";
+import { Status } from "@/status";
+import { Synnax } from "@/synnax";
+
+export const use = (id: ontology.CrudeID): label.Label[] => {
```
emilbon99, line 20: This hook is very complicated and should definitely be tested.

pluto/src/sync/Provider.tsx
emilbon99: Actually no... I think we should test this now. The point of this CDC refactor is to get our query infrastructure to be production grade. Code that is not tested should, by default, not be considered production grade

pluto/src/channel/Select.tsx
pjdotson: yes

console/src/channel/useListenForCalculationState.ts
pjdotson: Maybe let's keep it like this for right now, and then how the calculation state works
can be changed in the state / status refactor PR?

.github/workflows/test.alamos.ts.yaml
```diff
@@ -12,7 +12,6 @@ on:
       - pnpm-lock.yaml
       - pnpm-workspace.yaml
       - turbo.json
-      - vitest.workspace.ts
```
emilbon99, line 15: Is there a good reason why we removed this? I feel like it would be nice for CI to re-run when this config file changes.

.github/workflows/test.pluto.yaml
```diff
@@ -26,22 +25,50 @@ on:
   workflow_dispatch:
 
 jobs:
+  server:
```
emilbon99, line 28: Absolutely love building the server seven times :)

pluto/src/hardware/rack/use.ts
```diff
@@ -0,0 +1,42 @@
...
+// included in the file licenses/APL.txt.
+
+import { type rack } from "@synnaxlabs/client";
+import { useCallback, useState } from "react";
+
+import { useStateSynchronizer } from "@/hardware/rack/synchronizers";
+import { useAsyncEffect } from "@/hooks";
+import { Synnax } from "@/synnax";
+
+export const use = (key: rack.Key): rack.Rack | undefined => {
```
emilbon99, line 17: How much thought have we put into error handling for fetch functions like these? Shouldn't we be doing some `useQuery` style state?

pluto/src/hooks/ref.ts
```diff
@@ -47,7 +48,9 @@ export const useStateRef = <T extends state.State>(
  */
 export const useSyncedRef = <T>(value: T): RefObject<T> => {
   const ref = useRef<T>(value);
-  ref.current = value;
+  useEffect(() => {
```
emilbon99: This seems like a totally unnecessary use of useEffect. It's totally fine to mutate a ref in the body of a function component right?

emilbon99: This also kinda defeats the purpose of this hook.

pluto/src/hooks/useAsyncEffect.ts
emilbon99, line 1: Much better implementation.

pluto/src/hardware/rack/use.ts
```diff
@@ -0,0 +1,42 @@
...
+      const rack = await client.hardware.racks.retrieve(key);
+      if (signal.aborted) return;
+      setRack(rack);
+    },
+    [client, key],
+  );
+  const handleStateChange = useCallback(
+    (state: rack.State) => {
+      if (rack == null || client == null) return;
+      console.log("state change", state);
```
emilbon99: Remove this log

emilbon99 (changes requested):

.github/workflows/test.alamos.ts.yaml
pjdotson: This config file got removed

.github/workflows/test.pluto.yaml
pjdotson: Yeah we really need to have a discussion about general CI improvements because our workflow is getting kinda unwieldy.

pluto/src/hooks/ref.ts
pjdotson: See here under pitfall: https://react.dev/reference/react/useRef#usage

pluto/src/hardware/rack/use.ts
pjdotson: At that point we are just reimplementing `useQuery` and should just use that and let it refetch whenever we get a new state. Which I am fine with if you think makes sense, but basically the only useQuery errors we have encountered have been with the task schema parsing in the task dialogs. IMO we should keep data fetching a little split right now between useQuery and useAsyncEffect and later on down the road try to migrate to really take advantage of useQuery's features.

.github/workflows/test.alamos.ts.yaml
pjdotson: I only had it for my editor and then I was having too many problems with VSCode warnings.

PR #1281, opened 2025-06-11
SY-2553 - Zyn Validation Library

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2553](https://linear.app/synnax/issue/SY-2553/zyn-validation-library)

## Description

- Implements a zod-style validation library for go, called zyn.
- Refactors the ontology to use zyn instead of the previous internal schema package.

## Basic Readiness


synnax/pkg/distribution/ontology/core/resource.go
emilbon99, line 1: Add tests for these changes

synnax/pkg/distribution/ontology/core/resource.go
```diff
@@ -116,42 +143,17 @@ func (r Resource) GorpKey() ID { return r.ID }
...
-	D.Data[k] = v
-}
-
 // NewResource creates a new entity with the given schema and name and an empty set of
 // field data.
-func NewResource(schema *Schema, id ID, name string) Resource {
+func NewResource(schema *Schema, id ID, name string, data any) Resource {
+	dumped, err := schema.Dump(data)
+	if err != nil {
+		panic(err)
```
emilbon99: Double check if we should actually panic here.

synnax/pkg/distribution/ontology/ontology_suite_test.go
```diff
@@ -28,36 +32,36 @@ func TestOntology(t *testing.T) {
...
-	return e, nil
+func (s *sampleService) RetrieveResource(ctx context.Context, key string, tx gorp.Tx) (ontology.Resource, error) {
+	return core.NewResource(s.Schema(), newSampleType(key), "empty", Sample{Key: key}), nil
 }
 
-func (s *emptyService) OpenNexter() (iter.NexterCloser[ontology.Resource], error) {
+func (s *sampleService) OpenNexter() (iter.NexterCloser[ontology.Resource], error) {
 	return iter.NexterNopCloser(iter.All([]ontology.Resource{
-		schema.NewResource(s.Schema(), newEmptyID(""), "empty"),
+		lo.Must(s.RetrieveResource(context.TODO(), "", nil)),
```
emilbon99: Fix the `context.TODO()` here.

x/go/types/name.go
```diff
@@ -58,3 +59,85 @@ func PluralName[T any]() string {
 		return name + "s"
 	}
 }
+
+func PackageName(t reflect.Type) string {
```
pjdotson, line 68: add test

x/go/types/name.go
```diff
@@ -58,3 +59,85 @@ func PluralName[T any]() string {
...
+		if v.IsNil() {
+			return "func (nil)"
+		}
+		return "func"
+	case reflect.Interface:
+		if v.IsNil() {
+			if t.Name() != "" {
+				return t.String()
+			}
+			return "interface{} (nil)"
```
pjdotson:
```suggestion
			return "any (nil)"
```

x/go/types/name.go
```diff
@@ -58,3 +59,85 @@ func PluralName[T any]() string {
...
+		}
+		if t.Name() != "" {
+			return t.String()
+		}
+		return "interface{}"
```
pjdotson:
```suggestion
		return "any"
```

x/go/validate/errors.go
```diff
@@ -0,0 +1,105 @@
...
+type pathError struct {
+	Path []string `json:"path"`
+	Err  error    `json:"error"`
+}
+
+func (p pathError) joinPath() string {
+	return strings.Join(p.Path, ".")
+}
+
+type encodedPatherror struct {
```
pjdotson: typo

x/go/validate/errors_test.go
```diff
@@ -0,0 +1,60 @@
+package validate_test
```
pjdotson, line 10: header

x/go/signal/panicpolicy_string.go
pjdotson, line 1: whats the point of this?

client/ts/src/channel/channel.spec.ts
```diff
@@ -51,6 +51,24 @@ describe("Channel", () => {
...
+        await client.channels.create({
+          name: "test",
+          virtual: true,
+          dataType: DataType.FLOAT32,
+          expression: "test * 2",
+          requires: [],
+        });
+      } catch (e) {
+        expect(PathError.matches(e)).toBeTruthy();
+        expect((e as PathError).path).toEqual(["requires"]);
```
pjdotson, line 65: seems like it would be nice to have the matches function return a new PathError or some
type of type guard. that way if we ever did more advanced error manipulation you know
that e has all the same fields as path error.

x/go/zyn/zyn.go
```diff
@@ -0,0 +1,178 @@
...
+		[]DataType{ObjectT},
+		IntegerTypes,
+	)
+	TypesZ = Enum(Types...)
+)
+
+func Primitive() UnionZ { return Union(Number(), String(), Bool()) }
+
+// Z is a schema that provides methods for validating and converting data.
+type Z interface {
```
pjdotson: Have you tried allowing for generics in this interface types? it seems odd that someone
would be allow to do this:

```go
var dest2 int
Expect(zyn.Bool().Parse(0.0, &dest2)).To(Succeed())
```

x/go/zyn/zyn_test.go
```diff
@@ -0,0 +1,117 @@
+package zyn_test
```
pjdotson, line 10: header

x/go/zyn/zyn_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/synnaxlabs/x/zyn"
+)
+
+func dataTypeLiteralTest(literal zyn.Z, dt zyn.DataType) func() {
```
pjdotson: unused

x/go/zyn/enum_test.go
```diff
@@ -0,0 +1,129 @@
+package zyn_test
```
pjdotson, line 10: header

x/go/zyn/literal_test.go
```diff
@@ -0,0 +1,107 @@
+package zyn_test
```
pjdotson, line 10: header

x/go/zyn/union_test.go
```diff
@@ -0,0 +1,134 @@
+package zyn_test
```
pjdotson, line 10: header

x/go/zyn/zyn_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package zyn_test
```
pjdotson, line 10: header

synnax/pkg/distribution/ontology/core/resource.go
```diff
@@ -7,18 +7,29 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-package schema
+package core
```
pjdotson, line 10: package resource might be a better name? or maybe package ontology

x/go/zyn/zyn.go
pjdotson: Also I think `Schema` makes more sense as a type name than `Z`.

synnax/pkg/distribution/cluster/ontology.go
```diff
@@ -44,20 +45,12 @@ func OntologyID(key uuid.UUID) ontology.ID {
...
-		Fields: map[string]schema.Field{
-			"key": {Type: schema.String},
-		},
-	}
+	NodeZ = zyn.Object(map[string]zyn.Z{
+		"key":     zyn.Uint16().Coerce(),
+		"address": zyn.String(),
+		"state":   zyn.Uint32().Coerce(),
+	})
+	Z = zyn.Object(map[string]zyn.Z{"key": zyn.UUID()})
```
pjdotson: I know we append Z to all of the names of zod schemas, but I think that is a bad pattern
and we should instead name things schema and nodeSchema. These are more descriptive names.

synnax/pkg/distribution/ontology/search_indexing_test.go
```diff
@@ -7,138 +7,124 @@
...
-				Schema: schema_,
-				Data:   map[string]any{},
-			},
+		z := zyn.Object(nil)
+		//db = gorp.Wrap(memkv.New())
+		//otg = MustSucceed(ontology.Open(ctx, ontology.Config{
+		//	Instrumentation: alamos.New("test"),
+		//	DB:              db,
+		//	EnableSearch:    config.True(),
+		//}))
```
pjdotson: commented out code

synnax/pkg/distribution/ontology/core/resource.go
emilbon99: The parent package is `ontology`. This needs to be `core` so that search can be an independent package. The `core` package is not meant to be used outside of the `ontology`

synnax/pkg/distribution/cluster/ontology.go
emilbon99: My personal take is that using `Z` on all schemas is a widely used convention across the codebase that becomes very apparent and almost icon quickly. Kind of understanding how `x` is for utilities instead of calling it `util/ts` or `util/python`. Yes, we could be clinical and make sure all of our names are perfectly precise, but where are the good vibes in that :)

x/go/signal/panicpolicy_string.go
emilbon99: It's a generated file that turns enums into stringers

synnax/pkg/distribution/cluster/ontology.go
emilbon99: But I also get if you wanna be precise

client/ts/src/channel/channel.spec.ts
emilbon99: I've tried to implement type guards with the new error class mechanics and wasn't able to sort it easily.

emilbon99: https://linear.app/synnax/issue/SY-2642/make-x-error-matchers-implement-type-guards

x/go/zyn/zyn.go
emilbon99: Yes I've tried, but it's not really possible, mostly because of dealing with the mechanics of objects. Happy to talk this through in person

synnax/pkg/distribution/cluster/ontology.go
pjdotson: Personally I would prefer schema to have more explicit and immediately obvious names. I've also found that in a few places in the TS codebase
the Z at the end of a schema got missed, so I either wouldn't be able to find a schema
quickly or would use something and not realize it was a schema. this is obviously harder
to do if we default to naming things `Schema` and `NodeSchema`.

pjdotson: PS is there a reason to make all of these schemas / Zs public? It seems like they are
only ever used internally to packages.

x/go/zyn/zyn.go
pjdotson: Can you add tests for all of the Parse functions for where dest is an invalid type? I
don't see any tests for those.

pjdotson (changes requested):
All remaining comments are on outdated commits so please look through them all.

synnax/pkg/distribution/cluster/ontology.go
emilbon99: You're boring....

done

pjdotson (approved):

PR #1283, opened 2025-06-11
Verify form in OPC UA server before testing connection

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2555](https://linear.app/synnax/issue/SY-2555/verify-form-in-opc-ua-server-test-connection-before-running-query)

## Description

Verify rack location before testing for OPC UA server connection.

## Basic Readiness

### DIFF
--- a/console/src/hardware/opc/device/Connect.tsx
+++ b/console/src/hardware/opc/device/Connect.tsx
@@ -68,7 +68,7 @@ export const CONNECT_LAYOUT: Layout.BaseState = {
 
 const formSchema = z.object({
   name: Common.Device.nameZ,
-  rack: rack.keyZ,
+  rack: rack.keyZ.refine((k) => k > 0, "Must select a location to connect from"),
   connection: connectionConfigZ,
 });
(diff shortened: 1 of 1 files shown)


console/src/hardware/opc/device/Connect.tsx
```diff
@@ -87,7 +87,8 @@ const Internal = ({ initialValues, layoutKey, onClose, properties }: InternalPro
     onError: (e) => handleError(e, "Failed to test connection"),
     mutationFn: async () => {
       if (client == null) throw NULL_CLIENT_ERROR;
-      if (!methods.validate("connection")) throw new Error("Invalid configuration");
+      if (!methods.validate("connection") || !methods.validate("rack"))
+        throw new Error("Invalid configuration");
```
emilbon99: Why should we throw? Won't this send a notification? I think we should just return here as its not notification worth for the user not to select a field. 

Also, lower priority but why don't we just verify the entire form vs calling validate() on specific sections?

emilbon99 (changes requested):

PR #1303, opened 2025-06-16
SY-2618 - Standardize Status Specs

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2618](https://linear.app/synnax/issue/SY-2618/standardize-status-specs)

## Description

Standardize Status Specs

## Basic Readiness


synnax/pkg/service/hardware/task/service.go
```diff
@@ -77,7 +77,7 @@ type Service struct {
 	group           group.Group
 }
 
-const groupName = "Tasks"
+const groupName = "TaskStatuses"
```
pjdotson: revert?

synnax/pkg/service/hardware/task/task.go
```diff
@@ -66,31 +65,27 @@ func (t Task) String() string {
...
 
 // SetOptions implements the gorp.Entry interface.
-func (s State) SetOptions() []any { return []any{s.Task.Rack().Node()} }
+func (s Status) SetOptions() []any { return []any{s.Details.Task.Rack().Node()} }
 
-// CustomTypeName implements types.CustomTypeName to ensure that State struct does
+// CustomTypeName implements types.CustomTypeName to ensure that Status struct does
 // not conflict with any other types in gorp.
-func (s State) CustomTypeName() string { return "TaskState" }
+func (s Status) CustomTypeName() string { return "TaskState" }
```
pjdotson: TaskStatus right?

synnax/pkg/service/hardware/device/device.go
```diff
@@ -60,21 +59,14 @@ func (d Device) Validate() error {
 	return v.Error()
 }
 
-// State represents the state of a device.
-type State struct {
-	// Key is the key of the device.
-	Key string `json:"key" msgpack:"key"`
-	// Rack is the rack that the device is in.
+type StatusDetails struct {
 	Rack rack.Key `json:"rack" msgpack:"rack"`
```
pjdotson: Don't we want the key of the device here as well?

synnax/pkg/service/hardware/tracker/tracker.go
pjdotson, line 1: It would be good to add some type of verification of the Variant enum here

driver/opc/scan_task.cpp
```diff
@@ -127,12 +133,15 @@ void ScanTask::scan(const task::Command &cmd) const {
...
-        .task = task.key,
+    ctx->set_status({
         .key = cmd.key,
-        .variant = status::VARIANT_SUCCESS,
-        .details = util::DeviceProperties(args.connection, *scan_ctx->channels)
-                       .to_json(),
+        .variant = status::variant::SUCCESS,
+        .details =
+            synnax::TaskStatusDetails{
+                .task = task.key,
```
pjdotson, line 141: add running here

x/ts/src/status/status.ts
```diff
@@ -0,0 +1,77 @@
...
+  "disabled",
+]);
+
+// Represents one of the possible variants of a status message.
+export type Variant = z.infer<typeof variantZ>;
+
+const unknownOptional = z.unknown().optional();
+
+export const statusZ = <D extends z.ZodType = typeof unknownOptional>(
+  details: D = unknownOptional as unknown as D,
```
pjdotson: doesn't it make more sense here to have D extend z.ZodType? and then not give a
default parameter for details here. That way someone could set whatever they want
(number, string, object) ad details.)

client/ts/src/hardware/task/payload.ts
```diff
@@ -21,54 +21,94 @@ export const keyZ = z.union([
...
-  key: keyZ,
-  name: z.string(),
-  type: z.string(),
-  internal: z.boolean().optional(),
-  config: record.unknownZ.or(z.string().transform(decodeJSONString)),
-  state: stateZ.optional().nullable(),
-  snapshot: z.boolean().optional(),
-});
+export const statusZ = <StatusData extends z.ZodType>(
+  data: StatusData = z.unknown() as unknown as StatusData,
```
pjdotson: same thing here as in the general status package, its probably better not to have a
default and instead let any user manually specify this.

pluto/src/synnax/Provider.tsx
pjdotson, line 1: Think you want to revert most of the changes on this file, specifically the fact that
you removed the abort signal work

pluto/src/status/aether/aggregator.ts
```diff
@@ -29,29 +28,28 @@ export class Aggregator extends aether.Composite<typeof aggregatorStateZ> {
...
       ...p,
-      statuses: [...p.statuses, this.parse(spec)],
+      statuses: [...p.statuses, status.create(spec)],
     }));
   }
 }
 
-export interface Adder {
-  (spec: CrudeSpec): void;
+export interface Adder<D = undefined> {
```
pjdotson: it doesn't really make sense to be able to type the Adder here. It makes more sense that
it can just add status.Crude<unknown> or status.Crude<any>. The implementation of Adder
won't not work if you have the wrong details.

console/src/notifications/Notifications.tsx
```diff
@@ -20,8 +20,8 @@ export interface Sugared extends Status.NotificationSpec {
   content?: ReactElement;
 }
 
-export interface Adapter {
-  (status: Status.NotificationSpec, silence: (key: string) => void): null | Sugared;
+export interface Adapter<D = undefined> {
```
pjdotson, line 23: this is another type where adding the generic parameter doesn't make sense

console/src/hardware/task/Toolbar.tsx
```diff
@@ -172,39 +119,35 @@ const Content = () => {
...
-    },
-    [client?.key],
-  );
-  const handleStateUpdate = useCallback((state: task.State) => {
-    setTasks((prevTasks) => {
-      const tsk = prevTasks.find((t) => t.key === state.task);
-      if (tsk == null) return prevTasks;
-      updateTaskStatus(tsk, state);
-      return [...prevTasks];
+  useAsyncEffect(async () => {
```
pjdotson: keep the abort signal in here

pjdotson (changes requested):
Left a few comments here. Definitely an improvement and a move in the right direction so I think we should just get this merged in ASAP, but the tasks especially will need to be looked at again in the future.

synnax/pkg/service/hardware/tracker/tracker.go
emilbon99: https://linear.app/synnax/issue/SY-2641/validate-hardware-tracker-enum-variants

pluto/src/status/aether/aggregator.ts
```diff
@@ -29,23 +28,22 @@ export class Aggregator extends aether.Composite<typeof aggregatorStateZ> {
...
       ...p,
-      statuses: [...p.statuses, this.parse(spec)],
+      statuses: [...p.statuses, status.create(spec)],
     }));
   }
 }
 
 export interface Adder {
-  (spec: CrudeSpec): void;
+  <D = undefined>(spec: status.Crude<D>): void;
```
pjdotson, line 46: It still doesn't make sense to me that you have to now pass in a generic parameter to
calls to `addStatus`. Like this code in `synnax/Provider.tsx`:

```
        addStatus<StatusDetails>({
          variant: "warning",
          message: "Incompatible cluster version",
          description,
          details: {
            type: SERVER_VERSION_MISMATCH,
            oldServer,
            nodeVersion: connectivity.nodeVersion,
            clientVersion: connectivity.clientVersion,
          },
        });
```

we now have to define a type `StatusDetails` that is only used to pass in to the generic
parameter in addStatus? it seems like we are making extra work for the caller of this
function that did not exist before.

MODULE.bazel.lock
```diff
@@ -335,5 +335,582 @@
     "https://bcr.bazel.build/modules/zlib/1.3/MODULE.bazel": "6a9c02f19a24dcedb05572b2381446e27c272cd383aed11d41d99da9e3167a72"
   },
   "selectedYankedVersions": {},
-  "moduleExtensions": {}
+  "moduleExtensions": {
```
pjdotson, line 338: can you look into what these `moduleExtensions` are / why they appear? so we don't have
all of this diffs and presumably some type of dependency to install in CI for no reason.

PR #1306, opened 2025-06-16
SY-2449: Fix Line Plot Download as CSV Errors

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2449](https://linear.app/synnax/issue/SY-2449/fix-line-plot-download-as-csv)

## Description

Refactored download as csv into a hook and throw errors for null client and when trying to download no data.

Fixed the following issues with line plots (the last one is where our customer ran into issues):
  - added μ to default text atlas instead of u
  - Remove setState call during render in Channel.LinePlot
  - Rerendered the range annotation provider when the cursor moves above the range names. Previously, it only changed which range was selected when you move on the actual plot (not the whitespace above it).

I am not entirely sure that this fixes all of CSV downloading related issues, but it is a step in the right direction.

## Basic Readiness

### DIFF
--- a/console/package.json
+++ b/console/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@synnaxlabs/console",
   "private": true,
-  "version": "0.43.1",
+  "version": "0.43.2",
   "type": "module",
   "scripts": {
     "check-types": "tsc --noEmit",


pluto/src/aether/main.tsx
emilbon99, line 1: Could be worth it to check if this causes this hook to get re-evaluate repeatedly.

emilbon99 (approved):
See my comment, think about whether you want to address it, and then good to merge

PR #1310, opened 2025-06-17
SY-2627 Fix Zod Error on Invalid State Variant

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2627](https://linear.app/synnax/issue/SY-2627/fix-zod-error-on-invalid-state-variant)

## Description

Fixed the Zod errors on invalid state variant by piping states with an empty string to "info". Will put adding server-side validation and checking all driver calls to [SY-2618](https://linear.app/synnax/issue/SY-2618/standardize-status-specs)

## Basic Readiness

### DIFF
--- a/client/ts/package.json
+++ b/client/ts/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@synnaxlabs/client",
-  "version": "0.43.0",
+  "version": "0.43.1",
   "description": "The Synnax Client Library",
   "keywords": [
     "synnax",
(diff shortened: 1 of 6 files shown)


emilbon99 (approved):
Have you thought about what should happen in the null case? Also why do we need to type to a literal "info" instead of a status.Variant? If you've thought about this fine to merge, just want to bring it to attention.

pjdotson (commented):
> Have you thought about what should happen in the null case? Also why do we need to type to a literal "info" instead of a status.Variant? If you've thought about this fine to merge, just want to bring it to attention. 

the `nullToUndefined` converts `nil` states server-side to undefined in the client libraries. Changed the typing as well.

PR #1317, opened 2025-06-27
SY-2637 - Flux

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2637](https://linear.app/synnax/issue/SY-2637/standardize-queries)

## Basic Readiness


pluto/src/flux/sync/Provider.tsx
```diff
@@ -0,0 +1,103 @@
...
+export interface ProviderProps extends PropsWithChildren {}
+
+interface MutexValue {
+  streamer: framer.ObservableStreamer | null;
+}
+
+export const Provider = ({ children }: ProviderProps): ReactElement => {
+  const client = Synnax.use();
+  const handlersRef = useRef(new Map<FrameHandler, channel.Name>());
+  const streamerRef = useRef<sync.Mutex<MutexValue>>(sync.newMutex({ streamer: null }));
```
pjdotson, line 40: Why do we need this mutex ref? JS is single threaded and the abort controller mechanism
should make sure that when async stuff runs at the same time, cancellation stops any
updates from a previous scheduled one from going through.

pluto/src/flux/sync/Sync.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { newTestClient } from "@synnaxlabs/client";
+import { id } from "@synnaxlabs/x";
+import { renderHook, waitFor } from "@testing-library/react";
+import { act, useEffect, useState } from "react";
+import { describe, expect, it } from "vitest";
+
+import { Sync } from "@/flux/sync";
+import { newSynnaxWrapper } from "@/testutil/Synnax";
+
+describe("sync", () => {
```
pjdotson, line 19: This package needs to be better tested

pluto/src/flux/sync/listeners.ts
```diff
@@ -0,0 +1,74 @@
...
+  onChange: ListenerHandler<Value, Extra>;
+}
+
+export const useListener = (
+  listeners: ListenerSpec<MultiSeries, {}> | ListenerSpec<MultiSeries, {}>[],
+): void => {
+  const addListener = useAddListener();
+  const handleError = Status.useErrorHandler();
+  useEffect(() => {
+    const mu = new Mutex();
```
pjdotson: why do we need the mutex here?

pluto/src/flux/sync/listeners.ts
```diff
@@ -0,0 +1,74 @@
...
+import { useAddListener } from "@/flux/sync/Context";
+import { Status } from "@/status";
+
+export type ListenerArgs<Value, Extra> = {
+  changed: Value;
+} & Extra;
+
+export interface ListenerHandler<Value, Extra> {
+  (args: ListenerArgs<Value, Extra>): Promise<unknown>;
+}
```
pjdotson, line 24: having this `Extra` type doesn't make a lot of sense to me. It seems if List or Form or
whatever wants to expose more arguments, that should be handled in that file rather than
making more ungodly types. So the type signatures should look like this:

```ts
export const parsedHandler = <Z extends z.ZodType>(schema: Z, onChange: (value: z.output<Z>) => Promise<void>)
export const stringHandler (onChange: (value: string))
```

which makes this file easier to use.

freighter/go/freightfluence/receiver_test.go
```diff
@@ -74,7 +76,7 @@ var _ = Describe("Receiver", func() {
 			Expect(v).To(Equal(1))
 			cancel()
 			_, err = stream.Receive()
-			Expect(err).To(Equal(context.Canceled))
+			Expect(err).To(HaveOccurredAs(context.Canceled))
```
pjdotson, line 79: Is there a reason we have this `HaveOccuredAs` custom matcher instead of just using the
built-in `MatchError` matcher? it seems like for all of the changes here they would be
the same

x/ts/src/telem/series.spec.ts
```diff
@@ -1802,4 +1802,79 @@ describe("MultiSeries", () => {
...
+        expected: ["1", "2", "3", "4", "5", "6"],
+      },
+      {
+        name: "bigint",
+        series: [
+          [BigInt(1), BigInt(2)],
+          [BigInt(3), BigInt(4)],
+        ],
+        expected: ["1", "2", "3", "4"],
+      },
```
pjdotson, line 1901: add tests for decimal numbers and UUIDs

synnax/pkg/service/workspace/service.go
```diff
@@ -68,9 +73,20 @@ func NewService(ctx context.Context, configs ...Config) (*Service, error) {
...
+		return s, nil
+	}
+	s.shutdownSignals, err = signals.PublishFromGorp(ctx, cfg.Signals, signals.GorpPublisherConfigUUID[Workspace](cfg.DB))
+	if err != nil {
+		return nil, err
+	}
 	return s, nil
 }
 
+func (s *Service) Close() error {
```
pjdotson, line 86: Add test for changes with Close and signals

synnax/pkg/distribution/signals/gorp.go
```diff
@@ -110,7 +110,7 @@ func GorpPublisherConfigNumeric[K types.Numeric, E gorp.Entry[K]](db *gorp.DB, d
 	return GorpPublisherConfig[K, E]{
```
pjdotson: Is this function even used anywhere?

synnax/pkg/distribution/layer.go
```diff
@@ -96,6 +97,10 @@ type Config struct {
 	//
 	// [OPTIONAL] - Defaults to &binary.MsgPackCodec
 	GorpCodec binary.Codec
+	// EnableChannelSignals sets whether to enable CDC signal propagation for changes
+	// to channel data structures.
+	// [OPTIONAL] - Defaults to true.
```
pjdotson, line 103: You will hate me for this but please put an empty line like with all the other ones :)

synnax/pkg/distribution/channel/signals/signals.go
```diff
@@ -0,0 +1,31 @@
...
+	"context"
+	"io"
+
+	"github.com/synnaxlabs/synnax/pkg/distribution/channel"
+	"github.com/synnaxlabs/synnax/pkg/distribution/signals"
+	"github.com/synnaxlabs/x/gorp"
+	"github.com/synnaxlabs/x/telem"
+)
+
+func Publish(
```
pjdotson, line 22: Add tests

client/ts/src/access/policy/client.ts
```diff
@@ -57,10 +54,6 @@ export class Client {
...
 
-export const ontologyID = (key: Key): ontology.ID =>
-  new ontology.ID({ type: ONTOLOGY_TYPE, key });
+export const ontologyID = (key: Key): ontology.ID => ({ type: ONTOLOGY_TYPE, key });
 
-export const ALLOW_ALL_ONTOLOGY_ID = new ontology.ID({
-  type: ALLOW_ALL_ONTOLOGY_TYPE,
-  key: "",
-});
+export const ALLOW_ALL_ONTOLOGY_ID = { type: ALLOW_ALL_ONTOLOGY_TYPE, key: "" };
```
pjdotson: explicitly type this as an `ontology.ID`

client/ts/src/channel/payload.ts
```diff
@@ -23,7 +23,7 @@ export type KeysOrNames = Keys | Names;
 export type PrimitiveParams = Key | Name | Keys | Names;
 export type Params = Key | Name | Keys | Names | Payload | Payload[];
 
-export const channelZ = z.object({
+export const payloadZ = z.object({
```
pjdotson, line 26: IMO because we don't have a `Channel` class we should call these `channelZ` and
`Channel` instead of `payloadZ` and `payload`

client/ts/src/ranger/client.ts
```diff
@@ -261,9 +266,14 @@ export class Client implements AsyncTermSearcher<string, Key, Range> {
   async retrieve(range: CrudeTimeRange): Promise<Range[]>;
   async retrieve(range: Key | Name): Promise<Range>;
   async retrieve(range: Keys | Names): Promise<Range[]>;
-  async retrieve(ranges: Params | CrudeTimeRange): Promise<Range | Range[]> {
+  async retrieve(req: RetrieveRequest): Promise<Range[]>;
+  async retrieve(
```
pjdotson, line 270: add tests for this

client/ts/src/framer/streamer.ts
```diff
@@ -314,6 +314,10 @@ export class ObservableStreamer<V = Frame>
     this.closePromise = this.stream();
   }
 
+  async update(channels: channel.Params): Promise<void> {
```
pjdotson, line 317: Add tests

client/ts/src/ontology/group/group.ts
```diff
@@ -8,7 +8,7 @@
 // included in the file licenses/APL.txt.
 
 import { type Key, type Name, ONTOLOGY_TYPE } from "@/ontology/group/payload";
-import { ID as OntologyID } from "@/ontology/payload";
+import { type ID as OntologyID } from "@/ontology/payload";
```
pjdotson: since this is a type you can `import { type ontology } from @/ontology` which improves
the API slightly IMO

client/ts/src/ontology/payload.ts
```diff
@@ -67,65 +67,38 @@ export const resourceTypeZ = z.enum([
...
-    return this.toString() === new ID(other).toString();
-  }
-
-  get payload(): IDPayload {
-    return { type: this.type, key: this.key };
-  }
-
-  static readonly z = z.union([z.instanceof(ID), crudeIDZ.transform((v) => new ID(v))]);
-}
+export const idZ = z.object({ type: resourceTypeZ, key: z.string() }).or(stringIDZ);
```
pjdotson, line 75: should add parsing tests

client/ts/src/ontology/payload.ts
```diff
@@ -137,12 +110,12 @@ export const oppositeRelationshipDirection = (
...
 ): RelationshipDirection => (direction === "to" ? "from" : "to");
 
-export const relationshipSchemaZ = z.object({ from: ID.z, type: z.string(), to: ID.z });
-export interface Relationship extends z.infer<typeof relationshipSchemaZ> {}
-
-export const parseRelationship = (str: string): Relationship => {
-  const [from, type, to] = str.split("->");
-  return { from: new ID(from), type, to: new ID(to) };
-};
+export const relationShipZ = z.object({ from: idZ, type: z.string(), to: idZ }).or(
```
pjdotson: relationSSSSSSSSSSSSSShip

client/ts/src/ontology/payload.ts
```diff
@@ -137,12 +110,12 @@ export const oppositeRelationshipDirection = (
...
+  z.string().transform((v) => {
+    const [from, type, to] = v.split("->");
+    return { from: idZ.parse(from), type, to: idZ.parse(to) };
+  }),
+);
```
pjdotson, line 118: and add parsing tests

client/ts/src/ontology/payload.ts
```diff
@@ -67,65 +67,38 @@ export const resourceTypeZ = z.enum([
...
+
+export const ROOT_ID: ID = { type: BUILTIN_TYPE, key: "root" };
+
+export const idToString = (id: ID) => `${id.type}:${id.key}`;
 
-export const ROOT_ID = new ID({ type: BUILTIN_TYPE, key: "root" });
+export const idsEqual = (a: ID, b: ID) => a.type === b.type && a.key === b.key;
+
+export const parseIDs = (
+  ids: ID | ID[] | string | string[] | Resource | Resource[],
```
pjdotson, line 86: not really any point of accepting `ID` or `ID[]` right? since its a noop for `ID[]` and
you could just call `array.toArray(ID)`

client/ts/src/ontology/payload.ts
```diff
@@ -67,65 +67,38 @@ export const resourceTypeZ = z.enum([
...
+): ID[] => {
```
pjdotson, line 87: also add tests

pluto/src/state/state.ts
```diff
@@ -13,32 +13,35 @@ import { useState } from "react";
 export type State = primitive.Value | record.Unknown;
 export type SetFunc<S, PS = S> = (prev: PS) => S;
 
-export const isSetter = <S extends State>(arg: SetArg<S>): arg is SetFunc<S> =>
-  typeof arg === "function";
+export const isSetter = <S extends State, PS = S>(
```
pjdotson, line 16: This stuff with PS (should call it PreviousState, i thought it meant PureState at first)
is pretty odd. What is the point of trying to execute setter on state if the type of
state can change each time you call because of PS? It seems like at this point instead
of extending these `state` types more and making them harder to understand you should
manually write the code in the places where you have to use that.

pluto/src/state/state.ts
```diff
@@ -13,32 +13,35 @@ import { useState } from "react";
...
+  arg: SetArg<S, PS>,
+): arg is SetFunc<S, PS> => typeof arg === "function";
 
 export type SetArg<S extends State, PS = S> = S | SetFunc<S, PS>;
-export type Set<S extends State> = (value: SetArg<S>) => void;
-export type PureSet<S extends State> = (value: S) => void;
+export type Setter<S extends State> = (value: SetArg<S>) => void;
```
pjdotson: I also wonder if it would be better to just use the same types as React's `useState`
returns instead of writing our own types for these.

pluto/src/testutil/Synnax.tsx
```diff
@@ -33,29 +33,15 @@ const Context = createContext<ClientConnector>(() => () => {});
...
-          <Synnax.Provider
-            {...props}
-            connParams={isConnected ? TEST_CLIENT_PROPS : undefined}
-          />
-        </Status.Aggregator>
-      </AetherProvider>
-    </Context>
+export const newSynnaxWrapper =
+  (client: Synnax | null = null): FC<PropsWithChildren> =>
+  // eslint-disable-next-line react/display-name
```
pjdotson: not really any reason to put this eslint rule, just give it a display name. I believe if
you give the function a name in the body of the wrapper this won't appear.

pluto/src/synnax/Provider.tsx
```diff
@@ -46,6 +46,7 @@ export const useConnectionState = () => useContext().state;
 
 export interface ProviderProps extends PropsWithChildren {
   connParams?: SynnaxProps;
+  client?: Synnax;
```
pjdotson: this isn't used

pluto/src/flux/sync/types.ts
```diff
@@ -13,8 +13,9 @@ import { type Destructor, type observe } from "@synnaxlabs/x";
 export interface FrameHandler extends observe.Handler<framer.Frame> {}
 
 export interface Subscriber {
-  channels: channel.Name | channel.Names;
+  channel: channel.Name;
   handler: FrameHandler;
+  onOpen?: () => void;
```
pjdotson, line 18: Can we get rid of this? I don't get the use case for it.

pluto/src/flux/sync/Provider.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const handleError = Status.useErrorHandler();
+
+  useAsyncEffect(
+    async () => async () => {
```
pjdotson: why do you have `async () => ` twice?

pluto/src/flux/sync/listeners.ts
pjdotson: Like what is a specific scenario where this would be a problem? Also why make onChange
an async function at all?

pluto/src/flux/listeners.ts
pjdotson, line 1: If there is only one hook in this file then it should just be called
`useMountSynchronizers.ts`

pluto/src/flux/sync/listeners.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const useListener = (
+  listeners: ListenerSpec<MultiSeries, {}> | ListenerSpec<MultiSeries, {}>[],
+): void => {
+  const addListener = useAddListener();
+  const handleError = Status.useErrorHandler();
+  useEffect(() => {
+    const mu = new Mutex();
+    array.toArray(listeners).map(({ channel, onChange }) => {
+      addListener({
```
pjdotson, line 60: you never destroy these listeners.

pluto/src/flux/update.ts
```diff
@@ -0,0 +1,240 @@
...
+}
+
+/**
+ * Return type for the observable update hook.
+ *
+ * @template Data The type of data being updated
+ */
+export interface UseObservableUpdateReturn<Data extends state.State> {
+  /** Function to trigger an update (fire-and-forget) */
+  update: (value: Data, opts?: AsyncOptions) => void;
```
pjdotson: why would it need AsyncOptions here?

pluto/src/channel/queries.ts
```diff
@@ -0,0 +1,122 @@
...
+    message: "Expression must use at least one synnax channel",
+    path: ["requires"],
+  });
+
+const channelToFormValues = (ch: channel.Channel) => ({
+  ...ch.payload,
+  dataType: ch.dataType.toString(),
+});
+
+export interface FluxParams extends Flux.Params {
```
pjdotson: its really odd that you have to write extends Flux.Params for all of these. Why do we
need the [x: string]: primitive.Value in all of these interfaces for unknown strings - a) it's odd
to assume that all unknown keys are primitive.Value and b) it requires more work when
you are writing all of these queries packages.

pluto/src/flux/result.ts
```diff
@@ -0,0 +1,157 @@
...
+ * ```typescript
+ * const result: Result<User[]> = {
+ *   variant: "success",
+ *   data: [{ id: 1, name: "John" }],
+ *   error: null,
+ *   message: "Retrieved users"
+ * };
+ * ```
+ */
+export type Result<Data extends state.State> =
```
pjdotson, line 34: why does Data have to extend state.State, why not let it be anything?

pluto/src/flux/result.ts
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * @example
+ * ```typescript
+ * const loadingResult = pendingResult<User[]>("users", "retrieving");
+ * // Returns: { variant: "loading", data: null, error: null, message: "Retrieving users" }
+ * ```
+ */
+export const pendingResult = <Data extends state.State>(
+  name: string,
+  op: string,
```
pjdotson, line 68: having these separate name and op args for all of these functions are really weird. Why
not just call it message?

pluto/src/flux/result.ts
```diff
@@ -0,0 +1,157 @@
...
+ * @example
+ * ```typescript
+ * const errorResult = errorResult<User[]>("users", "retrieve", new Error("Network error"));
+ * // Returns: { variant: "error", data: null, error: Error, message: "Failed to retrieve users" }
+ * ```
+ */
+export const errorResult = <Data extends state.State>(
+  name: string,
+  op: string,
+  error: unknown,
```
pjdotson, line 127: shouldn't error be an error here?

pluto/src/flux/result.ts
```diff
@@ -0,0 +1,157 @@
...
+ *   message: "Retrieved users"
+ * };
+ * ```
+ */
+export type Result<Data extends state.State> =
+  | (status.Status<status.ExceptionDetails, "error"> & {
+      /** The data payload, null when in error state */
+      data: null;
+      /** The error that occurred during the operation */
+      error: unknown;
```
pjdotson: shouldn't this be type `Error`

pluto/src/flux/params.ts
```diff
@@ -0,0 +1,41 @@
...
+ * }
+ *
+ * const params: UserParams = {
+ *   userId: 123,
+ *   includeProfile: true,
+ *   searchTerm: "john"
+ * };
+ * ```
+ */
+export type Params = Record<string, primitive.Value>;
```
pjdotson: This type Params doesn't guarantee that when you do comparisons every value in the
object is a primitive.Value, making it kinda useless.

pluto/src/flux/listeners.ts
```diff
@@ -0,0 +1,67 @@
...
+  return useCallback(
+    (listeners?: Sync.Subscriber[]) => {
+      if (listeners == null || listeners.length === 0 || client == null) return;
+      ref.current.mounted = true;
+      const destructors = listeners.map(({ channel, handler }) =>
+        addListener({ channel, handler }),
+      );
+      ref.current.destructor = () => destructors.forEach((d) => d());
+    },
+    [addListener, client],
```
pjdotson: why do you care about the client here? isn't it better to just have the listeners
always mounted even when not connected?

pluto/src/flux/index.ts
```diff
@@ -0,0 +1,29 @@
...
+ * - Update operations with optimistic UI updates
+ * - Real-time synchronization with WebSocket-based data streams
+ *
+ * All operations are designed to work seamlessly with React hooks and provide
+ * consistent error handling, loading states, and data transformation capabilities.
+ *
+ * @module flux
+ * @version 1.0.0
+ * @since 1.0.0
+ */
```
pjdotson: what does 1.0.0 refer to?

pluto/src/channel/queries.ts
```diff
@@ -0,0 +1,122 @@
...
+  z.infer<typeof formSchema | typeof calculatedFormSchema>
+>) => {
+  const ch = await client.channels.create(value);
+  onChange(channelToFormValues(ch));
+};
+
+export const useForm = (args: Flux.UseFormArgs<FluxParams, typeof formSchema>) =>
+  Flux.createForm<FluxParams, typeof formSchema>({
+    name: "Channel",
+    schema: formSchema,
```
pjdotson, line 105: if we are passing in `formSchema` here, theres not any reason we should have to
explicitly declare `typeof formSchema` as well.

pluto/src/channel/queries.ts
```diff
@@ -0,0 +1,122 @@
...
+  .extend({
+    expression: z
+      .string()
+      .min(1, "Expression must not be empty")
+      .refine((v) => v.includes("return"), {
+        message: "Expression must contain a return statement",
+      }),
+  })
+  .refine((v) => v.requires?.length > 0, {
+    message: "Expression must use at least one synnax channel",
```
pjdotson: can just call this channel, not synnax channel

pluto/src/channel/queries.ts
```diff
@@ -0,0 +1,122 @@
...
+    retrieve,
+    update,
+    listeners: [],
+  })(args);
+
+export const useCalculatedForm = (
+  args: Flux.UseFormArgs<FluxParams, typeof calculatedFormSchema>,
+) =>
+  Flux.createForm<FluxParams, typeof calculatedFormSchema>({
+    name: "CalculatedChannel",
```
pjdotson: you need to specify in the docs for form what name is exactly used for, otherwise it is
odd to not have a space here, maybe odd to be capitalized, etc.

pluto/src/label/queries.ts
```diff
@@ -0,0 +1,131 @@
+// Copyright 2025 Synnax Labs, Inc.
```
pjdotson, line 1: add tests

pluto/src/flux/retrieve.ts
```diff
@@ -0,0 +1,387 @@
...
+ * // Usage with manual control
+ * const { data, variant, error, retrieve } = userRetrieve.useStateful();
+ * retrieve({ userId: 123 });
+ * ```
+ */
+export const createRetrieve = <RetrieveParams extends Params, Data extends state.State>(
+  factoryArgs: CreateRetrieveArgs<RetrieveParams, Data>,
+): CreateRetrieveReturn<RetrieveParams, Data> => ({
+  useObservable: (args: UseObservableRetrieveArgs<Data>) =>
+    useObservable({ ...factoryArgs, ...args }),
```
pjdotson: where are these observables used / what situation would they be used?

pluto/src/flux/retrieve.ts
```diff
@@ -0,0 +1,387 @@
...
+  useStateful: () => useStateful(factoryArgs),
```
pjdotson: I also don't see where useStateful is used

pluto/src/flux/retrieve.ts
```diff
@@ -0,0 +1,387 @@
...
+ * Creates a retrieve query system that provides hooks for fetching data.
+ *
+ * This function creates a set of React hooks that handle data retrieval with
+ * proper loading states, error handling, caching, and real-time updates. It provides
+ * multiple hook variants for different use cases:
+ *
+ * - `useObservable`: For external state management.
+ * - `useStateful`: For internal state management with manual control of query execution.
+ * - `useDirect`: For automatic retrieval based on parameters.
+ * - `useEffect`: For side-effect based retrieval.
```
pjdotson: These descriptions and names are really not that good (no need to overload what
useEffect means). you need to a) test and b) have better descriptions here.

pluto/src/flux/list.ts
```diff
@@ -0,0 +1,531 @@
...
+
+/**
+ * Function interface for getting items from a list by key(s).
+ *
+ * @template K The type of the key (must be a record key)
+ * @template E The type of the entity (must be keyed by K)
+ */
+interface GetItem<K extends record.Key, E extends record.Keyed<K>> {
+  /** Get a single item by key, returns undefined if not found */
+  (key?: K): E | undefined;
```
pjdotson: whats the point of even calling this when key is undefined?

pluto/src/flux/retrieve.spec.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { channel, DisconnectedError, newTestClient } from "@synnaxlabs/client";
+import { act, renderHook, waitFor } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+
+import { Flux } from "@/flux";
+import { newSynnaxWrapper } from "@/testutil/Synnax";
+
+const client = newTestClient();
+
+describe("retrieve", () => {
```
pjdotson, line 21: you need to add tests for all 4 types of retrieves you have

pluto/src/flux/update.ts
```diff
@@ -0,0 +1,240 @@
...
+  ) => UseObservableUpdateReturn<Data>;
+  /** Hook that provides update functions with internal state management */
+  useDirect: (args: UseDirectUpdateArgs<UpdateParams>) => UseDirectUpdateReturn<Data>;
+}
+
+/**
+ * Internal hook for observable updates with external state management.
+ * @internal
+ */
+const useObservable = <UpdateParams extends Params, Data extends state.State>({
```
pjdotson, line 125: this function needs to be tested

pluto/src/flux/pager.ts
```diff
@@ -0,0 +1,96 @@
...
+  onFetchMore: () => void;
+  /** Function to perform a search with the given term */
+  onSearch: (term: string) => void;
+}
+
+/**
+ * Arguments for the usePager hook.
+ */
+export interface UsePagerArgs
+  extends Pick<UseListReturn<PagerParams, any, any>, "retrieve"> {
```
pjdotson: No any!

pluto/src/flux/form.ts
```diff
@@ -0,0 +1,202 @@
...
+    const [result, setResult, resultRef] = useCombinedStateAndRef<
+      Result<z.infer<Schema> | null>
+    >(pendingResult(name, "retrieving"));
+
+    const handleResultChange: state.Setter<Result<z.infer<Schema> | null>> = (
+      setter,
+    ) => {
+      const nextStatus = state.executeSetter(setter, {
+        ...resultRef.current,
+        data: form.value() as any,
```
pjdotson: no any!

pluto/src/flux/pager.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Return type for the usePager hook, providing pagination utilities.
+ */
+export interface UsePagerReturn {
+  /** Function to fetch the next page of results */
+  onFetchMore: () => void;
+  /** Function to perform a search with the given term */
+  onSearch: (term: string) => void;
```
pjdotson: these should just be called fetchMore and search

pluto/src/flux/listeners.ts
```diff
@@ -0,0 +1,67 @@
...
+ *       // Handle real-time user updates
+ *       console.log("User updated:", frame.get("user_updates"));
+ *     }
+ *   }
+ * ];
+ *
+ * mountSynchronizers(listeners);
+ * ```
+ */
+export const useMountSynchronizers = (): ((listeners?: Sync.Subscriber[]) => void) => {
```
pjdotson: add test

pluto/src/flux/form.ts
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Arguments for using a form hook.
+ *
+ * @template FormParams The type of parameters for the form query
+ * @template Z The Zod schema type for form validation
+ */
+export interface UseFormArgs<FormParams extends Params, Z extends z.ZodObject> {
+  /** Initial values for the form fields */
+  initialValues?: z.infer<Z>;
+  /** Whether to automatically save form changes (not currently implemented) */
```
pjdotson: either implement it or get rid of this

pluto/src/flux/form.ts
```diff
@@ -0,0 +1,202 @@
...
+  autoSave?: boolean;
+  /** Parameters for the form query */
+  params: FormParams;
+  /** Callback function called after successful save */
+  afterSave?: (args: AfterSaveArgs<FormParams, Z>) => void;
```
pjdotson, line 80: this functionality isn't tested

pluto/src/flux/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export * as Flux from "@/flux/external";
```
pjdotson, line 10: You also need to test all of the files in this package with the listener functionality,
I only see it in a couple of the files.

pluto/src/ontology/external.ts
```diff
@@ -7,5 +7,4 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-export * from "@/ontology/synchronizers";
-export * from "@/ontology/useDependents";
+export * from "@/ontology/queries";
```
pjdotson, line 10: If you are going to stuff everything into one file the least you could do is remove the
external file

console/src/channel/Calculated.tsx
```diff
@@ -113,39 +56,21 @@ const GLOBALS: Variable[] = [
...
-        message: "Error creating calculated channel: ".concat(methods.value().name),
-        description: error.message,
-      });
-    },
-  });
 
-  const isIndex = Form.useFieldValue<boolean, boolean, typeof schema>("isIndex", {
-    ctx: methods,
-  });
+  const isIndex = Form.useFieldValue<boolean, boolean, typeof Channel.calculatedFormSchema>(
```
pjdotson: can you run all of your files through prettier? this isn't the only I've seen that goes
over the line length

pluto/src/hardware/device/synchronizers.ts
```diff
@@ -9,15 +9,30 @@
...
+    onChange: Sync.parsedHandler(device.keyZ, async (args) => {
+      onDelete(args.changed);
+    }),
+  });
 
 export const useStatusSynchronizer = (
   onStatusChange: (status: device.Status) => void,
 ): void =>
-  Sync.useParsedListener(device.STATUS_CHANNEL_NAME, device.statusZ, onStatusChange);
+  Sync.useListener({
```
pjdotson, line 29: When I look at stuff like this it makes me think that the older implementation was
better. We got rid of a custom hook in sync to make implementing all of these more verbose?

pluto/src/hardware/task/queries.ts
```diff
@@ -0,0 +1,82 @@
+// Copyright 2025 Synnax Labs, Inc.
```
pjdotson, line 1: add tests

pluto/src/ranger/queries.ts
```diff
@@ -0,0 +1,264 @@
+// Copyright 2025 Synnax Labs, Inc.
```
pjdotson, line 1: add tests

pluto/src/ranger/queries.ts
```diff
@@ -0,0 +1,264 @@
...
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { label, ontology, ranger } from "@synnaxlabs/client";
+import { type Optional, primitive } from "@synnaxlabs/x";
+import { z } from "zod/v4";
```
pjdotson: you can import right form "zod" now

pluto/src/workspace/queries.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright 2025 Synnax Labs, Inc.
```
pjdotson, line 1: add tests

pluto/src/ranger/queries.ts
```diff
@@ -0,0 +1,264 @@
...
+    }),
+  });
+
+export const useChildren = (
+  key: ranger.Key,
+): Flux.UseDirectRetrieveReturn<ranger.Range[]> => {
+  const res = Ontology.useChildren(ranger.ontologyID(key));
+  const client = Synnax.use();
+  if (res.variant !== "success") return { ...res, data: null };
+  if (client == null) return { ...res, data: [] };
```
pjdotson, line 60: shouldn't this have a null client error?

pluto/src/channel/queries.ts
```diff
@@ -0,0 +1,122 @@
+// Copyright 2025 Synnax Labs, Inc.
```
pjdotson, line 1: add tests

pluto/src/channel/queries.ts
```diff
@@ -0,0 +1,122 @@
...
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { channel, DataType } from "@synnaxlabs/client";
+import { z } from "zod/v4";
```
pjdotson: can import right from zod

console/src/range/overview/Snapshots.tsx
```diff
@@ -98,15 +98,15 @@ export interface SnapshotsProps {
 }
 
 export const Snapshots: FC<SnapshotsProps> = ({ rangeKey }) => {
-  const snapshots = Ontology.useChildren(ranger.ontologyID(rangeKey)).filter(
-    ({ data }) => data?.snapshot === true,
-  );
+  const { data: snapshots } = Ontology.useChildren(ranger.ontologyID(rangeKey));
+  if (snapshots == null) return null;
+  const filtered = snapshots.filter(({ data }) => data?.snapshot === true);
```
pjdotson, line 103: wouldn't it make sense to call data children and filtered snapshots?

pjdotson (commented):
It's good that stuff is moved lower down the stack, but the flux library needs to be way more heavily tested. Right now there really isn't any thought put into testing edge cases, and you mostly just run the nominal path for each function - not enough testing of where errors in implementation could occur. And for a lot of the resource-specific queries.ts files there are no tests.

pluto/src/flux/sync/Provider.tsx
emilbon99: The previous implementation of `Sync` opened a streamer regardless of whether any listeners were actually needed or not. This implementation updates the streamer only when a channel is requested, and closes the streamer when no channels are requested. 

To implement this behavior, updateStreamer needs to be called outside of an effect. Since its an async function, it races and multiple streamers may be opened at the same time if two react components mount listeners at the same time. Using a mutex makes sure we don't open redundant streamers.

freighter/go/freightfluence/receiver_test.go
emilbon99: Yes, `HaveOcurredAs` works with cockroachdb error wrapping and encoding, which we use extensively, `MatchError` does not.

client/ts/src/channel/payload.ts
emilbon99: We do have a channel class?

pluto/src/flux/sync/listeners.ts
emilbon99: `onChange` needs to be an async function because certain signals (task changes, workspace changes) don't send all their information through the channel (they just send the key), which means we need to fetch information in the callback using a client, hence the async function. 

We could use `then` and `catch` but why should we do that when we could use async.

For all practical purposes we probably don't need the mutex here but there is fundamentally a race condition where a stale value could complete its handler after the recent value and result in a stale modification to state. We could pass an abort controller but then its the callers responsibility to ensure they abort. I'm fine with that but it introduces an extra layer of complexity for the caller.

pluto/src/flux/sync/Sync.spec.tsx
emilbon99: lol

pluto/src/state/state.ts
emilbon99: I renamed the type arguments to make them easier to understand. The motivation for having the `PrevState` type arg is valid IMO. It's particularly useful in all of the aether components, where the caller can return next state as the input of a zod schema while ensuring that prev state is received as the output of a zod schema. 

It also makes sense in cases where prev state may not have been initialized, but the caller is expected to initialize it on the first and all subsequent calls to set state. 

This is a low priority refactor, and I don't think we should spend time on this.

pluto/src/state/state.ts
emilbon99: See my comments above. I think this is low prio.

pluto/src/testutil/Synnax.tsx
emilbon99: This comment isn't very productive IMO. It's a test util function.

emilbon99: Regardless, fixed.

pluto/src/flux/sync/Provider.tsx
emilbon99: This is an unmount effect that ensures the streamer gets closed on component unmount.

pluto/src/flux/result.ts
emilbon99: Because then you could provide a function, which would be incorrectly interpreted as a setter.

pluto/src/flux/result.ts
emilbon99: Because it lifts complexity upward. This ensures that all messages follow the same pattern.

pluto/src/flux/result.ts
emilbon99: Because we can't guarantee that the type of a thrown value is an error. We still need to return it to the caller.

pluto/src/flux/result.ts
emilbon99: Because we can't guarantee that the type of a thrown value is an error. We still need to return it to the caller.

pluto/src/flux/listeners.ts
emilbon99: Fair

pluto/src/flux/index.ts
emilbon99: GPT

pluto/src/channel/queries.ts
emilbon99: A. I think this comment is not productive.
B. If you provide one type argument you must provide all type arguments.

pluto/src/channel/queries.ts
emilbon99: Screenshot 2025-07-22 at 9 19 51 PM

It is documented and space added

pluto/src/flux/update.ts
emilbon99: This function is private and is used as an internal implementation of the two overloads for the update hooks. It is tested.

emilbon99: I'm not going to make it public just to test it directly.

pluto/src/hardware/device/synchronizers.ts
emilbon99: It adds 1 line of code? I don't think that's very verbose, and it uses a composition pattern that separates concerns instead of building a function inheritance tree.

pluto/src/ontology/external.ts
emilbon99: Fixed. Also, if you want me to split the file up, say that instead.

pluto/src/hardware/device/synchronizers.ts
emilbon99: Also, the main reason I did this is so that you could mount listeners as part of a list of listeners inside of the flux queries. Otherwise, the caller would need to mount a bunch of synchronizers explicitly which moves complexity up the stack.

pluto/src/flux/sync/listeners.ts
emilbon99: See my comment on this: https://github.com/synnaxlabs/synnax/pull/1317#discussion_r2224025748

We can talk about this in person.

pluto/src/flux/sync/types.ts
emilbon99: The point of this function is to call a callback when the listener has mounted, that way we can wait to do this like make mutations to items in tests. This ensures that we don't need wacky wait promises that will result in flakey tests.

pluto/src/flux/update.ts
emilbon99: Fair

pluto/src/flux/listeners.ts
emilbon99: This function is part of the private package API. Should we test it independently, or should we test the package interface itself? 

This is kind of like testing private functions inside of go, which we can but I don't think we should do. Shouldn't we be testing to the interface?

pluto/src/flux/form.ts
emilbon99: This is implemented. Comment is incorrect.

pluto/src/channel/queries.ts
emilbon99: Removed

pluto/src/flux/sync/Provider.tsx
emilbon99: I messed up here actually. Thanks for the correction.

pluto/src/flux/params.ts
emilbon99: Very good point. I didn't realize this. Corrected.

pluto/src/flux/listeners.ts
emilbon99: Corrected

pluto/src/channel/queries.ts
emilbon99: Added instructions for capitalization

console/src/range/overview/Snapshots.tsx
emilbon99: This has been fixed in :https://linear.app/synnax/issue/SY-2649/refactor-list-component

pluto/src/flux/list.ts
emilbon99: There was a point where I needed this due to hook mechanics but I don't anymore. Good catch.

pluto/src/flux/retrieve.ts
emilbon99: `useStateful` is used here: https://github.com/synnaxlabs/synnax/blob/sy-2649-refactor-list-component/console/src/hardware/common/device/use.ts

pluto/src/flux/retrieve.ts
emilbon99: useObservable is used internally. I removed it from the public API.

synnax/pkg/distribution/channel/signals/signals.go
emilbon99: This isn't worth testing right now IMO. All you are testing is one function call and its going to take several hundreds of lines of tests. Other things are higher priority.

synnax/pkg/service/workspace/service.go
emilbon99: Don't think its worth it to test signals right now. Low prio

pluto/src/flux/retrieve.ts
emilbon99: Why is `useEffect` not a good name? `useEffect` means execute an operation when parameters change. This is exactly what this function does.

pluto/src/channel/queries.ts
emilbon99: Tests implemented in https://linear.app/synnax/issue/SY-2649/refactor-list-component

pluto/src/workspace/queries.ts
emilbon99: Tests implemented in https://linear.app/synnax/issue/SY-2649/refactor-list-component

pluto/src/ranger/queries.ts
emilbon99: Tests implemented in https://linear.app/synnax/issue/SY-2649/refactor-list-component

pluto/src/hardware/task/queries.ts
emilbon99: Tests implemented in https://linear.app/synnax/issue/SY-2649/refactor-list-component

pluto/src/flux/index.ts
emilbon99: I added it to `form`. Retrieve and list already had listener tests, and update doesn't use them.

pluto/src/label/queries.ts
emilbon99: Tests implemented in https://linear.app/synnax/issue/SY-2649/refactor-list-component

pluto/src/flux/update.ts
emilbon99: Actually this is necessary, as it allows the caller to abort pending operations. I changed this options struct to `FetchOptions`

client/ts/src/ontology/payload.ts
emilbon99: This is a good point, but I"m going to leave as is in this PR. I don't want to type a long explanation, but has to do with async term searcher. I will remove in https://linear.app/synnax/issue/SY-2649/refactor-list-component

emilbon99: Also this comment isn't really meaningfully improving the quality of the codebase. It's one line of code, it doesn't matter

client/ts/src/ontology/payload.ts
emilbon99: This is tested via parseID

pluto/src/ranger/queries.ts
emilbon99: This function has been removed in https://linear.app/synnax/issue/SY-2649/refactor-list-component

pluto/src/flux/sync/listeners.ts
emilbon99: This is going to remain in place for now. I will remove in https://linear.app/synnax/issue/SY-2649/refactor-list-component

client/ts/src/ranger/client.ts
emilbon99: Tests are in https://linear.app/synnax/issue/SY-2649/refactor-list-component

pluto/src/ranger/queries.ts
```diff
@@ -0,0 +1,265 @@
...
+  ),
+};
+
+export const retrieve = Flux.createRetrieve<QueryParams, ranger.Range>({
+  name: "Range",
+  retrieve: async ({ client, params: { key } }) => await client.ranges.retrieve(key),
+  listeners: [SET_LISTENER_CONFIG],
+});
+
+export const rangeFormSchema = z.object({
```
pjdotson: makes more sense to just call this `formSchema`

pluto/src/hardware/task/queries.ts
```diff
@@ -0,0 +1,82 @@
...
+) =>
+  Flux.createRetrieve<QueryParams, task.Task<Type, Config, StatusData> | null>({
+    name: "Task",
+    retrieve: async ({ client, params: { key } }) => {
+      if (key == null) return null;
+      return await client.hardware.tasks.retrieve({ key, schemas });
+    },
+    listeners: [
+      {
+        channel: task.SET_CHANNEL_NAME,
```
pjdotson, line 72: we want the task_delete_channel_name too? along with the task state?

pluto/src/flux/retrieve.ts
```diff
@@ -0,0 +1,397 @@
...
+ *
+ *
+ * // Side effect only - trigger fetching but handle result elsewhere
+ * userRetrieve.useEffect({
+ *   params: { userId: 123 },
+ *   onChange: (result) => analyticsService.track('user_loaded', result)
+ * });
+ * ```
+ */
+export const createRetrieve = <RetrieveParams extends Params, Data extends state.State>(
```
pjdotson, line 387: Update the docs in this file now that there are only useDirect and useEffect

pluto/src/flux/useMountSynchronizers.ts
```diff
@@ -0,0 +1,76 @@
+import { type Destructor } from "@synnaxlabs/x";
```
pjdotson, line 10: header

pjdotson (changes requested):
Good to merge once you address the last comments.

pluto/src/hardware/task/queries.ts
emilbon99: This is addressed in SY-2649

PR #1323, opened 2025-07-11
SY-2638 Add CSV Data Exporting to Server

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2638](https://linear.app/synnax/issue/SY-2638/add-csv-data-exporting-to-server)

## Description

- Add an export/csv API to the HTTP server
- Add changes to Go and TypeScript Freighter implementations to allow for HTTP response body streaming.
- Add CSV Codec and CSV conversion options for frames

## Basic Readiness


synnax/pkg/api/api.go
```diff
@@ -177,15 +175,13 @@ type Layer struct {
 }
 
 // BindTo binds the API layer to the provided Transport implementation.
-func (a *Layer) BindTo(t Transport) {
+func (l *Layer) BindTo(t Transport) {
```
pjdotson, line 178: @emilbon99 Is there some good way to add `text/csv` to the content type - my first thought was
making a custom middleware and letting fhttp add it to the response headers, but I don't
know if you have a better idea.

client/ts/src/export/export.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+      expect(body).toContain(`${dataChannel.name},${indexChannel.name}`);
+      expect(body).toContain("10,1");
+      expect(body).toContain("20,2");
+      expect(body).toContain("30,3");
+      expect(body).toContain("40,4");
+      expect(body).toContain("50,5");
+    });
+    test("should throw an error if no keys are provided", async () => {
+      const res = await client.export.csv({ keys: [] });
+      await expect(res.text()).rejects.toThrow();
```
pjdotson, line 104: @emilbon99 The actual error on the server side doesn't really get propagated through (here it is
failing to open an iterator with no keys). IDK if you have any ideas to the best way to
send that through.

aspen/internal/kv/recovery.go
```diff
@@ -160,7 +160,7 @@ func runSingleNodeRecovery(
 		for {
 			resp, err := stream.Receive()
 			if err != nil {
-				if errors.Is(err, freighter.EOF) {
+				if errors.Is(err, freighter.ErrEOF) {
```
emilbon99: Screenshot 2025-07-15 at 10 40 29 AM

Do with this what you will. Errors are used as sentinel values.

freighter/go/errors.go
```diff
@@ -11,18 +11,23 @@ package freighter
...
-	// StreamClosed is returned when a caller attempts to send or receive a message
+	// ErrEOF is returned when either the receiving or sending end of a Stream exits
+	// normally.
+	ErrEOF = io.EOF
+	// ErrSecurity is returned when a security error occurs.
+	ErrSecurity = errors.New("[freighter] - security error")
+	// ErrStreamClosed is returned when a caller attempts to send or receive a message
 	// from a stream that is already closed.
-	StreamClosed = errors.New("[freighter] - stream closed")
+	ErrStreamClosed = errors.New("[freighter] - stream closed")
```
emilbon99, line 28: This is also a sentinel value

synnax/pkg/api/export.go
```diff
@@ -0,0 +1,105 @@
...
+
+type ExportCSVRequest struct {
+	Keys         channel.Keys           `json:"keys" msgpack:"keys"`
+	TimeRange    telem.TimeRange        `json:"time_range" msgpack:"time_range"`
+	ChannelNames map[channel.Key]string `json:"channel_names" msgpack:"channel_names"`
+}
+
+type ExportCSVResponse = *io.PipeReader
+
+func (es *ExportService) ExportCSV(ctx context.Context, req ExportCSVRequest) (ExportCSVResponse, error) {
```
emilbon99: This API design doesn't make sense IMO. Why don't we simply have a `read` API in framer that then returns a frame back to the caller? Then we have specific codecs that encode that frame into a CSV?

synnax/pkg/api/http/http.go
```diff
@@ -17,117 +17,87 @@ import (
...
+		UserDelete:             fhttp.UnaryServer[api.UserDeleteRequest, types.Nil](router, "/api/v1/user/delete"),
+		UserRetrieve:           fhttp.UnaryServer[api.UserRetrieveRequest, api.UserRetrieveResponse](router, "/api/v1/user/retrieve"),
+		ChannelCreate:          fhttp.UnaryServer[api.ChannelCreateRequest, api.ChannelCreateResponse](router, "/api/v1/channel/create"),
+		ChannelRetrieve:        fhttp.UnaryServer[api.ChannelRetrieveRequest, api.ChannelRetrieveResponse](router, "/api/v1/channel/retrieve"),
+		ChannelDelete:          fhttp.UnaryServer[api.ChannelDeleteRequest, types.Nil](router, "/api/v1/channel/delete"),
+		ChannelRename:          fhttp.UnaryServer[api.ChannelRenameRequest, types.Nil](router, "/api/v1/channel/rename"),
+		ChannelRetrieveGroup:   fhttp.UnaryServer[types.Nil, api.ChannelRetrieveGroupResponse](router, "/api/v1/channel/retrieve-group"),
+		ConnectivityCheck:      fhttp.UnaryServer[types.Nil, api.ConnectivityCheckResponse](router, "/api/v1/connectivity/check"),
+		ExportCSV:              fhttp.UnaryServer[api.ExportCSVRequest, api.ExportCSVResponse](router, "/api/v1/export/csv"),
```
emilbon99: Why didn't we make an `export` endpoint that is polymorphic with the content-type requested by the caller? Also why isn't the API just called `api/v1/frame/read`? Adding a separate `export` endpoint makes it confusing about what you're exporting? Are you exporting ranges, channels, users? This is a frame specific API so it should be in framer?

synnax/pkg/api/export.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, err
+	}
+	headerRecords := binary.NewCSVRecords(1, len(allKeys))
+	headerRecords[0] = lo.Map(channels, func(c channel.Channel, _ int) string {
+		if name, ok := req.ChannelNames[c.Key()]; ok {
+			return name
+		}
+		return c.Name
+	})
+	r, w := io.Pipe()
```
emilbon99: Using a pipe writer is a good design here, but I think we could make this implementation much more flexible in the file formats that it exports.

emilbon99 (changes requested):
Let's talk about implementation design choices

client/ts/src/channel/payload.ts
```diff
@@ -13,6 +13,12 @@ import { z } from "zod";
 import { nullableArrayZ } from "@/util/zod";
 
 export const keyZ = z.number();
+export const keyStringZ = z
```
emilbon99: Why does this need to be a separate exported type? Why can't it just be:

```
export const keyZ = z.number().or(
  .string()
  .refine((val) => !isNaN(Number(val)), {
    message: "Key must be  a number",
  })
  .transform((val) => Number(val)));
```

Like we do with a bunch of other key types

client/ts/src/framer/client.ts
```diff
@@ -46,15 +49,13 @@ export class Client {
...
-    unary: UnaryClient,
-    retriever: channel.Retriever,
-  ) {
-    this.streamClient = stream;
+  constructor(transport: Transport, retriever: channel.Retriever) {
+    this.streamClient = transport.stream;
     this.retriever = retriever;
-    this.deleter = new Deleter(unary);
+    this.deleter = new Deleter(transport.unary);
+    this.frameReader = new FrameReader(transport);
```
emilbon99: Both the property and the type should just be called `Reader`

client/ts/src/framer/reader.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  await client.write(testTimeRange.start, {
+    [indexChannel.key]: [1, 2, 3, 4, 5],
+    [dataChannel.key]: [10, 20, 30, 40, 50],
+  });
+  describe("csv", () => {
+    test("should export csv", async () => {
+      const res = await client.read({
+        keys: [indexChannel.key, dataChannel.key],
+        timeRange: testTimeRange,
+        contentType: "csv",
```
emilbon99: 3 Things:

1. When will there be a situation where the caller wants to read the file format as a CSV string instead of writing directly to a file?
2. `contentType` is a HTTP specific property name. Use `fileType` instead.
3. What happens if the caller wants to read a large file? Say 5GB?

client/ts/src/framer/reader.ts
```diff
@@ -0,0 +1,50 @@
...
+import { type UnaryClient } from "@synnaxlabs/freighter";
+import { TimeRange } from "@synnaxlabs/x";
+import { binary } from "@synnaxlabs/x/binary";
+import { z } from "zod";
+
+import { channel } from "@/channel";
+import { type Transport } from "@/transport";
+
+const contentTypeZ = z
+  .enum(["csv", "application/sy-framer"])
```
emilbon99: Application space content types should be prefixed with `application` right? i.e. `application/json`, `application/csv`

freighter/go/fhttp/unary.go
```diff
@@ -12,204 +12,230 @@ package fhttp
...
+
+func (us *unaryServer[RQ, RS]) encodeAndWrite(ctx *fiber.Ctx, v any) error {
+	contentType := ctx.Accepts(us.resEncoders.Keys()...)
+	getCodec, ok := us.resEncoders[contentType]
+	if !ok {
+		return fiber.ErrNotAcceptable
+	}
+	codec := getCodec()
+	ctx.Set(fiber.HeaderContentType, contentType)
+	debugging := contentType == "text/csv"
```
emilbon99: Get rid of this

freighter/go/fhttp/unary.go
```diff
@@ -12,204 +12,230 @@ package fhttp
...
+		fmt.Println("v", v)
+		fmt.Printf("v type: %T\n", v)
+	}
+	if uReader, ok := v.(UnaryReadable); ok {
+		r, w := io.Pipe()
+		reqCtx := ctx.Context()
+		go func() {
+			for {
+				v, err := uReader.Read()
+				if debugging {
```
emilbon99: Get rid of this. Also, I highly recommend using a debugger instead of pring debugging.

freighter/go/fhttp/unary.go
```diff
@@ -12,204 +12,230 @@ package fhttp
...
+	ctx.Set(fiber.HeaderContentType, contentType)
+	debugging := contentType == "text/csv"
+	if debugging {
+		fmt.Println("contentType", contentType)
+		fmt.Println("v", v)
+		fmt.Printf("v type: %T\n", v)
+	}
+	if uReader, ok := v.(UnaryReadable); ok {
+		r, w := io.Pipe()
+		reqCtx := ctx.Context()
```
emilbon99: You're already using this variable in the outer scope on line 142. Move it up.

freighter/go/fhttp/server.go
```diff
@@ -0,0 +1,75 @@
...
+	defaultEncoders = maps.Map[string, func() binary.Encoder]{}
+	defaultDecoders = maps.Map[string, func() binary.Decoder]{}
+	defaultCodecs   = maps.Map[string, func() binary.Codec]{
+		MIMEApplicationJSON:    func() binary.Codec { return binary.JSONCodec },
+		MIMEApplicationMsgPack: func() binary.Codec { return binary.MsgPackCodec },
+	}
+)
+
+type serverOptions struct {
+	reqDecoders maps.Map[string, func() binary.Decoder]
```
emilbon99: Why does this need to be a map. Can't it just be a function that takes in a string and returns a codec. You're unnecessarily narrowing the callers ability to resolve a custom codec.

freighter/go/fhttp/server.go
```diff
@@ -0,0 +1,75 @@
...
+		encodings.Add(contentType)
+	}
+	return freighter.Reporter{
+		Protocol:  "http",
+		Encodings: encodings.Elements(),
+	}
+}
+
+func init() {
+	for contentType, getCodec := range defaultCodecs {
```
emilbon99: This seems like a very bad use of an init function. Can you justify why you did it?

freighter/go/errors.go
emilbon99: I don't see an explanation for why you chose to call this `ErrStreamClosed` when almost all standard libraries, including go, don't prefix sentinel values with errors. We are coming up with our own standards...

x/go/maps/maps.go
```diff
@@ -0,0 +1,70 @@
...
+package maps
+
+import "maps"
+
+// Map is an alias for the built-in map type that provides additional functionality.
+type Map[K comparable, V any] map[K]V
+
+// Get returns the value associated with the given key. If the key is not found, the
+// zero value of V is returned and ok is false.
+func (m Map[K, V]) Get(key K) (V, bool) {
```
emilbon99: Why is this here? This method does nothing.

x/go/maps/maps.go
emilbon99, line 1: I really don't think we should be implementing our own maps package. The stl maps library covers 95% of the the functionality in this package. I has `maps.Keys`, `maps.Values`, `maps.All`,`maps.Clone` where are assembly optimized implementations of the methods you have here.

We should remove this package, just adds surface area.

x/go/telem/series.go
```diff
@@ -244,6 +244,11 @@ func (s Series) DataString() string {
 	}
 }
 
+// AsCSVStrings returns a slice of strings representing the series as a CSV row. If the
+// series is a variable density series, the strings will be the values of the series.
+// If the series is a fixed density series, the strings will be the values of the series
+// as strings. If the series is a time stamp series, the strings will be the value of
+// the underlying int64 values.
 func (s Series) AsCSVStrings() []string {
```
emilbon99: `As a CSV row`? I don't understand. Isn't this many CSV rows?

Also, this implementation is very very memory hungry. `fmt.Sprintf` is a very slow way to convert values to strings. You're also allocating multiple temporary buffers in order to convert the values here. Why don't you iterate over the raw series data sample by sample and convert that way. This is a hot path function, and we should make sure it doesn't chew through performance.

Also, why isn't this method just called DataStrings(). Converting samples to strings is not unique to a CSV.

emilbon99: This function would also be way faster if you accepted an io.Writer and wrote strings to that, which will bubble up to frame and eventually the codec. You can go from allocating thousands or even millions of strings across hundreds of arrays to allocating a single byte buffer that you flush a frame through.

Performance could go up 100000x

freighter/go/errors.go
emilbon99: Scratch that. This is fine

client/ts/src/framer/reader.ts
```diff
@@ -0,0 +1,50 @@
...
+  private readonly csvClient: UnaryClient;
+
+  constructor(transport: Transport) {
+    this.client = transport.unary;
+    this.csvClient = transport.withDecoder(binary.CSV_CODEC);
+  }
+
+  async read(props: ReadRequest): Promise<Response> {
+    const contentType = contentTypeZ.parse(props.contentType);
+    if (contentType === "csv") {
```
emilbon99:
```
const client = contentType === "csv" ? this.csvClient : this.client;
return await sendRequired(client, "/frame/read", props, readRequestZ)


```

freighter/go/fhttp/client.go
```diff
@@ -10,44 +10,43 @@
...
-type ClientFactoryConfig struct{ Codec httputil.Codec }
-
-var _ config.Config[ClientFactoryConfig] = ClientFactoryConfig{}
-
-func (cfc ClientFactoryConfig) Validate() error {
-	v := validate.New("[ws.streamClient]")
-	validate.NotNil(v, "Codec", cfc.Codec)
-	return v.Error()
+type ClientConfig struct {
+	binary.Encoder
```
emilbon99: This config opens rooms for bugs, why don't you just accept an `fhttputil.Encoder` and an `fhttputil.Decoder` that have `ContentType()` methods that match up with Accept and ContentType here. That way you have zero chance of a mismatch.

freighter/go/fhttp/fhttp.go
```diff
@@ -12,30 +12,16 @@ package fhttp
...
 	BindTo(*fiber.App)
 }
-
-type serverOptions struct {
-	codecResolver httputil.CodecResolver
-}
-
-type ServerOption func(*serverOptions)
-
-func WithCodecResolver(r httputil.CodecResolver) ServerOption {
```
emilbon99, line 40: This was a much nicer API than the one that uses a map IMO, and promotes a much more sustainable interface and pattern of abstraction.

emilbon99: Not to mention it uses a lot less code.

freighter/go/fhttp/stream_test.go
```diff
@@ -0,0 +1,49 @@
...
+		defer GinkgoRecover()
+		Expect(i.app.Listen(host.PortString())).To(Succeed())
+	}()
+	return server, client
+}
+
+func (i *streamImplementation) Stop() error { return i.app.Shutdown() }
+
+var _ = Describe("Stream", func() {
+	AssertStream(&streamImplementation{})
```
emilbon99, line 48: Just putting it out there that I think doing this is a pretty bad way to organize the tests and makes it more confusing than it was before. Also `Assert` is a bad name.

freighter/go/fhttp/unary.go
```diff
@@ -41,44 +41,128 @@ type unaryServer[RQ, RS freighter.Payload] struct {
...
+		go func() {
+			for {
+				v, err := uReader.Read()
+				if debugging {
+					fmt.Println("v", v)
+					fmt.Println("err", err)
+				}
+				if err != nil {
+					w.CloseWithError(err)
```
emilbon99: These errors aren't handled.

freighter/go/fhttp/unary.go
```diff
@@ -41,44 +41,128 @@ type unaryServer[RQ, RS freighter.Payload] struct {
...
+	debugging := contentType == "text/csv"
+	if debugging {
+		fmt.Println("contentType", contentType)
+		fmt.Println("v", v)
+		fmt.Printf("v type: %T\n", v)
+	}
+	if uReader, ok := v.(UnaryReadable); ok {
+		r, w := io.Pipe()
+		reqCtx := ctx.Context()
+		go func() {
```
emilbon99: Can you guarantee that this goroutine will exit? Whenever you spawn one you need to make sure its going to exit eventually.

freighter/go/context.go
```diff
@@ -62,30 +68,10 @@ type SecurityInfo struct {
...
-
-// Get returns the value of the given key, or nil if the key is not set.
-func (p Params) Get(k string) (any, bool) {
-	v, ok := p[k]
-	return v, ok
-}
-
-// Set sets the value of the given key.
-func (p Params) Set(k string, v any) { p[k] = v }
+type Params = maps.Map[string, any]
```
emilbon99: Just make params a standard map and get rid of all of the bad `Get` and `GetDefault` code I added years ago when I thought I was smart.

freighter/go/go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/samber/lo v1.50.0
 	github.com/synnaxlabs/alamos v0.0.0-00010101000000-000000000000
 	github.com/synnaxlabs/x v0.0.0-20220801122519-e4a5e96a532d
+	github.com/vmihailenco/msgpack/v5 v5.4.1
```
emilbon99, line 15: Why do we need a direct dependency on msgpack?

freighter/integration/http/http_test.go
emilbon99, line 1: This is a separate go.mod file, are we going to run these tests in CI?

freighter/ts/src/http.ts
```diff
@@ -47,8 +54,21 @@ export class HTTPClient extends MiddlewareCollector implements UnaryClient {
...
+    const c = new HTTPClient(this.endpoint, this.encoder, this.secure, decoder);
+    c.use(...this.middleware);
+    return c;
+  }
+
   get headers(): Record<string, string> {
-    return { [CONTENT_TYPE_HEADER_KEY]: this.encoder.contentType };
+    return {
+      Accept: this.decoder.contentType,
+      [HEADER_CONTENT_TYPE]: this.encoder.contentType,
```
emilbon99: Either use a constant for both or don't use a constant for either.

synnax/pkg/api/framer.go
```diff
@@ -57,41 +56,157 @@ type FrameService struct {
...
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	response := &frameUnaryReadable{}
+	response.metaData = frameReadMetadata{
+		channels:     channels,
+		channelNames: req.ChannelNames,
+	}
+	response.frameChannel = make(chan Frame, 1)
```
emilbon99: Why do you need both a frame channel and an error channel?

synnax/pkg/api/framer.go
```diff
@@ -57,41 +56,157 @@ type FrameService struct {
...
+	response.errorChannel = make(chan error, 1)
+	go func() {
+		iter, err := fs.internal.Iterator.Open(ctx, framer.IteratorConfig{
```
emilbon99: Why aren't we using stream iterators here? It would be much more performant. Didn't we have a conversation about this?

synnax/pkg/api/framer.go
```diff
@@ -418,35 +541,44 @@ func (s *FrameService) openWriter(
...
 
-type WSFramerCodec struct {
-	*codec.Codec
-	LowerPerfCodec xbinary.Codec
+type wsFramerCodec struct {
+	codec        *codec.Codec
+	lowPerfCodec xbinary.Codec
+}
+
+func NewStaticWSFramerCodec(
```
emilbon99, line 596: You shouldn't be returning private types from public functions. Do you have a linter set up?

x/go/binary/codec.go
```diff
@@ -76,3 +89,19 @@ func MarshalStringInt64(n int64) ([]byte, error) {
 func MarshalStringUint64(n uint64) ([]byte, error) {
 	return []byte(`"` + strconv.FormatUint(n, 10) + `"`), nil
 }
+
+// WrapStreamEncoder is a helper function for implementing Encoder.Encode. It calls
+// Encoder.EncodeStream and returns the data written to the buffer.
+func WrapStreamEncoder(enc Encoder, ctx context.Context, value any) ([]byte, error) {
```
emilbon99: `WrapStreamEncoder` implies that this function wraps an encoder type and returns another encoder. Also, why are you accepting an `Encoder` that already implements the `Encode` method as part of this function?

x/go/binary/codec.go
```diff
@@ -76,3 +89,19 @@ func MarshalStringInt64(n int64) ([]byte, error) {
...
+	if err := enc.EncodeStream(ctx, &buf, value); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+// WrapStreamDecoder is a helper function for implementing Decoder.DecodeStream. It
+// calls Decoder.DecodeStream and returns the error.
+func WrapStreamDecoder(dec Decoder, ctx context.Context, data []byte, value any) error {
+	return dec.DecodeStream(ctx, bytes.NewReader(data), value)
```
emilbon99: Same principle applies to my comment above.

x/go/binary/fallback.go
```diff
@@ -24,7 +24,7 @@ type decodeFallbackCodec struct {
 
 var _ Codec = (*decodeFallbackCodec)(nil)
 
-func NewDecodeFallbackCodec(base Codec, codecs ...Codec) Codec {
+func NewDecodeFallbackCodec(base Codec, codecs ...Codec) *decodeFallbackCodec {
```
emilbon99: Don't return private types from public functions

emilbon99: There's no reason we would ever need to access methods that are not on the codec interface for this type.

x/go/binary/string.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// StringCodec is a codec that encodes and decodes data to and from strings.
+var StringCodec = &stringCodec{}
+
+// StringUnmarshaller is an interface that can be implemented by types to provide a
+// custom string representation for unmarshalling string data.
+type StringUnmarshaller interface{ UnmarshalString(string) error }
+
+type stringCodec struct{}
```
emilbon99, line 28: Why does this codec exist? Are we ever planning on supporting text/plain?

x/go/encoding/csv/csv.go
emilbon99, line 1: Why do we have a separate encoding package? If we're going to have a separate encoding package, then why aren't the other encoders in it?

x/go/encoding/csv/csv.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/synnaxlabs/x/binary"
+	"github.com/synnaxlabs/x/encoding/csv/internal/csv"
+	"github.com/synnaxlabs/x/errors"
+)
+
+// Encoder is an encoder that encodes data to the CSV format.
+var Encoder = &encoder{}
+
+// Marshaler is a type that can marshal itself to a CSV representation.
+type Marshaler interface{ MarshalCSV() ([][]string, error) }
```
emilbon99, line 25: This type needs to be reworked. It has memory hungriness that could crash the server on large exports.

freighter/go/fhttp/stream_test.go
pjdotson: I don't agree with this comment (although would appreciate a better name for `Assert`).
It doesn't make sense to have code in the main `freighter` package test specific
implementations of Freighter. If Freighter is a contract that specific implementations
need to adhere to, those implementations should be tested in their own package. Imagine
a completely different module / third party package with its own Freighter 
implementation - they would have to write this whole similar test suite to test their
own functionality, versus using this provided testutil function.

synnax/pkg/api/framer.go
pjdotson: I have `staticcheck` set up in my IDE, it just doesn't catch this. But we
should integrate a standard linter that we can also run into CI instead of just relying
on whatever linters we have set up in our IDEs.

freighter/integration/http/http_test.go
pjdotson: Yes they are run in CI

x/go/encoding/csv/csv.go
pjdotson: I moved it here because it makes calling things much more friendly:

```go
binary.CSVEncoder.Encode()
```

vs.

```go
csv.Encoder.Encode()
```

Plus, this also lines up more with the stdlib, which is a convention we match already
with io, errors, fs, etc.

I didn't move the other encoders to separate encoding/json, encoding/msgpack, etc. packages yet because the test in codec_test.go
directly references all of those types, so i would have to rewrite the test files.

Making an issue for this:

https://linear.app/synnax/issue/SY-2676/move-xbinary-codecs-to-xencoding

x/go/binary/string.go
pjdotson: It's used in the freighter integration server for a simple encoding formula.

freighter/go/fhttp/fhttp.go
pjdotson: See other comment

freighter/go/fhttp/server.go
pjdotson: This needs to be a map so that the we can understand what types of content types are
supported by the server. That way we can properly parse the request header.

At the bottom is all of the code for how fiber resolves the Accept header on the http
request, including things like */* for any content type, the q-factor weighting, and
<media-type>/*. Unless we expect every CodecResolver to have all of that implementation
inside them, it makes much more sense to start with the list of encodings the server
supports, determine the specific encoding from the "Accept" header and then get a codec
by following. I switched this in the first place because the CodecResolvers didn't have
any of this functionality built in.

So the narrowing of the callers ability to resolve a codec assume that we would have
some instance where a caller wants to get a codec from the accept header in a different
way than a one-to-one mapping between content types and codecs, which seems very like a
very unlikely scenario.

```
// getOffer return valid offer for header negotiation
func getOffer(header string, isAccepted func(spec, offer, specParams string) bool, offers ...string) string {
	if len(offers) == 0 {
		return ""
	}
	if header == "" {
		return offers[0]
	}

	acceptedTypes := make([]acceptedType, 0, 8)
	order := 0

	// Parse header and get accepted types with their quality and specificity
	// See: https://www.rfc-editor.org/rfc/rfc9110#name-content-negotiation-fields
	forEachMediaRange(header, func(accept string) {
		order++
		spec, quality, params := accept, 1.0, ""

		if i := strings.IndexByte(accept, ';'); i != -1 {
			spec = accept[:i]

			// The vast majority of requests will have only the q parameter with
			// no whitespace. Check this first to see if we can skip
			// the more involved parsing.
			if strings.HasPrefix(accept[i:], ";q=") && strings.IndexByte(accept[i+3:], ';') == -1 {
				if q, err := fasthttp.ParseUfloat([]byte(utils.TrimRight(accept[i+3:], ' '))); err == nil {
					quality = q
				}
			} else {
				hasParams := false
				forEachParameter(accept[i:], func(param, val string) bool {
					if param == "q" || param == "Q" {
						if q, err := fasthttp.ParseUfloat([]byte(val)); err == nil {
							quality = q
						}
						return false
					}
					hasParams = true
					return true
				})
				if hasParams {
					params = accept[i:]
				}
			}
			// Skip this accept type if quality is 0.0
			// See: https://www.rfc-editor.org/rfc/rfc9110#quality.values
			if quality == 0.0 {
				return
			}
		}

		spec = utils.TrimRight(spec, ' ')

		// Get specificity
		var specificity int
		// check for wildcard this could be a mime */* or a wildcard character *
		if spec == "*/*" || spec == "*" {
			specificity = 1
		} else if strings.HasSuffix(spec, "/*") {
			specificity = 2
		} else if strings.IndexByte(spec, '/') != -1 {
			specificity = 3
		} else {
			specificity = 4
		}

		// Add to accepted types
		acceptedTypes = append(acceptedTypes, acceptedType{spec, quality, specificity, order, params})
	})

	if len(acceptedTypes) > 1 {
		// Sort accepted types by quality and specificity, preserving order of equal elements
		sortAcceptedTypes(&acceptedTypes)
	}

	// Find the first offer that matches the accepted types
	for _, acceptedType := range acceptedTypes {
		for _, offer := range offers {
			if len(offer) == 0 {
				continue
			}
			if isAccepted(acceptedType.spec, offer, acceptedType.params) {
				return offer
			}
		}
	}

	return ""
}
```

x/go/encoding/csv/csv.go
pjdotson: I rewrote how frames are encoded by directly using a writer. I personally think that
this type is fine for exports that are going to be smaller, as it allows us to take
advantage of properly adding escape characters to cells and ensuring all records have
the same length, which would need to be implemented manually using a writer or not done
at all.

x/go/telem/series.go
pjdotson: Rewrote to use a writer. Benchmarking gave roughly 4000x improvement in memory per op
since memory is now O(keys). It is a similar speed, but could be made ~2x faster if we
wanted to ~double memory usage per op.

client/ts/src/framer/reader.spec.ts
pjdotson: 1. I think it makes sense to keep the return type as `Response`, as if someone would
   want the actual values they would want it as a frame. This is similar to the pandas
   data frame `to_csv` method which takes in a file path instead of returning a csv
   string. If we come across a situation where someone wants the string, they can
   `await res.text()`.
2. Done
3. The `Response` type doesn't actually load the response body into memory, so this
   shouldn't be a problem on the client end. I will add a test on the server side for
   large exports.

client/ts/src/framer/reader.ts
pjdotson: I think it makes sense to just call things csv, json, etc. The `application/json` is
mostly used directly for HTTP request / responses and email attachments as a MIME type.
The API is a little cleaner IMO to just specify csv versus text/csv.

freighter/go/fhttp/unary.go
pjdotson: I'm not sure what the best approach is to trying to handle these errors (as in actually
communicating them to the client). From what i understand we basically have two options:
1. do all of the server-side processing first before sending the response. this can be
   problematic for handling very large amounts of data. however we could send an http
   bad status code and let the clients handle the error.
2. do the stream body response as implemented here. the problem with this is that all of
   the fields in the http response except body are already sent.
3. use http trailers:
   https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Trailer. These
   would work except for the fact that they are not supported by browsers.

synnax/pkg/api/framer.go
pjdotson: why would they be much more performant? I'm fine with rewriting if you think it is
important, but when I ran the benchmarking tests i wrote, the Read calls in cesium took
up 98% of total memory usage - it seems like the other stuff is all small potatoes.

client/ts/src/framer/reader.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export class Reader {
+  private readonly csvClient: UnaryClient;
+
+  constructor(transport: Transport) {
+    this.csvClient = transport.withDecoder(binary.CSV_CODEC);
+  }
+
+  async read(request: ReadRequest): Promise<Response> {
+    return sendRequired(this.csvClient, "/frame/read", request, readRequestZ);
```
emilbon99, line 36: Can we use this method as a replacement for `readLatest` and `read`? Why/why not?

freighter/go/context.go
```diff
@@ -45,70 +50,27 @@ const (
 // Context represents the metadata for a request that is passed to Middleware.
 type Context struct {
 	context.Context
-	// Role indicates the location of the middleware (client or server).
-	Role Role
-	// Variant indicates the variant of the middleware (unary or stream).
-	Variant Variant
+	Role
```
emilbon99, line 53: Why are these types embedded? Seems strange to embed a primitive value into a struct.

freighter/go/context.go
```diff
@@ -45,70 +50,27 @@ const (
...
+	Variant
 	// Protocol is the protocol that the request is being sent over.
 	Protocol string
 	// Target is the address the request is being sent to.
 	Target address.Address
-	// Sec is the security information for the requests/response connection.
-	Sec SecurityInfo
-	// Params is a set of arbitrary parameters that can be set by client side middleware,
-	// and read by server side middleware.
+	SecurityInfo
```
emilbon99, line 59: Why is this type being embedded? If so, where is the point in having a separate struct?

freighter/go/freightfluence/sender.go
```diff
@@ -54,42 +55,43 @@ func (s *Sender[M]) send(ctx context.Context) error {
...
 
 // TransformSender wraps freighter.StreamSenderCloser to provide a confluence compatible
-// interface for sending messages over a network freighter. TransformSender adds
-// a transform function to the Sender. This is particularly useful in cases
-// where network message types are different from the message types used by the
-// rest of a program.
+// interface for sending messages over a network freighter. TransformSender adds a
+// transform function to the Sender. This is particularly useful in cases where network
+// message types are different from the message types used by the rest of a program.
 type TransformSender[I Value, M freighter.Payload] struct {
```
emilbon99: This generic type should be renamed to P as well.

freighter/go/freightfluence/sender.go
```diff
@@ -99,34 +101,36 @@ o:
 }
 
 // MultiSender wraps a slice of freighter.StreamSender(s) to provide a confluence
-// compatible interface for sending messages over a network freighter. MultiSender
-// sends a copy of each message received from the Outlet.
+// compatible interface for sending messages over a network freighter. MultiSender sends
+// a copy of each message received from the Outlet.
 type MultiSender[M freighter.Payload] struct {
```
emilbon99: This generic type should be renamed to P as well.

freighter/go/freightfluence/sender.go
```diff
@@ -137,96 +141,102 @@ o:
...
+	c := errors.NewCatcher(errors.WithAggregation())
+	for _, s := range ms.Senders {
 		c.Exec(s.CloseSend)
 	}
 	return c.Error()
 }
 
+var _ TargetedSender[any] = (*MapTargetedSender[any])(nil)
+
 type MapTargetedSender[M freighter.Payload] map[address.Address]freighter.StreamSenderCloser[M]
```
emilbon99: This generic type should be renamed to P as well.

freighter/go/freightfluence/sender.go
```diff
@@ -286,37 +298,40 @@ o:
...
 }
 
-// MultiTransformSender wraps a slice of freighter.StreamSender(s) to provide a confluence
-// compatible interface for sending transformed messages over a network freighter.
-// MultiTransformSender transforms each input message and sends a copy to each sender.
+// MultiTransformSender wraps a slice of freighter.StreamSender(s) to provide a
+// confluence compatible interface for sending transformed messages over a network
+// freighter. MultiTransformSender transforms each input message and sends a copy to
+// each sender.
 type MultiTransformSender[I Value, M freighter.Payload] struct {
```
emilbon99: This generic type should be renamed to P as well.

freighter/go/testutil/unary.go
```diff
@@ -0,0 +1,93 @@
...
+	UnaryServer = freighter.UnaryServer[Request, Response]
+	UnaryClient = freighter.UnaryClient[Request, Response]
+)
+
+type UnaryImplementation interface {
+	Start(address.Address) (UnaryServer, UnaryClient)
+	Stop() error
+}
+
+func AssertUnary(impl UnaryImplementation) {
```
emilbon99, line 31: Assert is a very bad name for a test. This isn't an assertion, it's literally running a test. Call is `RunUnaryTest` or something.

freighter/ts/src/http.ts
```diff
@@ -75,24 +81,46 @@ export class HTTPClient extends MiddlewareCollector implements UnaryClient {
     });
   }
 
+  withDecoder(decoder: binary.Codec): HTTPClient {
```
emilbon99, line 84: If it's only for setting the decoder, why are you accepting a `Codec` interface, and not just a `decoder` interface.

freighter/ts/src/http.ts
```diff
@@ -75,24 +81,46 @@ export class HTTPClient extends MiddlewareCollector implements UnaryClient {
...
+      "Content-Type": this.encoder.contentType,
     };
   }
 
-  async send<RQ extends z.ZodType, RS extends z.ZodType = RQ>(
+  get secure(): boolean {
+    return this.endpoint.protocol === "https";
+  }
+
+  async send<RQ extends z.ZodType>(
```
emilbon99, line 101: The whole point of freighter is to abstract away the protocol being used, but here's we're explicitly returning an HTTP Response? Why?

freighter/ts/src/middleware.ts
```diff
@@ -22,29 +25,28 @@ export interface Context {
   params: Record<string, string>;
 }
 
-export const ROLES = ["client", "server"] as const;
-export type Role = (typeof ROLES)[number];
+export type Role = "client" | "server";
 
 /** Next executes the next middleware in the chain. */
 export type Next = (ctx: Context) => Promise<[Context, Error | null]>;
```
emilbon99, line 31: These should be renamed according to the same conventions used in freighter/go.

freighter/ts/src/unary.ts
```diff
@@ -15,22 +15,28 @@ import { type Middleware } from "@/middleware";
...
+   * @param reqSchema - The schema to validate the request against.
    * @param resSchema - The schema to validate the response against.
    */
-  send: <RQ extends z.ZodType, RS extends z.ZodType = RQ>(
+  send<RQ extends z.ZodType>(
     target: string,
-    req: z.input<RQ> | z.infer<RQ>,
+    req: z.input<RQ>,
+    reqSchema: RQ,
+  ): Promise<[Response, null] | [null, Error]>;
```
emilbon99, line 33: There's no way we can put HTTP response into the interface of a tool that is designed to abstract away protocols.

client/ts/src/framer/reader.ts
emilbon99: Why are we returning an HTTP response here? This interface doesn't seem pleasant to use, and it isn't consistent with any other of our API calls.

freighter/ts/src/http.ts
emilbon99: Also where is the `withCodec` and `withEncoder` methods that should be there as well?

freighter/ts/src/websocket.ts
```diff
@@ -194,10 +193,7 @@ export class WebSocketClient extends MiddlewareCollector implements StreamClient
 
   private buildURL(target: string, ctx: Context): string {
     const qs = buildQueryString(
-      {
-        [CONTENT_TYPE_HEADER_KEY]: this.encoder.contentType,
-        ...ctx.params,
-      },
+      { "Content-Type": this.encoder.contentType, ...ctx.params },
```
emilbon99, line 196: Why did we remove this? Now it's much easier to accidentally misspell this variable.

freighter/go/fhttp/server.go
emilbon99: Let's talk.

freighter/go/fhttp/stream_test.go
emilbon99: Fine. Call it `Run` though.

synnax/pkg/api/framer.go
emilbon99: I will rewrite.

emilbon99: The point is also that the API would be much more robust by using stream iterators.

emilbon99: Also we're just spawning go-routines without having control over them.

emilbon99:
Major issue with this is that it won't work with channels that have different indexes.

PR #1327, opened 2025-07-23
SY-2332: Add Verification to CI

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2332](https://linear.app/synnax/issue/SY-2332/add-channel-verification-to-ci)

## Description

Add tests and improve parts of the backend

## Basic Readiness


synnax/pkg/storage/mock/cluster_test.go
```diff
@@ -28,6 +28,6 @@ var _ = Describe("Storage", func() {
 		Expect(b.Close()).To(Succeed())
 	},
 		Entry("Memory-backed storage implementation"),
-		Entry("Stamp-backed storage implementation", storage.Config{InMemory: config.Bool(false), Dirname: "./tmp"}),
+		Entry("Stamp-backed storage implementation", storage.Config{InMemory: config.False(), Dirname: "./tmp"}),
```
emilbon99: I think this should be FS and not stamp backed

x/go/encoding/base64/base64.go
emilbon99, line 1: Seems kinda excessive that we need 3 files and a package for a function that is 1 line of code. Why don't we just use the STL?

emilbon99 (changes requested):

x/go/encoding/base64/base64.go
pjdotson: I agree that it seems a little excessive, but we use this pattern like ~15 times, and it
would otherwise be very verbose and harder to read to always call this:

```go
var foo = string(lo.Must(base64.StdEncoding.DecodeString("some string")))
```

instead of this:

```go
var foo = base64.MustDecode("some string")
```

PR #1350, opened 2025-08-15
Sy 2743 pluto value component tweak

nico-synnax:
# Issue Pull Request

## Linear Issue

[SY-2743](https://linear.app/synnax/issue/SY-2743/pluto-value-component-tweak)

## Description

Implemented UI staleness after not receiving tlm after a specified timeout period

## Basic Readiness

- ~[ ] I have added relevant tests to cover the changes to CI.~
- ~[ ] I have added needed QA steps to the [release candidate](/synnaxlabs/synnax/blob/main/.github/PULL_REQUEST_TEMPLATE/rc.md) template that cover these changes.~
- ~[ ] I have updated user-facing documentation to reflect the changes.~

### DIFF
--- a/pluto/src/vis/value/aether/value.ts
+++ b/pluto/src/vis/value/aether/value.ts
@@ -21,6 +21,8 @@ import { type FillTextOptions } from "@/vis/draw2d/canvas";
 import { render } from "@/vis/render";
 
 const FILL_TEXT_OPTIONS: FillTextOptions = { useAtlas: true };
+const STALENESS_COLOR: [number, number, number, number] = [50, 50, 255, 1];
+const STALENESS_TIMEOUT: number = 5000;
 
 const valueState = z.object({
   box: box.box,
@@ -54,6 +56,8 @@ interface InternalState {
   requestRender: render.Requestor | null;
   textColor: color.Color;
   fontString: string;
+  staleTimeout?: ReturnType<typeof setTimeout>;
+  isInitialized: boolean;
 }
 
 export class Value
@@ -68,11 +72,30 @@ export class Value
     const { internal: i } = this;
     i.renderCtx = render.Context.use(ctx);
     i.theme = theming.use(ctx);
-    if (color.isZero(this.state.color)) i.textColor = i.theme.colors.gray.l10;
-    else i.textColor = this.state.color;
+
+    // If not initialzied, set to STALENESS_COLOR
+    if (i.isInitialized === undefined) {
+      i.textColor = STALENESS_COLOR;
+    }
+    
     i.telem = telem.useSource(ctx, this.state.telem, i.telem);
     i.stopListening?.();
-    i.stopListening = i.telem.onChange(() => this.requestRender());
+    i.stopListening = i.telem.onChange(() => {
+      if (color.isZero(this.state.color)) i.textColor = i.theme.colors.gray.l10;
+      else i.textColor = this.state.color;
+
+      if (i.staleTimeout) clearTimeout(i.staleTimeout);
+      // If not initialized, set timout to 10ms for immediate staleness.
+      const timeoutDuration = i.isInitialized ? STALENESS_TIMEOUT : 20;
+      i.staleTimeout = setTimeout(() => {
+        i.textColor = STALENESS_COLOR;
+        i.isInitialized = true;
+        this.requestRender();
+      }, timeoutDuration);
+      
+      // Always re-render on new value
+      this.requestRender();
+    });
     i.fontString = theming.fontString(i.theme, { level: this.state.level, code: true });
     i.backgroundTelem = telem.useSource(
       ctx,
@@ -82,13 +105,13 @@ export class Value
     i.stopListeningBackground?.();
     i.stopListeningBackground = i.backgroundTelem.onChange(() => this.requestRender());
     i.requestRender = render.useOptionalRequestor(ctx);
-    this.requestRender();
   }
 
   afterDelete(): void {
     const { internal: i } = this;
     i.stopListening?.();
     i.stopListeningBackground?.();
+    if (i.staleTimeout) clearTimeout(i.staleTimeout);
     i.telem.cleanup?.();
     i.backgroundTelem.cleanup?.();
     if (i.requestRender == null)
--- a/pluto/src/vis/value/aether/value.ts
+++ b/pluto/src/vis/value/aether/value.ts
@@ -74,9 +74,8 @@ export class Value
     i.theme = theming.use(ctx);
 
     // If not initialzied, set to STALENESS_COLOR
-    if (i.isInitialized === undefined) {
+    if (i.isInitialized === undefined) 
       i.textColor = STALENESS_COLOR;
-    }
     
     i.telem = telem.useSource(ctx, this.state.telem, i.telem);
     i.stopListening?.();


pluto/src/telem/aether/transformers.ts
```diff
@@ -174,6 +174,8 @@ export const stringifyNumberProps = z.object({
   prefix: z.string().optional().default(""),
   suffix: z.string().optional().default(""),
   notation: notation.notationZ.optional().default("standard"),
+  staleness_timeout: z.number().optional().default(5),
+  staleness_color: color.colorZ.optional().default([204, 197, 0, 1]), // pluto-warning-m1: #ccc500
```
pjdotson: 3 things:
- we use camelCase and not snake_case for variable names in JavaScript.
- the staleness shouldn't really be set in the stringifyNumber state. This state is used
  for the operations to transform the raw value (1234.5678) to a certain notation and
  precision (123e4), not anything to do with the color. See my comment in TelemForm
- the default color shouldn't be hardcoded like that, and instead should be set as either

pluto/src/vis/value/TelemForm.tsx
```diff
@@ -107,6 +114,19 @@ export const TelemForm = ({ path }: TelemFormProps): ReactElement => {
             onChange={handleRollingAverageChange}
           />
         </Input.Item>
+        <Input.Item label="Staleness Timeout" align="start">
+          <Input.Numeric
+            value={stringifier.staleness_timeout ?? 2}
+            bounds={{ lower: 0, upper: 61 }}
+            onChange={handleTimeOutChange}
```
pjdotson: I would add the `endContent` prop here and set it to either "s" or "ms" to make this
look a little prettier. I would also get rid of / set a higher upper bound, not any
great reason to set it at 60.

pluto/src/vis/value/TelemForm.tsx
```diff
@@ -73,6 +74,12 @@ export const TelemForm = ({ path }: TelemFormProps): ReactElement => {
   const handlePrecisionChange = (precision: number): void =>
     handleChange({ stringifier: telem.stringifyNumber({ ...stringifier, precision }) });
 
+  const handleTimeOutChange = (staleness_timeout: number): void =>
+    handleChange({ stringifier: telem.stringifyNumber({ ...stringifier, staleness_timeout }) });
+
+  const handleStalenessColorChange = (staleness_color: color.Color): void =>
+    handleChange({ stringifier: telem.stringifyNumber({ ...stringifier, staleness_color }) });
+
```
pjdotson: Like i said in the other file, it doesn't make sense to track the staleness_timeout in
the stringifier. Instead I would edit the type `ValueTelemFormT` to have an additional
stalenessColor and stalenessTimeout field, and then use `<Form.Field />` components like
this section from `schematic/Forms.tsx`:

```ts
                <Form.Field<string>
                  path="units"
                  label="Units"
                  align="start"
                  padHelpText={false}
                >
                  {(p) => <Input.Text {...p} />}
                </Form.Field>
                <Form.NumericField
                  path="inlineSize"
                  label="Value Width"
                  hideIfNull
                  inputProps={valueWidthInputProps}
                />
                <Form.Field<Text.Level>
                  path="level"
                  label="Size"
                  hideIfNull
                  padHelpText={false}
                >
                  {({ value, onChange }) => (
                    <Select.Text.Level value={value} onChange={onChange} />
                  )}
 ```

File: pluto/src/vis/value/aether/value.ts [hunk 4]
```diff
@@ -29,6 +29,8 @@ const valueState = z.object({
   level: text.levelZ.optional().default("p"),
   color: color.colorZ.optional().default(color.ZERO),
   precision: z.number().optional().default(2),
+  staleness_timeout: z.number().optional().default(5),
+  staleness_color: color.colorZ.optional().default(color.ZERO),
```
pjdotson: camelCase not snake_case

pluto/src/vis/value/aether/value.ts
```diff
@@ -192,3 +251,4 @@ export class Value
 export const REGISTRY: aether.ComponentRegistry = {
   [Value.TYPE]: Value,
 };
+
```
pjdotson: Leaving this comment here, but you should set up Prettier and ESLint to run in your IDE
and autoformat on save. This will make your code style match the rest of the code base.
you can also run them on CLI tools. LMK if you need help getting these set up / using them.

pluto/src/vis/value/aether/value.ts
```diff
@@ -54,6 +57,10 @@ interface InternalState {
   requestRender: render.Requestor | null;
   textColor: color.Color;
   fontString: string;
+  staleTimeout?: ReturnType<typeof setTimeout>;
+  isInitialized: boolean;
+  timed_out: boolean;
+  dataCount: number;
```
pjdotson: instead of tracking all of this, i would add a `lastReceived` in the `InternalState`
here (which defaults to undefined). Then, to calculate if something is stale, it is
either undefined (is stale), or you check if sy.TimeStamp.now() is more than the
timeout. Should simplify your code decently.

pluto/src/vis/value/aether/value.ts
```diff
@@ -164,15 +214,24 @@ export class Value
...
         canvas.fillStyle = color.hex(textColor);
       }
     }
-    if (setDefaultFillStyle) canvas.fillStyle = color.hex(this.internal.textColor);
+    if (setDefaultFillStyle)
+      if (!this.internal.timed_out)
+        if (color.isZero(this.state.color)) canvas.fillStyle = color.hex(color.pickByContrast(theme.colors.border, theme.colors.gray.l11, theme.colors.gray.l0));
+        else canvas.fillStyle = color.hex(color.pickByContrast(theme.colors.border, theme.colors.gray.l11, this.state.color));
+      else
+        canvas.fillStyle = color.hex(this.state.staleness_color);
```
pjdotson: you should just reuse the same variable as above. BTW you should check how this
interfaces with the redlining.

pluto/src/vis/value/aether/value.ts
```diff
@@ -164,15 +214,24 @@ export class Value
...
-        );
+        // Use staleness color if timed out, otherwise use active color with contrast
+        let textColor;
+        if (this.internal.timed_out)
+          textColor = this.state.staleness_color;
+        else if (color.isZero(this.state.color)) 
+          textColor = color.pickByContrast(theme.colors.border, theme.colors.gray.l11, theme.colors.gray.l0);
+        else 
+          textColor = color.pickByContrast(theme.colors.border, theme.colors.gray.l11, this.state.color);
+
```
pjdotson: right here is where you can pick the default staleness color in a similar way to how the
default color for non stale values are picked.

pluto/src/vis/value/aether/value.ts
pjdotson: This should let you remove most of the code you added in the `afterUpdate` and
`afterDelete` functions.

pjdotson (changes requested):
Left a lot of comments here. Lets talk on a video call / pair program sometime tomorrow morning to get this PR in a good shape.

pluto/src/vis/value/aether/value.ts
sy-nico: The reason for all of this logic is because after initialization, the value element will always update with a "0.0" value even when stale. Because the core doesn't yet transmit NaN/Null as a valid message, this value needs to be a) ignored by count, b) ignored by time, or c) processed anyway and forced to stale after a short timeout (original implementation)

Until there is a way to transmit staleness with a NaN value, there's no getting around having to bypass the initial onChange(0.0) that occurs after initialization. 

Removing the count logic will result in the value initializing stale, immediately go non-stale 0.0, and then going stale after the 5s timeout.

pluto/src/vis/value/aether/value.ts
sy-nico: For the long-term, would it make sense to keep the default values in their respective form definition files? I'm thinking about what type of pattern would make it easier to eventually have a config page that the users can set up global default values (timeout, auto-start settings, precision, etc.).

pluto/src/vis/value/aether/value.ts
nico-synnax: Addressed in 4734f2. Would still like to talk through the logic and spot check it.

pluto/src/vis/value/aether/value.ts
sy-nico: Will find place where TS sends a default "0.0" and replace with "undefined". Then here we can simply ignore any undefined values

pluto/src/vis/value/aether/value.ts
sy-nico: Placed it in cells/registry.ts and schematic/registry.ts so the color selctors would default to the warning color.

pluto/src/telem/aether/pipeline.ts
```diff
@@ -38,6 +38,10 @@ export const sourcePipelinePropsZ = z.object({
   connections: z.array(connectionZ),
   outlet: z.string(),
   segments: z.record(z.string(), sourceSpecZ),
+  staleness: z.object({
+    stalenessTimeout: z.number(),
+    stalenessColor: z.tuple([z.number(), z.number(), z.number(), z.number()]),
+  }).optional(),
```
pjdotson: this should be removed from here.

pluto/src/telem/aether/transformers.ts
```diff
@@ -186,6 +186,7 @@ export class StringifyNumber extends UnarySourceTransformer<
   schema = StringifyNumber.propsZ;
 
   protected transform(value: number): string {
+    if (isNaN(value)) return ""; // Show empty instead of NaN
```
pjdotson: comments like this aren't really helpful, so they should be removed.

pluto/src/vis/value/TelemForm.tsx
```diff
@@ -60,8 +72,10 @@ export const TelemForm = ({ path }: TelemFormProps): ReactElement => {
         ...segments,
       },
       outlet: "stringifier",
+      // Add staleness as pipeline metadata
+      staleness: currentStaleness,
     });
-    onChange({ ...value, telem: t });
+    onChange({ ...value, telem: t, ...stalenessUpdate });
```
pjdotson: I don't think you should need to touch any of this code. There's no reason that the telem pipeline needs to be aware of the staleness.

pluto/src/telem/aether/transformers.ts
```diff
@@ -277,6 +278,7 @@ export class ScaleNumber extends UnarySourceTransformer<
   schema = ScaleNumber.propsZ;
 
   protected transform(value: number): number {
+    if (isNaN(value)) value = 0;
```
pjdotson: shouldn't you still return NaN here if value is NaN?

pluto/src/vis/value/TelemForm.tsx
```diff
@@ -107,6 +121,27 @@ export const TelemForm = ({ path }: TelemFormProps): ReactElement => {
...
+              onChange={onChange}
+              bordered
+            />
+          )}
+        </Form.Field>
+        <Form.NumericField
+          path="stalenessTimeout"
+          label="Staleness Timeout"
+          inputProps={{
+            bounds: { lower: 0, upper: Infinity },
```
pjdotson, line 136: add the end content here as well

pluto/src/vis/value/aether/value.ts
```diff
@@ -120,6 +144,24 @@ export class Value
       this.setState((p) => ({ ...p, width: Math.max(requiredWidth, p.minWidth) }));
   }
 
+  private getTextColor(): color.Color {
+    const { theme } = this.internal;
+    if (this.internal.isStale) return this.state.stalenessColor;
```
pjdotson: you need to check `color.isZero(this.state.stalenessColor)` and return the default staleness color if that is the case.

pluto/src/vis/value/aether/value.ts
```diff
@@ -89,13 +103,23 @@ export class Value
...
     i.telem.cleanup?.();
     i.backgroundTelem.cleanup?.();
     if (i.requestRender == null)
       i.renderCtx.erase(box.construct(this.state.box), xy.ZERO, ...CANVAS_VARIANTS);
     else i.requestRender("layout");
   }
 
+  private resetStaleTimeout(): void {
+    const { internal: i } = this;
+    if (i.staleTimeout) clearTimeout(i.staleTimeout);
```
pjdotson: two minor points:

1. If you check the type signature of `clearTimeout`, it accepts `undefined`, so you don't need this `if` check.
2. In general, it is better practice to pass explicit booleans to `if` statements, like this:
```ts
if (i.staleTimeout != null) clearTimeout(i.staleTimeout);
```
rather than relying on evaluating `undefined` to `false`.

pluto/src/vis/value/aether/value.ts
```diff
@@ -89,13 +103,23 @@ export class Value
     const { internal: i } = this;
     i.stopListening?.();
     i.stopListeningBackground?.();
+    if (i.staleTimeout) clearTimeout(i.staleTimeout);
```
pjdotson: same comments as below on this if statement.

pluto/src/vis/value/aether/value.ts
```diff
@@ -68,11 +72,21 @@ export class Value
     const { internal: i } = this;
     i.renderCtx = render.Context.use(ctx);
     i.theme = theming.use(ctx);
-    if (color.isZero(this.state.color)) i.textColor = i.theme.colors.gray.l10;
-    else i.textColor = this.state.color;
+
+    // Initialize as stale
```
pjdotson: don't need this comment.

pluto/src/vis/value/aether/value.ts
```diff
@@ -68,11 +72,21 @@ export class Value
...
+    i.isStale = true;
+
     i.telem = telem.useSource(ctx, this.state.telem, i.telem);
     i.stopListening?.();
-    i.stopListening = i.telem.onChange(() => this.requestRender());
+    i.stopListening = i.telem.onChange(() => {
+      const value = i.telem.value();
+      // Reset timeout on any data change - let the timeout handle staleness
```
pjdotson: don't need this comment

pluto/src/vis/value/aether/value.ts
```diff
@@ -54,6 +56,8 @@ interface InternalState {
   requestRender: render.Requestor | null;
   textColor: color.Color;
   fontString: string;
+  staleTimeout?: ReturnType<typeof setTimeout>;
```
pjdotson: It's a little nicer to have it defined as this:

```ts
staleTimeout?: NodeJS.Timeout;
```

It's the same type, but it is a lot easier to read in that case. I know I'm being a little anal on some of these comments, but mostly want to get you up to writing production quality code as quickly as possible.

PR #1356, opened 2025-08-19
SY-2769 - Integration Test Development

nico-synnax:
# Issue Pull Request

## Linear Issue

[SY-2769](https://linear.app/synnax/issue/SY-2769/integration-test-development)

## Description

Create integration test infrastructure using python client
- Configure multiple test sequence
- Sequences can run sequentially, randomly, or async
- TestCase framework handles client streaming and writing connections
- To write a test case, simply subscribe to channels, write a simple run() sequence, and add to a test config
- Easy to extend for different test case types.

Future test cases include:
- Scanning for devices
- Loading task configs
- Loading/Triggering automation configs 
- Verifying core connections
- Expanded benchmarking
- Disconnect/Reconnect tests
- Etc.

## Basic Readiness


.github/workflows/test.integration.yaml
```diff
@@ -1,93 +1,237 @@
-name: "Test - Integration"
+name: Test - Integration (End-to-End)
```
emilbon99: Just call this integration or E2E tests, no need for both

.github/workflows/test.integration.yaml
```diff
@@ -1,93 +1,237 @@
...
 
 on:
   push:
     branches:
-      - "leo/sy-685-integration-testing"
+      - "sy-2769-integration-test-development"
```
emilbon99, line 11: What do we plan on the long term configuration for this being? On pushes to RC and main? Workflow dispatch?

.github/workflows/test.integration.yaml
```diff
@@ -1,93 +1,237 @@
...
+            g++ \
+            cmake \
+            libsystemd-dev \
+            pkg-config \
+            libfreetype6-dev \
+            libpng-dev \
+            libjpeg-dev
+
+      - name: Setup Bazel
+        run: |
```
emilbon99: Is this necessary in the long term? I feel like there's no need to re-configure bazel every time.

.github/workflows/test.integration.yaml
```diff
@@ -1,93 +1,237 @@
...
+      - name: Test Conductor
+        working-directory: integration/test/py/framework
+        run: |
+          export PATH="$HOME/.local/bin:$PATH"
+          poetry run --directory ../../py python3 framework/Test_Conducter.py --name test_conductor --sequence ../testcases/basic_tests.json
 
-      - name: Diff Changes
-        uses: dorny/paths-filter@v3
-        id: filter
+      - name: Upload Benchmark Plot
```
emilbon99: Are benchmarks going in this file or the other workflow file?

client/py/synnax/util/normalize.py
```diff
@@ -8,10 +8,12 @@
...
 
 
-from typing import Any
+from typing import Any, TypeVar
 
+T = TypeVar("T")
 
-def normalize[T](*args: T | tuple[T] | list[T] | None) -> list[T]:
+
+def normalize(*args: T | tuple[T] | list[T] | None) -> list[T]:
```
emilbon99: Reason why you reverted this? For compatibility with older Python versions?

integration/py/pyproject.toml
emilbon99, line 1: What's the long term strategy for the directory conventions here? Is the test conductor inside of `test/py` only every going to run tests that use the python client, or will it also conduct tests that use the C++ client, typescript client, and even direct manipulation of the console.

integration/test/py/framework/TestCase.py
emilbon99, line 1: We've always used `snake_case` for all of our files in Python. Would like to stick to that unless there is a very good reason not to.

integration/test/py/framework/TestCase.py
```diff
@@ -0,0 +1,683 @@
...
+
+    Client Threads Monitoring:
+    - is_client_running() -> bool:
+    - get_client_status() -> str:
+    - get_streamer_status() -> str:
+    - get_writer_status() -> str:
+
+    """
+    
+    def __init__(self, SynnaxConnection: SynnaxConnection, name:str=None, expect: str = "PASSED", **params):
```
emilbon99: All variable names should use `snake_case`. This variable is confusing because it makes it seem like we're accepting a static class type instead of an instance of the class itself.

integration/test/py/framework/Test_Conducter.py
emilbon99, line 1: Should be called `test_conductor.py`

integration/test/py/framework/Test_Conducter.py
```diff
@@ -0,0 +1,962 @@
...
+
+@dataclass
+class TestDefinition:
+    """Data class representing a test case definition from the sequence file."""
+    case: str
+    name: str = None  # Optional custom name for the test case
+    params: Dict[str, Any] = field(default_factory=dict)
+    expect: str = "PASSED"  # Expected test outcome, defaults to "PASSED"
+
+class Test_Conductor:
```
emilbon99: Class names should be `PascalCase`

integration/test/py/framework/Test_Conducter.py
```diff
@@ -0,0 +1,962 @@
...
+    """Manages execution of test sequences with timeout monitoring and result collection."""
+    
+    def __init__(self,
```
emilbon99: Why does this one accept keyword args while test case accepts `SynnaxConnection` parameters? Feel like we should make sure we're standardized and keep defaults in one place.

integration/test/py/framework/Test_Conducter.py
```diff
@@ -0,0 +1,962 @@
...
+                    self.log_message(f"Tests randomized for execution")
+                self._execute_sequence(sequence, tests_to_execute)
+            
+            self.log_message(f"Completed sequence '{sequence['name']}'\n")
+        
+        self.is_running = False
+        self._print_summary()
+        return self.test_results
+    
+    def _display_name(self, obj) -> str:
```
emilbon99: If this is for `TestDefinition` and `TestResult` objects, then why isn't it defined as a sub-class method on them? Also, you just just overload the `__str__` method so that whenever you format one of those objects ijt autoamtically prints the pretty string.

integration/test/py/framework/utils.py
```diff
@@ -0,0 +1,180 @@
...
+#  included in the file licenses/APL.txt.
+
+import sys
+import re
+import platform
+import subprocess
+import gc
+
+# Also suppress stderr for WebSocket errors
+class WebSocketErrorFilter:
```
emilbon99, line 18: Is there a place where we're using raw websockets? Guess I don't understand why we need these filters.

integration/test/py/framework/utils.py
```diff
@@ -0,0 +1,180 @@
...
+            raise ValueError("Name must contain at least one alphanumeric character")
+        
+        sanitized = sanitized.strip('_-')
+        if not sanitized:
+            raise ValueError("Name cannot consist only of hyphens and underscores")
+            
+        return sanitized
+
+
+def get_machine_info():
```
emilbon99, line 71: Long term thought, but all of these should evenually go into a `x` package called `x/os` that just holds utilities for querying host info.

integration/test/py/testcases/example_basic.py
```diff
@@ -0,0 +1,65 @@
...
+import os
+import time
+
+# Set up the path before importing framework modules
+sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
+from framework.TestCase import TestCase, SynnaxConnection
+
+import synnax as sy
+
+class Check_Connection_Basic(TestCase):
```
emilbon99: Fix this class name.

integration/test/py/framework/Test_Conducter.py
```diff
@@ -0,0 +1,962 @@
...
+from enum import Enum, auto
+from dataclasses import dataclass, field
+from datetime import datetime
+import importlib.util
+import traceback
+import synnax as sy
+
+try:
+    # Import from the framework module to ensure we get the same class objects
+    sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
```
emilbon99: Can you explain why you're doing this? Why can't we do direct imports of these?

integration/py/pyproject.toml
emilbon99: Also the `test/py` folder deps need to be managed using `pyproject.toml` file

emilbon99 (changes requested):
Good start. Needs renaming of files, variables, directories to match naming conventions. Also make sure to run black formatter and isort in order to format and sort code.

What's the difference between a test case python file and a test case JSON file? Python is a template and the JSON file holds concrete usages of that template?

What's the long term plan for testing along with our other sdks and/or the console. If we're running true E2E tests, they shouldn't just be restricted to Python. We need to be running code from all of our SDKs at the same time.

.github/workflows/test.integration.yaml
sy-nico: Will resolve this last. We can start with PRs to RC. Let it soak for a bit and get an idea of the cost curve.

.github/workflows/test.integration.yaml
sy-nico: Will likely have a flat workspaces with a few directories - reports, media, logs. This will develop as we build out the tests.

.github/workflows/test.integration.yaml
sy-nico: First we check, then skip if is already installed.

integration/py/pyproject.toml
sy-nico: 1) Long term will have a test/cpp and test/ts. We will need to create an analagous framework for each. We will need to create test cases as quickly as we can respond to PR comments.

2) Also updated test-conductor to set it as an entrypoint in pyproject

integration/test/py/framework/utils.py
sy-nico: Yes, to filter out keepalive ping timeouts from the core

.github/workflows/test.integration.yaml
sy-nico: Only on PRs to main

.github/workflows/test.integration.yaml
```diff
@@ -1,93 +1,550 @@
...
+          path: |
+            integration/test/py/*.png
+            integration/test/py/framework/*.png
+          retention-days: 30
+
+      - name: Debug Test Environment
+        if: failure()
+        shell: bash
+        run: |
+          echo "🔍 Test failed - debugging info:"
```
pjdotson: Will do more detailed reviews of the full workflow on a later PR but we should really be keeping these scripts in their own `.sh` files rather than inlining it all in the workflow.

.github/workflows/test.integration.yaml
```diff
@@ -1,93 +1,550 @@
...
+          poetry run test-conductor --name test-conductor-${{ matrix.os }} --sequence ../test/py/testcases/basic_tests.json
+
+      - name: Upload Benchmark Plot
+        uses: actions/upload-artifact@v4
+        if: always() # Run even if tests fail
+        with:
+          name: benchmark-plots-${{ matrix.os }}
+          path: |
+            integration/test/py/*.png
+            integration/test/py/framework/*.png
```
pjdotson: Don't need to fix now, but when these are actually running we shouldn't be uploading large png files to the git repository and instead look at uploading to GDrive or DigitalOcean.

integration/test/py/pyproject.toml
```diff
@@ -0,0 +1,83 @@
...
+disallow_untyped_defs = true
+disallow_incomplete_defs = true
+check_untyped_defs = true
+disallow_untyped_decorators = true
+no_implicit_optional = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_no_return = true
+warn_unreachable = true
+strict_equality = true
```
pjdotson: for [tool.isort] and [tool.mypy], can you just copy what if in freighter/py?

integration/test/py/framework/utils.py
```diff
@@ -0,0 +1,223 @@
...
+    return sanitized
+
+
+def get_machine_info():
+    """Get machine information programmatically."""
+
+    """ 
+    TODO : 
+    Move host querying utilities into an "x" package named x/os.
+    """
```
pjdotson: instead of a TODO comment make a linear issue and then link to it.

integration/test/py/framework/test_conductor.py
```diff
@@ -0,0 +1,1031 @@
...
+
+            self.log_message(
+                f"==== SEQUENCE {seq_idx + 1}/{len(self.sequences)}: {sequence['name']} ===="
+            )
+            self.log_message(
+                f"Executing {len(sequence['tests'])} tests with {sequence['order']} order...\n"
+            )
+
+            # Prepare tests for execution (randomize if needed)
+            tests_to_execute = sequence["tests"].copy()
```
pjdotson, line 440: how does it randomize here?

integration/test/py/framework/test_case.py
```diff
@@ -0,0 +1,748 @@
...
+    - @property: should_continue
+    - get_state() -> Dict
+
+    Client Threads Monitoring:
+    - is_client_running() -> bool:
+    - get_client_status() -> str:
+    - get_streamer_status() -> str:
+    - get_writer_status() -> str:
+
+    """
```
pjdotson: the lower parts of this docstrings are kinda useless, any IDE will allow you to see all of the properties and methods.

pjdotson (changes requested):
Mostly looks good, left a few comments. Will review the workflow files more in depth when we actually start running them.

You might want to go in and pare down some of the comments / useless documentation GPT gave you, sometimes it's a lot to read through.

integration/test/py/framework/test_case.py
sy-nico: Going to keep these here. Since this is ultimately going to be a framework that users will be able to use for their own framework, I'm purposefully trying to over-comment. I know it's overly pendantic. I just want to make this a beginner-friendly tool for our users.

.github/workflows/test.integration.yaml
sy-nico: I'm going to work on this now. This will be extremely important for keeping development simple moving forward.

integration/test/py/framework/test_conductor.py
sy-nico: # Modify original list
if sequence["order"] == "random": random.shuffle(tests_to_execute)

.github/workflows/test.integration.yaml
sy-nico: Good call. I was originally thinking we just upload the reports and leave the the media files on the servers.

integration/test/py/framework/utils.py
sy-nico: https://linear.app/synnax/issue/SY-2811/move-test-framework-utils-to-x

.github/workflows/test.integration.yaml
sy-nico: Resolved and made massive improvements! Parallelized everything and properly set up the ability to skip a build and use artifacts from a previous successful build. We'll save so much on dev time. Now that it's set up, an easy improvement would be to add dispatch workflow input handling so we can manually trigger the tests whenever.

integration/scripts/build-macos-15.sh
pjdotson, line 1: For all of these script files, can you do the following:

1) move general things like the build and download artifacts to the root ./scripts directory
2) get rid of the useless comments at the top like this:
```sh
# build-macos-15.sh
# Builds Synnax driver and server binaries for macOS
# Used by GitHub Actions workflow: test.integration.yaml
```
3) Please get rid of all of the emojis. Also I would set up a system level prompt for whatever AI tool you are using that says stop using emojis.
4) some of the macos and linux prompts are identical, so they should just be one file.

integration/test/py/framework/test_case.py
pjdotson: Unresolving / I disagree with this.

If you want to make this user friendly, you do that by putting docstrings on the individual methods describing parameters. User's shouldn't have to be digging through the implementation of the code and seeing comments like this:

```py
        # Store for test cases to use
        self.synnax_connection = synnax_connection
```

But the properties / methods type signatures are really useless, as I said a user would see that in his IDE.

PS generally our convention on PRs is that if you think the reviewer may disagree / want to respond to your response, don't resolve the conversation.

.github/workflows/test.benchmark.yaml
```diff
@@ -0,0 +1,93 @@
+name: "Test - Benchmark"
+
+on:
+  push:
+    branches:
+      - "none"
+  workflow_dispatch:
```
pjdotson, line 4: can you remove the push - branches - none? that is kinda useless

.github/workflows/test.integration.yaml
```diff
@@ -1,93 +1,517 @@
...
+name: Test - Integration
+#run-name: Test - Integration (${{ github.run_id }})
 
 on:
-  push:
+  pull_request:
     branches:
-      - "leo/sy-685-integration-testing"
-  workflow_dispatch:
+      - main
```
pjdotson, line 7: do we still want this to run on PRs to main right now? and can you also add a workflow_dispatch send?

.github/workflows/test.integration.yaml
```diff
@@ -1,93 +1,517 @@
...
+          name: synnax-server-windows
+          path: synnax/synnax-v${{ steps.version.outputs.VERSION }}-windows.exe
+          retention-days: 7
+
+      - name: Debug Build Environment
+        if: failure()
+        shell: cmd
+        run: integration/scripts/debug-build-environment-windows.cmd
+
+  build-macos-15:
```
pjdotson: you can (and should) use macos-latest instead of macos-15 (they refer to the same thing): https://github.com/actions/runner-images. Can you also ctrl + f through the repo and replace any other usages of macos-15 with macos-latest?

.github/workflows/test.integration.yaml
sy-nico: Yes, this will run on Main. I am going to set to also run on this branch until it is ready to merge.

.github/workflows/test.integration.yaml
sy-nico: Updated in various places, but didn't go through whole repo. I figure this will be something we actively manage as we commission new dev machines.

integration/test/py/framework/test_case.py
sy-nico: Went ahead and implemented the feedback

integration/scripts/build-macos-15.sh
sy-nico: Implemented all feedback. Consolidated a few scripts to function-Unix.sh

PR #1376, opened 2025-08-25
SY-2605: Add Individual Visualization Mode Control

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2605](https://linear.app/synnax/issue/SY-2605/add-individual-line-plot-pausing)

## Description

- clean up code for lineplot and schematic types files
- add omit utility to x/ts
- add individual pausing of line plots via moving state out of slice state and into individual plot state
- add individual select / pan mode of schematics via moving state out of slice state and into individual schematic state

## Basic Readiness

### DIFF
--- a/console/src/schematic/selectors.ts
+++ b/console/src/schematic/selectors.ts
@@ -155,7 +155,15 @@ export const selectSelectedElementNames = (
 ): (string | null)[] => {
   const elements = selectSelectedElementsProps(state, layoutKey);
   return elements.map((element) => {
-    if (element.type === "node") return element.props.label?.label ?? null;
+    if (
+      element.type === "node" &&
+      "label" in element.props &&
+      typeof element.props.label === "object" &&
+      element.props.label != null &&
+      "label" in element.props.label &&
+      typeof element.props.label.label === "string"
+    )
+      return element.props.label.label;
     return null;
   });
 };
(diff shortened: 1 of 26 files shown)


console/src/schematic/selectors.ts
```diff
@@ -155,7 +155,15 @@ export const selectSelectedElementNames = (
 ): (string | null)[] => {
   const elements = selectSelectedElementsProps(state, layoutKey);
   return elements.map((element) => {
-    if (element.type === "node") return element.props.label?.label ?? null;
+    if (
+      element.type === "node" &&
```
emilbon99: I don't understand how this improves the code? If label is defined isn't it guaranteed to be a string? Why do we need to type check all of these things?

x/ts/src/omit.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { describe, expect, it } from "vitest";
+
+import { omit } from "@/omit";
```
emilbon99: I feel like this utility would go well inside of the record package. `omit.omit` doesn't have any built-in semantics that indicate that its supposed to be used for objects, so I feel like `record.omit` or `record.purge` would be better. We already have a `record.purgeUndefined` utility, so I think that would go well with that.

emilbon99 (approved):

PR #1378, opened 2025-08-26
SY-2804 - Implement Eager Cached Retrieve of List Queries

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2804](https://linear.app/synnax/issue/SY-2804/implement-cached-retrieve-of-list-queries)

## Basic Readiness


client/ts/src/channel/client.ts
```diff
@@ -135,6 +135,7 @@ export class Channel {
     expression = "",
     requires = [],
   }: New & {
+    internal?: boolean;
```
pjdotson, line 138: why are we allowing internal channels at all to be constructed here? shouldn't we just
remove the internal argument?

pluto/src/hooks/useDestructors.ts
```diff
@@ -1,5 +1,5 @@
 import { array, type Destructor } from "@synnaxlabs/x";
```
pjdotson, line 10: header

pjdotson (changes requested):
approved after changes

client/ts/src/channel/client.ts
emilbon99: This param gets specified when we retrieve internal channels from the cluster.

PR #1381, opened 2025-08-26
SY-2599: Add Download as CSV to line plot toolbar

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2599](https://linear.app/synnax/issue/SY-2599/add-download-data-button-to-plots)

## Description

Add a download as csv button to the line plot toolbar that downloads all visible data for a plot as a CSV.

## Basic Readiness


console/src/lineplot/useDownloadAsCSV.ts
```diff
@@ -65,22 +69,54 @@ const download = async ({
...
-  return ({ timeRange, lines, name }: DownloadArgs) =>
+  return ({ timeRanges: timeRange, lines, name }: DownloadAsCSVArgs) =>
     handleError(async () => {
       if (client == null) throw new DisconnectedError();
-      await download({ timeRange, lines, client, name });
+      await download({ timeRanges: timeRange, lines, client, name });
     }, "Failed to download CSV");
 };
+
+export const useDownloadPlotAsCSV = (): ((key: string) => void) => {
```
emilbon99: Minor, but might be worth taking in the key as an argument to the hook as opposed to the returned function. Would save you anonymous lambdas.

x/ts/src/telem/telem.ts
```diff
@@ -1500,6 +1500,31 @@ export class TimeRange implements primitive.Stringer {
...
   }
+
+  /**
+   * Simplify takes the list of `TimeRange`s, makes all of them valid, sorts them, and
+   * merges any overlapping ranges.
+   *
+   * @param ranges - The list of `TimeRange`s to simplify.
+   * @returns A list of simplified `TimeRange`s.
+   */
+  static simplify(ranges: TimeRange[]): TimeRange[] {
```
emilbon99, line 1511: This seems like a general purpose utility function that would be useful in a bunch of other places. Could go well into the `bounds` package where we have a function that simplifies a set of bounds. Then call that function from here. Do with this what you will

emilbon99 (changes requested):
Overall looks good. 2 Things:

1. I think we should send a notification after the export finishes successfully. I don't get any sort of visual feedback that the export was good, and I get this subconscious feeling that it failed when I don't get feedback. We should do this for all of our downloads/exports.
2. If we're gonna use the CSV Icon, I think we should use it for the context menu too.
Approved after comments addressed.

pjdotson:
@emilbon99 I'm going to tackle the feedback one in the other CSV PR just because I moved the core logic to a separate module in that one.

x/ts/src/telem/telem.ts
pjdotson: Decided against this just because of the oddness of converting the start and end from timerange to the lower and upper of bounds. we can move if we end up needing that functionality in bounds.

PR #1382, opened 2025-08-27
SY-2809 - Improve Modal Positioning and Add Hold Delay to Command + W for Window Closing

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2809](https://linear.app/synnax/issue/SY-2809/improve-modal-positioning-and-cmd-w)

## Description

Fixes three things:

1. Improves dialog positioning algorithms (pulled from sy-2617-annotations)
2. Modal positioning when the window gets smaller.
3. Adds a delay to Command + W when attempting to use it to close the window, which means you need to press and hold for at least 350ms before the window closes.

## Basic Readiness

### DIFF
--- a/x/ts/src/spatial/direction/direction.ts
+++ b/x/ts/src/spatial/direction/direction.ts
@@ -50,20 +50,10 @@ export const isDirection = (c: unknown): c is Direction => crude.safeParse(c).su
 export const signedDimension = (direction: CrudeDirection): SignedDimension =>
   construct(direction) === "x" ? "signedWidth" : "signedHeight";
 
-export interface IsX {
-  (direction: Direction): direction is "x";
-  (direction: CrudeDirection): direction is CrudeXDirection;
-}
-
-export const isX = ((direction: CrudeDirection): direction is CrudeXDirection => {
+export const isX = (direction: CrudeDirection): direction is CrudeXDirection => {
   if (direction === "center") return false;
   return construct(direction) === "x";
-}) as IsX;
+};
 
-export interface IsY {
-  (direction: Direction): direction is "y";
-  (direction: CrudeDirection): direction is CrudeYDirection;
-}
-
-export const isY = ((direction: CrudeDirection): direction is CrudeYDirection =>
-  construct(direction) === "y") as IsY;
+export const isY = (direction: CrudeDirection): direction is CrudeYDirection =>
+  construct(direction) === "y";


pluto/src/hooks/useResize.ts
```diff
@@ -88,3 +89,16 @@ const shouldResize = (
   if (triggers.includes("y") && box.height(prev) !== box.height(next)) return true;
   return false;
 };
+
+export const useWindowResize = (onResize: UseResizeHandler) => {
```
pjdotson, line 93: can you add a test for this?

pluto/src/dialog/position.ts
```diff
@@ -7,80 +7,249 @@
...
+  if (parsedXYLoc.success) return parsedXYLoc.data;
+  const parsedLoc = location.location.safeParse(initial);
+  if (parsedLoc.success) {
+    const isX = direction.construct(parsedLoc.data) === "x";
+    return isX
+      ? { x: parsedLoc.data as location.X, y: undefined }
+      : { x: undefined, y: parsedLoc.data as location.Y };
+  }
+  return initial as Partial<location.XY>;
 };
```
pjdotson, line 39: this function could be simplified with better type safety like this:

```ts
if (initial == null) return {x: undefined, y: undefined };
if (typeof initial !== "string") return initial;
if (direction.isX(location)) { // some new function in direction that uses a type guard.
    return {x: location; y: undefined}
}
return {x: undefined; y: location}
```

and isX looks like this:

```ts
const isX = (location: location.Outer): location is location.X
const isX = (location: location.Outer): location is location.Y
const isX = (location: location.Outer): boolean {}
```

pluto/src/dialog/position.ts
```diff
@@ -7,80 +7,249 @@
...
+const normalizePreference = (pref: LocationPreference): Preference => {
+  if (
+    typeof pref === "object" &&
+    pref != null &&
+    "targetCorner" in pref &&
+    "dialogCorner" in pref
+  )
+    return pref;
+  return {
+    targetCorner: pref as Location,
```
pjdotson: couldn't `pref` still be a `Preference` in this branch, if one of either `targetCorner`
or `dialogCorner` is undefined?

pluto/src/dialog/position.ts
```diff
@@ -7,80 +7,249 @@
...
+  initial?: LocationPreference;
+  prefer?: LocationPreference | LocationPreference[];
+  disable?: LocationPreference | LocationPreference[];
 }
 
-const FLOATING_PROPS: Partial<position.DialogProps> = {
-  alignments: ["end"],
-  disable: ["center"],
-  prefer: [{ y: "bottom" }],
+export const parseLocationOptions = (
```
pjdotson: if this is exported it should be tested

pjdotson (changes requested):

pluto/src/dialog/position.ts
emilbon99: No longer exported from package level.

PR #1391, opened 2025-08-29
SY-781: Highlight Context Menu Item on Right Click

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-781](https://linear.app/synnax/issue/SY-781/highlight-context-menu-item-on-right-click)

## Basic Readiness


pluto/src/triggers/Provider.tsx
```diff
@@ -78,6 +78,8 @@ const isInputOrContentEditable = (e: KeyboardEvent): boolean => {
   return false;
 };
 
+const listenerElement = window;
```
pjdotson, line 81: you can keep this if you want but it is kind of a stupid change.

pluto/src/select/Button.tsx
```diff
@@ -63,7 +63,7 @@ export const Button = <K extends record.Key = record.Key>({
   return (
     <CoreButton.Toggle
       {...rest}
-      onChange={onSelect}
+      onChange={() => onSelect()}
```
pjdotson, line 66: also a bit of a useless change

pjdotson (changes requested):
1. I think that the appearance of right clicked should be different than the regular selected appearance. Like maybe it should only be highlighted and not have a blue line to the left.
2. It is odd that when I have one element selected and a right click another element, the context menu for the newly clicked element only appears (which seems like the correct behavior). But, when I have two elements selected and right click a third, the context menu appears for all three. I think it should only appear for the third element then. (If the element was part of the first two elements, then it should appear for both of those elements). This matches the behavior in Finder and Linear.

PR #1395, opened 2025-08-29
SY-2824: Fixed Sorting Implementation in Frame Codec

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2824](https://linear.app/synnax/issue/SY-2824/fix-bad-sorting-in-high-performance-framer-codec)

## Description

- Fixed a race condition where mutating frame keys and/series in place would cause bugs in the frame codec implementation.
- Fixed an issue where rack, task, and device statuses would not have proper timestamps

## Basic Readiness


core/pkg/service/hardware/tracker/tracker.go
pjdotson, line 1: you should add tests for these changes in `tracker_test.go`

core/pkg/distribution/framer/codec/codec.go
```diff
@@ -48,6 +49,40 @@ func writeTimeRange(w *xbinary.Writer, tr telem.TimeRange) {
...
+
+func (s *sorter) sort() { sort.Sort(s) }
+
+// Len implements sort.Interface.
+func (s *sorter) Len() int { return s.offset }
+
+// Less implements sort.Interface.
+func (s *sorter) Less(i, j int) bool { return s.keys[i] < s.keys[j] }
+
+// Swap implements sort.Interface.
```
pjdotson, line 80: IMO we should avoid / remove comments like "X implements sort.Interface". That doesn't give any information to callers about what the function does, and if we want to verify a type implements an interface we should use an interface assertion.

core/pkg/distribution/framer/codec/codec.go
```diff
@@ -80,6 +115,9 @@ type Codec struct {
 	// channels used in dynamic codecs to retrieve information about channels
 	// when Update is called.
 	channels channel.Readable
+	// encodeSorter is used to sort source frames that are being encoded. Used instead
+	// of sorting the frame directly in order ot avoid excess heap allocations
```
pjdotson: typo

pjdotson (changes requested):

core/pkg/distribution/framer/codec/codec.go
emilbon99: I think it gives a lot of information about what the function does. if I see that a function implements an interface, then I know to navigate to the go doc for that interface.

pjdotson: I guess you can write whatever documentation you want on the code you write, but I'm just letting you know as someone reading it these comments are way less helpful then "Swap switches the values at indices i and j"

pjdotson (approved):

PR #1446, opened 2025-09-08
SY-2603: Add Child Range Information to Python Documentation

nico-synnax:
# Issue Pull Request

## Linear Issue
[SY-2603](https://linear.app/synnax/issue/SY-2603/add-child-range-information-to-python-documentation)

## Description

* Add Python client documentation for creating child ranges
* Add ranges test case

## Basic Readiness


emilbon99:
@nico-synnax Marking this as draft until ready to review.

docs/site/src/pages/reference/python-client/ranges.mdx
```diff
@@ -59,6 +59,36 @@ my_range = client.ranges.create(
...
 
+## Creating Child Ranges
+
+There are two ways to create child ranges: by creating a sub-range of a parent range, or
+specifying a parent range.
+
+### Creating a Sub-Range of a Parent
+
+```python
+child_range = parent_range.create_sub_range(
```
emilbon99: We should probably make a breaking change now to rename this to `create_child_range` since that's the proper nomenclature, then creating a passthrough method called `create_sub_range` that is marked as deprecated.

docs/site/src/pages/reference/python-client/ranges.mdx
```diff
@@ -59,6 +59,36 @@ my_range = client.ranges.create(
...
+    name="My Child Range",
+    time_range=sy.TimeRange(
+        start = datetime.strptime("2023-2-12 12:30:00", time_format),
+        end = datetime.strptime("2023-2-12 14:30:00", time_format),
+    ),
+)
+```
+
+### Defining a Parent by Ontology ID
```
emilbon99: Let's remove this documentation for now. The caller is going to see this random new word 'ontology' that we haven't explained anywhere else in the codebase.

docs/site/src/pages/reference/python-client/ranges.mdx
```diff
@@ -106,6 +136,14 @@ my_ranges = client.ranges.retrieve(names=["My Range", my_other_range.key])
 In these examples, Synnax will not raise an error if a range cannot be found. Instead,
 the missing range will be omitted from the returned list.
 
+### Retrieving Child Ranges
```
emilbon99, line 124: Instead of doing this, let's modify the range API to add a `children` or `retrieve_children()` method instead of referencing the ontology here.

docs/site/src/pages/reference/python-client/ranges.mdx
emilbon99: It would probably be worth extending the `concepts/ranges` docs to add explanations of what child ranges are and how they work as well, then reference that page from here.

docs/site/src/pages/reference/python-client/ranges.mdx
sy-nico: Added a @property for children, which returns self.retrieve_children()

docs/site/src/pages/reference/python-client/ranges.mdx
sy-nico: I added a link to the concepts/ranges#child-ranges section, which already exists. Should I expand on that section?

sy-nico: https://docs-git-sy-2603-add-child-range-information-to-p-a122f4-synnax.vercel.app/reference/python-client/ranges#creating-child-ranges

.gitignore
```diff
@@ -258,7 +258,7 @@ assets
 src/**.js
 
 # ||| Build Binary |||
-
 freighter/integration/integration
-integration/integration
+integration/integration/
+!integration/test/
```
pjdotson: why do you need either of these lines? What part of `.gitignore` is covering `integration/test`, and what is `integration/integration`?

client/py/synnax/ranger/client.py
```diff
@@ -448,6 +449,42 @@ def create_sub_range(
...
+        self,
+        *,
+        name: str,
+        time_range: TimeRange,
+        color: str = "",
+        key: RangeKey = UUID(int=0),
+    ) -> Range:
+        """
+        This method is deprecated in favor of create_child_range() and is
+        maintained for backwards compatibility.
```
pjdotson: Replace with:

This method is deprecated and will be removed in a future release. Use create_child_range instead.

client/py/synnax/ranger/client.py
```diff
@@ -448,6 +449,42 @@ def create_sub_range(
...
+    def children(self) -> list[Range]:
+        """Returns a list of child ranges of this range."""
+        return self.retrieve_children()
+
+    def retrieve_children(self) -> list[Range]:
+        """Retrieves all child ranges of this range."""
+        res = self._ontology.retrieve_children(self.ontology_id)
+        range_children = [r for r in res if r.id.type == "range"]
+        return self._client.retrieve(keys=[r.id.key for r in range_children])
+
```
pjdotson, line 485: I don't think there is a reason to have both `children` and a `retrieve_children` method. I would just keep `children`.

client/py/synnax/ranger/client.py
```diff
@@ -432,7 +433,7 @@ def write(
         start = self.time_range.start
         self.__frame_client.write(start, to, series)
 
-    def create_sub_range(
+    def create_child_range(
```
pjdotson, line 436: Add / update unit tests

client/py/synnax/ranger/client.py
```diff
@@ -448,6 +449,42 @@ def create_sub_range(
...
+            color=color,
+            key=key,
+        )
+
+    @property
+    def children(self) -> list[Range]:
+        """Returns a list of child ranges of this range."""
+        return self.retrieve_children()
+
+    def retrieve_children(self) -> list[Range]:
```
pjdotson: Add unit tests for whichever one of these you keep

integration/test/latency/bench_latency_response.py
```diff
@@ -45,13 +45,11 @@ def setup(self) -> None:
...
         )
 
-        # Just make sure to call super() last!
-        super().setup()
-
     def run(self) -> None:
         """
         Run the test case.
         """
+        time.sleep(5)
```
pjdotson: Why do you need to sleep here?

integration/test/latency/bench_latency_report.py
```diff
@@ -42,18 +43,15 @@ def setup(self) -> None:
...
-        super().setup()
-
     def run(self) -> None:
         """
         Run the test case.
         """
 
         # Wait for the "response" to start
-        time.sleep(3)
+        time.sleep(8)
```
pjdotson: instead of a nondeterministic sleep here you should communicate that response has started somehow.

integration/test/ranges_basic.py
pjdotson, line 1: Please don't use the integration testing for stuff like this. The only things that are tested here is the Python client and the core. These things are already tested in the Python client unit tests. Integration testing should be for testing how different parts of the system integrate together - do the python client, driver, console, and core all play together?

integration/test/framework/test_conductor.py
```diff
@@ -725,6 +726,46 @@ def _execute_single_test(self, test_def: TestDefinition) -> TestResult:
...
+        ]
+
+        try:
+            conductor_range = self.client.ranges.create(
+                name=self.name,
+                time_range=sy.TimeRange(
+                    start=self.start_time,
+                    end=datetime.now(),
+                ),
+                color="#000000",  # Do not modify this color
```
pjdotson: Why not just leave the color for the parent range blank?

integration/test/framework/test_conductor.py
```diff
@@ -725,6 +726,46 @@ def _execute_single_test(self, test_def: TestDefinition) -> TestResult:
...
+            )
+            for i, test in enumerate(self.test_results):
+                # Cycle through colors using modulo
```
pjdotson: Bad comment

integration/test/framework/test_conductor.py
```diff
@@ -725,6 +726,46 @@ def _execute_single_test(self, test_def: TestDefinition) -> TestResult:
 
         return result
 
+    def create_ranges(self) -> None:
+        """Create a range in Synnax with the given name and time span."""
+        color_wheel = [
```
pjdotson: you should define this list outside of the `create_ranges` method so the list isn't recreated every time.

.gitignore
sy-nico: integration/integration was there before. Without the added lines, the changes to files in the integration/ directory were being ignored

sy-nico: Something must have been cached improperly when I first made the edit because now things are working fine. I am reverting to the original state.

client/py/synnax/ranger/client.py
sy-nico: ~Already exists in ranges_basic.py. Do you want a separate test specifically for this function?~

client/py/synnax/ranger/client.py
sy-nico: Updated unit test to use create_child_range as well as create_sub_range until it is officially deprecated

integration/test/ranges_basic.py
sy-nico: This test fits here as well and is generally better as the test case data is stored on a the core for later reporting. Not every test is going to test client, driver, console and core all together. Most of the tests will cover only subsets. The more complex set of tests will cover all aspects, but may even need to be broken down into individual test_cases that run async

sy-nico: Another thought, these tests are going to be running across various OSes and hardware, there is definitely value to this test outside of the localized unit tests

integration/test/latency/bench_latency_response.py
sy-nico: No need for it. It has been removed.

integration/test/latency/bench_latency_report.py
sy-nico: I realized this sleep wasn't needed. There doesn't need to be an explicit signal that "report" is writing, since the "response" file is simply there to read any new frames. I have increased that read timeout from 2.5s to 3s.

integration/test/framework/test_conductor.py
```diff
@@ -82,6 +82,32 @@ def __str__(self) -> str:
...
+            "#0000FF",  # Blue (240°)
+            "#8000FF",  # Purple (270°)
+            "#FF00FF",  # Magenta (300°)
+            "#FF0080",  # Rose (330°)
+        ]
+    )
+
+    def __call__(self, index: int) -> str:
+        """Get color by index with automatic modulus to handle any index value."""
+        return self._colors[index % len(self._colors)]
```
pjdotson: You can keep this if you want and it won't cause any bugs, but a) we still need to create a separate version of this list for every test case (not a big deal but a bad practice), and b) it's just more work to maintain.

vs. having a 

COLORS: List[str] = ["#FF0000", ... ]

and later on down calling:

colors = COLORS[i % len(COLORS)].

integration/test/ranges_basic.py
pjdotson: This test does not fit here. The test belongs with the other tests that test functionality of ranges in the python client, in the `client/py/tests\test_ranger.py` file. Why do we need to store the test data on the core for later reporting here - if the test fails one of us can run this locally and fix whatever bug is in there? There is really no point in moving the test functionality to a place further away from the code, it just means the development cycle is worse (integration tests only run on PRs to `main`, not on all PRs that effect the Python client code).

Unit tests in the Python client folder will also be much easier to write then adding new integration tests cases like this, and the tests will run much faster.

To put it differently, if this test only tests the Python client and the core, then let's put it with the other tests that only test the Python client and the core.

I agree that it is helpful to test on various OSs - but that's an argument for rewriting the `test.client.yaml` file to run the tests on PRs / pushes to `main` on all OSs like the workflows that don't require setting up a service, not for moving Python client unit tests away from the rest of the client unit tests.

pjdotson (changes requested):

PR #1448, opened 2025-09-09
SY-2749: Tooltip Updates

nico-synnax:
# Issue Pull Request

## Linear Issue

[SY-2749](https://linear.app/synnax/issue/SY-2749/task-configuration-updates-channels-tooltips-and-more)

## Description

Update tooltips and color of unconfigured changes icon.

## Basic Readiness

- [ ] ~I have added relevant tests to cover the changes to CI.~
- [ ] ~I have updated user-facing documentation to reflect the changes.~

### DIFF
--- a/console/src/hardware/common/task/EnableDisableButton.tsx
+++ b/console/src/hardware/common/task/EnableDisableButton.tsx
@@ -31,7 +31,7 @@ export const EnableDisableButton = ({ path, ...rest }: EnableDisableButtonProps)
       tooltip={
         isSnapshot ? undefined : (
           <Text.Text level="small">
-            {value ? "Disable" : "Enable"} data acquisition
+            {value ? "Disable" : "Enable"} on reconfiguration
           </Text.Text>
         )
       }
--- a/console/src/hardware/common/task/fields/AutoStart.tsx
+++ b/console/src/hardware/common/task/fields/AutoStart.tsx
@@ -7,8 +7,11 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-import { Form } from "@synnaxlabs/pluto";
+import { Form, Tooltip } from "@synnaxlabs/pluto";
 
 export const AutoStart = () => (
-  <Form.SwitchField label="Auto Start" path="config.autoStart" hideIfNull />
+  <Tooltip.Dialog location={{ x: "center", y: "bottom" }}>
+    Automatically start the task when configured
+    <Form.SwitchField label="Auto Start" path="config.autoStart" hideIfNull />
+  </Tooltip.Dialog>
 );
--- a/console/src/hardware/common/task/fields/DataSaving.tsx
+++ b/console/src/hardware/common/task/fields/DataSaving.tsx
@@ -7,8 +7,11 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-import { Form } from "@synnaxlabs/pluto";
+import { Form, Tooltip } from "@synnaxlabs/pluto";
 
 export const DataSaving = () => (
-  <Form.SwitchField label="Data Saving" path="config.dataSaving" />
+  <Tooltip.Dialog location={{ x: "center", y: "bottom" }}>
+    Save task data on Synnax core
+    <Form.SwitchField label="Data Saving" path="config.dataSaving" />
+  </Tooltip.Dialog>
 );
--- a/pluto/src/tabs/Tabs.css
+++ b/pluto/src/tabs/Tabs.css
@@ -163,9 +163,11 @@
             & svg {
                 fill: var(--pluto-gray-l8);
             }
+            /* Indicates unconfigured changes */
             .pluto-icon--circle {
                 height: 0.9em;
                 width: 0.9em;
+                fill: var(--pluto-warning-m1);
             }
             &:has(.pluto-icon--circle) {
                 opacity: 1;


console/src/hardware/common/task/EnableDisableButton.tsx
```diff
@@ -31,7 +31,7 @@ export const EnableDisableButton = ({ path, ...rest }: EnableDisableButtonProps)
       tooltip={
         isSnapshot ? undefined : (
           <Text.Text level="small">
-            {value ? "Disable" : "Enable"} data acquisition
+            {value ? "Disable" : "Enable"} on reconfiguration
```
pjdotson, line 34: I think the previous tooltip was better, this field is only for data acquisition.

console/src/hardware/common/task/fields/AutoStart.tsx
```diff
@@ -7,8 +7,11 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-import { Form } from "@synnaxlabs/pluto";
+import { Form, Tooltip } from "@synnaxlabs/pluto";
 
 export const AutoStart = () => (
-  <Form.SwitchField label="Auto Start" path="config.autoStart" hideIfNull />
+  <Tooltip.Dialog location={{ x: "center", y: "bottom" }}>
+    Automatically start the task when configured
```
pjdotson, line 14: auto start automatically starts the task when the synnax driver boots up.

console/src/hardware/common/task/fields/DataSaving.tsx
```diff
@@ -7,8 +7,11 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-import { Form } from "@synnaxlabs/pluto";
+import { Form, Tooltip } from "@synnaxlabs/pluto";
 
 export const DataSaving = () => (
-  <Form.SwitchField label="Data Saving" path="config.dataSaving" />
+  <Tooltip.Dialog location={{ x: "center", y: "bottom" }}>
+    Save task data on Synnax core
```
pjdotson, line 14: I think this tooltip is a little useless

pluto/src/tabs/Tabs.css
```diff
@@ -163,9 +163,11 @@
             & svg {
                 fill: var(--pluto-gray-l8);
             }
+            /* Indicates unconfigured changes */
             .pluto-icon--circle {
                 height: 0.9em;
                 width: 0.9em;
+                fill: var(--pluto-warning-m1);
```
pjdotson, line 170: I don't think we should color it yellow, IMO it draws too much attention to it and is not what you see in other programs like VSCode

pjdotson (changes requested):
I don't think any of these changes are really that high value add. The things that would be good to have tooltips for would be stuff like the meat and potatoes of the NI Task forms (in `AnalogRead.tsx`, `AIChannelForm.tsx`, etc.). Those ones can have a lot of options that are a little confusing since they are cut and paste from the DAQmx APIs.

P.S. The base branch should be `rc`, not `main`.

console/src/hardware/common/task/EnableDisableButton.tsx
sy-nico: Yes, but it is not clear that it takes effect after reconfiguration

console/src/hardware/common/task/fields/AutoStart.tsx
sy-nico: Would it be better to say "... and on driver boot up" ?

pluto/src/tabs/Tabs.css
sy-nico: We're not like VS Code. This yellow is standard for other "warning' notification states that are not errors (red) or nominal (blue). Without the color the indicator is imperceptible to anyone.

console/src/hardware/common/task/fields/DataSaving.tsx
sy-nico: Will remove

console/src/hardware/common/task/EnableDisableButton.tsx
pjdotson: Right, but every other field on the form only takes place upon reconfiguration as well. It is very odd to only specify this on one tooltip. If you think the reconfiguration behavior is unclear, maybe we need something like a pop up if you click away from the task pane that warns that you need to configure for changes to take effect? But we also have something similar when closing the task tab as well.

console/src/hardware/common/task/fields/AutoStart.tsx
pjdotson: See note above. I think like "Automatically start the task when the Synnax driver boots up" makes sense.

pluto/src/tabs/Tabs.css
```diff
@@ -163,9 +163,11 @@
             & svg {
                 fill: var(--pluto-gray-l8);
             }
+            /* Indicates unconfigured changes */
```
pjdotson, line 166: If you think that what this is applying to is unclear, it makes more sense to add a CSS selector for unsaved changes on the close icon in `Selector.tsx` and apply it here (which could also fix a potential bug where the icon for the tab is a circle anyways.

pluto/src/tabs/Tabs.css
pjdotson: Right, but the point is this isn't a warning but a nominal state - you start editing it, which is a normal thing to do, and this pops up. The warning / error comes when someone tries to close the tab, which we have pop up as a modal.

PR #1451, opened 2025-09-11
SY-2932: Create Consolidated Status Service

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2932](https://linear.app/synnax/issue/SY-2932/create-consolidated-status-service)

## Basic Readiness

### DIFF
--- a/core/pkg/service/status/writer.go
+++ b/core/pkg/service/status/writer.go
@@ -46,24 +46,24 @@ func (w Writer) SetWithParent(
 	ctx context.Context,
 	s *Status,
 	parent ontology.ID,
-) (err error) {
+) error {
 	hasParent := !parent.IsZero()
 	if !hasParent {
 		parent = w.group.OntologyID()
 	}
-	if err = w.validate(*s); err != nil {
-		return
+	if err := w.validate(*s); err != nil {
+		return err
 	}
 	exists, err := gorp.NewRetrieve[string, Status]().WhereKeys(s.Key).Exists(ctx, w.tx)
 	if err != nil {
-		return
+		return err
 	}
 	if err = gorp.NewCreate[string, Status]().Entry(s).Exec(ctx, w.tx); err != nil {
-		return
+		return err
 	}
 	otgID := OntologyID(s.Key)
 	if err = w.otgWriter.DefineResource(ctx, otgID); err != nil {
-		return
+		return err
 	}
 	// Status already exists and parent provided = delete incoming relationships and define new parent
 	// Status already exists and no parent provided = do nothing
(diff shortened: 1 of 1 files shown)


core/pkg/api/status.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Set creates or updates statuses in the cluster.
+func (s *StatusService) Set(
+	ctx context.Context,
+	req StatusSetRequest,
+) (res StatusSetResponse, err error) {
+	// For status setting, we use Create action for new statuses
+	// and Update action for existing ones. Since Set can do both,
+	// we'll use Create permission.
+	if err := s.access.Enforce(ctx, access.Request{
```
pjdotson, line 91: Do you think we should switch the action types to just be read / write access? I can't really think of why you would need separate create and update, although I do see a point of a separate permissions for writing data and deleting data e.g. from channels.

core/pkg/service/status/writer.go
```diff
@@ -0,0 +1,150 @@
...
+// and no parent is provided, the existing parent relationship will be preserved. If an empty
+// parent is provided, the status will be created under the top level "Statuses" group.
+func (w Writer) SetWithParent(
+	ctx context.Context,
+	s *Status,
+	parent ontology.ID,
+) (err error) {
+	hasParent := !parent.IsZero()
+	if !hasParent {
+		parent = w.group.OntologyID()
```
pjdotson, line 52: why are we making a status group here?

core/pkg/service/status/writer.go
pjdotson, line 1: can you remove all the naked returns in this file

core/pkg/service/status/status_suite_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"context"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var ctx = context.Background()
```
pjdotson, line 20: Declaring the single context for all tests is a bad pattern that we should avoid going forward, instead we should set it in a `BeforeEach` clause in each test

core/pkg/service/status/status_test.go
```diff
@@ -0,0 +1,281 @@
...
+	"github.com/synnaxlabs/x/gorp"
+	xio "github.com/synnaxlabs/x/io"
+	"github.com/synnaxlabs/x/kv/memkv"
+	"github.com/synnaxlabs/x/query"
+	xstatus "github.com/synnaxlabs/x/status"
+	"github.com/synnaxlabs/x/telem"
+	. "github.com/synnaxlabs/x/testutil"
+)
+
+var _ = Describe("Status", Ordered, func() {
```
pjdotson, line 31: We should try to avoid using these `Ordered` labels so we can make test more parallelizable and run faster.

emilbon99: Disagree. The only place you can use a `BeforeAll` in ginkgo is when you have an ordered container. Removing this means we would need to open expensive services such as pebble for **every** test. Removing the ordered specified would hurt performance.

core/pkg/service/status/status_suite_test.go
emilbon99: I'm not going to change until we change all at once

core/pkg/api/status.go
emilbon99: This seems like something that's beyond the scope of this feature.

core/pkg/service/status/writer.go
emilbon99: I mean generally speaking all resources in the ontology have a root parent group

PR #1454, opened 2025-09-12
SY-2936: Improve Form Context Handling

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2936](https://linear.app/synnax/issue/SY-2936/fix-form-context)

## Description

Makes it so that calls to `Form.useContext()` no longer return a zero value and instead now throw an error. Also added tests that error messages are actually good.

## Basic Readiness

### DIFF
--- a/pluto/src/hooks/useRequiredContext.ts
+++ b/pluto/src/hooks/useRequiredContext.ts
@@ -15,6 +15,8 @@ export const useRequiredContext = <T>(
 ): NonNullable<T> => {
   const value = use(context);
   if (value == null)
-    throw new NotFoundError(`useRequiredContext: context value is null`);
+    throw new NotFoundError(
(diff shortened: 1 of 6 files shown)


pluto/src/hooks/useRequiredContext.ts
```diff
@@ -15,6 +15,8 @@ export const useRequiredContext = <T>(
 ): NonNullable<T> => {
   const value = use(context);
   if (value == null)
-    throw new NotFoundError(`useRequiredContext: context value is null`);
+    throw new NotFoundError(
```
pjdotson, line 18: Can you write some tests for `useRequiredContext`?

pjdotson (changes requested):
Good but I think the error messages should be "Form.useContext must be ..." instead of "useContext must be ...".

Approved after this error message is changed and `useRequiredContext` tests are added.

PR #1461, opened 2025-09-15
SY-2947: Add Primitive Zero Handling Check to Flux List Get Item

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2947](https://linear.app/synnax/issue/SY-2947/flux-make-it-so-that-key-based-retrieves-do-not-retrieve-on-zero)

## Description

See comment on non-test diff for explanation of this change.

## Basic Readiness

### DIFF
--- a/pluto/src/flux/list.ts
+++ b/pluto/src/flux/list.ts
@@ -556,9 +556,15 @@ export const createList =
 
     const getItem = useCallback(
       ((key?: K | K[]) => {
-        if (key == null) return undefined;
         if (Array.isArray(key))
           return key.map((k) => getItem(k)).filter((v) => v != null);
+        // Zero-value keys that are not null or undefined are common as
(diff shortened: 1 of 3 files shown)


pluto/src/flux/list.ts
```diff
@@ -556,9 +556,15 @@ export const createList =
 
     const getItem = useCallback(
       ((key?: K | K[]) => {
-        if (key == null) return undefined;
         if (Array.isArray(key))
           return key.map((k) => getItem(k)).filter((v) => v != null);
```
pjdotson, line 560: should this be filter for `!primitive.IsZero(v)` instead?

emilbon99: No. This is for the value type, which we don't filter out unless it's null or undefined

PR #1470, opened 2025-09-17
SY-2952: Log Warning instead of returning error

nico-synnax:
# Issue Pull Request

## Linear Issue
[SY-2952](https://linear.app/synnax/issue/SY-2952/labjack-device-search-bug)

## Description
Log warning for Labjack Factory when a task exists, but the device acquisition for that task fails. Do not return error.

## Basic Readiness

- [ ] ~I have added relevant tests to cover the changes to CI.~
- [ ] ~I have updated user-facing documentation to reflect the changes.~

### DIFF
--- a/driver/labjack/factory.cpp
+++ b/driver/labjack/factory.cpp
@@ -35,7 +35,13 @@ common::ConfigureResult configure_read(
     }
     auto [dev, d_err] = devs->acquire(cfg.device_key);
     if (d_err) {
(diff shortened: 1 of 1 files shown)


pjdotson (changes requested):
I don't quite understand why we are doing this. Certainly if a device cannot be found, the configure command for the read and write tasks should return errors, right - we don't successfully configure them?

PR #1473, opened 2025-09-18
SY-2956: Fix `deep.get` and `deep.set` when keys contain periods

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2956](https://linear.app/synnax/issue/SY-2956/fix-deepget-and-deepset-when-keys-contain-periods)

## Description

Fix an issue where forms failed to work because `deep.get` and `deep.set` did not properly work when there are periods in the key.

## Basic Readiness


x/ts/src/deep/path.spec.ts
```diff
@@ -168,6 +168,28 @@ describe("path", () => {
     });
   });
 
+  describe("path includes a .", () => {
+    it("should get a value in case the path includes a zero", () => {
```
emilbon99: in case the path includes a zero?

x/ts/src/deep/path.ts
```diff
@@ -174,24 +181,64 @@ const getIndex = (part: string): number | null => {
...
-    if (v == null) throw new Error(`Path ${path} does not exist. ${part} is null`);
+
+  let i = 0;
+  while (i < parts.length - 1) {
+    // Prefer existing keys that may contain periods
+    const best = findBestKey(result, parts.slice(i, parts.length - 1));
+    let part: string;
+    if (best != null) [part, i] = [best[0], i + best[1]];
+    else {
+      // If no existing key, use the shortest possible (single part)
```
emilbon99: I don't understand the logic of these comments from 219 to 214. Isn't the best key possible one that doesn't contain periods, so why are you preferring keys that do contain periods? A lot of this feels like its GPT generated and needs to be cleaned up.

emilbon99 (changes requested):
Approved, but my guess is that this was claud generated and it could be cleaned up a lot.

PR #1474, opened 2025-09-19
Sy 2951 fix queries for visualizations

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-####]()

## Basic Readiness


core/pkg/distribution/ontology/signals/signals.go
```diff
@@ -34,18 +34,25 @@ func Publish(
 	resourceObserver := observe.Translator[iter.Nexter[ontology.Change], []change.Change[[]byte, struct{}]]{
 		Observable: otg.ResourceObserver,
 		Translate: func(nexter iter.Nexter[ontology.Change]) []change.Change[[]byte, struct{}] {
-			return iter.MapToSlice(ctx, nexter, func(ch ontology.Change) change.Change[[]byte, struct{}] {
+			return iter.MapToSliceWithFilter(ctx, nexter, func(ch ontology.Change) (change.Change[[]byte, struct{}], bool) {
+				if ch.Variant == change.Set {
+					v, err := signals.MarshalJSON(ch.Value)
+					if err != nil {
+						return change.Change[[]byte, struct{}]{}, false
```
pjdotson, line 43: This error gets silently ignored? It seems like if a non-JSON gets sent down this channel we should log an error somehow right?

pjdotson: And should add a test for this case.

x/go/iter/iter.go
```diff
@@ -138,3 +138,14 @@ func MapToSlice[I, O any](ctx context.Context, n Nexter[I], f func(I) O) []O {
 	}
 	return values
 }
+
+func MapToSliceWithFilter[I, O any](ctx context.Context, n Nexter[I], f func(I) (O, bool)) []O {
```
pjdotson, line 146: add test and comment

.github/workflows/test.pluto.yaml
```diff
@@ -126,4 +126,4 @@ jobs:
         run: pnpm lint:pluto
 
       - name: Test
-        run: pnpm test:pluto
+        run: pnpm test:pluto -- --no-file-parallelism
```
pjdotson, line 129: Isn't this going to make CI test longer? and it feels like putting a bandaid on the real issue.

pluto/src/select/Frame.tsx
```diff
@@ -221,17 +228,14 @@ export const Frame = <
...
   virtual = false,
   ...rest
 }: FrameProps<K, E>): ReactElement => {
-  let child: ReactElement;
-  if (multiple === true)
-    child = <MultipleProvider value={value} onChange={onChange} {...rest} />;
-  else child = <SingleProvider value={value} onChange={onChange} {...rest} />;
+  const Provider = (multiple ? MultipleProvider : SingleProvider) as FC<
+    MultipleProviderProps<K> | SingleProviderProps<K>
+  >;
```
pjdotson: why do you need the type assertion here? these should be avoided if possible

pluto/src/index.ts
```diff
@@ -32,7 +32,10 @@ export * from "@/hooks";
 export { Icon } from "@/icon";
 export { Input } from "@/input";
 export { Label } from "@/label";
+export { LinePlot } from "@/lineplot";
+export { Range } from "@/lineplot/range";
```
pjdotson: it's a little weird to have exports `Range` and `Ranger`, maybe we should only be able to access this export through calling `LinePlot.Range`?

pluto/src/tree/core.ts
```diff
@@ -96,13 +96,24 @@ export const moveNode = <K extends record.Key = string>({
...
+  parent?: K | null;
 }
 
 export const removeNode = <K extends record.Key = string>({
   tree,
   keys,
+  parent,
 }: RemoveNodeProps<K>): Node<K>[] => {
   keys = array.toArray(keys);
+  if (parent !== undefined) {
```
pjdotson, line 108: add tests

pluto/src/tree/Tree.tsx
```diff
@@ -166,27 +160,29 @@ export const Tree = <K extends record.Key, E extends record.Keyed<K>>({
...
+        virtual={virtual}
       >
-        {({ index, ...rest }) => children({ index, ...nodes[index], ...rest })}
-      </List.Items>
-    </Select.Frame>
+        <List.Items<K, E>
+          full="y"
+          className={CSS(CSS.B("tree"), className, showRules && CSS.M("show-rules"))}
+          {...rest}
+          displayItems={Infinity}
```
pjdotson, line 181: Why do you need this `displayItems={Infinity}` here? and if you have this you should remove the `displayItems` option from `TreeProps`

console/src/store.ts
```diff
@@ -166,7 +166,7 @@ const createStore = async (): Promise<RootStore> => {
     preloadedState: initialState,
     middleware: (def) => new Tuple(...def(), ...BASE_MIDDLEWARE, persistMiddleware),
     reducer,
-    enablePrerender: true,
+    enablePrerender: false,
```
pjdotson: revert

console/src/ontology/Tree.tsx
```diff
@@ -242,52 +265,39 @@ const Internal = ({ root, emptyContent }: InternalProps): ReactElement => {
...
       const isSingle = !Array.isArray(id);
       const ids = array.toArray(id);
-      const resources = resourceStore.getItem(ids.map((id) => ontology.idToString(id)));
-      if (isSingle) {
-        if (resources[0] == null)
-          throw new Error(`Resource ${ontology.idToString(id)} not found`);
+      const stringIDs = ontology.idToString(ids);
+      const resources = resourceStore.get(stringIDs);
+      if (isSingle)
+        // if (resources[0] == null)
```
emilbon99: Mark: fix this

.github/workflows/test.pluto.yaml
emilbon99: Yes, it will make CI longer. It is a `bandaid` for the real issue. I've spent a cumulative of 10 hours over the last week trying to find the root cause.

We have three options:

1. I keep going until I find the root cause. Doesn't seem like we have bandwidth rn.
2. We don't merge this PR, but the problems are still present in the code that's currently on RC, so that doesn't address the underlying issue either.
3. We remove query related tests that could potentially fail, removing useful coverage for important sections of code.

pluto/src/index.ts
emilbon99: Removed.

pluto/src/tree/Tree.tsx
emilbon99: Will resolve, but moving to a separate PR for clarity: https://linear.app/synnax/issue/SY-2972/fix-list-displayitems

core/pkg/distribution/ontology/signals/signals.go
emilbon99: I added a `DPanic` log here. It's not possible to pass anything down this channel that can't be marshaled to JSON, so it's impossible to test. 

The ontology signals tests have been updated to test the happy path.

pluto/src/tree/core.ts
emilbon99: Will add, but moving to separate PR: https://linear.app/synnax/issue/SY-2973/tree-test-cases

pjdotson (approved):

PR #1479, opened 2025-09-22
SY-2961: Fix Cesium metadata parsing and `memFS` bugs

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2961](https://linear.app/synnax/issue/SY-2961/fix-cesium-metadata-parsing-and-memfs-bugs)

## Description

- Fixed a bug in `x/io/fs` where the truncation bit did not work.
- Fixed a bug in the Cesium metadata parsing where JSON files sometimes did not fully clear previous contents after being opened, leading to invalid JSON in a buffer being encoded.

## Basic Readiness

### DIFF
--- a/console/src/layout/useDropOutside.ts
+++ b/console/src/layout/useDropOutside.ts
@@ -61,11 +61,9 @@ const useDropOutsideMacOS = ({
         if (winLabel == null || winLabel !== Drift.MAIN_WINDOW) return;
         const win = await Window.getByLabel(winLabel);
         if (win == null) return;
-        const sf = await win.scaleFactor();
-        const rawCursor = xy.construct(x, y);
-        const cursor = xy.scale(rawCursor, sf);
+        const cursor = xy.construct(x, y);
         if (windowsContain(cursor)) return;
-        const dropped = onDrop(dragging.current, rawCursor);
+        const dropped = onDrop(dragging.current, cursor);
         drop({ target, dropped });
       }, "Failed to drop outside");
     });
--- a/drift/src/configureStore.ts
+++ b/drift/src/configureStore.ts
@@ -67,8 +67,12 @@ const configureStoreInternal = async <
   ...opts
 }: ConfigureStoreOptions<S, A, M, E>): Promise<EnhancedStore<S, A | Action>> => {
   await runtime.configure();
-
-  const store: EnhancedStore<S, A | Action> = base<S, A, M, E>({
+  // This needs to remain as a `let` definition, because we need to be
+  // able to return the dynamically assigned store instance within the
+  // receivePreloadedStateAndListen.
+  let store: EnhancedStore<S, A | Action> | undefined;
+  // eslint-disable-next-line prefer-const
+  store = base<S, A, M, E>({
     ...opts,
     preloadedState: await receivePreloadedStateAndListen(
       debug,


x/go/io/fs/mem.go
```diff
@@ -298,35 +286,31 @@ func (f *memNode) ModTime() time.Time {
 }
 
 func (f *memNode) Mode() os.FileMode {
+	// TODO: constant
```
emilbon99: This TODO comment needs to be addressed or the comment needs to be improved.

x/go/io/fs/fs_test.go
```diff
@@ -47,391 +38,371 @@ var _ = Describe("FS", func() {
...
+						f := MustSucceed(fs.Open("test_file.txt", os.O_CREATE))
 						Expect(f.Close()).To(Succeed())
 
-						file, err := fs.Open("test_file.txt", os.O_RDONLY)
-						Expect(err).ToNot(HaveOccurred())
-						n, err := file.Write([]byte("tacocat"))
-						Expect(err).ToNot(BeNil())
-						Expect(n).To(Equal(0))
+						file := MustSucceed(fs.Open("test_file.txt", os.O_RDONLY))
+						Expect(file.Write([]byte("tacocat"))).Error().To(HaveOccurred())
```
emilbon99: This test should assert on the actual error type returned

x/go/io/fs/fs_test.go
```diff
@@ -47,391 +38,371 @@ var _ = Describe("FS", func() {
...
 						f = MustSucceed(fs.Open("test_file.txt", os.O_WRONLY))
-						_, err = f.Write([]byte("ipromise"))
-						Expect(err).ToNot(HaveOccurred())
-						_, err = f.Write([]byte("n"))
-						Expect(err).ToNot(HaveOccurred())
-						_, err = f.Write([]byte("e"))
-						Expect(err).ToNot(HaveOccurred())
-						_, err = f.Write([]byte("w"))
-						Expect(err).ToNot(HaveOccurred())
+						Expect(f.Write([]byte("ipromise"))).To(Equal(8))
```
emilbon99, line 279: Shouldn't these be wrapped in MustSucceed calls?

emilbon99 (changes requested):
Seems like a lot of the test cases just assert on the return value of certain methods while not checking the error first i.e. `Expect(w.Write()).To(Equal(8))`. Does this correctly catch and assert on error messages, or will we get obscure failures where 0 != 8 without getting an understanding of the error message first?

x/go/io/fs/fs_test.go
pjdotson: Theres no need, if `f.Write` returns an error then the error message will show that an unexpected error occurred.

emilbon99: Dang I wish I had known this for the tens of thousands of other assertions I wrote

x/go/io/fs/fs.go
```diff
@@ -27,35 +27,37 @@ type File interface {
...
 
+type FileInfo = fs.FileInfo
+
 const defaultPerm = 0o755
 
 type FS interface {
 	// Open opens a file according to the provided flag. The provided flag can be OR-ed
 	// in out of the flags in os, e.g. os.O_CREATE|os.O_WRONLY.
-	Open(name string, flag int) (File, error)
+	Open(string, int) (File, error)
```
emilbon99: It is not clear to me that this int is the flag without reading the documentation, and the argument should be the first source of documentation, not the comments.

x/go/io/fs/permissions.go
```diff
@@ -27,23 +26,11 @@ const (
...
 	OS_OTH_W   = OS_WRITE << OS_OTH_SHIFT
 	OS_OTH_X   = OS_EX << OS_OTH_SHIFT
 	OS_OTH_RW  = OS_OTH_R | OS_OTH_W
 	OS_OTH_RWX = OS_OTH_RW | OS_OTH_X
-
-	OS_ALL_R   = OS_USER_R | OS_GROUP_R | OS_OTH_R
-	OS_ALL_W   = OS_USER_W | OS_GROUP_W | OS_OTH_W
-	OS_ALL_X   = OS_USER_X | OS_GROUP_X | OS_OTH_X
-	OS_ALL_RW  = OS_ALL_R | OS_ALL_W
-	OS_ALL_RWX = OS_ALL_RW | OS_ALL_X
```
emilbon99, line 46: Why were these removed? Just because they are unused variables does not mean they should be removed.

x/go/io/fs/fs.go
```diff
@@ -27,35 +27,37 @@ type File interface {
...
-	Truncate(size int64) error
-	Stat() (os.FileInfo, error)
+	Truncate(int64) error
+	Stat() (FileInfo, error)
 	Sync() error
 }
 
+type FileInfo = fs.FileInfo
+
 const defaultPerm = 0o755
```
emilbon99: Instead of using this, you should be using the appropriate permission bits as defined in `x/go/io/fs/permissions`.

PR #1480, opened 2025-09-22
SY-2963: Improve Organization and Display of Telem Related Text

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-2963](https://linear.app/synnax/issue/SY-2963/improve-display-of-time-related-text)

## Description

Consolidates text for `TimeRange`, `TimeSpan`, and `TimeStamp` into it's own `Text` package within `Telem`. Also does some minor refactoring to x telem `toString()` methods for simplicity.

## Basic Readiness

### DIFF
--- a/x/ts/src/telem/telem.spec.ts
+++ b/x/ts/src/telem/telem.spec.ts
@@ -950,7 +950,7 @@ describe("TimeSpan", () => {
 
       // Minutes without seconds (>= 5 minutes)
       [TimeSpan.seconds(300), "5m"],
-      [TimeSpan.seconds(330), "5m"],  // seconds dropped
+      [TimeSpan.seconds(330), "5m"], // seconds dropped
       [TimeSpan.minutes(30), "30m"],
       [TimeSpan.minutes(59), "59m"],
 
@@ -964,7 +964,7 @@ describe("TimeSpan", () => {
 
       // Hours without minutes (>= 3 hours)
       [TimeSpan.minutes(180), "3h"],
-      [TimeSpan.minutes(195), "3h"],  // minutes dropped
+      [TimeSpan.minutes(195), "3h"], // minutes dropped
       [TimeSpan.hours(12), "12h"],
       [TimeSpan.hours(23), "23h"],
 
@@ -976,7 +976,7 @@ describe("TimeSpan", () => {
 
       // Days without hours (>= 2 days)
       [TimeSpan.hours(48), "2d"],
-      [TimeSpan.hours(50), "2d"],  // hours dropped
+      [TimeSpan.hours(50), "2d"], // hours dropped
       [TimeSpan.days(3), "3d"],
       [TimeSpan.days(6), "6d"],
 
@@ -988,7 +988,7 @@ describe("TimeSpan", () => {
 
       // Weeks without days (>= 2 weeks)
       [TimeSpan.days(14), "2w"],
-      [TimeSpan.days(15), "2w"],  // days dropped
+      [TimeSpan.days(15), "2w"], // days dropped
       [TimeSpan.days(21), "3w"],
       [TimeSpan.days(28), "4w"],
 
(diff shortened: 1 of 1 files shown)


x/ts/src/telem/telem.ts
```diff
@@ -834,6 +839,73 @@ export class TimeSpan
...
   }
 
+  private toSemanticString(): string {
+    const absValue = this.valueOf() < 0n ? -this.valueOf() : this.valueOf();
+    const span = new TimeSpan(absValue);
+    const isNegative = this.valueOf() < 0n;
+
+    if (span.valueOf() === 0n) return "0s";
+
+    if (span.lessThan(TimeSpan.SECOND)) return "< 1s";
```
pjdotson, line 851: seems like this should still show the milliseconds or microseconds or nanoseconds, right?

x/ts/src/telem/telem.ts
```diff
@@ -834,6 +839,73 @@ export class TimeSpan
...
+
+    const totalDays = span.days;
+    const totalHours = span.hours;
+    const totalMinutes = span.minutes;
+    const totalSeconds = span.seconds;
+
+    const years = Math.floor(totalDays / 365);
```
pjdotson: this should be 365.2425

x/ts/src/telem/telem.ts
```diff
@@ -834,6 +839,73 @@ export class TimeSpan
...
+    const months = Math.floor(totalDays / 30);
```
pjdotson: This should be 30.436875

x/ts/src/telem/telem.ts
```diff
@@ -834,6 +839,73 @@ export class TimeSpan
...
+      }
+    } else if (minutes >= 1) {
+      result = `${minutes}m`;
+      if (minutes < 5) {
+        const remainingSeconds = Math.floor(totalSeconds - minutes * 60);
+        if (remainingSeconds > 0) result += ` ${remainingSeconds}s`;
+      }
+    } else result = `${seconds}s`;
+
+    return isNegative ? `-${result}` : result;
```
pjdotson: you could avoid a bunch of the nesting if you start with result as either "" or "-" and then have early returns.

x/ts/src/telem/telem.ts
```diff
@@ -805,9 +804,15 @@ export class TimeSpan
...
    * Returns a string representation of the TimeSpan.
    *
+   * @param format - Optional format for the string representation. Defaults to "full".
+   *   - "full": Shows all non-zero units with full precision (e.g., "2d 3h 45m 12s 500ms")
+   *   - "semantic": Shows 1-2 most significant units (e.g., "2d 3h")
    * @returns A string representation of the TimeSpan.
    */
-  toString(): string {
+  toString(format: TimeSpanStringFormat = "full"): string {
+    if (format === "semantic") return this.toSemanticString();
```
pjdotson, line 815: when is this semantic string ever used?

pjdotson (changes requested):

x/ts/src/telem/telem.ts
emilbon99: This is going to be used in the status explorer

x/ts/src/telem/telem.ts
emilbon99: I don't know yet. This is for presentation to the user as an upcoming duration or duration after (i.e. 5s ago). I don't think they can perceive less than a millisecond. I'm thinking I'll leave as is for now, and then change presentation if necessary in future.

x/ts/src/telem/telem.ts
emilbon99: Originally I changed this code but now I disagree with this change. I think that for user-presentational formatting, using a value of 365 is easier to reason about. When I think about 'the number of days in a year' I think '365' regardless of whether that's actually the case. This means when I do (TimeSpan.days(365).toString("semantic") I would expect 1 year to have passed. Because this is _not_ involved in any precise calculations I don't think changing this constant is necessary.

x/ts/src/telem/telem.ts
emilbon99: See my comment above

x/ts/src/telem/telem.ts
pjdotson: Copy

x/ts/src/telem/telem.ts
pjdotson: It seems like the answer to this is to use rounding instead of floors, right? Like if you type in `TimeSpan.days(364)` we get 12m 4days or something like that, which is a little odd.

emilbon99: I mean you would get 12 months? Not 12 months 4 days

emilbon99: The output is `12mo`

emilbon99: I tried rounding and it introduced a whole host of other problems. I can spend half the afternoon solving this or we can accept that using 30 is not such a big deal and focus on more important things

pjdotson: Sounds good

PR #1481, opened 2025-09-22
SY-2937: Setpoint Node Support and Test Case

nico-synnax:
# Issue Pull Request

## Linear Issue
[SY-2937(https://linear.app/synnax/issue/SY-2937/write-console-integration-tests)

## Description

Console UI Tests:
- Added support for "Setpoint" nodes
- Create SchematicNode class for future nodes to inheret from
- Created a new unit test
- Console:create_a_channel
- Dropdown selection function
- Test_case:get_value() function from synnax client

## Basic Readiness

- [ ] ~I have updated in-code documentation to reflect the changes.~
- [x] ~I have updated user-facing documentation to reflect the changes.~


integration/test/framework/test_case.py
```diff
@@ -503,6 +503,25 @@ def read_tlm(
...
+            if latest_value is not None and len(latest_value) > 0:
+                return float(latest_value)
+
+            # If read_latest is empty, read recent time range
+            now = sy.TimeStamp.now()
+            recent_range = sy.TimeRange(now - sy.TimeSpan.SECOND * 3, now)
+            frame = self.client.read(recent_range, channel_name)
+            if len(frame) > 0:
+                return float(frame[-1])
+
```
pjdotson, line 519: why do we need this case of reading the last 3 seconds?

integration/test/console/schematic.py
```diff
@@ -9,127 +9,236 @@
...
 
-class Schematic(Console):
-    """
-    Parent class for schematic tests
-    """
 
-    def setup(self) -> None:
-        super().setup()
-        self.create_page("Schematic")
+class SchematicNode(ABC):
```
pjdotson: `SchematicSymbol` seems like a better name

integration/test/console/schematic.py
```diff
@@ -9,127 +9,236 @@
...
+        edit_off_icon = self.page.get_by_label("pluto-icon--edit-off")
+        if edit_off_icon.count() > 0:
+            edit_off_icon.click()
 
-    def _get_node(self, node_id: str) -> Any:
-        if not node_id:
-            raise ValueError("node_id cannot be empty")
-        node = self.page.get_by_test_id(node_id)
-        node.locator("div").first.click()
+    def _click_node(self) -> Any:
```
pjdotson: do we need to return `Any` here?

integration/test/console/schematic.py
```diff
@@ -9,127 +9,236 @@
...
 
-    def add_schematic_value(
+    def set_label(self, label: str) -> None:
+        self._click_node()
+        self.page.get_by_text("Style").click()
+        label_input = (
+            self.page.locator("text=Label").locator("..").locator("input").first
+        )
+        label_input.fill(label)
+        self.label = label
```
pjdotson: just leaving a small comment to say that this is exactly how you should be writing functions to test this and this is great.

integration/test/console/console.py
pjdotson, line 1: It seems like these should be part of a `channel.py` file

integration/test/console/console.py
```diff
@@ -135,3 +136,71 @@ def ESCAPE(self) -> None:
     @property
     def ENTER(self) -> None:
         self.page.keyboard.press("Enter")
+
+    def create_a_channel(
+        self,
+        channel_name: str,
+        virtual: bool = False,
+        is_index: bool = False,
+        data_type: str = "Timestamp",
```
pjdotson: This should be a `sy.CrudeDataType`

integration/test/console/console.py
```diff
@@ -135,3 +136,71 @@ def ESCAPE(self) -> None:
...
+        channel_found = False
+        item_selector = self.page.locator(".pluto-list__item").all()
+        for item in item_selector:
+            if item.is_visible() and input_text in item.inner_text().strip():
+                item.click()
+                channel_found = True
+                break
+
+        if not channel_found:
+            raise RuntimeError(f"Could not find channel '{input_text}' in dropdown")
```
pjdotson, line 207: why is this channel_found if it is a general utility for selecting from a dropdown?

integration/test/console/schematic_set_output.py
```diff
@@ -0,0 +1,57 @@
...
+            CHANNEL_NAME,
+            data_type="Float64",
+            is_index=False,
+            index=INDEX_NAME,
+        )
+
+        setpoint_node = self.add_to_schematic("Setpoint", CHANNEL_NAME)
+        setpoint_node.move(-200, 0)
+
+        value_node = self.add_to_schematic("Value", CHANNEL_NAME)
```
pjdotson: It seems like I would rather call self.create_setpoint and self.create_value, and these internally call add_to_schematic?

integration/test/console/schematic_set_output.py
```diff
@@ -0,0 +1,57 @@
...
+        value_node.move(200, 0)
+
+        set_p_value = 47.23333333
```
pjdotson: why does it need to be precise down to like 8 digits?

integration/test/cases/console/schematic_set_output.py
```diff
@@ -0,0 +1,57 @@
...
+        value_node.move(200, 0)
+
+        set_p_value = 47.23333333
+        self._log_message(f"Verifying setpoint value: {set_p_value}")
+        setpoint_node.set_value(set_p_value)
+        actual_value = self.get_value(CHANNEL_NAME)
+
+        assert (
+            actual_value == set_p_value
+        ), f"Setpoint value mismatch!\nActual: {actual_value}\nExpected: {set_p_value}"
```
pjdotson, line 51: You shouldn't be using assertions in production code, they are mostly for debugging.

integration/test/console/schematic_edit_value_props.py
```diff
@@ -7,22 +7,20 @@
...
     Add a value component and edit its properties
     """
 
     def run(self) -> None:
 
         self._log_message("Checking default properties of schematic value")
-        # Create a schematic value with default properties
-        node_id = self.add_schematic_value(f"{self.name}_uptime")
-        default_props = self.get_schematic_value_props(node_id)
+        node = self.add_to_schematic("Value", f"{self.name}_uptime")
```
pjdotson: why do we want this _uptime suffix?

integration/test/console/schematic.py
pjdotson, line 1: you should split this file up some. I would make a schematic directory and then have a schematic.py, schematic_symbol.py, setpoint.py, value.py, etc. Generally a good rule of thumb is 1 class for file (we break this other places in the code base and it makes it hard to read, smaller files are better).

integration/test/console/schematic.py
```diff
@@ -9,127 +9,236 @@
...
+        self._click_node()
+
+        # Fill the input and set the value
+        value_input = self.node.locator("input[type='number'], input").first
+        value_input.fill(str(value))
+        set_button = self.node.locator("button").filter(has_text="Set")
+        set_button.click()
+
+
+class ValueNode(SchematicNode):
```
pjdotson: this class should just be called `Value`

integration/test/console/schematic.py
```diff
@@ -9,127 +9,236 @@
...
+    stale_timeout: int
+
+    def __init__(self, page: Page, node_id: str, channel_name: str):
+        super().__init__(page, node_id, channel_name)
+
+    def edit_properties(
+        self,
+        channel_name: Optional[str] = None,
+        properties: Optional[Dict[str, Any]] = None,
+    ) -> Dict[str, Any]:
```
pjdotson: instead of these `Dict[str, Any]` you should be able to define the properties - the only ones they can set are channel name, notation, precision like above.

integration/test/console/schematic.py
```diff
@@ -194,82 +303,158 @@ def get_schematic_value_props(self, node_id: str) -> Dict[str, Any]:
...
+    ) -> SchematicNode:
+        """Factory method to create node objects"""
+        if node_type.lower() == "value":
+            return ValueNode(self.page, node_id, channel_name)
+        elif node_type.lower() == "setpoint":
+            return SetpointNode(self.page, node_id, channel_name)
+        else:
+            raise ValueError(f"Unknown node type: {node_type}")
+
+    def add_to_schematic(
```
pjdotson: It seems like `create_node` and `add_to_schematic` should be one function called `create_symbol`, right?

integration/test/console/schematic.py
```diff
@@ -194,82 +303,158 @@ def get_schematic_value_props(self, node_id: str) -> Dict[str, Any]:
...
+        self, node_type: str, channel_name: str, properties: Dict[str, Any] = {}
+    ) -> SchematicNode:
```
pjdotson: But the problem with this api is this. You have this code in one of the tests:

```py
        setpoint_node = self.add_to_schematic("Setpoint", CHANNEL_NAME)
        setpoint_node.move(-200, 0)

        value_node = self.add_to_schematic("Value", CHANNEL_NAME)
        value_node.move(200, 0)
```

It would be way better to call this:

```py
        setpoint = self.create_setpoint( CHANNEL_NAME)
        setpoint.move(-200, 0)

        value = self.create_value(CHANNEL_NAME)
        value.move(200, 0)
```

integration/test/console/schematic.py
```diff
@@ -194,82 +303,158 @@ def get_schematic_value_props(self, node_id: str) -> Dict[str, Any]:
...
+    def create_node(
+        self, node_type: str, node_id: str, channel_name: str
+    ) -> SchematicNode:
+        """Factory method to create node objects"""
+        if node_type.lower() == "value":
+            return ValueNode(self.page, node_id, channel_name)
+        elif node_type.lower() == "setpoint":
+            return SetpointNode(self.page, node_id, channel_name)
+        else:
+            raise ValueError(f"Unknown node type: {node_type}")
```
pjdotson: See comment below, but it doesn't make sense that the general class will have to call all of the individual functions to create symbols here.

integration/test/console/schematic.py
```diff
@@ -194,82 +303,158 @@ def get_schematic_value_props(self, node_id: str) -> Dict[str, Any]:
...
+        }
+
+        if handle not in handle_positions:
+            raise ValueError(
+                f"Invalid handle: {handle}. Must be one of {list(handle_positions.keys())}"
+            )
+
+        return handle_positions[handle]
+
+    def click_on_pane(self) -> None:
```
pjdotson: please change this, a better pattern would just be to select the fit to view button

integration/test/console/schematic.py
```diff
@@ -194,82 +303,158 @@ def get_schematic_value_props(self, node_id: str) -> Dict[str, Any]:
...
+            self.set_label(channel_name)
+
+            # Navigate to Properties > Control tab
+            self.page.get_by_text("Properties").click()
+            self.page.get_by_text("Control").last.click()
+            self.set_channel("Command Channel", channel_name)
+
+        return properties
+
+    def set_control_authority(self, authority: int) -> None:
```
pjdotson: The control authority is not a property of the setpoint symbol but rather the whole schematic. This should not be defined here.

pjdotson (changes requested):
Something that you should be thinking about is how to better organize this directory for the code to run these tests and the actual tests. It seems like maybe we should have something like this:

integration
  - client
  - console
    - schematic
    - lineplot
    - etc.
  - core
  - driver
    - NI
    - OPC
    - etc.
  - framework
    - test_case
    - test_conductor
  - tests
    - press_sequence
    - ni_setup
    - etc.

this obviously will evolve but it would be good to start thinking about it now.

integration/test/framework/test_case.py
sy-nico: Arbitrary value chosen. 
I initially only included the first part of the function, but "None" kept being returned after setting the submitting the value, so I added the recent_range as a fallback.

integration/test/console/schematic.py
sy-nico: SB None

integration/test/console/schematic.py
sy-nico: * My first implementatio had the factory method within add_to_schematic, but I was unsure if it was more appropriate to keep it separate as we add support for the rest of the SchematicSymbols.

* Do you agree with "create_schematic_symbol" instead?

integration/test/console/schematic.py
nico-synnax: Clicking the fit-to-view button does not reset the bottom panel

Before:
image

After:
image


Is there another way that we could reset the Schematic symbols panel?
image

integration/test/console/schematic.py
sy-nico: Deleted set_control_authority()

integration/test/console/console.py
sy-nico: Was originally created for channels ans broadened for any drop down. Renaming to "item_found"

integration/test/console/schematic_set_output.py
sy-nico: Leftover from when I thought I could pull the text within a a Value Symbol. The intent was to set the value to something precise, set the value symbol precision to "2" and then verify that the text was "47.23" exactly.

However, since the text within a value node is drawn on the canvas, I am unable to query that data directly.

Regardless, I am updating that number to 47.23

integration/test/console/schematic_edit_value_props.py
sy-nico: Each test case automatically creates 3 channels _time, _uptime, and _state. Rather than creating a new channel, we can just use one that should already exist. We don't want to try to find an existing channel (such as tc_state), since those are subject to change.

integration/test/cases/console/schematic_set_output.py
sy-nico: Assertions are the standard way of validating results in test code (which this is meant to be)

integration/test/console/console.py
sy-nico: The entire test_case or a specific function? The intent of console.py is to provide a base class for testing specific user actions on the console (like a story of actions) as well as to test out the interface between the console and core

sy-nico: Paired this. Scoped to new ticket but this will buy us a lot in the future.
https://linear.app/synnax/issue/SY-2965/console-test-case-handling

integration/test/console/schematic.py
sy-nico: Going to instead just select the "Symbols" tab

integration/test/cases/latency/bench_report.py
pjdotson, line 1: If this file is in the `latency` folder it doesn't need to be called `bench_latency_report.py`, could just be called `bench_report.py`.

integration/test/cases/latency/bench_response.py
pjdotson, line 1: See above comment

integration/test/console/console.py
```diff
@@ -13,12 +13,15 @@
...
 
+import synnax as sy
 from playwright.sync_api import Browser, BrowserType, Page, sync_playwright
 
 
 class Console(TestCase):
     """
     Console TestCase implementation using Playwright
+
+    SY-2965: Break this out to Console & ConsoleCase
```
pjdotson: Could you remove this comment?

integration/test/console/schematic/symbol.py
pjdotson, line 1: This file should just be called `symbol.py` and the class should just be called `Symbol`

integration/test/console/schematic/symbol.py
```diff
@@ -0,0 +1,158 @@
...
+        self.symbol = self.page.get_by_test_id(self.symbol_id)
+        self.set_label(channel_name)
+
+    def _disable_edit_mode(self) -> None:
+        edit_off_icon = self.page.get_by_label("pluto-icon--edit-off")
+        if edit_off_icon.count() > 0:
+            edit_off_icon.click()
+
+    def _click_symbol(self) -> None:
+        self.symbol.click(force=True)
```
pjdotson, line 45: what does this `force=True` do?

integration/test/console/schematic/symbol.py
```diff
@@ -0,0 +1,158 @@
...
+            self.page.wait_for_timeout(500)
+
+            # Iterate through dropdown items
+            channel_found = False
+            item_selector = self.page.locator(".pluto-list__item").all()
+            for item in item_selector:
+                if item.is_visible() and channel_name in item.inner_text().strip():
+                    item.click()
+                    channel_found = True
+                    break
```
pjdotson, line 97: It seems like this snippet here is going to be widely use and should be put into a common function for the whole console

integration/test/console/schematic/schematic_symbol.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def set_channel(self, input_field: str, channel_name: str) -> None:
+        if channel_name is not None:
+            channel_button = (
+                self.page.locator(f"text={input_field}")
+                .locator("..")
+                .locator("button")
+                .first
+            )
+            # Click on the selector and fille channel_name
```
pjdotson: typo

integration/test/console/schematic/schematic_symbol.py
```diff
@@ -0,0 +1,158 @@
...
+                    channel_found = True
+                    break
+
+            if not channel_found:
+                raise RuntimeError(
+                    f"Could not find channel '{channel_name}' in dropdown"
+                )
+
+    def get_properties(self) -> Dict[str, Any]:
+        return {}
```
pjdotson: Why does this function exist?

integration/test/console/schematic/schematic_symbol.py
```diff
@@ -0,0 +1,158 @@
...
+        grid_tolerance = 25
+        if (
+            abs(final_x - target_x) > grid_tolerance
+            or abs(final_y - target_y) > grid_tolerance
+        ):
+            raise RuntimeError(
+                f"Symbol {self.symbol_id} moved to ({final_x}, {final_y}) instead of ({target_x}, {target_y})"
+            )
+
+    def set_value(self, value: Any = None) -> None:
```
pjdotson: This seems specific to a setpoint, right?

integration/test/console/schematic/symbol.py
sy-nico: Marking as resolved because I started on the common function in the refactored Console class. It still needs some work, so I am adding a reminder comment on the next PR to clean it up.

https://github.com/synnaxlabs/synnax/blob/bbe2c4aabdece57b161fbd9f3d1b73ce33543fbd/integration/test/console/console.py#L77

[Link to comment on SY-2965 PR](https://github.com/synnaxlabs/synnax/pull/1485#issuecomment-3325067141)

integration/test/console/schematic/schematic_symbol.py
sy-nico: Removed

integration/test/console/schematic/schematic_symbol.py
sy-nico: Yes and this was also moved on [SY-2965](https://github.com/synnaxlabs/synnax/blob/bbe2c4aabdece57b161fbd9f3d1b73ce33543fbd/integration/test/console/schematic/setpoint.py#L44).
Moved it to setpoint.py for this PR

integration/test/console/schematic/symbol.py
sy-nico: This option bypasses the standard actionability checks that playwright does before interacting with an element (**such as ensuring the element is visible**).

When two elements are created and kept in their default positions, they will be centered on the screen and overlapping. Even if it is clickable, for some reason, Playwright might not consider it so. 

Setting to `true` tells playwright to click on the element anyway.

The alternative is to move a symbol every time it's created to a non-default position on the canvas. With this approach, we would still need to be cautious that we don't move a new node over top of an existing node.

integration/test/console/console.py
```diff
@@ -135,3 +136,72 @@ def ESCAPE(self) -> None:
...
+
+    def _select_from_dropdown(self, input_field: str, input_text: str) -> None:
+
+        channel_button = (
+            self.page.locator(f"text={input_field}")
+            .locator("..")
+            .locator("button")
+            .first
+        )
+        channel_button.click()
```
pjdotson, line 191: Don't call this `channel_button`

integration/test/console/schematic/setpoint.py
```diff
@@ -0,0 +1,56 @@
...
+
+            # Navigate to Properties > Control tab
+            self.page.get_by_text("Properties").click()
+            self.page.get_by_text("Control").last.click()
+            self.set_channel("Command Channel", channel_name)
+            applied_properties["channel"] = channel_name
+
+        return applied_properties
+
+    def set_value(self, value: Any = None) -> None:
```
pjdotson, line 44: Could just make the type of value be a `float` here

pjdotson (changes requested):
2 minor comments and the changing the folder structure (unless you do it on the other PR) and then you can merge it.

For merging
1) make sure to squash and merge
2) make sure the detailed description is something readable and not just all of the individual commit messages.

integration/test/console/console.py
sy-nico: Outdated as of SY-2965

Image

integration/test/console/schematic/setpoint.py
sy-nico: Resolved/Outdated as of SY-2965
image

PR #1485, opened 2025-09-23
SY-2965: Console Test Case - Component  Modularization

nico-synnax:
# Issue Pull Request

## Linear Issue

[SY-2965](https://linear.app/synnax/issue/SY-2965/console-test-case-modularization)

## Description
Split functionality fo playwright console control and console_test_case interface

## Basic Readiness


integration/console/channels.py
```diff
@@ -0,0 +1,72 @@
...
+#  included in the file licenses/APL.txt.
+
+from typing import TYPE_CHECKING
+
+import synnax as sy
+from playwright.sync_api import Page
+
+from ..console_page import ConsolePage
+
+if TYPE_CHECKING:
```
pjdotson, line 24: what is the point of this TYPE_CHECKING code? never seen it before.

integration/test/console/console_page.py
```diff
@@ -0,0 +1,169 @@
...
+        self.page.screenshot(
+            path=path,
+            clip=cast(FloatRect, clip_area),
+            animations="disabled",
+            omit_background=False,
+            type="png",
+            scale="device",
+        )
+
+    def move(self, direction: str) -> None:
```
pjdotson: for something like this, it makes more sense to have direction only be one of those options instead of having this valid_directions code below and having to raise a ValueError

integration/test/console/plot/plot.py
```diff
@@ -7,205 +7,151 @@
...
+            "Ranges": [],
+            "X1": None,
+        }
 
-        # Close any open dialogs first
-        self.page.keyboard.press("Escape")
-        time.sleep(0.2)
+    def add_Y(self, axis: str, channel_ids: Union[str, List[str]]) -> None:
+        if axis not in ("Y1", "Y2"):
+            raise ValueError(f"Invalid axis: {axis}. Must be 'Y1' or 'Y2'")
```
pjdotson: instead of raising this value error, you should just more strictly type the axis input

integration/test/console/plot/plot.py
```diff
@@ -7,205 +7,151 @@
...
+            self.console._select_from_dropdown_item(
+                channel, "input[placeholder*='Search']"
+            )
+            self.data[axis].append(channel)
+
+        self.console.ESCAPE
 
+    def add_ranges(self, ranges: List[str]) -> None:
+        """Add time ranges to the plot."""
+        valid_ranges = {"30s", "1m", "5m", "15m", "30m"}
```
pjdotson: same thing here with this valid_ranges pattern. don't make ranges a List[str]

integration/test/console/plot/plot.py
```diff
@@ -7,205 +7,151 @@
...
+        """Set X1 axis configuration."""
         self.set_axis("X1", config)
 
     def set_Y1_axis(self, config: Dict[str, Any]) -> None:
+        """Set Y1 axis configuration."""
         self.set_axis("Y1", config)
 
     def set_Y2_axis(self, config: Dict[str, Any]) -> None:
+        """Set Y2 axis configuration."""
         self.set_axis("Y2", config)
```
pjdotson: It does not make sense to have any of these set_X1, set_Y1, set_Y2 functions. just let people call set_axis and type the axis argument to only be these three values.

integration/console/plot.py
```diff
@@ -7,205 +7,151 @@
...
-                    f'Warning: Failed to set axis property "{key}" to "{value}": {e}'
-                )
-        self.ENTER
+                input_field = self.page.locator(selector)
+                if input_field.count() > 0:
+                    input_field.clear(timeout=5000)
+                    input_field.fill(str(value), timeout=5000)
+                    return
+            except Exception:
+                continue
```
pjdotson, line 124: why are we silently failing here?

integration/test/console/plot.py
```diff
@@ -7,205 +7,151 @@
...
+                if input_field.count() > 0:
+                    input_field.clear(timeout=5000)
+                    input_field.fill(str(value), timeout=5000)
+                    return
+            except Exception:
+                continue
+
+        raise RuntimeError(f"Could not find input field for {key}")
+
+    def _set_label_direction(self, value: Any) -> None:
```
pjdotson: why not have value be horizontal or vertical or x or y instead of Any?

integration/test/console/plot.py
```diff
@@ -7,205 +7,151 @@
...
+        """Set label direction button."""
+        direction = "arrow-up" if str(value).lower() == "up" else "arrow-right"
+        selector = f"label:has-text('Label Direction') + div button:has([aria-label='pluto-icon--{direction}'])"
+        self.page.locator(selector).click(timeout=5000)
+
+    def _set_label_size(self, value: Any) -> None:
```
pjdotson: again, lets remove this `Any` here and just only allow you to set sizes as xs, s, ...

integration/console/log.py
pjdotson, line 1: just remove this folder until we actually need it.

integration/console/schematic/schematic.py
```diff
@@ -7,62 +7,68 @@
...
-        setpoint_id = self._add_symbol_to_schematic("Setpoint")
+        """Create a setpoint symbol on the schematic."""
+        setpoint_id = self._add_symbol("Setpoint")
         setpoint = Setpoint(self.page, setpoint_id, channel_name)
         setpoint.edit_properties(channel_name=channel_name)
-
-        self._log_message(f"Added setpoint with channel {channel_name}")
         return setpoint
 
     def create_value(
```
pjdotson, line 74: This is fine right now but the better solution to think about how we get there is calling like create_symbol and you pass in a value or a setpoint class. Might be tricky to get that to work, but otherwise you need a separate function for each symbol type which is obviously not the right way to do this long-term.

integration/test/console/schematic/setpoint.py
```diff
@@ -40,3 +40,17 @@ def edit_properties(
             applied_properties["channel"] = channel_name
 
         return applied_properties
+
+    def set_value(self, value: Any = None) -> None:
```
pjdotson: shouldn't value be a number here?

integration/test/console/table.py
pjdotson, line 1: again, just remove this until we actually test tables.

integration/test/console/console_case.py
```diff
@@ -0,0 +1,143 @@
...
+
+import synnax as sy
+from playwright.sync_api import Browser, BrowserType, Page, sync_playwright
+
+
+class ConsoleCase(TestCase):
+    """
+    Console TestCase implementation using Playwright
+
+    SY-2965: Break this out to Console & ConsoleCase
```
pjdotson: remove this comment

integration/test/console/console_page/__init__.py
pjdotson, line 1: I think a better name for this folder / class is layout instead of console_page to match the directory structure in the console more.

PS when you are naming things (one of the [two hard problems](https://martinfowler.com/bliki/TwoHardThings.html) in computer science), you shouldn't generally have redundant parts of names like console/console_page. Page works just fine because you import it from console, unless you are another component already in the console. The exception to this is things like console/console.py where you have a class console. This is a pattern we follow in every part of our code base, so there is plenty of examples to look at.

integration/test/console/channels.py
```diff
@@ -0,0 +1,72 @@
...
+
+class Channels(ConsolePage):
+    """Console channels management interface"""
+
+    def __init__(self, page: Page, console: "Console"):
+        super().__init__(page, console)
+
+    def create(
+        self,
+        name: str,
```
pjdotson: for long list of arguments you should make sure that arguments have to be passed by name by making the second argument after `self` be `"*"`?

integration/test/console/console_case.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def teardown(self) -> None:
+        self.browser.close()
+
+    def determine_browser(self) -> BrowserType:
+        """
+        Provide random coverage for all browsers.
+        """
+
+        browsers = ["chromium", "firefox", "webkit"]
```
pjdotson: you should just remove this assignment of the list then

integration/test/console/console_case.py
```diff
@@ -0,0 +1,143 @@
...
+        self,
+        channel_name: str,
+        virtual: bool = False,
+        is_index: bool = False,
+        data_type: sy.CrudeDataType = sy.DataType.TIMESTAMP,
+        index: str = "",
+    ) -> bool:
+        """
+        Will be deprecated.
+        Keeping around to remind me to check if exists in channels.py
```
pjdotson: Don't understand what you mean by this comment

integration/test/cases/console/tabs_open_close.py
```diff
@@ -60,26 +64,26 @@ def run(self) -> None:
...
         if self.page.get_by_text("New Component").count() > 0:
             self._log_message("All pages closed - 'New Component' screen visible")
         else:
-            self._log_message(
+            self.fail(
                 "FAILED: Pages still be open - 'New Component' screen not visible"
             )
-            self.fail()
 
     def open_page(self, page_name: str, inputs_items: list[str] = []) -> None:
```
pjdotson: A random thought - it might be good for integration tests where there are multiple ways of doing things like creating a page to have different private methods, then the public open page either randomly uses the new page button or the command palette.

integration/console/channels.py
sy-nico: This is used to avoid circular imports while maintaining type correct type annotations. There are a few other ways to do this, but we would lose type safety or have to import at runtime (slower)

integration/test/console/console_page/__init__.py
sy-nico: will rename to "page.py", but keeping the class name as "ConsolePage" to prevent conflicts with playwright.Page

integration/test/console/channels.py
sy-nico: I deliberately placed it after "name", but I will add it after "self" to better match channel/client.

integration/test/console/console_case.py
sy-nico: Just a reminder note 

Context: This function was moved from from ConsoleCase to ChannelClient and the ConsoleCase.create_a_channel() function no longer exists.

integration/test/cases/console/tabs_open_close.py
sy-nico: Love this idea

integration/console/schematic/schematic.py
sy-nico: Yeah I see what you're saying. I'll do that refactor when I add support for "button" symbols. After the work done in this PR, we are moving closer to that direction.

integration/console/log.py
sy-nico: Keeping log so that I can have more than two page types for the snap_pages_to_grid test. It's also fairly enough to configure, so that functionality is now being tested for.

integration/console/schematic/__init__.py
```diff
@@ -12,4 +12,4 @@
 from .symbol import Symbol
 from .value import Value
 
-__all__ = ["Schematic", "Symbol", "Value", "Setpoint"]
+__all__ = ["Plot", "Schematic", "Setpoint", "Symbol", "Value"]
```
pjdotson: why is Plot getting exported in the schematic folder?

integration/console/channels.py
pjdotson: Why do we have circular imports at all? The code sounds like it needs to be refactored to avoid it. Happy to talk about ways we can do that on a call.

PR #1496, opened 2025-09-28
SY-2969: Add more Console Test Cases

nico-synnax:
- PR opened on mobile app. Description pending


integration/console/schematic/button.py
```diff
@@ -0,0 +1,129 @@
...
+        mode_options = ["Fire", "Momentary", "Pulse"]
+        for option in mode_options:
+            try:
+                button = self.page.get_by_text(option).first
+                if button.count() > 0:
+                    class_name = button.get_attribute("class") or ""
+                    if "pluto-btn--filled" in class_name:
+                        props["mode"] = str(option).lower()
+                        break
+            except:
```
pjdotson: don't hide this exception here

integration/console/channels.py
```diff
@@ -197,16 +199,18 @@ def delete(self, names: ChannelNames) -> None:
...
-        if not self.existing_channel(name):
-            raise ValueError(f"Channel {name} does not exist")
+        exists, all_channels = self.existing_channel(name)
+        if not exists:
+            raise ValueError(f"Channel {name} does not exist in {all_channels}")
 
-        # Find the channel in the list and right-click it
-        self.show_channels()
+        # Find the channel in the list and delete it
+        # self.show_channels()
```
pjdotson: remove commented out code

integration/framework/test_case.py
```diff
@@ -449,13 +450,51 @@ def add_channel(
 
         self.tlm[tlm_name] = initial_value
 
-    def subscribe(self, channels: Union[str, List[str]]) -> None:
-        """Subscribe to channels. Can take either a single channel name or a list of channels."""
+    def subscribe(
+        self, channels: Union[str, List[str]], timeout: Optional[int] = 10
```
pjdotson: make this timeout a `sy.TimeSpan` instead of an int

integration/tests/console/simple_press_valves.py
pjdotson, line 1: where are the actual assertions on stuff like this? It's nice to have a control sequence example, but where do you test say that a valve looks filled when a last command is opened?

Remember that when you are devving this you are looking at the console run each time. During actual integration we won't be looking at the console or these tests unless the workflow fails. So these test cases should have a lot more assertions, calling `self.fail`, etc. than this one currently does.

pjdotson (changes requested):
3 super minor things and one longer one. If the longer one is going to take more than 20 minutes to do, than push it off for a separate ticket so we can focus on doing QA.

PR #1514, opened 2025-09-30
SY-3022: Get Machine Info - Win11 Compatibility

nico-synnax:
# Issue Pull Request

## Linear Issue
[SY-3022](https://linear.app/synnax/issue/SY-3022/update-get-machine-info-for-win11)

## Basic Readiness

### DIFF
--- a/integration/tests/console/plot_abcd.py
+++ b/integration/tests/console/plot_abcd.py
@@ -7,8 +7,6 @@
 #  License, use of this software will be governed by the Apache License, Version 2.0,
 #  included in the file licenses/APL.txt.
 
-import synnax as sy
-
 from console.case import ConsoleCase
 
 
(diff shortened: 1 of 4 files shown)


integration/tests/console/plot_abcd.py
```diff
@@ -27,6 +29,8 @@ def setup(self) -> None:
         )
 
     def run(self) -> None:
+
+        sy.sleep(3)
```
pjdotson: this is fine to merge for now but I would really appreciate going through at some point in time and removing all of the sleeps from the test cases.

pjdotson (changes requested):
Good to merge after comment is (or isn't addressed).

integration/tests/console/plot_abcd.py
sy-nico: Updated to instead remove 3s of sleep from the parallel threads (rather than adding 3s to this one).
 
Test run:
https://github.com/synnaxlabs/synnax/actions/runs/18118645212/job/51559237350

PR #1533, opened 2025-10-04
SY-3018: Test NI Analog Task Forms & Benchmark NI DO Driver Loopback Latency

nico-synnax:
# Issue Pull Request

## Linear Issue

[SY-3018](https://linear.app/synnax/issue/SY-3018/test-obscure-ni-channel-types)

## Description

- New ni do driver latency loopback tesst
  - Write metadata to range 
- Test to verify NI Analog Task forms (did not create tasks)
  - Split amongs multiple test cases
- increase tc timeout
- Fix element get/set functions in console.py
 - no longer using sleep
 - faster playwright tests)

## Basic Readiness


integration/framework/test_case.py
```diff
@@ -755,8 +762,9 @@ def fail(self, message: Optional[str] = None) -> None:
     def execute(self) -> None:
         """Execute complete test lifecycle: setup -> run -> teardown."""
         try:
+            gc.disable()
```
pjdotson: Are we sure we want to be disabling garbage collection here for all tests? It seems like the default should be to keep it on.

integration/framework/utils.py
```diff
@@ -198,6 +199,14 @@ def get_memory_info() -> str:
     raise RuntimeError(f"Unable to get memory information for {platform.system()}")
 
 
+def get_cpu_cores() -> str:
+    """Get the number of CPU cores."""
+    try:
+        return f"{multiprocessing.cpu_count()} cores"
+    except Exception:
```
pjdotson: why are you silently ignoring this exception?

integration/console/console.py
```diff
@@ -358,9 +384,46 @@ def get_input_field(self, input_label: str) -> str:
...
+        for option in button_options:
+            try:
+                button = self.page.get_by_text(option).first
+                if button.count() > 0:
+                    button.wait_for(state="attached", timeout=300)
+                    class_name = button.get_attribute("class") or ""
+                    if "pluto-btn--filled" in class_name:
+                        return option
+            except:
+                continue
```
pjdotson: seems like you shouldn't be ignoring this exception

integration/console/task/ni_ai.py
```diff
@@ -0,0 +1,66 @@
...
+
+from playwright.sync_api import Page
+
+from .task import Task
+
+if TYPE_CHECKING:
+    from console.console import Console
+
+
+class NiAi(Task):
```
pjdotson: you should name these classes `NIAnalogInput`, it's a lot easier to read than `NiAi` (which sounds like some chemical compound, or even worse, the new National Instruments AI agent)

integration/console/task/ni_ao.py
```diff
@@ -0,0 +1,112 @@
...
+
+        Returns:
+            The created channel instance
+
+        Raises:
+            ValueError: If channel type is not supported for AO tasks
+        """
+        # Validate channel type
+        if type not in AO_CHANNEL_TYPES:
+            raise ValueError(
```
pjdotson, line 67: you can remove this check by better typing the `type` parameter

integration/console/task/ni.py
```diff
@@ -0,0 +1,246 @@
...
+    "Thermocouple": Thermocouple,
+    "Torque Bridge Table": TorqueBridgeTable,
+    "Torque Bridge Two-Point Linear": TorqueBridgeTwoPointLinear,
+    "Velocity IEPE": VelocityIEPE,
+    "Voltage": Voltage,
+}
+
+
+class Task(ConsolePage):
+    """NI Task automation interface for managing analog channels."""
```
pjdotson, line 25: If this stuff is NI specific, you should have a `integration/console/task/ni` folder

integration/console/task/channels/function_generator.py
```diff
@@ -0,0 +1,120 @@
...
+            waveform_options = ["Sine", "Triangle", "Square", "Sawtooth"]
+            for option in waveform_options:
+                try:
+                    button = console.page.get_by_text(option).first
+                    if button.count() > 0:
+                        class_name = button.get_attribute("class") or ""
+                        if "pluto-btn--filled" in class_name:
+                            values["Waveform"] = option
+                            break
+                except:
```
pjdotson: why are we ignoring errors here?

integration/console/task/channels/function_generator.py
```diff
@@ -0,0 +1,120 @@
...
+            else:
+                # Handle normally
+                try:
+                    actual_value = self.console.get_input_field(key)
+                except:
+                    actual_value = self.console.get_dropdown_value(key)
+
+            assert (
+                actual_value == expected_value
+            ), f"Channel {self.name} Form value '{key}' - Expected: {expected_value} - Actual: {actual_value}"
```
pjdotson: should we have this assert here? seems like if this is worth checking you should directly raise an exception

integration/console/task/channels/analog.py
```diff
@@ -0,0 +1,133 @@
...
+        min_val: Optional[float] = None,
+        max_val: Optional[float] = None,
+        custom_scale: Optional[
+            Literal[
+                "None",
+                "Linear",
+                "Map",
+                "Table",
+            ]
+        ] = None,
```
pjdotson, line 51: I feel like all of these options don't exist for all channel types

integration/console/task/channels/analog.py
```diff
@@ -0,0 +1,133 @@
...
+            console.select_from_dropdown(terminal_config)
+            values["Terminal Configuration"] = terminal_config
+        else:
+            try:
+                values["Terminal Configuration"] = console.get_dropdown_value(
+                    "Terminal Configuration"
+                )
+            except:
+                # Many AI types do not have a terminal config option.
+                pass
```
pjdotson: shouldn't silently ignore

integration/console/task/channels/analog.py
```diff
@@ -0,0 +1,133 @@
...
+            console.fill_input_field("Minimum Value", str(min_val))
+            values["Minimum Value"] = str(min_val)
+        elif type != "Microphone":
+            values["Minimum Value"] = console.get_input_field("Minimum Value")
+
+        if max_val is not None:
+            console.fill_input_field("Maximum Value", str(max_val))
+            values["Maximum Value"] = str(max_val)
+        elif type != "Microphone":
+            values["Maximum Value"] = console.get_input_field("Maximum Value")
```
pjdotson, line 108: why is this base class responsible for logic that is dependent on the type like microphone?

integration/tests/console/task/ni_ao_forms.py
```diff
@@ -0,0 +1,148 @@
...
+
+import platform
+import random
+
+import synnax as sy
+
+from console.case import ConsoleCase
+
+
+class Ni_Ao_Forms(ConsoleCase):
```
pjdotson: use PascalCase and not snake case (NIAnalogOutputForms)

integration/tests/console/task/ni_ao_forms.py
```diff
@@ -0,0 +1,148 @@
...
+        # Talks to NI MAX sim devices
+        rack_name = f"TestRack_{random.randint(100, 999)}"
+        device_name = "E203"
+        sy.sleep(5)
+        self._log_message("Creating NI Analog Write Task")
+        console.ni_ao.new()
+
+        # Check simple functionality
+        console.ni_ao.set_parameters(
+            task_name="DO_Test_task",
```
pjdotson: this is an ao test, right?

integration/tests/latency/driver_ni_do.py
```diff
@@ -0,0 +1,487 @@
...
+    get_cpu_cores,
+    get_machine_info,
+    get_memory_info,
+    get_synnax_version,
+)
+
+matplotlib.use("Agg")  # Use non-interactive backend
+
+
+class DriverNiDo(TestCase):
```
pjdotson: `DriverNIDigitalOutput`

integration/tests/latency/driver_ni_digital_write.py
```diff
@@ -0,0 +1,487 @@
...
+
+class DriverNiDo(TestCase):
+    """
+    Send a command to an NI DO channel and measure the latency between the
+    core and loop-back (python) timestamp.
+    """
+
+    def setup(self) -> None:
+        if platform.system().lower() != "windows":
+            self.auto_pass(msg="Windows DAQmx drivers required")
```
pjdotson, line 39: could we not install the linux drivers as well? not sure if that juice is worth the squeeze though

integration/tests/console/task/ni_ai_forms.py
```diff
@@ -0,0 +1,806 @@
...
+
+import platform
+import random
+
+import synnax as sy
+
+from console.case import ConsoleCase
+
+
+class Ni_Ai_Forms(ConsoleCase):
```
pjdotson: `NIAnalogReadForms`

integration/tests/console/task/ni_analog_read_forms.py
```diff
@@ -0,0 +1,806 @@
...
+            self.verify_pressure_bridge_table_inputs(device_name)
+        if mode == "d":
+            self.verify_pressure_bridge_two_point_linear_inputs(device_name)
+            self.verify_resistance_inputs(device_name)
+            self.verify_rtd_inputs(device_name)
+        if mode == "e":
+            self.verify_strain_gauge_inputs(device_name)
+            self.verify_temperature_built_in_sensor_inputs(device_name)
+            self.verify_thermocouple_inputs(device_name)
+        if mode == "f":
```
pjdotson, line 76: this mode stuff is really confusing. If they are just random things, why not just select a random integer instead of cutting a character off of the end of the name? Then we have these add modes: A,B,C,etc. with no semantic meaning.

integration/tests/console/task/ni_analog_read_forms.py
pjdotson, line 1: Seems like the really useful test would be to pull the device config from synnax via the client library and then check if the given field in the config matches what you tried to verify.

sy-nico: I would like to pair this. I am _mostly_ set up to do this. The EC2 instances have a bunch of simulated modules. I just need to spend the time to go through and define the proper configuration settings for each channel type.

I stopped short of this because it adds a layer of complexity. 

We will eventually need to verify configuration parameters are correct (as well as verifying the expected and generated sine wave data). The question is if we should do it in this PR or in a new one.

This wouldn't replace the form validation, as each channel form should maintain its state prior to configuration.

sy-nico: Spinning off the next set of tickets which will require time mapping everything out to configured devices. This PR will be to ensure we build the test code right.

A comprehensive suite of tests will require the following:
* Form functionality verification (this PR)
   * Checking input option can be set
   * Checking drop down elements can be selected 
* Task Status assertions 
   * A small "No Device" test was included in this PR.
   * Force errors when selecting a dropdown item that is not compatible with the channel type
   * Example is selecting "RSE" for accels. Our form allows this, but the task will throw a UnsupportedValueError. 
* Information synchronization to the Console if a client updates a task config.
* Verifying driver and sim data (sine wave analysis) at different settings

All of these will build off of the new code being added by this PR.

[SY-3096: Nominal NI Task Tests (Console)](https://linear.app/synnax/issue/SY-3096/nominal-ni-task-tests)
[SY-3097: NI Driver Tests (Client)](https://linear.app/synnax/issue/SY-3097/ni-driver-tests-client)
[SY-3098: Invalid NI Task Config Console Tests](https://linear.app/synnax/issue/SY-3098/invalid-ni-task-configuration-console-tests)
[SY-3099: Export/Load task configs test (Console)](https://linear.app/synnax/issue/SY-3098/invalid-ni-task-configuration-console-tests)

integration/console/task/ni_ai.py
sy-nico: OxygenPotassium 

Renaming to NIAnalogRead/Write for consistency with Synnax

integration/console/task/ni.py
sy-nico: I plan to move it once I start adding more task types. Until then, I will keep it at the current level to avoid unnecessary nesting.

integration/console/task/channels/analog.py
sy-nico: Because it's the only analog channel type that does not have a min/max input. The form for "microphone" does not confirm to the inputs that are used for the "Sound Pressure" types, so we will need to refactor there first before updating the test case
image

integration/tests/latency/driver_ni_digital_write.py
sy-nico: Yes we _could_ but I agree the juice is not worth the squeeze

- The majority of NI hardware users are running DAQmx on windows
- There's some overhead with getting it set up
- For now, we can mature the tests on windows. Once the tests and sim device configs have settled, we can import them into a linux machine.

integration/console/task/channels/analog.py
sy-nico: The way Synnax handles different scaling options is odd. Most all of the analog input types (with exception Temps I believe) have an option for custom scales. There is no distinction between a "Table" and "two-point linear" as different channel types. 

For example, "Force Bridge" _is_ the channel type with the ability to configure a custom scale (tabe, linear map ranges, polynomial)

image
image

https://linear.app/synnax/issue/SY-3105/fix-ni-channel-types-and-scales

integration/console/task/channels/function_generator.py
sy-nico: This file was deleted as "function generator" is an a mode of the analog output channels, not a channel type.

integration/tests/console/task/ni_analog_read_forms.py
sy-nico: This was done to split the test case into smaller chunks and run them concurrently. Running all of the functions sequentially will take about 5-6 minutes on the AWS servers. The split is arbitrary, but we don't want to test randomly since we want to ensure all channel types are covered. 

We need to balance minimizing the total time to run the tests and the number of playwright instances that the AWS servers can launch and handle before tests start to fail

Since I have to rename them all something separate, stripping off the letter to use as the "mode" means I can break things apart efficiently while preserving the names that are used for the test_case range and channel names
image

integration/console/task/ni_analog_read.py
```diff
@@ -0,0 +1,66 @@
...
+    from console.console import Console
+
+
+class NIAnalogRead(Task):
+    """NI Analog Read/Input Task automation interface."""
+
+    def __init__(self, page: Page, console: "Console") -> None:
+        super().__init__(page, console)
+        self.page_type = "NI Analog Read Task"
+        self.pluto_label = ".ni_ai_somethingsomething"
```
pjdotson: ??

integration/console/task/task.py
```diff
@@ -0,0 +1,244 @@
...
+    "Resistance": Resistance,
+    "RTD": RTD,
+    "Strain Gauge": StrainGauge,
+    "Temperature Built-In Sensor": TemperatureBuiltInSensor,
+    "Thermocouple": Thermocouple,
+    "Torque Bridge Table": TorqueBridgeTable,
+    "Torque Bridge Two-Point Linear": TorqueBridgeTwoPointLinear,
+    "Velocity IEPE": VelocityIEPE,
+    "Voltage": Voltage,
+}
```
pjdotson: This type of pattern is bad. This file / class is for a general NI analog task, yet it
needs to be aware of the specific types for analog read channels. Sure the code is fine
and works, but you have these circular dependencies that makes it harder to understand
the code. The NIAnalogRead class should be the one responsible for validating that a
channel type is a valid analog read type, not this task.

integration/console/task/ni.py
pjdotson: I know this is sort of the opposite advice I've given previously, but when its obvious
what direction we need to go in, then it's fine to cut to the chase and do the nesting
now. So we know we are probably eventually going to have a Task type for all tasks, then
like an NITask that is a subclass of that task, then AnalogRead, AnalogWrite,
DigitalRead, etc. So for right now lets have `integration/console/task/ni`, which has a
main class `Task` (this file), and then the `analog_read.py` and `analog_write.py` files
that have `AnalogRead` and `AnalogWrite` classes that inherit from `Task`. In the
future, a lot of the functionality in this file will get moved to another class that
contains code for LabJack, Modbus, etc.

integration/console/task/ni_analog_write.py
```diff
@@ -0,0 +1,100 @@
...
+}
+
+
+class NIAnalogWrite(Task):
+    """NI Analog Write/Output Task automation interface."""
+
+    def __init__(self, page: Page, console: "Console") -> None:
+        super().__init__(page, console)
+        self.page_type = "NI Analog Write Task"
+        self.pluto_label = ".ni_do_somethingsomething"
```
pjdotson: ??

integration/console/task/analog_write.py
```diff
@@ -0,0 +1,100 @@
...
+        result = super().new()
+        self.console.task = self
+        return result
+
+    def add_channel(
+        self,
+        name: str,
+        type: str,
+        device: str,
+        dev_name: Optional[str] = None,
```
pjdotson, line 46: why do we need device and dev_name separate here? all channels in an analog write task
need to be on the same device

integration/framework/utils.py
```diff
@@ -198,6 +199,11 @@ def get_memory_info() -> str:
     raise RuntimeError(f"Unable to get memory information for {platform.system()}")
 
 
+def get_cpu_cores() -> str:
```
pjdotson: This is a minor point but it makes more sense to for this to return an `int` and let the
caller format it how they want?

integration/tests/latency/driver_ni_digital_write.py
```diff
@@ -0,0 +1,487 @@
...
+        if not devices:
+            raise RuntimeError("No devices found")
+
+        device = None
+        for device in devices:
+            # Sim device must be set up in NI MAX
+            if device.location == "SYMod1":
+                dev = device
+                self.log(f"Found NI DO device: {dev.location}")
+                break
```
pjdotson: why not use the location key in the retrieve call versus this logic?

integration/tests/latency/driver_ni_digital_write.py
```diff
@@ -0,0 +1,487 @@
...
+        with tsk.run():
+            with client.open_streamer([do_1_state.key, do_state_time.key]) as stream:
+                with client.open_writer(sy.TimeStamp.now(), do_1_cmd.key) as writer:
+
+                    start_time = sy.TimeStamp.now()
+                    now_time = sy.TimeStamp.now()
+                    cmd_state: bool = False
+
+                    self.log("Begin latency test")
+                    while (now_time - start_time) < sy.TimeSpan.SECOND * 3:
```
pjdotson: could use `start_time.since` instead of tracking `now_time` as well

integration/tests/latency/driver_ni_digital_write.py
```diff
@@ -0,0 +1,487 @@
...
+        assert stats_loop["std"] <= 4, "Loop std latency is greater than 4 ms"
+
+        assert stats_driver["p90"] <= 6, "Driver p90 latency is greater than 6 ms"
+        assert stats_loop["p90"] <= 6, "Loop p90 latency is greater than 6 ms"
+
+        assert stats_driver["p95"] <= 8, "Driver p95 latency is greater than 8 ms"
+        assert stats_loop["p95"] <= 8, "Loop p95 latency is greater than 8 ms"
+
+        assert stats_driver["p99"] <= 10, "Driver p99 latency is greater than 10 ms"
+        assert stats_loop["p99"] <= 10, "Loop p99 latency is greater than 10 ms"
```
pjdotson, line 301: less than not greater than in these logs

.github/workflows/test.integration.yaml
pjdotson, line 1: What do you think about also running the integration tests on PRs that modify the
`integration` folder as well? that way we make sure changes to this folder still result
in all tests passing.

pjdotson (changes requested):
Looks pretty good, just a few comments. Make sure that you backmerge `rc` as well.

integration/console/task/analog_write.py
sy-nico: because the dev_name is used to name the device when configuring it (using the modal). The parameters are different and have different validation rules as well. I would agree that this can be simplified, but it falls under the umbrella of refactoring how we present/name devices to the user (for example, we should be using location as the name)

integration/tests/latency/driver_ni_digital_write.py
sy-nico: Python assertion message strings describe the failure case

.github/workflows/test.integration.yaml
sy-nico: I looked into making that happen, but you cannot have duplicate keys in the yamls. We would have to enable the test on both branches for the same set of changes. I could add some logic into the workflow to check conditions, but I need to be cautious about impacting the current ref_run logic

PR #1538, opened 2025-10-05
SY-3082: Fix C++ Compiler Warnings

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-3082](https://linear.app/synnax/issue/SY-3082/fix-all-c-compiler-warnings)

## Description

Does two things:

1. Disables all compiler warnings that come from external dependencies. There's nothing we can do about them, and they clog local and CI build logs.
2. Fixes all compiler warnings in the current build.

## Basic Readiness

### DIFF
--- a/driver/sequence/plugins/channel_write_test.cpp
+++ b/driver/sequence/plugins/channel_write_test.cpp
@@ -311,7 +311,7 @@ TEST_F(SetAuthorityTest, MultipleChannelsDifferentAuth) {
 
     // Create a map of channel keys to their authorities for easier verification
     std::map<synnax::ChannelKey, telem::Authority> auth_map;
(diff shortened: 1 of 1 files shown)


pjdotson (changes requested):
Is there some way we could add a CI check so that the build fails when there are compiler warnings?

driver/opc/dev/client.cpp
```diff
@@ -14,7 +14,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 
-static UA_StatusCode node_iter(
+[[maybe_unused]] static UA_StatusCode node_iter(
```
pjdotson: what is the point of this? / if it is unused, can we get rid of this?

PR #1541, opened 2025-10-05
SY-3085: Add Memory Checks (ASAN) To C++ Code

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-####]()

## Basic Readiness

### DIFF
--- a/.github/workflows/test.x.cpp.yaml
+++ b/.github/workflows/test.x.cpp.yaml
@@ -71,4 +71,10 @@ jobs:
         run: scripts/check_clang_format.sh x/cpp
 
       - name: Test
-        run: bazel test --test_output=all //x/cpp/... --nocache_test_results
+        run: |
+          if [ "${{ matrix.os }}" == "ubuntu-latest" ]; then
+            bazel test --config=sanitize --test_output=all //x/cpp/... --nocache_test_results
+          else
+            bazel test --test_output=all //x/cpp/... --nocache_test_results
+          fi
+        shell: bash
--- a/x/cpp/binary/binary.h
+++ b/x/cpp/binary/binary.h
@@ -36,40 +36,58 @@ public:
 
     /// @brief Writes a byte to the buffer
     /// @param value The byte to write
-    void uint8(const uint8_t value) { buf[offset++] = value; }
-
-    /// @brief Writes a 32-bit unsigned integer to the buffer
-    /// @param value The uint32 to write
-    void uint32(const uint32_t value) {
-        this->uint8(value);
-        this->uint8(value >> 8);
-        this->uint8(value >> 16);
-        this->uint8(value >> 24);
+    /// @return The number of bytes written (0 or 1)
+    size_t uint8(const uint8_t value) {
+        if (offset >= buf.size()) return 0;
+        buf[offset++] = value;
+        return 1;
     }
 
-    /// @brief Writes a 64-bit unsigned integer to the buffer
+    /// @brief Writes a 32-bit unsigned integer to the buffer (little-endian)
+    /// @param value The uint32 to write
+    /// @return The number of bytes written (0-4)
+    size_t uint32(const uint32_t value) {
+        size_t written = 0;
+        written += this->uint8(value);
+        written += this->uint8(value >> 8);
+        written += this->uint8(value >> 16);
+        written += this->uint8(value >> 24);
+        return written;
+    }
+
+    /// @brief Writes a 64-bit unsigned integer to the buffer (little-endian)
     /// @param value The uint64 to write
-    void uint64(const uint64_t value) {
-        this->uint8(value);
-        this->uint8(value >> 8);
-        this->uint8(value >> 16);
-        this->uint8(value >> 24);
-        this->uint8(value >> 32);
-        this->uint8(value >> 40);
-        this->uint8(value >> 48);
-        this->uint8(value >> 56);
+    /// @return The number of bytes written (0-8)
+    size_t uint64(const uint64_t value) {
+        size_t written = 0;
+        written += this->uint8(value);
+        written += this->uint8(value >> 8);
+        written += this->uint8(value >> 16);
+        written += this->uint8(value >> 24);
+        written += this->uint8(value >> 32);
+        written += this->uint8(value >> 40);
+        written += this->uint8(value >> 48);
+        written += this->uint8(value >> 56);
+        return written;
     }
 
     /// @brief Writes a 64-bit signed integer to the buffer
     /// @param value The int64 to write
-    void int64(const int64_t value) { this->uint64(static_cast<uint64_t>(value)); }
+    /// @return The number of bytes written (0-8)
+    size_t int64(const int64_t value) {
+        return this->uint64(static_cast<uint64_t>(value));
+    }
 
     /// @brief Writes raw bytes to the buffer
     /// @param data The bytes to write
     /// @param size The number of bytes to write
-    void write(const void *data, const size_t size) {
-        std::memcpy(buf.data() + offset, data, size);
-        offset += size;
+    /// @return The number of bytes actually written
+    size_t write(const void *data, const size_t size) {
+        const size_t available = buf.size() - offset;
+        const size_t to_write = size < available ? size : available;
+        std::memcpy(buf.data() + offset, data, to_write);
+        offset += to_write;
+        return to_write;
     }
 
     /// @brief Returns the buffer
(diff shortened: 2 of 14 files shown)


.github/workflows/test.client.yaml
```diff
@@ -33,6 +33,7 @@ on:
       - pnpm-lock.yaml
       - pnpm-workspace.yaml
       - scripts/check_clang_format.sh
+      - scripts/sanitizers/**
```
pjdotson, line 36: You need to update the changes step as well with this path

.github/workflows/test.driver.yaml
```diff
@@ -137,6 +139,9 @@ jobs:
         run: scripts/check_clang_format.sh driver
 
       - name: Test Driver
-        run:
-          bazel test --test_output=all //driver/... --jobs=1
-          --test_tag_filters=-hardware --nocache_test_results
+        run: |
+          SUPPRESSIONS_PATH="${{ github.workspace }}/scripts/sanitizers/lsan_suppressions.txt"
+          bazel test --config=asan-leak --test_output=all //driver/... \
+            --test_env=LSAN_OPTIONS="suppressions=${SUPPRESSIONS_PATH}" \
```
pjdotson: what is the point of making this a variable?

.github/workflows/test.freighter.cpp.yaml
```diff
@@ -73,4 +75,10 @@ jobs:
         run: scripts/check_clang_format.sh freighter/cpp
 
       - name: Test
-        run: bazel test --test_output=all //freighter/cpp/... --nocache_test_results
+        run: |
+          if [ "${{ matrix.os }}" == "ubuntu-latest" ]; then
```
pjdotson: This should be two separate steps that run based of matrix.os rather than making this shell script. And are we confident that we only can / should check on ubuntu - there won't be differences on different OSs?

.github/workflows/test.x.cpp.yaml
```diff
@@ -71,4 +73,10 @@ jobs:
         run: scripts/check_clang_format.sh x/cpp
 
       - name: Test
-        run: bazel test --test_output=all //x/cpp/... --nocache_test_results
+        run: |
+          if [ "${{ matrix.os }}" == "ubuntu-latest" ]; then
```
pjdotson: See comment in `test.freighter.cpp.yaml`

x/cpp/binary/binary.h
pjdotson, line 1: are there any performance implications to the changes here?

driver/sequence/plugins/channel_write.cpp
```diff
@@ -84,17 +79,32 @@ xerrors::Error plugins::ChannelWrite::before_all(lua_State *L) {
                 lua_touserdata(cL, lua_upvalueindex(1))
             );
             const char *channel_name = lua_tostring(cL, 1);
-            const auto [channel, err] = op->resolve(channel_name);
-            if (err) {
-                luaL_error(cL, err.message().c_str());
+            const auto [channel, found] = op->resolve(channel_name);
+            if (!found) {
+                lua_pushfstring(cL, "Channel %s not found", channel_name);
```
pjdotson, line 84: Would be nice to be able to add tests for some of the changes like this one if its not too big of a lift.

driver/opc/util/telem.cpp
pjdotson, line 1: Seems like this file should just be deleted if all it does is calls functions from other files

driver/opc/errors/errors.h
```diff
@@ -0,0 +1,28 @@
...
+/// external
+#include "open62541/types.h"
+
+/// module
+#include "x/cpp/xerrors/errors.h"
+
+/// internal
+#include "driver/errors/errors.h"
+
+namespace opc::errors {
```
pjdotson, line 21: shouldn't these namespaces just be `errors`, `conn`, etc.? Or does that cause some weird conflicting issues with namespaces with the same name so we have to do it like this

driver/opc/conn/conn.h
```diff
@@ -0,0 +1,158 @@
...
+
+std::pair<std::shared_ptr<UA_Client>, xerrors::Error>
+connect(const Config &cfg, std::string log_prefix);
+
+xerrors::Error
+reconnect(const std::shared_ptr<UA_Client> &client, const std::string &endpoint);
+
+class Pool {
+public:
+    class Conn {
```
pjdotson: I think this class should be called `Connection`, makes it easier to read / understand than `Conn`

driver/opc/connection/BUILD.bazel
pjdotson, line 1: I think this folder should be called `connection`

driver/opc/connection/connection_test.cpp
```diff
@@ -166,6 +168,8 @@ TEST(ConnTest, serverStopDuringConnection) {
     UA_ReadResponse res = UA_Client_Service_read(client.get(), req);
     EXPECT_NE(res.responseHeader.serviceResult, UA_STATUSCODE_GOOD);
     UA_ReadResponse_clear(&res);
+
+    // node_id automatically cleaned up by RAII destructor
```
pjdotson: don't need this comment

driver/opc/types/types.cpp
```diff
@@ -143,4 +169,14 @@ static const std::map<UA_NodeClass, std::string> NODE_CLASS_MAP = {
...
+xerrors::Error WriteRequestBuilder::add_value(
+    const UA_NodeId &node_id,
+    const ::telem::Series &series
+) {
+    auto [variant, err] = opc::telem::series_to_variant(series);
+    if (err) return err;
+    add_value(node_id, variant);
+    return xerrors::NIL;
 }
+} // namespace opc
```
pjdotson: get rid of this comment / set your editor to remove comments like this

driver/opc/testutil/BUILD.bazel
pjdotson, line 1: It would nice to be able to move the stuff from `read.cpp` into another folder with a better name - personally I've found that having modules / files / classes called 'util' is a code smell because it doesn't actually give you any meaning.

driver/opc/scan_task.h
```diff
@@ -12,35 +12,98 @@
...
+            {"data_type", data_type.name()},
+            {"name", name},
+            {"node_id", node_id},
+            {"node_class", node_class},
+            {"is_array", is_array}
+        };
+    }
+};
+
+struct DeviceProperties {
```
pjdotson: The Node and DeviceProperties seem like they don't belong in scan_task.h and belong in either `opc.h` or `types.h`

pjdotson (changes requested):
Something that might be worth doing is consolidating files like `opc/telem/telem.h` and `opc/telem/telem.cpp` into one file that contains both the implementation and documentation. I know it's not possible in all cases, but that is the default that we should be doing IMO because it is a little bit of duplicated code to have to describe the function signature and documentation in one place and the actual implementation in another.

x/cpp/binary/binary.h
emilbon99: No.

driver/opc/util/telem.cpp
emilbon99: Good catch

driver/opc/testutil/BUILD.bazel
emilbon99: I moved into `testutil`.

driver/opc/errors/errors.h
emilbon99: The conventions we use around namespaces are somewhat lax right now. Sometimes we use this convention to intentionally narrow packages with generic names but with parent package specific functionality. 

IMO we should spend time in a dedicated issue to think about namespace conventions before spending time refactoring these.

x/cpp/binary/binary.h
emilbon99: No

.github/workflows/test.freighter.cpp.yaml
emilbon99: So we can check ASAN and UBSAN on windows and macos but we can only check LSAN on ubuntu. It's not an absolute guarantee that code will be leak free on other OSes, but adding these checks improves our confidence by an order of magnitude. 

Spending a lot of time looking into leak checkers in CI for Windows and MacOS isn't worth the energy right now IMO. 

I split up steps.

driver/opc/types/types.cpp
emilbon99: So I removed. In the future adding cpplint will check for this.

driver/opc/errors/errors.h
pjdotson: Sounds good to me

PR #1550, opened 2025-10-08
SY-2984: OPC Read Robustification - Data Validation and Improved Reporting

nico-synnax:
# Issue Pull Request

## Linear Issue

[SY-2984](https://linear.app/synnax/issue/SY-2984/fix-opc-ua-driver-segfault)

## Description

* Make the OPC UA driver robust against invalid data, preventing segmentation faults
* Make task robust to invalid data (skipping writes on invalid frames)
  * One channel will invalidate a frame. If it's a transient error, the task will resume writing
  * This behavior creates a clear separation between "A Synnax Task crashed" and "Synnax Tasks are robust to bad writes and/or configs"

## Basic Readiness

- [ ] I have added relevant tests to cover the changes to CI. (Will add integration test when sy-3013 is merged to rc)


driver/opc/util/telem.cpp
```diff
@@ -116,11 +116,31 @@ std::pair<UA_Variant, xerrors::Error> series_to_variant(const telem::Series &s)
 }
 
 size_t write_to_series(telem::Series &s, const UA_Variant &v) {
```
emilbon99: This function needs to be refactored to return an error instead of logging directly. If you look in `pipeline/acquisition.cpp` line you'll see that we automatically warn and continue on errors that are sub-errors of the `TEMPORARY_HARDWARE_ERROR` class. This will handle skipping writing the frame and move on to the next cycle. 

i.e. it does the hard work for you.

emilbon99: Also worth it to check out the implementation in common/read_task.h line 138 which supplements this temporary error handling..

sy-nico: Implemented but have not yet tested. Will mark 'resolved' after doing so.

driver/opc/read_task.h
pjdotson, line 1: You should add unit tests in the `opc_read_task_test.cpp` file to test these changes. Will probably require you modifying the code for the mock OPC UA server.

driver/opc/util/telem.cpp
```diff
@@ -115,12 +115,51 @@ std::pair<UA_Variant, xerrors::Error> series_to_variant(const telem::Series &s)
...
+            )
+        };
+    } catch (...) {
+        return {
+            0,
+            xerrors::Error(
+                xerrors::VALIDATION,
+                "unknown exception during cast/write"
+            )
+        };
```
pjdotson, line 153: why do we need this second catch here?

driver/opc/util/telem.cpp
```diff
@@ -115,12 +115,51 @@ std::pair<UA_Variant, xerrors::Error> series_to_variant(const telem::Series &s)
...
-        return s.write(s.data_type().cast(ua_datetime_to_unix_nano(*dt)));
+        return {s.write(s.data_type().cast(ua_datetime_to_unix_nano(*dt))), xerrors::NIL};
+    }
+
+    if (v.type == nullptr || v.data == nullptr) {
+        return {
+            0,
+            xerrors::Error(
+                xerrors::VALIDATION,
+                "variant has null type or data"
```
pjdotson: this makes more sense to first check if the type is null, then the message is variant has null type, then check if the data is null, that way the message is more specific.

driver/opc/read_task_test.cpp
pjdotson, line 1: These tests don't test any of the changed code in `read_task.h`. They are only testing the `util:write_to_series` function. You should have those added tests in `driver/opc/util/util_test.cpp`, and then edit this file with tests for changes you have made in `read_task.h`

driver/opc/read_task.h
sy-nico: Reopened because I just modified the mock server to inject/test invalid data

driver/opc/read_task_test.cpp
sy-nico: Implemented feedback. Leaving open until we ensure I have added the right set of tests

driver/opc/mock/server.h
```diff
@@ -145,6 +146,53 @@ struct ServerConfig {
         };
         return cfg;
     }
+
+    // Create a configuration with nodes that return invalid/null data for testing error
+    // handling
+    static ServerConfig create_with_invalid_data() {
```
pjdotson, line 152: any reason not to just put this on the same server? that way we only have to write one server mock

driver/opc/read_task_test.cpp
```diff
@@ -446,3 +446,555 @@ TEST_F(TestReadTask, testConnectionPoolConcurrentTasks) {
...
+
+    // Let it run for a bit to collect some data
+    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+
+    rt->stop("stop_cmd", true);
+
+    // Verify task lifecycle worked correctly
+    ASSERT_GE(ctx->states.size(), 2);
+    EXPECT_EQ(ctx->states[0].variant, status::variant::SUCCESS);
+    EXPECT_EQ(ctx->states[1].variant, status::variant::SUCCESS);
```
pjdotson, line 507: would be good to test on the message as well

driver/opc/read_task_test.cpp
```diff
@@ -446,3 +446,555 @@ TEST_F(TestReadTask, testConnectionPoolConcurrentTasks) {
...
+    ASSERT_EVENTUALLY_GE(mock_factory->writes->size(), 1);
+
+    rt->stop("stop_cmd", true);
+
+    // Check all integer channel data
+    const auto &fr = mock_factory->writes->at(0);
+    ASSERT_TRUE(fr.contains(this->int8_channel.key));
+    ASSERT_TRUE(fr.contains(this->int16_channel.key));
+    ASSERT_TRUE(fr.contains(this->int32_channel.key));
+    ASSERT_TRUE(fr.contains(this->int64_channel.key));
```
pjdotson, line 606: shouldn't you test on the value for stuff like this?

pjdotson (changes requested):

driver/opc/mock/server.h
sy-nico: When using a single server, I was getting test failures. Likely, the malformed/invalid data nodes on the same OPC UA server instance as valid nodes were causing issues between tests.

*  A connection that encounters invalid data might become corrupted and affect subsequent operations on that same connection
* The open62541 server implementation may not handle nodes with nullptr types or data gracefully when they coexist with valid
* Clear separation of concerns - tests that verify error handling use dedicated corrupted servers

Here is the run with the failure: https://github.com/synnaxlabs/synnax/actions/runs/18505756441

PR #1551, opened 2025-10-08
SY-3110: TypeScript Circular Dependencies and Lint Improvements

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-3110](https://linear.app/synnax/issue/SY-3110/fix-react-minified-error-130)

## Description

This PR is an attempt at fixing issues that may have caused the React Minified Error 130. I only have medium confidence that these fixes are related, but spent a lot of time trying to replicate and failed. Most of the stuff online that I saw for this error had to do with import / export issues, so I have added a circular dependency check to CI and fixed circular dependencies to attempt to fix this error.

Summary:
- Add a few ESLint rules (`react/jsx-no-undef`, `react/jsx-no-constructed-context-values`, `react/jsx-no-useless-fragment`) that ChatGPT recommended but definitely do not address this issue
- Move ESLint configuration file to TypeScript to take some advantage of that typing, which necessitated moving a few random files to TypeScript
- Fixed circular dependencies in our TypeScript codebase and added a circular dependency check to CI.

## Basic Readiness

- [x]  I have added needed QA steps to the [release candidate](/synnaxlabs/synnax/blob/main/.github/PULL_REQUEST_TEMPLATE/rc.md) template that cover these changes.


console/src/schematic/Schematic.tsx
```diff
@@ -251,7 +251,7 @@ export const Loaded: Layout.Renderer = ({ layoutKey, visible }) => {
   const calculateCursorPosition = useCallback(
     (cursor: xy.Crude) =>
       Diagram.calculateCursorPosition(
-        box.construct(ref.current),
+        box.construct(ref.current ?? box.ZERO),
```
emilbon99, line 254: This change is interesting. Might be worth it to think what the implications of passing in `box.ZERO` here are. Not super importnat.

client/ts/src/framer/payload.ts
emilbon99, line 1: This is kind of a bummer. I assume the dep has something to do with the codec? If so, might be worth it to create `payload.ts` file that contains all of the RPC data structures that we'll send (iterator, writer, streamer) and then import those where needed.

Would be more consistent with rest of code base.

PR #1552, opened 2025-10-08
SY-3114: Build Binaries Workflow

sy-nico:
# Pull Request
[SY-3114](https://linear.app/synnax/issue/SY-3114/build-binaries-workflow)

Used to quickly build any binary for any OS on any branch - hassle-free!
This is a baby step towards "chaining" or "importing" workflow steps (not sure what is possible yet) so that we can use built artifacts efficiently.

- Easily extended for future workflow_calls
- Input Validation 
   - Fails if no OSes selected 
   - Fails if no binaries selected
- Core build tags automatically generated based on inputs
- Builds for Mac OS

### DIFF
--- a/.github/workflows/build.binaries.yaml
+++ b/.github/workflows/build.binaries.yaml
@@ -137,20 +137,30 @@ jobs:
           bazel --output_user_root=C:/tmp build --enable_platform_specific_config -c opt
           --announce_rc //driver
 
+      - name: Rename Driver Binary
+        if: inputs.driver
+        shell: bash
+        run: |
+          mv bazel-bin/driver/driver${{ matrix.executable }} bazel-bin/driver/synnax-driver-v${{ needs.setup.outputs.VERSION }}${{ matrix.executable }}
+
       - name: Upload Driver Binary as Artifact
         if: inputs.driver
         uses: actions/upload-artifact@v4
         with:
-          name: synnax-driver-${{ matrix.binary-suffix }}
-          path: bazel-bin/driver/driver${{ matrix.executable }}
+          name:
+            synnax-driver-v${{ needs.setup.outputs.VERSION }}-${{ matrix.binary-suffix
+            }}
+          path:
+            bazel-bin/driver/synnax-driver-v${{ needs.setup.outputs.VERSION }}${{
+            matrix.executable }}
           retention-days: 7
           if-no-files-found: error
 
-      - name: Move Driver to Assets
+      - name: Copy Driver to Assets
         if: inputs.driver && inputs.core
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


scripts/verify_build_config.sh
```diff
@@ -0,0 +1,62 @@
...
+echo ""
+echo "Binaries:"
+if [ "$BUILD_DRIVER" = "true" ]; then
+    echo "  ✓ Driver"
+fi
+if [ "$BUILD_CONSOLE" = "true" ]; then
+    echo "  ✓ Console"
+fi
+if [ "$BUILD_CORE" = "true" ]; then
+    echo "  ✓ Core"
```
pjdotson, line 75: would be nice to echo an X for the ones that aren't being built so we have that output

.github/workflows/build.binaries.yaml
```diff
@@ -0,0 +1,230 @@
...
+        uses: actions/checkout@v5
+
+      - name: Get Version
+        id: version
+        working-directory: core
+        run: |
+          if [ "${GITHUB_REF}" == "refs/heads/main" ]; then
+            echo "VERSION=$(cat pkg/version/VERSION)" >> $GITHUB_OUTPUT
+          else
+            echo "VERSION=$(cat pkg/version/VERSION)-rc" >> $GITHUB_OUTPUT
```
pjdotson: you should improve this by setting the version to be vX.Y.Z for main, and vX.Y.Z-branch-name for all other branches. Often times I find myself editing this line if I am debugging why builds are failing on a custom branch

.github/workflows/build.binaries.yaml
```diff
@@ -0,0 +1,230 @@
...
+
+      - name: Determine Executable Name
+        id: executable
+        shell: bash
+        run: |
+          if [ "${{ matrix.os }}" = "windows-latest" ]; then
+            echo "EXECUTABLE=.exe" >> $GITHUB_OUTPUT
+          else
+            echo "EXECUTABLE=" >> $GITHUB_OUTPUT
+          fi
```
pjdotson: this is something that should be determined by the matrix, no need to have a step from this. IE you call matrix.executable

.github/workflows/build.binaries.yaml
```diff
@@ -0,0 +1,230 @@
...
+          if-no-files-found: error
+
+      - name: Move Driver to Assets
+        if: inputs.driver && inputs.core
+        shell: bash
+        run: |
+          mv bazel-bin/driver/driver${{ steps.executable.outputs.EXECUTABLE }} core/pkg/service/hardware/embedded/assets/driver${{ steps.executable.outputs.EXECUTABLE }}
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
```
pjdotson, line 167: should only do this for inputs.core

PR #1565, opened 2025-10-17
SY-3060: Add Support for NI Counter Tasks

sy-nico:
# Issue
[SY-3060](https://linear.app/synnax/issue/SY-3060/add-ni-counter-input-frequency-support)

## Description
Add support for National Instruments Counter Tasks

- Driver & tests
- Console & tests
- Integration Task Form tests

- Counter Input/Read
  - Frequency
  - Edge Count
  - Period
  - Puls Width
  - Semi Period
  - Two Edge Separation

## Basic Readiness

- [x] User facing documentation being updated in [SY-3170](https://linear.app/synnax/issue/SY-3170/synnax-client-library-task-support)

<h2>Greptile Overview</h2>

Updated On: 2025-10-23 23:02:35 UTC

<h3>Greptile Summary</h3>

This iteration adds function documentation and type hints to the NI Counter Task integration test infrastructure. The changes introduce three new channel classes (`EdgeCount`, `PulseOutput`, `TwoEdgeSeparation`) and a `CounterRead` task interface for automating console UI testing of National Instruments counter hardware.

Key changes:
- **EdgeCount** and **TwoEdgeSeparation** properly inherit from the `Counter` base class, leveraging shared initialization logic for common fields (channel type, device, port, min/max values)
- **PulseOutput** implements manual initialization without calling the base class constructor, duplicating common setup logic
- **CounterRead** task provides type-safe channel registration with validation against `COUNTER_READ_CHANNEL_TYPES` dictionary
- All classes include comprehensive docstrings with parameter descriptions and type hints
- Form assertion logic uses bare `except` clause in `pulse_output.py:115` (matches pattern in base class `counter.py:87`)

- This PR is safe to merge - adds well-structured integration test infrastructure with proper type safety
- Score reflects complete implementation of integration test channels with proper inheritance patterns, comprehensive type hints and documentation, and validation logic; minor style issue with bare except clause is consistent with existing codebase patterns
- No files require special attention - all implementations follow established patterns

File Analysis

| Filename | Score | Overview |
|----------|-------|----------|
| integration/console/task/channels/edge_count.py | 5/5 | New EdgeCount channel implementation with proper Counter base class inheritance and complete form field handling |
| integration/console/task/channels/pulse_output.py | 4/5 | PulseOutput channel with manual initialization instead of Counter base class, includes bare except clause in assert_form method |
| integration/console/task/channels/two_edge_separation.py | 5/5 | TwoEdgeSeparation channel with clean Counter inheritance and helper function for edge configuration |
| integration/console/task/counter_read.py | 5/5 | Counter read task interface with type validation and proper channel registration |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Test as Integration Test
    participant CR as CounterRead Task
    participant CH as Channel (EdgeCount/PulseOutput/TwoEdgeSeparation)
    participant Base as Counter Base Class
    participant Console as Console Interface
    participant UI as Playwright UI

    Test->>CR: add_channel(name, type, device, **kwargs)
    CR->>CR: Validate type in COUNTER_READ_CHANNEL_TYPES
    CR->>CH: Instantiate channel class
    
    alt EdgeCount/TwoEdgeSeparation
        CH->>Base: super().__init__(console, name, device, type, **kwargs)
        Base->>Console: click_btn("Channel Type")
        Console->>UI: Select channel type
        Base->>Console: get_dropdown_value("Device")
        Base->>Console: fill_input_field("Port", port)
        Base->>Base: Store form_values
        Base-->>CH: Return control
        CH->>Console: Configure specific fields (active_edge, count_direction, etc.)
        Console->>UI: Fill form fields
    else PulseOutput
        CH->>CH: Manual initialization (no base class call)
        CH->>Console: click_btn("Channel Type")
        Console->>UI: Select "Pulse Output"
        CH->>Console: Configure all fields manually
        Console->>UI: Fill form fields
    end
    
    CH-->>CR: Return channel instance
    CR-->>Test: Return channel
    
    Test->>CH: assert_form()
    CH->>Console: get_input_field(key) / get_dropdown_value(key)
    Console->>UI: Read form values
    UI-->>Console: Return values
    Console-->>CH: Return values
    CH->>CH: Assert expected == actual
```


integration/console/task/channels/counter_input/two_edge_separation.py
```diff
@@ -0,0 +1,74 @@
...
+            name=name,
+            device=device,
+            type="Two Edge Separation",
+            **kwargs,
+        )
+
+        # Scaled Units
+        if units is not None:
+            console.click_btn("Scaled Units")
+            console.select_from_dropdown(units)
```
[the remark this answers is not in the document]
sy-nico: Select_from_dropdown() will not fail silently if the item is not available. The test case will fail loudly if the element was not found. 
Also, the function needs be allow any strings because every drop down in the console is different.

integration/framework/test_conductor.py
```diff
@@ -347,6 +348,55 @@ def load_test_sequence(
...
+        expanded_tests = []
+        for combo in combinations:
+            # Create parameter dict from combination
+            combo_params = dict(zip(matrix_keys, combo))
+
+            # Merge with base params (matrix params override base params)
+            merged_params = {**test_def.params, **combo_params}
+
+            # Generate name from matrix values
+            matrix_suffix = "_".join(str(v) for v in combo)
```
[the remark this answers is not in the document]
sy-nico: Fine as is. Simplicity here is better.

console/src/hardware/ni/task/types/v1.ts
```diff
@@ -241,9 +241,111 @@ export const ZERO_AI_CHANNELS: Record<v0.AIChannelType, AIChannel> = {
 };
 export const ZERO_AI_CHANNEL: AIChannel = ZERO_AI_CHANNELS[v0.AI_VOLTAGE_CHAN_TYPE];
 
+// ==================== Counter Input Channels ====================
+
+const ciChanExtensionShape = { device: Common.Device.keyZ };
+interface CIChanExtension extends z.infer<z.ZodObject<typeof ciChanExtensionShape>> {}
+const ZERO_CI_CHAN_EXTENSION: CIChanExtension = { device: "" };
```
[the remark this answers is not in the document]
sy-nico: This is the established pattern in the codebase, so we should leave the CI/CO extension pattern as-is to match the existing AI pattern. The duplication is intentional for consistency across different channel types. No action needed for this comment - the current implementation follows the established codebase pattern.

console/src/hardware/ni/task/CounterRead.tsx
```diff
@@ -0,0 +1,243 @@
...
+    // Initialize shared index for counter channels (use same index as analog input)
+    let shouldCreateIndex = primitive.isZero(dev.properties.analogInput.index);
+    if (!shouldCreateIndex)
+      try {
+        await client.channels.retrieve(dev.properties.analogInput.index);
+      } catch (e) {
+        if (NotFoundError.matches(e)) shouldCreateIndex = true;
+        else throw e;
+      }
+    if (shouldCreateIndex) {
```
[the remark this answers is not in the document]
sy-nico: No concern. The index channel can not be deleted if the data channels are still tied to it. Updated to counterInput.index

console/src/hardware/ni/task/CounterRead.tsx
```diff
@@ -0,0 +1,243 @@
...
+    for (const channel of deviceChannels) {
+      const exKey = dev.properties.counterInput.channels[channel.port.toString()];
+      if (primitive.isZero(exKey)) toCreate.push(channel);
+      else
+        try {
+          await client.channels.retrieve(exKey.toString());
+        } catch (e) {
+          if (QueryError.matches(e)) toCreate.push(channel);
+          else throw e;
+        }
```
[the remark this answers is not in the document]
sy-nico: No concern. Port collision detection happens at the driver level when a task is being configured. This already matches the pattern of AnalogRead and DigitalRead.

integration/tests/console/task/ni_counter_write_forms.py
```diff
@@ -0,0 +1,98 @@
...
+        self.log(f"Asserting {total} channel forms in random order")
+        for ch in ch_names:
+            console.ni_co.assert_channel(ch)
+
+    def create_test_rack(self, rack_name: str, device_name: str) -> None:
+        rack = self.client.hardware.racks.create(name=rack_name)
+        self.client.hardware.devices.create(
+            [
+                sy.Device(
+                    key=f"130227d9-03bb-47e4-b370-0d590add1bc5",
```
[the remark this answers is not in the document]
sy-nico: This is just a random device key. There is no need to extract to a shared constant.

integration/console/task/counter_write.py
```diff
@@ -0,0 +1,103 @@
...
+        return super().new()
+
+    def add_channel(
+        self,
+        name: str,
+        type: str,
+        device: str,
+        dev_name: Optional[str] = None,
+        **kwargs: Any,
+    ) -> NIChannel:
```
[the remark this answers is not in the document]
sy-nico: Updated helper so this is now consistent.

console/src/hardware/ni/task/CounterWrite.tsx
```diff
@@ -0,0 +1,261 @@
...
+  typeof counterWriteConfigZ,
+  typeof counterWriteStatusDataZ
+> = ({ deviceKey, config }) => {
+  const cfg =
+    config != null
+      ? counterWriteConfigZ.parse(config)
+      : ZERO_COUNTER_WRITE_PAYLOAD.config;
+  return {
+    ...ZERO_COUNTER_WRITE_PAYLOAD,
+    config: { ...cfg, device: deviceKey ?? (cfg as any).device },
```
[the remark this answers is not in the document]
sy-nico: CounterWrite should have been using v0 schema. This has been fixed.

integration/console/task/channels/counter.py
```diff
@@ -0,0 +1,105 @@
...
+
+        self.form_values = values
+
+    def assert_form(self) -> None:
+        """Assert that form values match expected values."""
+        for key, expected_value in self.form_values.items():
+            try:
+                actual_value = self.console.get_input_field(key)
+            except Exception:
+                actual_value = self.console.get_dropdown_value(key)
```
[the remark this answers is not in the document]
sy-nico: No concern. There are only 2 ways to select a value on the task forms. If this fails, the test will (and should) fail.

sy-nico:
@greptile /rerun

integration/console/task/counter_write.py
```diff
@@ -0,0 +1,104 @@
...
+            channel_class=COUNTER_WRITE_CHANNEL_TYPES[type],
+            **kwargs,
+        )
+
+    def set_parameters(
+        self,
+        task_name: Optional[str] = None,
+        data_saving: Optional[bool] = None,
+        auto_start: Optional[bool] = None,
+        state_update_rate: Optional[float] = None,
```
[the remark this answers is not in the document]
sy-nico: The current implementation is correct. Python's function signature semantics ensure that explicit parameters take precedence over `**kwargs`. When `state_update_rate` is passed as a keyword argument, it binds to the explicit parameter and does not appear in `**kwargs. If someone attempts to pass it both ways (e.g., `set_parameters(state_update_rate=1.0, **{"state_update_rate": 2.0})`), Python raises a `TypeError: got multiple values for keyword argument 'state_update_rate'` automatically. The explicit parameter was intentionally promoted from `**kwargs` (as mentioned in the Greptile summary) to improve API discoverability - users can now see `state_update_rate` in IDE autocomplete and type hints. No additional validation is needed since Python's parameter binding handles this correctly.

sy-nico:
@greptile /rerun

pluto/src/select/Dialog.tsx
```diff
@@ -31,9 +31,14 @@ export interface DefaultEmptyContentProps extends Status.SummaryProps {
   resourceName?: string;
 }
 
+const pluralize = (word: string): string => {
```
pjdotson: instead of this custom function, use the `pluralize` function from npm (pluralization
functions, like auth and JSON parsers,  are one thing where you should never code your
own - just too many edge cases in the English language.)

driver/ni/channel/channels.h
```diff
@@ -289,14 +335,17 @@ struct Output : virtual Base {
...
 
     /// @brief binds remotely fetched information to the channel.
     void bind_remote_info(const synnax::Channel &state_ch, const std::string &dev_loc) {
         this->state_ch = state_ch;
         this->dev_loc = dev_loc;
     }
+
+    /// @brief returns true if this channel supports runtime command/state control.
+    [[nodiscard]] virtual bool supports_runtime_control() const { return true; }
 };
```
pjdotson, line 337: I feel like it doesn't make sense for Counter Output. to inherit from this class. Like
there should be one class that has a state and command channel and one that doesn't.

The other thought I had is that maybe counter output channels should have a state and
command, but when you pass a 1 to the cmd channel, it starts the counter output for that
channel in particular, and
when you pass a zero, it stops it. I am not sure how possible that is with the DAQmx
API, but I think you should consider it.

driver/ni/channel/channels.h
```diff
@@ -356,8 +405,8 @@ struct Analog : virtual Base {
 
     explicit Analog(xjson::Parser &cfg):
         port(cfg.required<int>("port")),
-        min_val(cfg.optional<float>("min_val", 0)),
-        max_val(cfg.optional<float>("max_val", 0)),
+        min_val(cfg.optional<double>("min_val", 0)),
+        max_val(cfg.optional<double>("max_val", 0)),
```
pjdotson, line 397: why is this a double now / add a test for this?

driver/ni/ni.h
```diff
@@ -139,5 +141,10 @@ class Factory final : public task::Factory {
 
     common::ConfigureResult
     configure_scan(const std::shared_ptr<task::Context> &ctx, const synnax::Task &task);
+
+    common::ConfigureResult configure_counter_write(
```
pjdotson: why does this have to be a separate function?

driver/ni/hardware/hardware.h
```diff
@@ -128,6 +128,57 @@ struct AnalogReader final : Base, Reader<double> {
 
     int64 update_skew(const size_t &n_requested);
 };
+
+/// @brief a hardware interface for counter input tasks.
+struct CounterReader final : Base, Reader<double> {
```
pjdotson: Can you explain / justify why we should have float64 as the type for the reader? I am
just wondering if an int type would be better, or if it should be different for
different channel types.

driver/ni/hardware/daqmx.cpp
```diff
@@ -147,4 +151,104 @@ int64 AnalogReader::update_skew(const size_t &n_requested) {
...
+    if (const auto err = this->dmx->SetReadOverWrite(
+            this->task_handle,
+            DAQmx_Val_OverwriteUnreadSamps
+        ))
+        return err;
+    return Base::start();
+}
+
+int64 CounterReader::update_skew(const size_t &n_requested) {
+    uInt64 next_total_samples_acquired;
```
pjdotson: why isn't this shared code with the stuff for the analog reader?

integration/console/task/channels/counter.py
```diff
@@ -0,0 +1,127 @@
...
+#  licenses/BSL.txt.
+#
+#  As of the Change Date specified in that file, in accordance with the Business Source
+#  License, use of this software will be governed by the Apache License, Version 2.0,
+#  included in the file licenses/APL.txt.
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from console.console import Console
```
pjdotson, line 17: When are we going to remove this circular dependency in the Console integration code? I
feel like this should be done sooner rather than later so refactoring doesn't get
tougher and tougher.

integration/console/task/channels/counter.py
```diff
@@ -0,0 +1,127 @@
...
+                actual_value = self.console.get_input_field(key)
+            else:
+                actual_value = self.console.get_dropdown_value(key)
+
+            assert (
+                actual_value == expected_value
+            ), f"Channel {self.name} Form value '{key}' - Expected: {expected_value} - Actual: {actual_value}"
+
+    @staticmethod
+    def _is_numeric_string(value: str | bool) -> bool:
```
pjdotson: I feel like I've seen this method elsewhere. This shouldn't be a static method anyways
and should just be a function in a util file or something.

integration/tests/console/task/no_device.py
```diff
@@ -81,6 +81,7 @@ def configure_without_channels(self) -> None:
         console.ni_ai.configure()
 
         # Assert error notification
+        sy.sleep(1)
```
pjdotson: I feel like you need to refactor this pattern from sleep once / check once to something
like a loop where you poll every 100ms and have a timeout where you fail after 1 second.
Look for functions in Go test files that use the Ginkgo `Eventually()` function to see
more of what I am talking about. It makes the test a little more robust and should also
allow them to move a bit quicker as we won't have to wait the whole 1 second and these
things can start to add up. You probably want to write a helper function for this or use
something in `pytest`.

integration/tests/test_schema.json
```diff
@@ -41,6 +41,21 @@
...
+                  "additionalProperties": {
+                    "type": "array",
+                    "items": {
+                      "anyOf": [
+                        { "type": "string" },
+                        { "type": "number" },
+                        { "type": "boolean" }
+                      ]
+                    },
+                    "minItems": 1
```
pjdotson: The minimum items should be 2 unless you remove the `parameters` property as well and do
everything in a matrix.

driver/ni/ni.h
pjdotson, line 1: Where are the tests for counter task?

console/src/hardware/ni/task/CIChannelForm.tsx
```diff
@@ -0,0 +1,707 @@
...
+>({
+  fieldKey: "measMethod",
+  fieldProps: { label: "Measurement Method" },
+  inputProps: {
+    resourceName: "Measurement Method",
+    data: [
+      { key: "LowFreq1Ctr", name: "1 Counter (Low Frequency)" },
+      { key: "HighFreq2Ctr", name: "2 Counters (High Frequency)" },
+      { key: "LargeRng2Ctr", name: "2 Counters (Large Range)" },
+      { key: "DynamicAvg", name: "Dynamic Averaging" },
```
pjdotson, line 189: You should format these as "One Counter", "Two Counters" instead of numbers. Generally
using words instead of numbers for smaller numbers is better writing style (exceptions
apply, yada yada, it's a bit of a matter of taste).

console/src/hardware/ni/task/SelectCOChannelTypeField.tsx
```diff
@@ -0,0 +1,56 @@
...
+      const parentPath = path.slice(0, path.lastIndexOf("."));
+      const prevParent = get<COChannel>(parentPath).value;
+      const schema = CO_CHANNEL_SCHEMAS[value];
+      set(parentPath, {
+        ...deep.overrideValidItems(next, prevParent, schema),
+        type: next.type,
+      } as COChannel);
+    },
+  },
+  inputProps: {
```
pjdotson: There seems to only be one channel type here? why do we even have this selection field

console/src/hardware/ni/task/CounterWrite.tsx
```diff
@@ -0,0 +1,172 @@
...
+  if (item == null) return null;
+  const { port, type, configured = false } = item;
+  const Icon = CO_CHANNEL_TYPE_ICONS[type];
+
+  // CO Pulse Output channels don't have cmd/state channels - config only
+  return (
+    <Select.ListItem {...props}>
+      <Flex.Box direction="x" gap="small" align="center">
+        <Text.Text>{port}</Text.Text>
+        <Icon />
```
pjdotson: this icon is too dark and should have a tooltip - look at the other forms

console/src/hardware/ni/task/types/v0.ts
```diff
@@ -952,6 +952,500 @@ export const AI_CHANNEL_TYPE_ICONS: Record<AIChannelType, Icon.FC> = {
   [AI_VOLTAGE_CHAN_TYPE]: Icon.Units.Voltage,
 };
 
+// ==================== Counter Input Channels ====================
```
pjdotson: why is this comment here (doesn't exist for other ones)? Why is all of this code
duplicated in both `v0.ts` and `v1.ts` - no need to write it twice?

console/src/hardware/ni/task/types/v0.ts
```diff
@@ -1186,6 +1680,60 @@ export const ZERO_ANALOG_READ_PAYLOAD: AnalogReadPayload = {
...
+      portToIndexMap.set(port, i);
+      return;
+    }
+    const index = portToIndexMap.get(port) as number;
+    const code = "custom";
+    const message = `Port ${port} has already been used on another channel`;
+    issues.push({ path: [index, "port"], code, message, input: channels });
+    issues.push({ path: [i, "port"], code, message, input: channels });
+  });
+};
```
pjdotson: I feel like this function is duplicated in two or three separate places and it should
only be written once (and then tested!).

console/src/hardware/ni/task/types/v1.ts
```diff
@@ -15,25 +15,29 @@ import * as v0 from "@/hardware/ni/task/types/v0";
...
+    channels.forEach(({ device, port }, i) => {
+      if (!deviceToPortMap.has(device)) deviceToPortMap.set(device, new Map());
+      const portToIndexMap = deviceToPortMap.get(device) as PortToIndexMap;
+      if (!portToIndexMap.has(port)) {
+        portToIndexMap.set(port, i);
+        return;
+      }
+      const index = portToIndexMap.get(port) as number;
+      const code = "custom";
+      const message = `${portTypeLabel} port ${port} has already been used on another channel on the same device`;
```
pjdotson: why do you need this port type label at all? and if it `""` this error message will look weird.

driver/ni/channel/channels.h
sy-nico: Resolved/OutOfDate:

Removed support for Counter Write and can implement in [SY-3237](https://linear.app/synnax/issue/SY-3237/counter-channel-support-for-remainder-api-channels)

driver/ni/channel/channels.h
sy-nico: Changing to `<double>` was a correction
The member variables are declared as `double` (lines 390-391, 458-459), and the NI-DAQmx API expects `float64` parameters (which is `double`). Parsing as` <float>` then storing in `double` would introduce unnecessary precision loss. The existing tests in `channels_tests.cpp` already verify this works correctly with double-precision values like `0.000001` seconds.

driver/ni/ni.h
sy-nico: Resolved/OutOfDate:

Removed support for Counter Write and can implement in [SY-3237](https://linear.app/synnax/issue/SY-3237/counter-channel-support-for-remainder-api-channels)

driver/ni/hardware/hardware.h
sy-nico: `float64` (double) is the correct type because the `CounterReader` uses `ReadCounterF64` to read data from all counter input channel types. While edge count channels return integer counts, most other counter measurements (frequency, period, pulse width, velocity, position, duty cycle) return floating-point values. DAQmx's `ReadCounterF64` handles all counter types uniformly, converting integer counts to doubles when needed. This unified approach allows a single `Reader<double>` interface to work for all counter input tasks.

integration/console/task/channels/counter.py
sy-nico: [SY-3078](https://linear.app/synnax/issue/SY-3078/enable-workspaces-and-multiple-pages-of-any-type-for-the-console-test)

For now, I added `from __future__ import annotations` which makes all type annotations strings by default. Combined with the `TYPE_CHECKING` block, this breaks the circular dependency at runtime (the import only happens during type checking) while still satisfying mypy. Also fixed the return type annotations in the `has_terminal_config` methods to avoid `no-any-return` errors.

integration/tests/console/task/no_device.py
sy-nico: Refactored the check_for_notifications function

console/src/hardware/ni/task/SelectCOChannelTypeField.tsx
sy-nico: Resolved/Outdated

Counter Write task support removed

console/src/hardware/ni/task/CounterWrite.tsx
sy-nico: Resolved/Out of Date: Counter Write task support removed

driver/ni/write_task.h
```diff
@@ -69,13 +69,19 @@ struct WriteTaskConfig : common::BaseWriteTaskConfig {
         state_rate(telem::Rate(cfg.required<float>("state_rate"))) {
         cfg.iter("channels", [&](xjson::Parser &ch_cfg) {
             auto ch = channel::parse_output(ch_cfg);
-            if (ch != nullptr && ch->enabled)
+            if (ch != nullptr && ch->enabled && ch->cmd_ch_key != 0)
```
pjdotson: seems like you should return an error for a cmd channel key of zero. you also should edit the test here

driver/task/common/write_task.h
```diff
@@ -255,8 +255,7 @@ class WriteTask final : public task::Task {
         const auto sink_started = !this->state.error(this->sink->internal->start());
         if (sink_started) {
             this->cmd_write_pipe.start();
-            if (!this->sink->internal->writer_config().channels.empty())
-                this->state_write_pipe.start();
+            this->state_write_pipe.start();
```
pjdotson: can you add a test for this change?

driver/ni/hardware/hardware.h
```diff
@@ -109,24 +109,48 @@ struct DigitalReader final : Base, Reader<uint8_t> {
...
         TaskHandle task_handle
     );
     ReadResult read(size_t samples_per_channel, std::vector<double> &data) override;
+};
 
-    xerrors::Error start() override;
-
-    int64 update_skew(const size_t &n_requested);
+/// @brief a hardware interface for counter input tasks.
+struct CounterReader final : SkewTrackingReader<double> {
```
pjdotson, line 148: is double the right type for the counter?

console/src/hardware/ni/task/types/v1.ts
pjdotson, line 1: I don't see why we take all of this code defined here - seems like almost all of this should be in v0.ts as this is v0 of the counter input channels

console/src/hardware/ni/task/CIChannelForm.tsx
```diff
@@ -0,0 +1,707 @@
...
+  type: CIChannelType;
+  prefix: string;
+}
+
+export const CIChannelForm = ({ type, prefix }: CIChannelFormProps) => {
+  const Form = CHANNEL_FORMS[type];
+  return (
+    <>
+      <Flex.Box x wrap>
+        <Device.Select path={`${prefix}.device`} />
```
pjdotson, line 700: Do you have any idea what path we should move towards - allowing all tasks to contain
multiple devices or not? Originally all tasks were only done on one device, but we
allowed for multi-device tasks because of some limit that came up when deploying to a
customer that @emilbon99 could say more about. Obviously all channels on a task have to
be on the same rack.

console/src/hardware/ni/task/createChannel.ts
```diff
@@ -73,3 +75,32 @@ export const createAOChannel = (channels: AOChannel[], key?: string): AOChannel
...
+  override: Partial<CIChannel>,
+  keyToCopy?: string,
+): CIChannel => {
+  const key = id.create();
+  let template: CIChannel;
+  if (channels.length === 0) template = deep.copy(zeroChannel);
+  else if (keyToCopy == null) template = deep.copy(channels[0]);
+  else {
+    const channel = channels.find(({ key }) => key === keyToCopy);
+    if (channel == null) return { ...deep.copy(zeroChannel), key } as CIChannel;
```
pjdotson: type assertions like `as CIChannel` are a bad practice in TypeScript and have been a
source of bugs for us. please try to rewrite this function to avoid it.

console/src/hardware/ni/task/CounterRead.tsx
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    if (toCreate.length > 0) {
+      devModified = true;
+      const channels = await client.channels.create(
+        toCreate.map((c) => ({
+          name: `${dev.properties.identifier}_ctr_${c.port}`,
+          dataType: "float64",
```
pjdotson, line 208: is float64 the right type for these channels?

integration/tests/console_task_tests.json
```diff
@@ -17,7 +29,7 @@
       ]
     },
     {
-      "sequence_name": "NI Form Input Verification 2",
+      "sequence_name": "NI Analog Forms 2",
```
pjdotson, line 32: a little confused why we have two different sequences for NI Analog Forms

integration/tests/console_task_tests.json
```diff
@@ -2,7 +2,19 @@
...
   "sequences": [
     {
-      "sequence_name": "NI Form Input Verification 1",
+      "sequence_name": "NI Counter Forms",
+      "sequence_order": "asynchronous",
+      "tests": [
+        {
+          "case": "console/task/ni_counter_read_forms",
+          "parameters": {
+            "mode": ["a", "b", "c"]
```
pjdotson: If mode is just one of three possible things to call it, it would be better to just call
it "variant" and have variant 1, 2, and 3. I think the convention of just calling things
a b and c is a bad practice and we should use either numbers or a name with a meaning.

integration/console/console.py
```diff
@@ -219,59 +222,75 @@ def check_for_error_screen(self) -> None:
             self.page.get_by_text("Try again").click()
             self.page.wait_for_timeout(200)
 
-    def check_for_notifications(self) -> list[dict[str, Any]]:
+    def check_for_notifications(self, timeout: float = 1.0) -> list[dict[str, Any]]:
```
pjdotson: make this a `CrudeTimeSpan`

integration/console/task/counter_read.py
```diff
@@ -0,0 +1,129 @@
+#  Copyright 2025 Synnax Labs, Inc.
+#
+#  Use of this software is governed by the Business Source License included in the file
+#  licenses/BSL.txt.
+#
+#  As of the Change Date specified in that file, in accordance with the Business Source
+#  License, use of this software will be governed by the Apache License, Version 2.0,
+#  included in the file licenses/APL.txt.
+
+from typing import TYPE_CHECKING, Any, Optional, Type
```
pjdotson: You should avoid using `Optional[T]` and instead use `T | None = None`. I'd like to set
up a linter for Python eventually to catch stuff like this but we should just try
writing better / more modern code until we get the chance to do that.

integration/console/task/channels/counter.py
pjdotson, line 1: I would organize this folder into a `channels/analog-input` and `channels/counter-input`
so there isn't dozens of files in it.

pjdotson (changes requested):
Looks good and is really close

driver/ni/write_task.h
sy-nico: This was meant to be reverted after removing counter write task support. Should be original code.

driver/task/common/write_task.h
sy-nico: This should not have been changed. Change reverted.

driver/ni/hardware/hardware.h
sy-nico: Yes, `double` is correct here. The DAQmx API uses `ReadCounterF64` for counter measurements, which returns `float64*` (double). Most counter input types (frequency, period, pulse width, duty cycle, velocity, position) naturally return floating-point values. Even `ci_edge_count`, which logically returns integers, is accessed via the F64 API in DAQm

console/src/hardware/ni/task/CIChannelForm.tsx
sy-nico: Yes, we should allow multi-device tasks for all tasks. This is standard for NI tasks. The channels will specify what module/port they run on. However, I don't believe you can have cross-chassis tasks (because the task will use a single clock) without some extra set up that we don't need to implement.

Context:
The limit was that you can only have 3 independent read tasks sharing the chassis' clock. This resulted in the need to group channels into a single task to free up resources. 
(Volt, TC, RTD, PT, Curr ) -- consolidated --> (Volt+PT, RC+RTD, Curr)

console/src/hardware/ni/task/createChannel.ts
sy-nico: Fixed. Made `createCounterChannel` generic like `createAnalogChannel` (added `<C extends CIChannel>`) so TypeScript can properly infer types. Removed the `as CIChannel` type assertions which were code smells indicating improper type inference. Now CIChannel follows the exact same pattern as AIChannel/DIChannel.

console/src/hardware/ni/task/CounterRead.tsx
sy-nico: `float54`is correct for counter channels. This matches the C++ driver which uses `double` (float64) and the DAQmx `ReadCounterF64` API.

integration/tests/console_task_tests.json
sy-nico: Load distribution of
- 4 concurrent playwright test
- python test infrastructure
- synnax console
- git runner infrastructure

integration/tests/console_task_tests.json
sy-nico: I used "set" and updated to numerics

console/src/hardware/ni/task/types/v1.ts
sy-nico: v0 already has the base counter channels definitions and v1 adds the device field extension to mirror the analog channel pattern and respect the versioning system.

pjdotson: Right, but my point is there is no need to mirror the analog channel pattern. We need a v0 and a v1 for those because we released single-device tasks, and then released multi-device tasks which required migrations. For this stuff we are already releasing it as multi-device for v0, so lets keep everything in one file

console/src/hardware/ni/task/CIChannelForm.tsx
pjdotson: Copy. Could you make an issue to migrate the other NI tasks like this?

integration/console/task/analog_write.py
```diff
@@ -12,10 +12,9 @@
 from playwright.sync_api import Page
 
 from console.task.channels.analog import Analog
-from console.task.channels.current import Current
-from console.task.channels.voltage import Voltage
+from console.task.channels.analog_input import Current, Voltage
```
pjdotson: why are the current and voltage channels in analog input? shouldn't they be analog output?

console/src/hardware/ni/task/CIChannelForm.tsx
sy-nico: [SY-3254](https://linear.app/synnax/issue/SY-3254/upgrade-all-ni-tasks-to-a-multi-device-pattern)

integration/console/task/analog_write.py
sy-nico: The analog write tasks should have their own channels. I could get away by using the same imports due to the structure, however, Voltage and Current writes don't have `terminal config` or `shunt resistor`, so in order to be correct, AI and AO should have separate channels

console/src/hardware/ni/task/types/v1.ts
sy-nico: Understood. We are using v0 because this is the first version of Counter Read tasks.
Implemented the changes and v1.ts should have exclusively migrations for analog read tasks

pjdotson (approved):

PR #1579, opened 2025-10-24
[SY-3087] PR1: Add foundational x/go utilities for Arc-based calculated channels

emilbon99:
## Summary

First PR in a series to introduce Arc-based calculated channel support. This PR adds foundational utilities in `x/go/` that provide the core data structures and operations needed by the Arc compiler and runtime.

### New Package: `telem/op`
- Mathematical and logical operations on telemetry series
- Type-safe operation definitions with code generation
- Benchmarks for performance validation

### Enhanced Packages
- **computron**: Computation utilities for channel evaluation
- **gorp**: Improved CRUD operations for data persistence
- **telem**: Extended frame/series handling with better alignment
- **signal/routine**: Signal processing enhancements for reactive execution
- **maps/set**: Collection utilities for managing channel sets
- **unsafe**: Performance-critical low-level utilities

## Dependencies

This PR has no dependencies and can be merged directly into `rc`.

## Testing

- ✅ All x/go tests pass
- ✅ Builds successfully

## Related Issues

Part of #3087

### DIFF
--- a/x/go/telem/series.go
+++ b/x/go/telem/series.go
@@ -153,12 +153,15 @@ func SetValueAt[T types.Numeric](s Series, i int, v T) {
 	f(s.Data[i*int(s.DataType.Density()):], v)
 }
 
-// CopyValue copies the sample from src at the index srcIdx to the index srcIdx in src.
+// CopyValue copies the sample from src at the index srcIdx to the index dstIdx in dst.
 // dst and src must have the same DataType, and that DataType cannot be of variable
 // density.
 func CopyValue(dst, src Series, dstIdx, srcIdx int) {
-	if dst.DataType != src.DataType || dst.DataType.IsVariable() || src.DataType.IsVariable() {
-		panic("cannot copy values from non-variable series")
+	if dst.DataType != src.DataType {
+		panic("cannot copy values across series with different data types")
+	}
+	if dst.DataType.IsVariable() {
+		panic("cannot copy values from series with variable density data types")
 	}
 	den := int(dst.DataType.Density())
 	copy(dst.Data[dstIdx*den:(dstIdx+1)*den], src.Data[srcIdx*den:(srcIdx+1)*den])
--- a/x/go/signal/signal_test.go
+++ b/x/go/signal/signal_test.go
@@ -281,6 +281,21 @@ var _ = Describe("Signal", func() {
 			Expect(ctx.Wait()).To(BeNil())
 		})
 
+		It("Should wrap an error panic with routine key", func() {
+			ctx, _ := signal.Isolated()
+			originalErr := errors.New("original panic error")
+
+			ctx.Go(func(ctx context.Context) error {
+				panic(originalErr)
+			}, signal.RecoverWithErrOnPanic(), signal.WithKey("test-routine"))
+
+			err := ctx.Wait()
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("routine test-routine recovered"))
+			Expect(err.Error()).To(ContainSubstring("original panic error"))
+			Expect(errors.Is(err, originalErr)).To(BeTrue())
+		})
+
 		It("Should try to restart when instructed to", func() {
 			var (
 				counter = 0


x/go/telem/series_factory.go
```diff
@@ -291,3 +291,242 @@ func UnmarshalF[T Sample](dt DataType) func(b []byte) T {
...
+		data[i] = start + T(i)*spacing
+	}
+	return NewSeries(data)
+}
+
+// NewSeriesFromAny creates a single-value Series from a value of type any, casting it
+// to the specified DataType. This function preserves numeric precision by avoiding
+// unnecessary intermediate conversions. Supports numeric types, strings, TimeStamp,
+// JSON, and bytes. Panics if the value cannot be converted to the target DataType.
+func NewSeriesFromAny(value any, dt DataType) Series {
```
emilbon99, line 310: Since I already know you're going to complain about this, I created an issue to think more carefully through the patterns we use to construct series in go.

https://linear.app/synnax/issue/SY-3175/improve-series-factories

x/go/signal/routine.go
```diff
@@ -354,7 +354,7 @@ func (r *routine) maybeRecover(panicReason any) error {
 	case recoverErr:
 		r.state.state = Failed
 		if err, ok := panicReason.(error); ok {
-			return errors.Wrap(err, "routine recovered")
+			return errors.Wrapf(err, "routine %s recovered", r.key)
```
pjdotson, line 357: Do we have a test for this? not a huge deal

x/go/testutil/must.go
pjdotson, line 1: would be nice to have tests for these as well but isn't a huge deal

x/go/telem/series.go
```diff
@@ -127,6 +153,17 @@ func SetValueAt[T types.Numeric](s Series, i int, v T) {
 	f(s.Data[i*int(s.DataType.Density()):], v)
 }
 
+// CopyValue copies the sample from src at the index srcIdx to the index srcIdx in src.
```
pjdotson: typo

x/go/telem/series.go
```diff
@@ -127,6 +153,17 @@ func SetValueAt[T types.Numeric](s Series, i int, v T) {
...
+// dst and src must have the same DataType, and that DataType cannot be of variable
+// density.
+func CopyValue(dst, src Series, dstIdx, srcIdx int) {
+	if dst.DataType != src.DataType || dst.DataType.IsVariable() || src.DataType.IsVariable() {
+		panic("cannot copy values from non-variable series")
```
pjdotson: should be two different panics there, one based off of data type mismatch and one based off of variable series

x/go/telem/op/op_test.go
```diff
@@ -0,0 +1,445 @@
...
+				output := telem.Series{DataType: telem.Uint8T}
+
+				op.EqualI32(a, b, &output)
+
+				expected := []uint8{1, 0, 1, 0}
+				Expect(telem.UnmarshalSlice[uint8](output.Data, telem.Uint8T)).To(Equal(expected))
+			})
+		})
+
+		Context("Different Length Series - Last Value Repetition", func() {
```
pjdotson, line 63: Can you explain to me why we want this behavior? it's weird / unintuitive and a little unexpected behavior

x/go/telem/op/op_test.go
```diff
@@ -0,0 +1,445 @@
...
+				output := telem.Series{DataType: telem.Uint8T}
+
+				op.OrU8(a, b, &output)
+
+				expected := []uint8{0xFF, 0xFF, 0xFF}
+				Expect(telem.UnmarshalSlice[uint8](output.Data, telem.Uint8T)).To(Equal(expected))
+			})
+		})
+
+		Context("XOR Operation", func() {
```
pjdotson, line 328: when i see stuff like this it makes me wonder whether we should have a completely different `bool` and `uint8` series.

x/go/telem/op/op_test.go
emilbon99: This is how you handle two series that come in at different rates. Let's say you have a series that is comes in at 1hz and one that comes in at 10Hz. for every 10 values you have one from the other series. the alignment strategy is to repeat the latest value from the other telemetry source. Right so if you're adding ox_pt_1 which comes in at 1hz and fuel_pt_1 which comes in at 10 hz, you'll be adding the last value of ox_pt_1 to the remaining 9 values of fuel_pt_1.

x/go/telem/op/op_test.go
emilbon99: Having a separate bool series would cause dramatically more problems than it would solve.

x/go/testutil/must.go
emilbon99: it's hard to test test assertions

PR #1598, opened 2025-10-28
SY-2986: Views

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-2986](https://linear.app/synnax/issue/SY-2986/views)

## Description

Add a v1 implementation of views to the Core and the Console. Implement views for ranges and statuses.

## Basic Readiness

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR implements a comprehensive Views system for filtering and organizing resources (ranges and statuses) in both the Core backend and Console UI. The implementation follows a clean 3-tier architecture with proper separation between the service layer, API layer, and client code.

**Major Changes:**
- Added complete backend implementation in Go with CRUD operations, search, filtering, and pagination
- Integrated views into the ontology system for proper resource relationships and change tracking
- Implemented access control enforcement on all view operations (create, retrieve, delete)
- Created TypeScript client with Zod schema validation and proper type safety
- Built React UI components using the View.Provider pattern for stateful view management
- Refactored Range and Status explorers to use the new view system
- Added comprehensive test coverage for both backend and frontend

**Implementation Quality:**
- Proper error handling with transaction rollbacks
- Builder pattern for flexible query construction
- Real-time synchronization through Flux store and WebSocket listeners
- Combines static views (like "All Ranges") with user-created views seamlessly
- Access control integrated at the API boundary

**Minor Issues Found:**
- Debug `console.log` statements left in `pluto/src/view/queries.ts` (lines 50, 56)
- Duplicate null check in `console/src/view/Provider.tsx` (line 72)
- Test assertion could be more specific in `view_test.go` (line 223)

- This PR is safe to merge after removing debug statements
- High-quality implementation with proper architecture, comprehensive tests, and access control. Score reduced from 5 to 4 solely due to debug console.log statements that should be removed before production deployment. The core logic is sound, error handling is proper, and the integration across the stack is well-executed.
- pluto/src/view/queries.ts needs debug statements removed (lines 50, 56)

File Analysis

| Filename | Score | Overview |
|----------|-------|----------|
| core/pkg/service/view/service.go | 5/5 | implements service layer with proper dependency injection, config validation, and ontology integration |
| core/pkg/service/view/writer.go | 5/5 | handles create/update/delete operations with proper validation and ontology relationships |
| core/pkg/api/view.go | 5/5 | exposes view operations via API with access control enforcement on all operations |
| client/ts/src/view/client.ts | 5/5 | TypeScript client with proper schema validation and overloaded methods for single/multiple operations |
| pluto/src/view/queries.ts | 4/5 | implements Flux patterns for view state management with listeners and synchronizers; contains debug console.log statements |
| console/src/view/Provider.tsx | 4/5 | main view provider component combining static and remote views with selection state; has duplicate null check |
| console/src/range/Explorer.tsx | 5/5 | updated range explorer to use new View.Provider pattern with filter menu |
| console/src/status/Explorer.tsx | 5/5 | updated status explorer to use new View.Provider pattern with filter menu |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Console as Console UI
    participant Client as TS Client
    participant API as API Layer
    participant Service as View Service
    participant Writer as View Writer
    participant DB as Gorp DB
    participant Ontology as Ontology System

    Note over Console,Ontology: View Creation Flow
    Console->>Client: create(view)
    Client->>API: POST /view/create
    API->>API: Enforce Access Control
    API->>Service: NewWriter(tx)
    Service->>Writer: Create(ctx, view)
    Writer->>Writer: validate(view)
    Writer->>Writer: Generate UUID if needed
    Writer->>DB: gorp.Create().Exec()
    Writer->>Ontology: DefineResource(otgID)
    Writer->>Ontology: DefineRelationship(group, view)
    Writer-->>API: view with key
    API-->>Client: ViewCreateResponse
    Client-->>Console: created view

    Note over Console,Ontology: View Retrieval Flow
    Console->>Client: retrieve({types, searchTerm})
    Client->>API: POST /view/retrieve
    API->>Service: NewRetrieve()
    Service->>Service: Apply filters (types, keys, search)
    alt Has Search Term
        Service->>Ontology: SearchIDs(searchTerm)
        Ontology-->>Service: matching IDs
        Service->>Service: Convert IDs to keys
    end
    Service->>DB: gorp.Retrieve().Exec()
    DB-->>Service: views[]
    Service-->>API: views[]
    API->>API: Enforce Access Control
    API-->>Client: ViewRetrieveResponse
    Client-->>Console: views[]

    Note over Console,Ontology: Real-time Synchronization
    DB->>Service: Change detected
    Service->>Ontology: Publish signal
    Ontology->>Client: WebSocket update
    Client->>Console: Update Flux store
    Console->>Console: Re-render affected components
```

### DIFF
--- a/console/src/view/View.css
+++ b/console/src/view/View.css
@@ -14,8 +14,8 @@
 }
 
 .console-view__views {
-    padding: 0rem 2rem;
-    height: 8rem;
+    padding: 1.5rem 2rem;
+    flex-shrink: 0;
 }
 
 .console-view__view-item {
--- a/console/src/view/Provider.tsx
+++ b/console/src/view/Provider.tsx
@@ -95,20 +95,22 @@ export const Provider = ({ resourceType, children }: ProviderProps): ReactElemen
   );
 
   return (
-    <Context value={contextValue}>
-      <Selector
-        showEditButton={staticViewKeys.includes(selected) ? true : canUpdateView}
-        editable={editable}
-        onEditableClick={() => setEditable((prev) => !prev)}
-        resourceType={resourceType}
-        onFetchMore={handleFetchMore}
-        staticViews={staticViews}
-        onSelect={setSelected}
-        listProps={combinedProps}
-        selected={selected}
-      />
-      <Fragment key={selected}>{children}</Fragment>
-    </Context>
+    <Flex.Box full="y" empty>
+      <Context value={contextValue}>
+        <Selector
+          showEditButton={staticViewKeys.includes(selected) ? true : canUpdateView}
+          editable={editable}
+          onEditableClick={() => setEditable((prev) => !prev)}
+          resourceType={resourceType}
+          onFetchMore={handleFetchMore}
+          staticViews={staticViews}
+          onSelect={setSelected}
+          listProps={combinedProps}
+          selected={selected}
+        />
+        <Fragment key={selected}>{children}</Fragment>
+      </Context>
+    </Flex.Box>
   );
 };
 


pluto/src/hooks/useInactivity.ts
```diff
@@ -0,0 +1,44 @@
...
+// included in the file licenses/APL.txt.
+
+import { type RefObject, useCallback, useEffect, useRef, useState } from "react";
+
+export interface UseInactivityReturn<E extends HTMLElement> {
+  visible: boolean;
+  ref: RefObject<E | null>;
+}
+
+export const useInactivity = <E extends HTMLElement>(
```
emilbon99: This is a problem that could be solved with a few lines of CSS...

console/src/view/View.tsx
```diff
@@ -0,0 +1,328 @@
...
+    <Dialog.Trigger hideCaret tooltip="Filter">
+      <Icon.Filter />
+    </Dialog.Trigger>
+    <Dialog.Dialog
+      background={1}
+      bordered={false}
+      pack={false}
+      style={{ padding: "1rem" }}
+    >
+      <>{children({ request, onRequestChange })}</>
```
emilbon99: You can't render an FC like this, it's illegal react.

console/src/view/View.tsx
```diff
@@ -0,0 +1,328 @@
...
+export interface ViewProps<
+  K extends record.Key,
+  E extends record.Keyed<K>,
+  R extends Request,
+> extends Pick<
+    Flux.UseListReturn<R, K, E>,
+    "data" | "getItem" | "subscribe" | "retrieve"
+  > {
+  onCreate: () => void;
+  filters?: React.FC<FiltersProps<R>>;
```
emilbon99: These can't be FC's and should be passed as render props.

console/src/view/View.tsx
```diff
@@ -0,0 +1,328 @@
...
+  initialRequest: R;
+  shownFilters?: React.FC<FiltersProps<R>>;
+  resourceType: string;
```
emilbon99: Shouldn't `resourceType` be typed as an ontology resource type?

console/src/status/Explorer.tsx
```diff
@@ -7,32 +7,39 @@
...
-      getItem={getItem}
-      subscribe={subscribe}
-      retrieve={retrieve}
-      enableSearch
-      enableFilters
+    <View<status.Key, status.Status, status.MultiRetrieveArgs>
+      {...listProps}
+      resourceType="status"
+      item={item}
+      initialRequest={{}}
```
emilbon99: Passing this as a new object will cause it to rerender every time

console/src/status/Explorer.tsx
```diff
@@ -7,32 +7,39 @@
...
+      filters={FilterContextMenu}
+      shownFilters={CoreFilters}
+      onCreate={() => placeLayout(CREATE_LAYOUT)}
```
emilbon99: Passing this without memoization will cause the entire component to rerender every time

console/src/view/View.tsx
```diff
@@ -0,0 +1,328 @@
...
+}
+
+const Item = (props: ItemProps) => {
+  const { getItem } = List.useUtilContext<string, view.View>();
+  const { update: del } = PView.useDelete();
+  const view = getItem?.(props.itemKey);
+  if (view == null) return null;
+  return (
+    <Flex.Box x pack>
+      <Button.Button level="p" onClick={() => props.onSelectView(view)}>
```
emilbon99: No need for these levels.

emilbon99: Destructure props

console/src/view/View.tsx
```diff
@@ -0,0 +1,328 @@
...
+    message={`No ${plural(resourceType)} created.`}
+    action={`Create a ${resourceType}`}
+    onClick={onCreate}
+  />
+);
+
+interface FiltersContentProps<R extends Request> {
+  request: R;
+  onRequestChange: state.Setter<R>;
+  children: React.FC<FiltersProps<R>>;
```
emilbon99: Don't pass components like this. Use render rpops

console/src/view/View.tsx
emilbon99: No need for a fragment.

console/src/view/View.tsx
```diff
@@ -0,0 +1,328 @@
...
+      bordered={false}
+      pack={false}
+      style={{ padding: "1rem" }}
+    >
+      <>{children({ request, onRequestChange })}</>
+    </Dialog.Dialog>
+  </Dialog.Frame>
+);
+
+interface SearchBoxProps {
```
emilbon99: This component should just extend `Input.TextProps`. `searchTerm` and `onChange` or unnecessary.

console/src/view/services/ontology.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const { icon } = VIEW_REGISTRY[type];
+    if (icon == null) return <Icon.View />;
+    const I = Icon.createComposite(icon, { topRight: Icon.View });
+    compositeIconRegistry[type] = I;
+    return <I />;
+  },
+  hasChildren: false,
+};
+
+const compositeIconRegistry: Record<string, Icon.FC> = {};
```
emilbon99: How is this pattern supposed to work? This is also pretty much the only place in the codebase where we mutate globals.

console/src/view/types.ts
```diff
@@ -0,0 +1,21 @@
...
+import { type Icon } from "@synnaxlabs/pluto";
+
+import { type Layout } from "@/layout";
+
+export interface RegistryItem {
+  icon: Icon.FC;
+  layout: Layout.BaseState;
+}
+
+export interface Registry {
```
emilbon99: This could just be a record.

pluto/src/view/queries.ts
```diff
@@ -0,0 +1,92 @@
...
+    const { types = [], keys = [] } = query;
+    const typesSet = types.length > 0 ? new Set(types) : undefined;
+    const keysSet = keys.length > 0 ? new Set(keys) : undefined;
+    return store.views.get((v) => {
+      if (typesSet != null && !typesSet.has(v.type)) return false;
+      if (keysSet != null && !keysSet.has(v.key)) return false;
+      return true;
+    });
+  },
+  retrieve: async ({ client, query }) => await client.views.retrieve(query),
```
emilbon99: You need to do a cached retrieve here. See thew way other queries are implemented.

pluto/src/view/queries.ts
```diff
@@ -0,0 +1,92 @@
...
+  },
+});
+
+export type DeleteParams = view.Key | view.Key[];
+
+export const { useUpdate: useDelete } = Flux.createUpdate<DeleteParams, FluxSubStore>({
+  name: RESOURCE_NAME,
+  verbs: Flux.DELETE_VERBS,
+  update: async ({ client, data }) => {
+    await client.views.delete(data);
```
emilbon99, line 102: This needs to properly update the store.

pluto/src/icon/Icon.tsx
```diff
@@ -42,7 +42,9 @@ const SUB_POSITIONS: Record<location.CornerXYString, { x: number; y: number }> =
 };
 
 const createSubIcon = (key: location.CornerXYString, Icon: FC): ReactElement | null => {
+  // TODO: fix
   if (Icon == null) return null;
```
emilbon99, line 48: This TODO needs to be addressed

emilbon99 (changes requested):
Main question here is: why is it your preference to follow a pattern of dependency injected, render prop based composition as opposed to react's natural component based composition?

Generally speaking render props are no longer recommended in react. The only place we really should be using them is in places where they are unavoidable i.e. the list items. 

It make much more sense to compose components using context-based structural APIs (such as the refactored select component).  Even things like form fields should be refactored away from render props.

I also think that you're inverting dependencies in an unhealthy way. The current implementation is using a `View` component for everything and then passing dependencies leads to us needing to modify the `View` component in the future whenever we want to make changes to explorers for different items. There will be fundamentally different structures between explorers for different resource types. Why does `View` need to be this god level component that takes in tons of dependency injections. Why can't view be a utility package (much like all of our base components) which you then import into your type specific packages and use as a collection of components (i.e. View.Frame, View.Controls, View.Filters). This is much more sustainable in the long run, as we realized with things like the dialog and the list. 

Composition via using multiple components in a tree is superior to composition via dependency injecting render props.

Not to mention render props can't be memoized efficiently by react.

emilbon99:
1. Where are views for ranges?
2. The border no longer appears on the filter menu.

Screenshot 2025-12-06 at 5 23 11 PM

3. The fact that the search input gets shifted over when you add a filter is strange. Why isn't the filter button on the right side?

4. Should we be displaying the 'create view' button and the '+' button when not in edit mode? That will save you a lot of real estate. 

Screenshot 2025-12-06 at 5 25 20 PM

5. The 'delete' button on a view is something that is almost never clicked? Why is it visible on every view by default. Adds clutter. Make it a context menu item. There is also no confirmation button for deleting a view. 

6. There is no visual separation between the view list and the actual statuses. Why not?

Screenshot 2025-12-06 at 5 28 33 PM

7. Why is the list of views below the filter set? This violates hierarchy. A view applies a particular set of filters therefore the filters should be under the views.

8. See how the schematic chips use text that have a lower visual contrast for the item that is not currently selected? This improves visual distinction and reduces clutter. 

Screenshot 2025-12-06 at 5 29 57 PM
2-7102a97102a4" />

10. If you're worried about the fact that the controls menu overlaps with the statuses, why not make the list of view chips stay appeared but you hide the search and filter dialog. I think it would be pretty common/useful to want to switch the view of statuses that I'm looking at outside of edit mode. This means you also solve the problem with overlapping controls. 

11. Why isn't the 'add view' button next to the list of views and then we have a default view called 'All Statuses' or 'All Ranges'. This would mean you no longer need the weird set of spectacles.

console/src/label/filter/Chips.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { type label } from "@synnaxlabs/client";
+import { Flex, Form, Icon, Label, Tag, Text } from "@synnaxlabs/pluto";
+import { type ReactElement } from "react";
+
+import { View } from "@/view";
+
+export const Chips = (): ReactElement | null => {
+  const { editable } = View.useContext();
+  const { set } = Form.useContext();
+  const hasLabels = Form.useFieldValue<label.Key[]>("query.hasLabels", {
```
emilbon99, line 24: You could use `Form.useField` here then it would come with an onChange handler that you can use instead of the form `set` field.

console/src/label/filter/Chips.tsx
```diff
@@ -0,0 +1,55 @@
...
+      hasLabels.filter((k) => k !== key),
+      { notifyOnChange: true },
+    );
+  };
+  return (
+    <Flex.Box x pack background={0}>
+      <Text.Text
+        bordered
+        size="small"
+        style={{ padding: "0 1rem", boxShadow: "var(--pluto-shadow-v1)" }}
```
emilbon99: Move this into a CSS file.

console/src/label/filter/MenuItem.tsx
```diff
@@ -0,0 +1,34 @@
...
+        value={value}
+        onChange={onChange}
+        location={labelLocation}
+        triggerProps={triggerProps}
+      />
+    )}
+  </Form.Field>
+);
+
+const triggerProps = { hideTags: true, variant: "text" } as const;
```
emilbon99: These should be all caps, properly typed, and moved to be above the `MenuItem` component.

console/src/range/Explorer.tsx
```diff
@@ -25,19 +29,25 @@ export const EXPLORER_LAYOUT: Layout.State = {
...
-    <List
-      data={data}
-      getItem={getItem}
-      subscribe={subscribe}
-      retrieve={retrieve}
-      enableAddButton
-      enableSearch
-      enableFilters
-    />
+    <View.Frame {...listProps} resourceType="range" onCreate={handleCreate}>
```
emilbon99: This API is nice.

console/src/view/FilterMenu.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <Dialog.Dialog background={1} pack={false} style={style}>
+        <Menu.Menu level="small" gap="small">
+          {children}
+        </Menu.Menu>
+      </Dialog.Dialog>
+    </Dialog.Frame>
+  );
+};
+
+const style = { padding: "1rem" } as const;
```
emilbon99: Properly typed, scream case, and put above component. Please go back and find all the places where you might bed oing this.

console/src/view/Frame.tsx
```diff
@@ -0,0 +1,132 @@
...
+    autoSave: true,
+    beforeSave: async ({ value }) => {
+      const { key, query } = value();
+      // type assertion because the current implementation of the query client doesn't
+      // support custom typing yet.
+      retrieve((p) => {
+        const nextQuery = { ...p, ...(query as Q), offset: 0, limit: 25 };
+        return nextQuery;
+      });
+      if (key == null || key === "") return false;
```
emilbon99: These two lines can be condensed to `return primitive.isNonZero`

console/src/view/Frame.tsx
```diff
@@ -0,0 +1,132 @@
...
+      name: "",
+      query: { hasLabels: [] },
+    },
+    autoSave: true,
+    beforeSave: async ({ value }) => {
+      const { key, query } = value();
+      // type assertion because the current implementation of the query client doesn't
+      // support custom typing yet.
+      retrieve((p) => {
+        const nextQuery = { ...p, ...(query as Q), offset: 0, limit: 25 };
```
emilbon99: Condense this

console/src/view/Frame.tsx
```diff
@@ -0,0 +1,132 @@
...
+        return nextQuery;
+      });
+      if (key == null || key === "") return false;
+      return true;
+    },
+  });
+  const canEditView = Access.useUpdateGranted({
+    type: "view",
+    key: form.value().key ?? "",
```
emilbon99: Is there a case where this key would change? If so, using form.value().key is not going to work. Need a field value hook

console/src/view/Frame.tsx
```diff
@@ -0,0 +1,132 @@
...
+  });
+  const [editable, setEditable] = useState(canEditView);
```
emilbon99: What happens if `canEditView` changes?

console/src/view/Frame.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const contextValue = useMemo(
+    () => ({ editable, resourceType, search, save }),
+    [editable, resourceType, search, save],
+  );
+  const canCreate = Access.useCreateGranted({ type: resourceType, key: "" });
```
emilbon99: Create a `TYPE_ONTOLOGY_ID` for views

console/src/view/Frame.tsx
```diff
@@ -0,0 +1,132 @@
...
+              tooltip={`Create a ${resourceType}`}
+            >
+              <Icon.Add />
+            </Button.Button>
+          )}
+          {canEditView && (
+            <Button.Toggle
+              size="small"
+              value={editable}
+              onChange={() => setEditable(!editable)}
```
emilbon99: Small, but would be better practice to use setEditable((p) => !p)

console/src/view/Items.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { plural } from "pluralize";
+import { type ReactElement } from "react";
+
+import { useContext } from "@/view/context";
+
+export interface ItemsProps<K extends record.Key = record.Key>
+  extends List.ItemsProps<K> {}
+
+export const Items = <K extends record.Key>(props: ItemsProps<K>): ReactElement => (
+  <List.Items<K> emptyContent={<DefaultEmptyContent />} grow {...props} />
```
emilbon99: Any reason this can't get put into a variable called `emptyContent` that statically renders the JSX

console/src/view/Search.tsx
```diff
@@ -0,0 +1,38 @@
...
+// included in the file licenses/APL.txt.
+
+import { Icon, Input } from "@synnaxlabs/pluto";
+import { plural } from "pluralize";
+import { type ReactElement, useCallback, useState } from "react";
+
+import { useContext } from "@/view/context";
+
+export const Search = (): ReactElement | null => {
+  const { editable, resourceType, search } = useContext("View.Search");
```
emilbon99: Why doesn't this leverage `Form.useFieldValue`?

console/src/view/Views.tsx
```diff
@@ -0,0 +1,165 @@
...
+      }
+      const v = getItem(view);
+      if (v == null) return;
+      set("", v);
+      save();
+    },
+    [set, reset, getItem],
+  );
+  const { update: del } = View.useDelete();
+  const handleDelete = useCallback(
```
emilbon99: Instead of using `handleError`, use the `beforeUpdate` item on the `View.useDelete` hook. This will do all of the async error handling for you.

console/src/view/Views.tsx
```diff
@@ -0,0 +1,165 @@
...
+        align="center"
+      >
+        {editable && (
+          <Button.Button onClick={handleCreate} tooltip="Create a view" size="small">
+            <Icon.Add />
+          </Button.Button>
+        )}
+        <Menu.ContextMenu
+          {...contextMenuProps}
+          menu={({ keys }) => (
```
emilbon99: There is no other place where we put this menu inlien. We should extract into a separte component IMO

console/src/view/Views.tsx
```diff
@@ -0,0 +1,165 @@
...
+          </List.Items>
+        </Menu.ContextMenu>
+      </Flex.Box>
+    </Select.Frame>
+  );
+};
+
+interface ViewItemProps extends List.ItemProps<view.Key> {}
+
+const ViewItem = ({ itemKey, onContextMenu }: ViewItemProps): ReactElement | null => {
```
emilbon99: Why is this called `ViewItem`?

console/src/view/Views.tsx
```diff
@@ -0,0 +1,165 @@
...
+                Delete
+              </Menu.Item>
+              <Menu.Divider />
+              <CMenu.ReloadConsoleItem />
+            </Menu.Menu>
+          )}
+        >
+          <List.Items<view.Key> x gap="medium">
+            {({ key, ...rest }) => (
+              <ViewItem key={key} {...rest} onContextMenu={contextMenuProps.open} />
```
emilbon99: Why does this need an `onContextMenu` prop? I don't see anywhere else in the codebase where we follow this pattern for context menus. Seems like we always put an `onContextMenu` prop on a parent element.

Generally speaking, not making the value of this render prop a constant i.e. `listItem = Component.renderProp(Item)` is fairly expensive.

console/src/view/Views.tsx
```diff
@@ -0,0 +1,165 @@
...
+  if (query.variant !== "success") return null;
+  return (
+    <Flex.Box pack onContextMenu={onContextMenu}>
+      <Select.Button itemKey={itemKey} size="small" justify="between">
+        <Text.MaybeEditable
+          id={itemKey}
+          value={query.data.name}
+          color={9}
+          onChange={handleRename}
+          style={{ padding: "0rem 1rem" }}
```
emilbon99: Put this into a CSS file

console/src/view/Views.tsx
```diff
@@ -0,0 +1,165 @@
...
+    },
+    [itemKey, rename],
+  );
+  if (query.variant !== "success") return null;
+  return (
+    <Flex.Box pack onContextMenu={onContextMenu}>
+      <Select.Button itemKey={itemKey} size="small" justify="between">
+        <Text.MaybeEditable
+          id={itemKey}
+          value={query.data.name}
```
emilbon99: Destructure

console/src/view/Views.tsx
```diff
@@ -0,0 +1,165 @@
...
+        </Menu.ContextMenu>
+      </Flex.Box>
+    </Select.Frame>
+  );
+};
+
+interface ViewItemProps extends List.ItemProps<view.Key> {}
+
+const ViewItem = ({ itemKey, onContextMenu }: ViewItemProps): ReactElement | null => {
+  const query = View.useRetrieve({ key: itemKey });
```
emilbon99: Why aren't we leveraging `List.useItem` if we already have a parent item query. Would save a lot of extra data fetching and would mean you don't need to handle a query result on everything.

core/pkg/api/grpc/grpc.go
```diff
@@ -121,11 +121,21 @@ func New(channelSvc *channel.Service) (api.Transport, []fgrpc.BindableTransport)
 	a.AccessAssignRole = fnoop.UnaryServer[api.AccessAssignRoleRequest, types.Nil]{}
 	a.AccessUnassignRole = fnoop.UnaryServer[api.AccessUnassignRoleRequest, types.Nil]{}
 
-	// arc
+	// STATUS
+	a.StatusSet = fnoop.UnaryServer[api.StatusSetRequest, api.StatusSetResponse]{}
```
emilbon99: These noops need to be removed.

emilbon99: It's why the C++ tests are failing.

console/src/view/View.css
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Use of this software is governed by the Business Source License included in the file
+ * licenses/BSL.txt.
+ *
+ * As of the Change Date specified in that file, in accordance with the Business Source
+ * License, use of this software will be governed by the Apache License, Version 2.0,
+ * included in the file licenses/APL.txt.
+ */
+
+.console-view {
```
emilbon99: Inline this style and use .console-view__toolbar:not(:empty) {
  /* styles here */
}

core/pkg/service/view/retrieve.go
```diff
@@ -0,0 +1,83 @@
...
+		if err != nil {
+			return err
+		}
+		keys, err := KeysFromOntologyIDs(ids)
+		if err != nil {
+			return err
+		}
+		r = r.WhereKeys(keys...)
+	}
+	if err := r.gorp.Exec(ctx, tx); err != nil {
```
emilbon99: Pointless control flow

core/pkg/service/view/service.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	viewSignals, err := signals.PublishFromGorp(
+		ctx,
+		s.cfg.Signals,
+		signals.GorpPublisherConfigUUID[View](s.cfg.DB),
+	)
+	if err != nil {
+		return nil, err
+	}
+	s.shutdownSignals = xio.MultiCloser{viewSignals}
```
emilbon99: No reason that this needs to be an `xio.MultiCloser`

core/pkg/service/view/writer.go
```diff
@@ -0,0 +1,97 @@
...
+		if err := w.Delete(ctx, key); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (w Writer) validate(v View) error {
+	vld := validate.New("view.View")
+	validate.NotEmptyString(vld, "Name", v.Name)
```
emilbon99: At some point we should decide whether these should be snake case.

pluto/src/button/Button.css
```diff
@@ -78,6 +78,9 @@
         padding: 0 1rem;
     }
     &.pluto-btn--filled {
+        .pluto-text {
```
emilbon99, line 81: Somewhat concerned about side effects of this change.

pluto/src/channel/queries.ts
```diff
@@ -375,7 +375,7 @@ export const useCalculatedForm = Flux.createForm<
   typeof calculatedFormSchema,
   FluxSubStore
 >({
-  name: "Calculated Channel",
+  name: "calculated channel",
```
emilbon99, line 378: Why is `Channel Group` capitalized but `calculated channel` isn't?

pluto/src/ranger/queries.ts
```diff
@@ -25,12 +25,12 @@ export interface AliasFluxStore extends Flux.UnaryStore<ranger.Key, ranger.Alias
 export const RANGE_KV_FLUX_STORE_KEY = "rangeKV";
 export const RANGE_ALIASES_FLUX_STORE_KEY = "rangeAliases";
 
-const RESOURCE_NAME = "Range";
-const PLURAL_RESOURCE_NAME = "Ranges";
-const KV_RESOURCE_NAME = "Meta-Data Item";
-const PLURAL_KV_RESOURCE_NAME = "Meta-Data";
-const PLURAL_CHILDREN_RESOURCE_NAME = "Child Ranges";
-const PARENT_RESOURCE_NAME = "Parent Range";
+const RESOURCE_NAME = "range";
```
emilbon99, line 28: All these name changes should have gone in a separate PR you naughty boy. Practice what you preach!

client/ts/src/view/payload.ts
```diff
@@ -0,0 +1,25 @@
...
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { record } from "@synnaxlabs/x";
+import z from "zod";
+
+export const keyZ = z.uuid();
+export type Key = z.infer<typeof keyZ>;
+
+export const viewZ = z.object({
```
emilbon99, line 16: The `query` type inside of the view should probably accept an optional schema to improve the typing quality. Like we do with tasks. Up to you whether you want to do it in this issue or a different one.

pluto/src/view/queries.ts
```diff
@@ -0,0 +1,174 @@
...
+  FluxSubStore
+>({
+  name: RESOURCE_NAME,
+  retrieve: retrieveSingle,
+  mountListeners: ({ store, query: { key }, onChange }) => [
+    store.views.onSet(onChange, key),
+  ],
+});
+
+const zeroValues = {
```
emilbon99: `SCREAM_CASE`. Also, why isn't our eslint or formatting checker flaggint hat there should be a space between 132 and 133?

pluto/src/view/queries.ts
```diff
@@ -0,0 +1,174 @@
...
+  ],
+});
+
+export interface RenameParams extends Pick<View.View, "key" | "name"> {}
+
+export const { useUpdate: useRename } = Flux.createUpdate<RenameParams, FluxSubStore>({
+  name: RESOURCE_NAME,
+  verbs: Flux.RENAME_VERBS,
+  update: async ({ client, data: { key, name }, store, rollbacks }) => {
+    const v = await retrieveSingle({ client, store, query: { key } });
```
emilbon99, line 168: Should rename the ontology resource here as well.

emilbon99:
Screenshot 2025-12-07 at 3 35 22 PM
When you're out of edit mode and there are no label filters you get a double border.

core/pkg/service/view/writer.go
pjdotson: https://linear.app/synnax/issue/SY-3390/make-sure-all-backend-validations-are-snake-case

client/ts/src/view/payload.ts
pjdotson: Want to do it in a different one, I don't love what we are doing with the tasks because
I believe there are some places where we are still doing type assertions instead of
schema validations. Would love to think this out much more intelligently and find a
sustainable way of doing this for devices, tasks, statuses, and views: https://linear.app/synnax/issue/SY-3391/improve-typescript-schema-validations-for-tasks-views-statuses-and

pluto/src/channel/queries.ts
pjdotson: Because "Channel Group" is the proper name of a specific unique resource (the top level
group that holds all of the channels). "calculated channel" is not a proper name, so it
is lowercase. In general we should use lower case / sentence case
unless for a proper noun.

pluto/src/ranger/queries.ts
pjdotson: You are right about that, sorry...

pluto/src/view/queries.ts
pjdotson: Prettier doesn't have that ability, there is a ESLint Stylistic rule that we could set
up but might take a little configuration and finagling: https://linear.app/synnax/issue/SY-3392/add-padding-line-between-statements-eslint-rule

pluto/src/button/Button.css
pjdotson: what specifically are you concerned about? the color of the text in the view buttons
wasn't changing because up getting wrapped in a `<p />` tag due to the
`<Text.MaybeEditable />`. Is there any other places
we even use this pattern - I would think that normally we don't wrap what comes under a
button in a text tag at all?

console/src/view/Search.tsx
pjdotson: I didn't want to actually update the view for searches, it didn't seem like something
that belonged in the view. I'm thinking that views should mostly be sorting + filtering,
it would be odd if we both had the same view pulled up and you started typing something
to search and that appeared on my screen.

console/src/view/Frame.tsx
pjdotson: is there a way to do that when I am not in the form context yet, or do I have to rewrite
this into two separate components and make the `useUpdateGranted` call in the child component?

console/src/view/Frame.tsx
pjdotson: do we have a way to hook into these changes, IE call `setEditable` whenever
`canEditView` changes without using a `useEffect`?

console/src/view/Frame.tsx
emilbon99: `Form.useField()` and `Form.useFieldValue` both take in a second set of options where one arc is ctx. Example is:
```tsx
  const isIndex = Form.useFieldValue<boolean, boolean, typeof Channel.formSchema>(
    "isIndex",
    { ctx: form },
  );
```

pjdotson:
@greptile /rerun

pjdotson:
@emilbon99 could you look at an issue w/ large number of items before you review? scrollbar for list doesn't appear and I couldn't figure out why.

emilbon99:
I'll take a look now

PR #1604, opened 2025-10-29
SY-3181: Implement Arc Toolbar in Console

emilbon99:
# Issue

[SY-3181](https://linear.app/synnax/issue/SY-3181)

## Description

Implements the Arc toolbar UI component in the Console application, providing a user-facing interface for the Arc programming language integration. This adds a top-level toolbar for Arc-related functionality and reorganizes Arc editor components for better modularity.

## Changes

- Added new `Toolbar` component (`console/src/arc/Toolbar.tsx`) with Arc-specific actions and controls
- Reorganized Arc editor components into `editor/` subdirectory for better separation of concerns
- Added Arc list item component for Arc-related UI elements
- Integrated toolbar with Console's main layout and state management
- Updated Arc module exports and external interfaces

## Basic Readiness

### DIFF
--- a/console/src/arc/Toolbar.tsx
+++ b/console/src/arc/Toolbar.tsx
@@ -27,7 +27,7 @@ import { array } from "@synnaxlabs/x";
 import { useCallback, useMemo, useState } from "react";
 import { useDispatch } from "react-redux";
 
-import { createEditor } from "@/arc/editor";
+import { Editor } from "@/arc/editor";
 import { remove } from "@/arc/slice";
 import { EmptyAction, Menu, Toolbar } from "@/components";
 import { CSS } from "@/css";
@@ -37,7 +37,7 @@ import { Ontology } from "@/ontology";
 
 const EmptyContent = () => {
   const placeLayout = Layout.usePlacer();
-  const handleClick = () => placeLayout(createEditor());
+  const handleClick = () => placeLayout(Editor.create());
   return (
     <EmptyAction
       message="No existing Arcs."
@@ -95,7 +95,7 @@ const Content = () => {
           message: "Failed to open Arc editor",
           description: `Arc with key ${key} not found`,
         });
-      placeLayout(createEditor({ key, name: arc.name }));
+      placeLayout(Editor.create({ key, name: arc.name }));
     },
     [getItem, addStatus, placeLayout],
   );
@@ -108,7 +108,7 @@ const Content = () => {
     handleError(async () => {
       const name = await rename({}, { icon: "Arc", name: "Arc.Create" });
       if (name == null) return;
-      placeLayout(createEditor({ name }));
+      placeLayout(Editor.create({ name }));
     }, "Failed to create arc");
   }, [rename, handleError, placeLayout]);
 
@@ -242,7 +242,7 @@ const ArcListItem = ({ onToggleDeploy, onRename, ...rest }: ArcListItemProps) =>
           />
         </Flex.Box>
         <Text.Text level="small" color={10}>
-          {arc?.status?.message ?? (isDeployed ? "Deployed" : "Not deployed")}
+          {arc?.status?.message ?? (isDeployed ? "Started" : "Stopped")}
         </Text.Text>
       </Flex.Box>
       <Button.Button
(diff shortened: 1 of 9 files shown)


client/ts/src/arc/payload.ts
```diff
@@ -47,20 +47,27 @@ export const keyZ = z.uuid();
...
+  status: statusZ.optional().nullable(),
 });
 
-export interface Arc extends z.infer<typeof arcZ> {
-  status?: Status;
-}
+export interface Arc extends z.infer<typeof arcZ> {}
 
-export const newZ = arcZ.partial({ key: true });
+export const newZ = arcZ.partial({ key: true }).omit({ status: true });
```
pjdotson, line 70: In the future do you think it makes sense that we could also create the status if you pass in the status field?

pluto/src/arc/queries.ts
```diff
@@ -79,8 +108,19 @@ export const useList = Flux.createList<ListQuery, arc.Key, arc.Arc, FluxSubStore
...
+        return { ...arc, status: arc.status ?? p.status };
+      }),
+    ),
     store.arcs.onDelete(onDelete),
+    store.statuses.onSet((status) =>
+      onChange(status.key, (p) => {
+        if (p == null) return p;
+        return { ...p, status: arc.statusZ.parse(status) };
+      }),
+    ),
```
pjdotson, line 123: shouldn't there be a statuses on delete as well?

console/src/arc/editor/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  type CreateArg,
+  createEditor,
+  EDIT_LAYOUT_TYPE,
+  type EditLayoutType,
+  Editor,
+  HAUL_TYPE,
+  Loaded,
+  SELECTABLE,
+} from "@/arc/editor/Editor";
+export { Toolbar } from "@/arc/editor/toolbar";
```
pjdotson: should have an `external.ts` file here to match the rest of the codebase

console/src/layout/types/v7.ts
```diff
@@ -0,0 +1,48 @@
...
+import * as v5 from "@/layout/types/v5";
+
+const VERSION = "7.0.0";
+
+const ZERO_NAV_STATE: NavState = {
+  main: {
+    drawers: {
+      left: {
+        activeItem: null,
+        menuItems: ["channel", "range", "workspace", "device", "task", "user", "arc"],
```
pjdotson, line 24: seems like we should refactor this in the future so that the possible menu items aren't stored in the slice, and just the active menu item.

pluto/src/arc/queries.ts
```diff
@@ -150,3 +190,32 @@ export const { useRetrieve, useRetrieveObservable } = Flux.createRetrieve<
     return [store.arcs.onSet(onChange, query.key)];
   },
 });
+
+export const { useUpdate: useToggleDeploy } = Flux.createUpdate<arc.Key, FluxSubStore>({
+  name: RESOURCE_NAME,
+  verbs: Flux.UPDATE_VERBS,
+  update: async ({ client, store, data: key, rollbacks }) => {
+    const arc = await retrieveSingle({ client, store, query: { key } });
+    const updated = await client.arcs.create({ ...arc, deploy: !arc.deploy });
```
pjdotson, line 199: why do we need to make two API calls here? And why did you decide on this instead of having a command channel like for tasks?

console/src/arc/Toolbar.tsx
```diff
@@ -0,0 +1,336 @@
...
+      {canDeploy && (
+        <PMenu.Item itemKey="deploy">
+          <Icon.Play />
+          Deploy
+        </PMenu.Item>
+      )}
+      {canUndeploy && (
+        <PMenu.Item itemKey="undeploy">
+          <Icon.Pause />
+          Undeploy
```
pjdotson: I feel like calling it "stop" or "pause" makes more sense then "undeploy". And why even call it "deploying" vs just "starting"?

client/ts/src/arc/payload.ts
emilbon99: Not really. The arc status is an internal status so there's no reason it would ever be set externally.

pluto/src/arc/queries.ts
emilbon99: The only time an arc status will get deleted is if the arc itself is deleted.

pluto/src/arc/queries.ts
emilbon99: Because it results in substantially less backend code. I think we're going to need to rethink/refine our command strategy in the future. This is also not two API hits in the vast majority of cases, as retrieveSingle will already have a cached arc inside.

console/src/arc/Toolbar.tsx
emilbon99: Renamed

console/src/layout/types/v7.ts
emilbon99: Yeah, the original intention behind this was to allow users to move the toolbars between different locations.

PR #1605, opened 2025-10-29
SY-3195: Fix Undefined Arc Multiply Symbol

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-3195](https://linear.app/synnax/issue/SY-3195/arc-multiply-symbol-is-undefined)

## Description

Fixes naming across symbols in the arc registry.

## Basic Readiness

### DIFF
--- a/client/py/examples/dev/calc_channel_stress.py
+++ b/client/py/examples/dev/calc_channel_stress.py
@@ -73,7 +73,6 @@ calc_avg_sum_div_50_sine = client.channels.create(
     expression="return ("
     + " + ".join([f"sine_50_channel_{i + 1}" for i in range(50)])
     + ") / 50.0",
-    requires=[ch.key for ch in sine_50_channels],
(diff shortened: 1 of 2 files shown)


arc/runtime/op/registry.go
```diff
@@ -112,7 +112,7 @@ var (
 			telem.Uint16T:  op.AddU16,
 			telem.Uint8T:   op.AddU8,
 		},
-		"sub": {
+		"subtract": {
```
pjdotson, line 115: These keys should be consistent - if we are fully writing out subtract and multiply, it doesn't make sense to have greater than be "gt".

emilbon99: Yes, but unfortunately we already have arcs in production that have this issue. So if we want to change this we'll have to build migrations.

emilbon99:
That is what the tests that are running arc are checking.

PR #1609, opened 2025-10-30
SY-3204: Support Deeply Nested Calculations in Iterators

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-3204](https://linear.app/synnax/issue/SY-3204/fix-nested-calculations-in-iterators)

## Basic Readiness

### DIFF
--- a/core/pkg/service/framer/calculation/service_test.go
+++ b/core/pkg/service/framer/calculation/service_test.go
@@ -83,7 +83,7 @@ var _ = Describe("Calculation", Ordered, func() {
 		streamer := MustSucceed(
 			dist.Framer.NewStreamer(
 				ctx,
-				framer.StreamerConfig{Keys: []channel.Key{streamKeys[1]}, SendOpenAck: config.True()},
+				framer.StreamerConfig{Keys: streamKeys, SendOpenAck: config.True()},
 			),
 		)
 		_, sOutlet := confluence.Attach(streamer, 1, 1)


core/pkg/service/framer/iterator/iterator_test.go
```diff
@@ -205,6 +205,270 @@ var _ = Describe("StreamIterator", Ordered, func() {
...
+					// Create calc_circ_b that depends on calc_circ_a (doesn't exist yet, but Arc allows it)
+					calcB := &channel.Channel{
+						Name:       "calc_circ_b",
+						DataType:   telem.Float32T,
+						Expression: "return calc_circ_a + 1",
+					}
+					Expect(dist.Channel.Create(ctx, calcB)).To(Succeed())
+
+					// Create calc_circ_a that depends on calc_circ_b (creating the cycle)
+					calcA := &channel.Channel{
```
pjdotson, line 396: should this fail on creation or on opening the iterator? would be good to give the user the feedback earlier but I don't know how complicated it would be to do that

emilbon99: I agree that it should error on creating the channel, but implementing that would be pretty complex ATM, and I think we should delay that until next release. Issue: https://linear.app/synnax/issue/SY-3208/error-on-circular-calculation-dependencies-on-channel-creation

PR #1617, opened 2025-10-31
SY-3205: Fix Freighter/X Cpp ASan Failures (MacOS & Windows)

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-3205](https://linear.app/synnax/issue/SY-3205/asan-test-failures-windows-and-macos)

## Description
* Remove ASan from Windows Freighter and X C++ tests
* Fix some tests
* Suppress warnings on external deps that were treated as errors in CI
* Remove no-cache test results flag
* Run tests on all OSes when triggered manually via workflow dispatch

## Basic Readiness

### DIFF
--- a/x/cpp/xtest/xtest.h
+++ b/x/cpp/xtest/xtest.h
@@ -356,7 +356,7 @@ inline void eventually_nil(
 /// @param line The source line number
 /// @return The first element of the pair (the result value) if successful
 template<typename Pair>
-auto assert_nil_p_helper(Pair &&pair_result, const char *file, int line) ->
+auto assert_nil_p(Pair &&pair_result, const char *file, const int line) ->
     typename std::remove_reference<decltype(pair_result.first)>::type {
     if (pair_result.second) {
         ADD_FAILURE_AT(file, line)


.github/workflows/test.freighter.cpp.yaml
```diff
@@ -75,16 +75,19 @@ jobs:
...
         if: matrix.os == 'ubuntu-latest'
-        run:
-          bazel test --config=asan-leak --test_output=all //freighter/cpp/...
-          --nocache_test_results
+        run: bazel test --config=asan-leak --test_output=all //freighter/cpp/...
+        shell: bash
+
+      - name: Test (Windows)
+        if: matrix.os == 'windows-latest'
+        run: bazel test --test_output=all //freighter/cpp/...
```
pjdotson, line 92: why don't we run config=asan with Windows?

.github/workflows/test.x.cpp.yaml
```diff
@@ -73,15 +73,19 @@ jobs:
...
       - name: Test (Ubuntu)
         if: matrix.os == 'ubuntu-latest'
-        run:
-          bazel test --config=asan-leak --test_output=all //x/cpp/...
-          --nocache_test_results
+        run: bazel test --config=asan-leak --test_output=all //x/cpp/...
+        shell: bash
+
+      - name: Test (Windows)
+        if: matrix.os == 'windows-latest'
```
pjdotson, line 89: same comment as above

.github/workflows/test.x.cpp.yaml
```diff
@@ -44,11 +44,11 @@ jobs:
...
           - os:
-              ${{ (github.ref != 'refs/heads/main' && github.base_ref != 'main') &&
-              'macos-latest' }}
+              ${{ github.event_name != 'workflow_dispatch' && (github.ref !=
+              'refs/heads/main' && github.base_ref != 'main') && 'macos-latest' }}
           - os:
-              ${{ (github.ref != 'refs/heads/main' && github.base_ref != 'main') &&
-              'windows-latest' }}
+              ${{ github.event_name != 'workflow_dispatch' && (github.ref !=
+              'refs/heads/main' && github.base_ref != 'main') && 'windows-latest' }}
```
pjdotson, line 51: Can you make an issue to edit this for all CI files?

sy-nico: [SY-3238](https://linear.app/synnax/issue/SY-3238/update-workflow-dispatch-manual-triggerbehavior)

.github/workflows/test.freighter.cpp.yaml
sy-nico: I attempted to enable ASAN on Windows but encountered several blockers that made it impractical:

1. MSVC vs Clang toolchain: GitHub Actions Windows runners use MSVC by default. MSVC's /fsanitize=address support is less mature than Clang's -fsanitize=address, and Bazel's integration is more complex.

2. Configuration conflicts: Mixing MSVC-specific ASAN flags with the existing GCC/Clang configuration in .bazelrc required extensive workarounds (see commits da5df3e and 8f760c1 where I attempted this).

3. Runtime requirements: MSVC ASAN requires specific runtime DLLs that must be properly located, adding deployment complexity to CI.

Alternative considered but not pursued: Using Clang-cl (Clang with MSVC-compatible interface) instead of MSVC would enable better ASAN support, but would require reconfiguring Bazel's C++ toolchain for Windows, which is a larger scope change.

Current approach: Tests still run on Windows without ASAN, providing basic correctness validation. Linux (with LSan) and macOS (with ASAN) provide memory error detection, which catches the vast majority of issues since the codebase is cross-platform.

[SY-3240](https://linear.app/synnax/issue/SY-3240/consider-clang-cl-for-windows-builds)

.github/workflows/test.x.cpp.yaml
sy-nico: See above

sy-nico:
@pjdotson , Not going to worry about the failing integration test builds. That is being resolved elsewhere.

x/cpp/xtest/xtest_test.cpp
```diff
@@ -107,7 +107,9 @@ TEST_F(XTestTest, TestMustSucceedSuccess) {
     auto successful_op = []() -> std::pair<int, xerrors::Error> {
         return {42, xerrors::NIL};
     };
-    EXPECT_EQ(ASSERT_NIL_P(successful_op()), 42);
```
emilbon99, line 110: What's the reasoning behind this change? ASSERT_NIL_P already does the error check wiht nice logging

emilbon99: Also, the whole point of this file is to test the macro, so if you get rid of the macro the test is useless

PR #2092, opened 2026-03-15
SY-3919: Schematic Off-Page Navigation

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-3919](https://linear.app/synnax/issue/SY-3919/schematic-page-node-hyperlinking)

## Basic Readiness

<h3>Greptile Summary</h3>

This PR implements off-page reference navigation for schematics, allowing users to click (single or double) an `OffPageReference` symbol to open a linked schematic in a new tab. It adds a `useRetrieveChildren` workspace hook that walks the ontology graph to discover sibling schematics, a redesigned `OffPageReferenceForm` with page/click-mode/color controls, contrast-aware CSS variable rendering for the symbol, and an end-to-end 3-schematic navigation cycle in the integration suite.

Safe to merge; only P2 findings, well-tested across unit, component, and integration layers.

No P0 or P1 findings. The `findWorkspaceAncestor` early-return issue is a theoretical edge case that doesn't affect today's single-parent ontology structure, and the `ctx` closure note in `useHandlePageChange` is informational. Test coverage is solid.

pluto/src/workspace/queries.ts — `findWorkspaceAncestor` early-return logic worth hardening before the ontology model allows multi-parent resources.

| Filename | Overview |
|----------|----------|
| console/src/schematic/Schematic.tsx | Adds `useHandleNodeClickAction` hook that handles both single- and double-click navigation for off-page reference nodes, with proper guard against edit mode and null client |
| pluto/src/workspace/queries.ts | Adds `useRetrieveChildren` hook using recursive `collectChildren` and `findWorkspaceAncestor`; `findWorkspaceAncestor` has a subtle early-return bug when a resource has multiple group-type parents |
| pluto/src/schematic/symbol/Forms.tsx | Rewrites `OffPageReferenceForm` to include page selection dropdown (siblings via `useRetrieveChildren`), click mode toggle, and auto-sets color to primary on first page selection |
| pluto/src/schematic/symbol/Primitives.tsx | Switches off-page reference rendering to CSS variables `--off-page-color` and `--off-page-text-color` for contrast-aware text, and moves `HandleBoundary` after the SVG filter def block |
| pluto/src/schematic/symbol/Symbols.tsx | Adds `page` and `dblClickNav` props to `OffPageReferenceProps`; adds `--linked` modifier class and tooltip via `offPageReferenceTooltip` helper |
| pluto/src/vis/diagram/Diagram.tsx | Exposes `onNodeClick` and `onNodeDoubleClick` from ReactFlow props through `DiagramProps` interface |
| integration/console/schematic/off_page_reference.py | New `OffPageReference` integration test symbol with `set_label`, `set_properties` (page dropdown), and `double_click` methods |
| integration/tests/console/schematic/schematic_lifecycle.py | Adds `test_off_page_navigation` that creates a 3-schematic navigation cycle and verifies tab-switching behavior; cleanup is handled via `off_page_schematic_names` teardown list |
| pluto/src/workspace/queries.spec.ts | Comprehensive tests for `useRetrieveChildren`: single type, multiple types, group-nested, deeply-nested, empty results, and no-client cases |
| console/src/schematic/slice.spec.ts | Adds off-page reference unit tests covering prop storage, update, clearing, and prop preservation via `selectNodeProps` |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant U as User
    participant D as Diagram (ReactFlow)
    participant H as useHandleNodeClickAction
    participant S as Redux Store
    participant Q as useRetrieveChildren
    participant C as Synnax Client
    participant L as Layout.Placer

    U->>D: click / double-click node
    D->>H: onNodeClick / onNodeDoubleClick(nodeId)
    H->>S: selectNodeProps(layoutKey, nodeId)
    S-->>H: props {key, page, dblClickNav}
    H->>H: guard: editable? retrieve null? wrong click mode?
    H->>C: retrieveSingle({ key: page })
    C-->>H: Schematic data
    H->>L: placeLayout(create({...data}))
    L-->>U: opens schematic tab

    Note over Q,C: Form side - sibling discovery
    Q->>C: ontology.retrieveParents(schematicID)
    C-->>Q: workspace parent
    Q->>C: ontology.retrieveChildren(workspaceID, {types:[schematic,group]})
    C-->>Q: sibling list (excl. self)
    Q-->>D: siblings[] for Page dropdown
```

1. `console/src/schematic/Schematic.tsx`, line 51 ([link](https://github.com/synnaxlabs/synnax/blob/8d0bdfa7e1237471e84a19c5c0616ae0198fd42b/console/src/schematic/Schematic.tsx#L51)) 

   **Broad `any` type on `schematics.retrieve` return value**

   The return type of `client.schematics.retrieve` is declared as `Promise<any>`, and `s.data` is spread without any shape validation. If the server response shape ever changes (or the retrieve fails partially, returning a partial object), the resulting `create(...)` call could silently produce an invalid layout state.

   Consider typing the minimal expected shape of `s` (e.g., `{ key: string; name: string; data: Record<string, unknown> }`) so TypeScript can surface any future API drift at compile time.
</details>

<sub>Reviews (6): Last reviewed commit: ["SY-3919: Fix search bug and formatting"](https://github.com/synnaxlabs/synnax/commit/7119078b94cedc861b833e67fe4949d7c58ca339) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=25189815)</sub>


pluto/src/schematic/queries.ts
```diff
@@ -143,6 +146,27 @@ export const { useUpdate: useSnapshot } = Flux.createUpdate<
...
+      setPages(
+        resources
+          .filter((r) => r.id.key !== exclude)
+          .map((r) => ({ key: r.id.key, name: r.name })),
+      );
+    },
+    [client, exclude],
+  );
+  return pages;
+};
```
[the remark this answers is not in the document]
sy-nico: It's a "theoretically stale" observation, not a practical bug. The existing Flux.createUpdate patterns they reference are for data that's displayed persistently (like the workspace resource tree), not for a transient dropdown that remounts frequently.

The tests added account for this and are passing.

core/pkg/service/workspace/writer.go
```diff
@@ -98,3 +112,31 @@ func (w Writer) Delete(
...
+		if err != nil {
+			return err
+		}
+		childKeys = append(childKeys, k)
+	}
+	for _, deleter := range w.childDeleters {
+		if err := deleter(ctx, w.tx, childKeys...); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
sy-nico: Refactored `ChildDeleter` to a struct with an `ontology.Type` field. `deleteChildren` now groups children by type and only passes matching keys to each deleter. Also added a mixed-type cascade test (schematic + line plot in the same workspace).

core/pkg/service/layer.go
```diff
@@ -252,6 +254,30 @@ func OpenLayer(ctx context.Context, cfgs ...LayerConfig) (l *Layer, err error) {
...
+		Delete: func(ctx context.Context, tx gorp.Tx, keys ...uuid.UUID) error {
+			return l.Log.NewWriter(tx).Delete(ctx, keys...)
+		},
+	})
+	l.Workspace.RegisterChildDeleter(workspace.ChildDeleter{
+		Type: table.OntologyType,
+		Delete: func(ctx context.Context, tx gorp.Tx, keys ...uuid.UUID) error {
+			return l.Table.NewWriter(tx).Delete(ctx, keys...)
+		},
+	})
```
pjdotson: a few comments on this workspace child deleter stuff:

1) `workspace.ChildDeleter` should probably be an interface rather than a struct, an
   ideally one that we can just make the `schematic.Service`, `table.Service`, etc.
   satisfy. Then, they are registered when the service is opened.
2) what about groups that are children of workspaces?
3) any worries about the same visualization being shared across multiple workspaces?

pluto/src/schematic/symbol/Forms.tsx
```diff
@@ -1120,15 +1123,49 @@ export const TextBoxForm = (): ReactElement => {
...
+        <Form.Field<Text.Level>
+          hideIfNull
+          path="label.level"
+          label="Label Size"
+          padHelpText={false}
+        >
+          {({ value, onChange }) => (
+            <Select.Text.Level value={value} onChange={onChange} />
+          )}
+        </Form.Field>
```
pjdotson, line 1258: should make this render prop a constant

pluto/src/schematic/symbol/Forms.tsx
```diff
@@ -1120,15 +1123,49 @@ export const TextBoxForm = (): ReactElement => {
...
-    </Flex.Box>
-    <OrientationControl path="" hideOuter />
-  </FormWrapper>
-);
+export const OffPageReferenceForm = ({ layoutKey }: SymbolFormProps): ReactElement => {
+  const pages = usePages(layoutKey);
+  return (
+    <FormWrapper x align="stretch">
+      <Flex.Box x grow align="stretch">
+        <Form.Field<string> path="label.label" label="Label" padHelpText={false} grow>
```
pjdotson: should use `Form.TextField`

pluto/src/schematic/symbol/Forms.tsx
```diff
@@ -1120,15 +1123,49 @@ export const TextBoxForm = (): ReactElement => {
...
+        <Form.Field<string>
+          path="page"
+          label="Page"
+          padHelpText={false}
+          hideIfNull={false}
+          grow
+          className={CSS.BE("symbol-form", "page-field")}
+        >
+          {({ value, onChange }) => (
+            <Select.Static
```
pjdotson, line 1232: this isn't static, this should be dynamic, right?

pluto/src/schematic/queries.ts
```diff
@@ -143,6 +146,35 @@ export const { useUpdate: useSnapshot } = Flux.createUpdate<
   },
 });
 
+export type Page = record.KeyedNamed;
+
+export const usePages = (exclude?: string): Page[] => {
```
pjdotson: this should be leveraging the existing `Flux.Retrieve` architecture

console/src/schematic/Schematic.tsx
```diff
@@ -76,6 +80,46 @@ import { Workspace } from "@/workspace";
...
 
+export const navigateToLinkedSchematic = async (
+  client: { schematics: { retrieve: (arg: { key: string }) => Promise<any> } },
+  page: string,
+  placeLayout: Layout.Placer,
+  addStatus: Status.Adder,
+  label?: string,
+): Promise<void> => {
+  try {
+    const s = await client.schematics.retrieve({ key: page });
```
pjdotson: we shouldn't be hitting the `client` retrieves raw, we should be using Flux

console/src/schematic/Schematic.tsx
```diff
@@ -76,6 +80,46 @@ import { Workspace } from "@/workspace";
...
+  addStatus: Status.Adder;
+}
+
+export const handleNodeDoubleClickAction = (args: HandleNodeDoubleClickArgs): void => {
+  const { editable, client, storeState, layoutKey, nodeId, placeLayout, addStatus } =
+    args;
+  if (editable || client == null) return;
+  const props = selectNodeProps(storeState, layoutKey, nodeId);
+  if (props?.key !== "offPageReference" || !props.page) return;
+  const page = props.page as string;
```
pjdotson: use tests (typeof props.page) instead of type assertions

console/src/schematic/Schematic.tsx
```diff
@@ -76,6 +80,46 @@ import { Workspace } from "@/workspace";
...
+  const label = (props.label as { label?: string } | undefined)?.label;
+  void navigateToLinkedSchematic(client, page, placeLayout, addStatus, label);
```
pjdotson: should be using `handleError` for stuff like this instead of calling `void`

pjdotson (changes requested):
- the off page reference should have a little better highlighting / grayed in when we hover over it (should probably look more like a button)
- you should only have to click the off page refernce once
- capitalization on "Workspace schematic" in select is weird, i.e. "Search Workspace schematics..."

pluto/src/schematic/symbol/Forms.tsx
sy-nico: `Select.Static` is correct here. "static" refers to the options being passed in via the `data` prop rather than the Select fetching its own data. The actual data is fetched dynamically by the `usePages` hook and passed in as `data={pages}`.

pluto/src/schematic/queries.ts
sy-nico: I don't know if I agree with this with my limited ts dev knowledge. But the pages list is effectively static in practice. Once a workspace is set up, users aren't continuously adding/removing schematics while configuring off-page references. `Flux.Retrieve` would add real-time subscription overhead that isn't needed for this use case.

console/src/schematic/Schematic.tsx
sy-nico: This is a one-shot retrieval in a non-hook context (can't use `useRetrieve`), and errors are already handled with `addStatus`. It seemed like overkill for a single navigation fetch...
(See previous comment)

core/pkg/service/layer.go
sy-nico: 1. Refactored `ChildDeleter` to an interface with `Type()` and 
   `DeleteChildren()`. `schematic.Service`, `lineplot.Service`, `log.Service`, 
   and `table.Service` now implement it directly, and `layer.go` just passes 
   the services.

2. Groups are a distribution-layer resource, not service-layer like the 
   visualization types. Adding workspace-aware deletion to `group.Service` 
   would cross the distribution/service layer boundary, so I would say out of scope for this change.

3. `usePages` already scopes the dropdown to the current workspace's 
   schematics, so the UI naturally discourages cross-workspace references. 
   No server-side guard currently, but in practice each workspace is a 
   self-contained environment. Same thing in terms of changes there. I would cut another issue and have to think about this more.

core/pkg/service/workspace/writer_test.go
```diff
@@ -49,13 +51,45 @@ var _ = Describe("Writer", func() {
...
+			s2 := schematic.Schematic{Name: "schematic_2", Data: "{}"}
+			Expect(schematicSvc.NewWriter(tx).Create(ctx, ws.Key, &s2)).To(Succeed())
+
+			Expect(svc.NewWriter(tx).Delete(ctx, ws.Key)).To(Succeed())
+
+			var res schematic.Schematic
+			Expect(gorp.NewRetrieve[uuid.UUID, schematic.Schematic]().WhereKeys(s1.Key).Entry(&res).Exec(ctx, tx)).ToNot(Succeed())
+			Expect(gorp.NewRetrieve[uuid.UUID, schematic.Schematic]().WhereKeys(s2.Key).Entry(&res).Exec(ctx, tx)).ToNot(Succeed())
+		})
+		It("Should cascade delete mixed resource types", func() {
```
pjdotson: add tests for children under a group under a workspace

core/pkg/service/lineplot/service.go
```diff
@@ -67,6 +69,11 @@ func NewService(cfgs ...ServiceConfig) (*Service, error) {
 	return s, nil
 }
 
+// DeleteChildren implements workspace.ChildDeleter.
+func (s *Service) DeleteChildren(ctx context.Context, tx gorp.Tx, keys ...uuid.UUID) error {
```
pjdotson: this method name probably makes more sense as `Delete`, from the point of view of the
line plot service, this doesn't have anything to do with children.

core/pkg/service/workspace/service.go
```diff
@@ -86,11 +87,21 @@ func OpenService(ctx context.Context, configs ...ServiceConfig) (*Service, error
 
 func (s *Service) Close() error { return s.shutdownSignals.Close() }
 
+// RegisterChildDeleter registers a deleter that will be called when a
+// workspace is deleted. This allows child services (schematics, line plots,
+// etc.) to clean up their records without circular imports.
+func (s *Service) RegisterChildDeleter(d ChildDeleter) {
```
pjdotson: instead of having the `RegisterChildDeleter` method, just have them be part of the
workspace service configuration.

core/pkg/service/workspace/writer.go
```diff
@@ -98,3 +118,37 @@ func (w Writer) Delete(
...
+		if err != nil {
+			return err
+		}
+		byType[child.ID.Type] = append(byType[child.ID.Type], k)
+	}
+	// Children with no registered deleter are silently skipped. Ensure a
+	// ChildDeleter is registered for every type that can be a workspace child.
+	for _, deleter := range w.childDeleters {
+		keys, ok := byType[deleter.Type()]
+		if !ok || len(keys) == 0 {
```
pjdotson: don't really need `ok` since `keys` will be `nil` if the type isn't in there

integration/tests/console/schematic/schematic_lifecycle.py
```diff
@@ -368,6 +373,79 @@ def _test_snapshot_rename_synchronization(self) -> None:
...
+        schematic.disable_edit()
+        schematic.fit_view()
+        self._dblclick_node(page, f"Go to {names[0]}")
+
+        # Verify we navigated back to snap
+        self.console.layout.wait_for_tab(names[0])
+        assert self.console.layout.get_tab(
+            names[0]
+        ).is_visible(), f"Expected to navigate back to '{names[0]}'"
+        self.log("Off-page navigation cycle verified")
```
pjdotson, line 443: should be asserting after every click, for instance if the navigation is completely broken
and we never moved tabs this test could pass.

console/src/schematic/Schematic.tsx
```diff
@@ -76,6 +80,56 @@ import { Workspace } from "@/workspace";
 
 export const HAUL_TYPE = "schematic-element";
 
+export const navigateToLinkedSchematic = async (
+  client: { schematics: { retrieve: (arg: { key: string }) => Promise<any> } },
+  page: string,
+  placeLayout: Layout.Placer,
+): Promise<void> => {
+  const s = await client.schematics.retrieve({ key: page });
```
pjdotson: should be using Flux instead of these `client.schematics.retrieve` calls

pluto/src/schematic/symbol/Symbols.tsx
```diff
@@ -1123,26 +1123,40 @@ export const Light = ({
   );
 };
 
+export const offPageReferenceTooltip = (
+  page?: string,
+  dblClickNav?: boolean,
+): string | undefined => {
+  if (page == null || page.length === 0) return undefined;
+  const mode = dblClickNav !== false ? "Double" : "Single";
+  return `${mode}-click to navigate`;
```
pjdotson, line 1136: why give an option, why not just always be single click?

pjdotson (changes requested):

pluto/src/schematic/symbol/Symbols.tsx
sy-nico: Because operators will have a a preference. You wouldn't believe the amount of conversations that are had and standardizations across departments and programs. Each set of operators will have their explicit reasons for preferring one over the other. This is staying as-is, but let me know if you want a rename or a different style.

console/src/schematic/Schematic.tsx
sy-nico: See replies on [this thread](https://github.com/synnaxlabs/synnax/pull/2092#discussion_r2949992289) and [this thread](https://github.com/synnaxlabs/synnax/pull/2092#discussion_r2942827316). Same reasoning applies here.

core/pkg/service/workspace/service.go
sy-nico: Moved to `ServiceConfig.ChildDeleters`. Child services are opened before workspace so they can be passed in the config.

core/pkg/service/workspace/writer.go
sy-nico: fixed

integration/tests/console/schematic/schematic_lifecycle.py
sy-nico: `wait_for_tab` calls `wait_for(state="visible")`  which raises a `TimeoutError` if the tab doesn't appear. Since the schematics are closed after creation, the tabs only reappear if navigation actually works. But I've added explicit assert is_visible() after each step for clarity.

core/pkg/service/lineplot/writer_test.go
```diff
@@ -28,6 +28,16 @@ var _ = Describe("Writer", func() {
...
 		})
 	})
+	Describe("Service Delete", func() {
+		It("Should delete a LinePlot via the service", func() {
+			plot := lineplot.LinePlot{Name: "test", Data: "data"}
+			Expect(svc.NewWriter(tx).Create(ctx, ws.Key, &plot)).To(Succeed())
+			Expect(svc.Delete(ctx, tx, plot.Key)).To(Succeed())
+			var res lineplot.LinePlot
+			Expect(gorp.NewRetrieve[uuid.UUID, lineplot.LinePlot]().
+				WhereKeys(plot.Key).Entry(&res).Exec(ctx, tx)).ToNot(Succeed())
```
pjdotson: should test against `query.ErrNotFound`

core/pkg/service/log/writer_test.go
```diff
@@ -28,6 +28,16 @@ var _ = Describe("Writer", func() {
...
 		})
 	})
+	Describe("Service Delete", func() {
+		It("Should delete a Log via the service", func() {
+			l := log.Log{Name: "test", Data: "data"}
+			Expect(svc.NewWriter(tx).Create(ctx, ws.Key, &l)).To(Succeed())
+			Expect(svc.Delete(ctx, tx, l.Key)).To(Succeed())
+			var res log.Log
+			Expect(gorp.NewRetrieve[uuid.UUID, log.Log]().
+				WhereKeys(l.Key).Entry(&res).Exec(ctx, tx)).ToNot(Succeed())
```
pjdotson: should test against `query.ErrNotFound`

core/pkg/service/schematic/writer_test.go
```diff
@@ -53,6 +53,16 @@ var _ = Describe("Writer", func() {
...
 	})
 
+	Describe("Service Delete", func() {
+		It("Should delete a Schematic via the service", func() {
+			s := schematic.Schematic{Name: "test", Data: "data"}
+			Expect(svc.NewWriter(tx).Create(ctx, ws.Key, &s)).To(Succeed())
+			Expect(svc.Delete(ctx, tx, s.Key)).To(Succeed())
+			var res schematic.Schematic
+			Expect(gorp.NewRetrieve[uuid.UUID, schematic.Schematic]().
+				WhereKeys(s.Key).Entry(&res).Exec(ctx, tx)).ToNot(Succeed())
```
pjdotson: should test against `query.ErrNotFound`

core/pkg/service/table/writer_test.go
```diff
@@ -28,6 +28,16 @@ var _ = Describe("Writer", func() {
...
 		})
 	})
+	Describe("Service Delete", func() {
+		It("Should delete a Table via the service", func() {
+			t := table.Table{Name: "test", Data: "data"}
+			Expect(svc.NewWriter(tx).Create(ctx, ws.Key, &t)).To(Succeed())
+			Expect(svc.Delete(ctx, tx, t.Key)).To(Succeed())
+			var res table.Table
+			Expect(gorp.NewRetrieve[uuid.UUID, table.Table]().
+				WhereKeys(t.Key).Entry(&res).Exec(ctx, tx)).ToNot(Succeed())
```
pjdotson: should test against `query.ErrNotFound`

core/pkg/service/workspace/service.go
```diff
@@ -42,6 +43,7 @@ func (c ServiceConfig) Override(other ServiceConfig) ServiceConfig {
 	c.Ontology = override.Nil(c.Ontology, other.Ontology)
 	c.Group = override.Nil(c.Group, other.Group)
 	c.Signals = override.Nil(c.Signals, other.Signals)
+	c.ChildDeleters = override.Nil(c.ChildDeleters, other.ChildDeleters)
```
pjdotson: should use `override.Slice`

core/pkg/service/workspace/writer_test.go
```diff
@@ -49,13 +55,109 @@ var _ = Describe("Writer", func() {
...
+				schematic.OntologyID(s.Key),
+			)).To(Succeed())
+
+			Expect(svc.NewWriter(tx).Delete(ctx, ws.Key)).To(Succeed())
+
+			// The schematic is a grandchild (workspace -> group -> schematic),
+			// so it should still exist since deleteChildren only traverses
+			// direct children.
+			var sRes schematic.Schematic
+			Expect(gorp.NewRetrieve[uuid.UUID, schematic.Schematic]().WhereKeys(s.Key).Entry(&sRes).Exec(ctx, tx)).To(Succeed())
```
pjdotson: Isn't this the exact opposite behavior we want? now we still have the same problem of
floating groups.

core/pkg/service/workspace/writer_test.go
```diff
@@ -49,13 +55,109 @@ var _ = Describe("Writer", func() {
...
+
+			s1 := schematic.Schematic{Name: "schematic_1", Data: "{}"}
+			Expect(schematicSvc.NewWriter(tx).Create(ctx, ws.Key, &s1)).To(Succeed())
+			s2 := schematic.Schematic{Name: "schematic_2", Data: "{}"}
+			Expect(schematicSvc.NewWriter(tx).Create(ctx, ws.Key, &s2)).To(Succeed())
+
+			Expect(svc.NewWriter(tx).Delete(ctx, ws.Key)).To(Succeed())
+
+			var res schematic.Schematic
+			Expect(gorp.NewRetrieve[uuid.UUID, schematic.Schematic]().WhereKeys(s1.Key).Entry(&res).Exec(ctx, tx)).ToNot(Succeed())
```
pjdotson: for all of these `ToNot(Succeed())` you should assert against the specific error types

core/pkg/service/workspace/service.go
```diff
@@ -25,10 +25,11 @@ import (
...
 type ServiceConfig struct {
-	Signals  *signals.Provider
-	DB       *gorp.DB
-	Ontology *ontology.Ontology
-	Group    *group.Service
+	Signals       *signals.Provider
+	DB            *gorp.DB
+	Ontology      *ontology.Ontology
+	Group         *group.Service
+	ChildDeleters []ChildDeleter
```
pjdotson: FWIW the changes in `core` and the Console changes probably belong in two separate PRs.
Up to you if you want to split but smaller PRs are better because it decreases
integration time / distance between the code on a feature branch and `rc`, makes PRs
easier to review as they are smaller, and means that if one half is a lot more
problematic than the other half it can get in and benefits from that PR can be used by
everyone else.

pluto/src/schematic/queries.ts
pjdotson: this isn't exactly the point. There is no real-time subscription overhead if we aren't
adding/removing schematics, because nothing is changing. But if we are adding
schematics, then we do want this to be updated. Plus, this it literally the only hook in
any `queries.ts` file that uses this pattern instead of a Flux retrieve.

pluto/src/schematic/queries.ts
```diff
@@ -143,6 +146,35 @@ export const { useUpdate: useSnapshot } = Flux.createUpdate<
...
+  const client = Synnax.use();
+  const [pages, setPages] = useState<Page[]>([]);
+  useAsyncEffect(
+    async (signal) => {
+      setPages([]);
+      if (client == null || exclude == null) return;
+      const parents = await client.ontology.retrieveParents(
```
pjdotson: this doesn't work if the schematic is a grandchild of a workspace: ws -> group -> schematic

pluto/src/schematic/symbol/Primitives.css
```diff
@@ -263,6 +263,37 @@
     }
 }
 
+.pluto-symbol__container:not(.pluto--editable) .pluto-arrow.pluto--linked {
+    cursor: pointer;
+
+    &:hover .wrapper {
+        filter: url("#goo") brightness(1.5);
```
pjdotson: this is very odd. I would just change the off-page reference so that it wraps a button
and uses that coloring instead.

pluto/src/schematic/symbol/Forms.tsx
```diff
@@ -1120,15 +1130,63 @@ export const TextBoxForm = (): ReactElement => {
...
+          {({ value, onChange }) => (
+            <Select.Buttons
+              value={value ? "double" : "single"}
+              onChange={(v: string) => onChange(v === "double")}
+              keys={["single", "double"] as const}
+            >
+              <Select.Button itemKey="single">Single</Select.Button>
+              <Select.Button itemKey="double">Double</Select.Button>
+            </Select.Buttons>
+          )}
```
pjdotson: 1) this could be memoized
2) I don't think we should even allow someone to select double, just adds extra
   confusion. Imagine you are at a company, you have the expectation that clicking an
   off-page-reference takes you to it. But then you go to someone else's schematic, and
   it's configured for double-click. Making it configurable here just adds more
   variability in expected behavior, which makes our software harder to use, for the
   sake of adding extra configurability that people are not likely to care as much about
   as, say, what color something is.

pluto/src/schematic/symbol/Forms.tsx
```diff
@@ -50,7 +51,16 @@ import { type StateIndicator as BaseStateIndicator } from "@/vis/stateIndicator"
...
-export interface SymbolFormProps extends Pick<Tabs.TabsProps, "actions"> {}
+const SelectTextLevel = ({
+  value,
+  onChange,
+}: Input.Control<Text.Level>): ReactElement => (
+  <Select.Text.Level value={value} onChange={onChange} />
+);
+
+export interface SymbolFormProps extends Pick<Tabs.TabsProps, "actions"> {
+  layoutKey?: string;
```
pjdotson: layoutKey is a bit odd to call, we don't have the concept of layouts in Pluto, only in
the Console. I would probably call it `schematicKey` and type it as `schematic.Key`

console/src/schematic/Schematic.tsx
```diff
@@ -76,6 +80,56 @@ import { Workspace } from "@/workspace";
 
 export const HAUL_TYPE = "schematic-element";
 
+export const navigateToLinkedSchematic = async (
+  client: { schematics: { retrieve: (arg: { key: string }) => Promise<any> } },
```
pjdotson: ps don't type things like this, should just type it as a synnax client. `any` is never good.

pjdotson (changes requested):

pluto/src/schematic/symbol/Forms.tsx
sy-nico: updated to `schematicKey`/`schematic.Key`

core/pkg/service/workspace/service.go
sy-nico: I will be splitting these up. I'm going to address each piece of feedback, and then split up the changes.

core/pkg/service/lineplot/writer_test.go
sy-nico: implemented

core/pkg/service/log/writer_test.go
sy-nico: implemented

core/pkg/service/schematic/writer_test.go
sy-nico: implemented

core/pkg/service/table/writer_test.go
sy-nico: implemented

core/pkg/service/workspace/service.go
sy-nico: implemented

pluto/src/schematic/symbol/Primitives.css
sy-nico: New behavior is transparent when no schematic is selected. Once selected, it changes to filled and the default primary color (blue). The color can then be changed and the text color will adapt to black/white.

pluto/src/schematic/symbol/Forms.tsx
sy-nico: 1. Implemented
2. Discussed offline and I'm going to leave as-is. This is something that system owners will feel very strongly about and have their own valid reasons for preferring single or double click. The same logic extends to valve controls... but that's a whole other can of worms that is related to the `Fire/Momentary/Pulse` selector for buttons.

pluto/src/schematic/queries.ts
sy-nico: Fair enough, switched to using flux

sy-nico: I also thought it would be a bigger change

pluto/src/schematic/queries.ts
sy-nico: Created a `collectSchematics` function to traverse the workspace and groups recursively. Added tests.

core/pkg/service/workspace/writer_test.go
sy-nico: Fixed. `deleteChildren` now recursively traverses into groups via `collectDescendants`, so nested resources at any depth get cascade-deleted. Updated the test to assert the schematic is gone.

console/src/schematic/Schematic.tsx
pjdotson: can you refactor this to use a flux retrieve call as well?

pluto/src/schematic/queries.ts
```diff
@@ -143,6 +143,60 @@ export const { useUpdate: useSnapshot } = Flux.createUpdate<
...
+  if (parents.length === 0) return [];
+  return await collectSchematics(client, parents[0].id, exclude);
+};
+
+const { useRetrieve: useRetrievePages } = Flux.createRetrieve<
+  PagesQuery,
+  Page[],
+  FluxSubStore
+>({
+  name: "schematic pages",
```
pjdotson: why does it add this new thing of "searching for pages", "schematic pages", etc.? Can't
we get rid of this terminology and just say "schematic"? Every "page" is just a "schematic"?

pluto/src/schematic/queries.ts
```diff
@@ -143,6 +143,60 @@ export const { useUpdate: useSnapshot } = Flux.createUpdate<
...
+  mountListeners: ({ store, onChange }) => [
+    store.relationships.onSet(() => onChange(undefined)),
+    store.resources.onSet(() => onChange(undefined)),
+  ],
+});
+
+export const usePages = (exclude?: string): Page[] => {
+  const result = useRetrievePages({ exclude });
+  return result.data ?? [];
+};
```
pjdotson: this seems like a really unhelpful function to add to a component library.

pluto/src/schematic/queries.ts
```diff
@@ -143,6 +143,60 @@ export const { useUpdate: useSnapshot } = Flux.createUpdate<
...
+  PagesQuery,
+  Page[],
+  FluxSubStore
+>({
+  name: "schematic pages",
+  retrieve: retrievePages,
+  mountListeners: ({ store, onChange }) => [
+    store.relationships.onSet(() => onChange(undefined)),
+    store.resources.onSet(() => onChange(undefined)),
+  ],
```
pjdotson: Unfortunately you are going to need to do more work here. IE what happens if a schematic
elsewhere in the workspace gets removed from that workspace and moved to a different
one? You need to test for all of these optimistic updates and make sure you have logic
for them. The ranges and status `queries.ts` files have logic for handling similar
issues like this that you should look at as a guide.

pluto/src/schematic/queries.ts
```diff
@@ -143,6 +143,60 @@ export const { useUpdate: useSnapshot } = Flux.createUpdate<
...
+  for (const child of children)
+    if (child.id.type === "schematic" && child.id.key !== exclude)
+      pages.push({ key: child.id.key, name: child.name });
+    else if (child.id.type === "group")
+      pages.push(...(await collectSchematics(client, child.id, exclude)));
+
+  return pages;
+};
+
+const retrievePages = async ({
```
pjdotson: The better problem here is to make this a `retrieveComponent`s hook in the
workspace/queries.ts file, with a filter for what type of visualization it is. I could
see similar logic being used in the future if we wanted to embed a table or plot inside
a schematic, for instance.

integration/tests/console/schematic/schematic_lifecycle.py
```diff
@@ -367,6 +372,85 @@ def _test_snapshot_rename_synchronization(self) -> None:
...
+            f"Expected to navigate back to '{names[0]}'"
+        )
+        self.log("Off-page navigation cycle verified")
+
+        # Cleanup
+        for name in names:
+            try:
+                self.console.layout.close_tab(name)
+            except Exception:
+                pass
```
pjdotson: should be failing the test if we hit an exception here

integration/console/context_menu.py
```diff
@@ -64,14 +63,11 @@ def click_option(self, text: str, *, exact: bool = True) -> None:
...
-            # Fixed-position menus may extend beyond the viewport, causing
-            # both click() and click(force=True) to fail. dispatch_event
-            # fires the click via the DOM and does not require the element
-            # to be within the viewport.
-            option.dispatch_event("click")
+        # Fixed-position menus may extend beyond the viewport, causing
+        # both click() and click(force=True) to fail. dispatch_event
+        # fires the click via the DOM and does not require the element
+        # to be within the viewport.
+        option.dispatch_event("click")
```
pjdotson, line 70: shouldn't we want to be failing if the option is not visible in the viewport?

integration/console/schematic/off_page_reference.py
```diff
@@ -0,0 +1,91 @@
...
+            Dictionary of applied properties.
+        """
+        self.click()
+        applied_properties: dict[str, Any] = {}
+
+        self.page.get_by_text("Properties").click()
+
+        if page_name is not None:
+            # Use specific CSS class selector — click_btn("Page") matches
+            # "Off Page" text elsewhere on the page.
```
pjdotson, line 74: couldn't you use a match exactly instead?

console/src/schematic/Schematic.tsx
```diff
@@ -75,6 +79,60 @@ import { Workspace } from "@/workspace";
...
+  client: SchematicRetriever | null;
+  storeState: StoreState;
+  layoutKey: string;
+  nodeId: string;
+  placeLayout: Layout.Placer;
+  handleError: Status.ErrorHandler;
+  dblClick: boolean;
+}
+
+export const handleNodeClickAction = (args: HandleNodeClickArgs): void => {
```
pjdotson: it is probably better to make this a hook `useHandleNodeClickAction`, that internally
calls, `useErrorHandler`, `Synnax.use`, `useStore`, `Layout.usePlacer`. that way there
are less args to pass to this

console/src/schematic/slice.spec.ts
```diff
@@ -550,4 +555,573 @@ describe("Schematic Slice", () => {
...
+      const client = {
+        schematics: { retrieve: vi.fn().mockResolvedValue(mockSchematic) },
+      };
+      const placeLayout = vi.fn();
+
+      await navigateToLinkedSchematic(client, "target-key", placeLayout);
+
+      expect(client.schematics.retrieve).toHaveBeenCalledWith({ key: "target-key" });
+      expect(placeLayout).toHaveBeenCalledTimes(1);
+    });
```
pjdotson: this isn't really a helpful test if we are just passing mocks

console/src/schematic/slice.spec.ts
```diff
@@ -550,4 +555,573 @@ describe("Schematic Slice", () => {
...
+      };
+      const placeLayout = vi.fn();
+
+      await expect(
+        navigateToLinkedSchematic(client, "bad-key", placeLayout),
+      ).rejects.toThrow("not found");
+
+      expect(placeLayout).not.toHaveBeenCalled();
+    });
+  });
```
pjdotson: None of these tests are that helpful as they are just testing mocks

console/src/schematic/Schematic.tsx
```diff
@@ -404,6 +496,8 @@ export const Loaded: Layout.Renderer = ({ layoutKey, visible }) => {
           editable={canEdit}
           triggers={triggers}
           onDoubleClick={handleDoubleClick}
+          onNodeClick={handleNodeClick}
+          onNodeDoubleClick={handleNodeDoubleClick}
```
pjdotson, line 481: So now every node click and node double click has to do the exact same thing? What if I
want to make double clicking a valve a different thing? The better API would be an
`onOffPageReference` click prop, that way we only have to pass one thing down.

pluto/src/schematic/symbol/registry.ts
```diff
@@ -1387,6 +1387,7 @@ const offPageReference: Spec<OffPageReferenceProps> = {
   defaultProps: (t) => ({
     color: t.colors.gray.l11,
     orientation: "right",
+    page: "",
```
pjdotson, line 1390: just a note, you also need to write a migration in the schematic `console` migrations
that adds this `page: ""` prop to all existing off-page reference. This sucks but is
needed to fix issues where existing schematics won't have the page form field show up.

You may be able to fix this by adding a `defaultValue` to the form field as well

pluto/src/schematic/symbol/Primitives.tsx
```diff
@@ -2993,7 +3000,15 @@ export const OffPageReference: React.FC<OffPageReferenceProps> = ({
...
-        <div className="outline" style={{ backgroundColor: color.cssString(colorVal) }}>
+        <div
+          className="outline"
+          style={
+            {
+              "--off-page-color": color.cssString(resolvedColor),
+              "--off-page-text-color": color.cssString(textColor),
+            } as React.CSSProperties
+          }
+        >
```
pjdotson, line 3011: is there a reason the color prop can't just be passed to `Text.MaybeEditable`?

pluto/src/schematic/symbol/Forms.tsx
```diff
@@ -1138,15 +1149,97 @@ export const TextBoxForm = (): ReactElement => {
...
+          {({ value, onChange }) => (
+            <Select.Static
+              value={value}
+              onChange={(v: string) => handlePageChange(value, onChange, v)}
+              data={pages}
+              resourceName="workspace schematic"
+              emptyContent="No other schematics in this workspace"
+              allowNone
+            />
+          )}
```
pjdotson, line 1240: you probably want to make this a `useCallback` so it rerenders less

pluto/src/schematic/symbol/Forms.tsx
```diff
@@ -1138,15 +1149,97 @@ export const TextBoxForm = (): ReactElement => {
...
+        </Form.Field>
+        <Form.Field<boolean>
+          path="dblClickNav"
+          label="Click Mode"
+          padHelpText={false}
+          hideIfNull={false}
+          defaultValue
+        >
+          {ClickModeSelect}
```
pjdotson, line 1249: You should `const clickModeSelectRender = Component.renderProp(ClickModeSelect)`. there
are slightly odd issues with passing the `key` prop if you don't do that.

pjdotson (changes requested):

pjdotson:
@sy-nico let's go over any comments you have questions about synchronously

pluto/src/schematic/queries.ts
sy-nico: Renamed to drop the "pages" terminology. Also, the Flux retrieve is now `useRetrieveSiblings`, and the resource name is just `"schematic"`.

pluto/src/schematic/queries.ts
sy-nico: Removed it. `Forms.tsx` now calls `useRetrievePages` directly.

pluto/src/schematic/queries.ts
sy-nico: The dropdown only needs current data when the user opens it to select a target. The current listeners already trigger a re-fetch on any relationship/resource change, so the list will be up to date by the time the user interacts with it. I don't think optimistic updates are necessary here since this isn't a persistently visible list.

pluto/src/schematic/queries.ts
sy-nico: Moved the retrieval logic to `Workspace.useRetrieveChildren` in `workspace/queries.ts`, which accepts a `types` filter (e.g. `["lineplot", "table"]`) so it can be reused for embedding different visualization types.

console/src/schematic/Schematic.tsx
sy-nico: Implemented

console/src/schematic/slice.spec.ts
sy-nico: Removed. This function is now internal to `useHandleNodeClickAction` and no longer exported or unit tested directly.

console/src/schematic/slice.spec.ts
sy-nico: Removed along with all the `handleNodeClickAction` unit tests.

console/src/schematic/Schematic.tsx
sy-nico: The hook centralizes the single/double click routing via `dblClickNav` in one place. A symbol-level callback would require threading click handling through the symbol rendering pipeline, which is a larger change with no other symbols needing it yet.

pluto/src/schematic/symbol/registry.ts
sy-nico: Implemented

pluto/src/schematic/symbol/Primitives.tsx
sy-nico: I tried this, but `Text.MaybeEditable` doesn't accept a `color` prop, and the text color is derived from contrast against `--off-page-color` which is set on the same parent element. Splitting them between CSS variables and inline styles was messier.

pluto/src/schematic/symbol/Forms.tsx
sy-nico: Implemented

pluto/src/schematic/symbol/Forms.tsx
sy-nico: Implemented

PR #2115, opened 2026-03-24
SY-3974: Prevent Orphaned Visualizations on Workspace Delete

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-3974](https://linear.app/synnax/issue/SY-3974/prevent-orphaned-visualizations-on-workspace-delete)

## Description
Originally discovered during [SY-3919](https://github.com/synnaxlabs/synnax/pull/2092)
When a workspace is deleted, child resources (schematics, lineplots, logs, tables) inside groups are orphaned because deleteChildren only traverses direct children. Add recursive traversal through groups so all nested resources are cascade-deleted regardless of nesting depth.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes orphaned child visualizations (schematics, line plots, logs, tables) when a workspace is deleted, by introducing a `ChildDeleter` interface and a recursive `collectDescendants` walk that finds all nested resources regardless of group depth. The workspace service initialization in `layer.go` is reordered so that the child services exist before they are injected as `ChildDeleters`, and each child service gets a thin `Delete` shim to satisfy the interface.

Key changes and findings:
- **Visualization cascade works correctly** — `collectDescendants` recursively traverses groups and collects all non-group resource keys; each child service then deletes them via gorp and ontology in a single pass.
- **Groups themselves are not cleaned up** — the traversal recurses into groups but never adds them to `byType`, so no `ChildDeleter` ever removes them. After a workspace delete, every intermediate `Group` record persists in the gorp table as an orphan (only the workspace→group ontology *relationships* are removed). This is the primary issue that needs to be addressed before merging.
- **Magic string `"group"`** — `collectDescendants` compares `child.ID.Type == "group"` instead of using the typed constant `ontology.TypeGroup`.
- **Test coverage is otherwise solid** — the new tests cover direct children, all four resource types, and group-nested children, though none assert that the groups themselves are gone after deletion.

- PR correctly fixes orphaned visualizations but leaves orphaned group records in the database on every workspace delete — a resource-leak bug in the same code path the PR is meant to fix.
- The primary fix (cascade-deleting schematics, line plots, logs, and tables) is correct and well-tested. However, the groups that contain those resources are never deleted — their gorp entries accumulate as orphans indefinitely. Because the PR explicitly targets cascade cleanup on workspace deletion, this gap is a meaningful omission rather than an edge case, warranting a concrete fix before merge.
- core/pkg/service/workspace/writer.go — the `collectDescendants` function skips adding groups to `byType`, so no deleter ever removes them.

| Filename | Overview |
|----------|----------|
| core/pkg/service/workspace/writer.go | Core logic change: adds recursive `collectDescendants` and `deleteChildren` to cascade-delete child resources on workspace deletion. Works correctly for schematics/lineplots/logs/tables, but groups encountered during traversal are never deleted themselves — their gorp entries leak. |
| core/pkg/service/workspace/service.go | Adds `ChildDeleters` slice to `ServiceConfig` and moves `Workspace` service initialization after child services in layer.go so the deleters can be injected. Change is correct and consistent. |
| core/pkg/service/layer.go | Moves workspace service initialization after the child services (Schematic, LinePlot, Log, Table) so they can be passed as `ChildDeleters`. Initialization order change is safe. |
| core/pkg/service/schematic/service.go | Adds a thin `Delete` shim on `*Service` that satisfies the `workspace.ChildDeleter` interface by delegating to `NewWriter(tx).Delete`. |
| core/pkg/service/workspace/writer_test.go | Good coverage added for direct children, nested-in-group children, all four resource types, and empty workspaces. The nested-group test only verifies the schematic is gone — the group itself is not checked, which masks the group-orphan bug. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["Writer.Delete(workspaceKey)"] --> B["deleteChildren(key)"]
    B --> C["collectDescendants(workspaceID, byType)"]
    C --> D{"child.ID.Type == 'group'?"}
    D -- Yes --> E["recurse: collectDescendants(groupID, byType)"]
    E --> D
    D -- No --> F["byType[child.Type] append(key)"]
    C --> G["For each ChildDeleter\nSchematic / LinePlot / Log / Table"]
    G --> H["deleter.Delete(ctx, tx, keys...)"]
    H --> I["otg.DeleteResource(childID)\n✅ removes child from gorp + ontology"]
    B --> J["gorp.Delete Workspace record"]
    J --> K["otg.DeleteResource(workspaceID)\n✅ removes workspace→group relationships"]
    K --> L["⚠️ Group gorp entries remain\n(orphaned — never deleted)"]

    style L fill:#f96,stroke:#c00,color:#000
    style I fill:#6f6,stroke:#060,color:#000
    style K fill:#6f6,stroke:#060,color:#000
```

<sub>Reviews (1): Last reviewed commit: ["SY-3974: Prevent orphaned visualizations..."](https://github.com/synnaxlabs/synnax/commit/6ebdc85fe4bbbadc993c05fad0b5216b1669051c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26246857)</sub>


core/pkg/service/lineplot/service.go
```diff
@@ -92,6 +92,10 @@ func (s *Service) NewWriter(tx gorp.Tx) Writer {
 	}
 }
 
+func (s *Service) Delete(ctx context.Context, tx gorp.Tx, keys ...uuid.UUID) error {
```
pjdotson, line 95: you should add individual tests for this in `lineplot_test`, `log_test`, etc.

core/pkg/service/lineplot/service.go
```diff
@@ -92,6 +92,10 @@ func (s *Service) NewWriter(tx gorp.Tx) Writer {
...
+	return s.NewWriter(tx).Delete(ctx, keys...)
```
pjdotson, line 96: and add a doc comment for public functions

core/pkg/service/workspace/writer.go
```diff
@@ -98,3 +111,49 @@ func (w Writer) Delete(
...
+		Exec(ctx, w.tx); err != nil {
+		return err
+	}
+	for _, child := range children {
+		if child.ID.Type == "group" {
+			if err := w.collectDescendants(ctx, child.ID, byType); err != nil {
+				return err
+			}
+			continue
+		}
```
pjdotson, line 151: instead of this special case for group right here, I would a) pass the group service as
a `ChildDeleter` like lineplot, schematic, etc., and b) just call `collectDescendants`
no matter what the ontology type is.

core/pkg/service/workspace/writer_test.go
```diff
@@ -49,13 +56,88 @@ var _ = Describe("Writer", func() {
...
+				ctx,
+				ontology.ID{Type: "group", Key: g.Key.String()},
+				ontology.RelationshipTypeParentOf,
+				schematic.OntologyID(s.Key),
+			)).To(Succeed())
+
+			Expect(svc.NewWriter(tx).Delete(ctx, ws.Key)).To(Succeed())
+
+			var sRes schematic.Schematic
+			Expect(gorp.NewRetrieve[uuid.UUID, schematic.Schematic]().WhereKeys(s.Key).Entry(&sRes).Exec(ctx, tx)).To(HaveOccurredAs(query.ErrNotFound))
```
pjdotson, line 106: should test that the group gets deleted as well

pjdotson (changes requested):

sy-nico:
Obsolete after the ongoing metadata improvements

PR #2150, opened 2026-04-02
SY-4004: Use Oracle to Generate Retrieve Queries

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4004](https://linear.app/synnaxlabs/issue/SY-4004)

## Description

Add a new `go/query` Oracle plugin that generates gorp `Retrieve` query wrappers from schema definitions, replacing hand-written retrieve files across all core services.

**New Oracle annotations:**
- `@retrieve` on structs to generate `retrieve.gen.go` with `WhereKeys`, `Entry`, `Entries`, and `Exec` methods
- `@retrieve custom` for services (rack, ranger) that need additional hand-written query methods beyond the generated ones
- `@filter` / `@filter required` on fields to generate `Where<Field>` filter methods
- `@search` to generate `Search` method integration

**Generated retrieve files** for: arc, device, lineplot, log, rack, ranger, schematic, table, task, user, workspace. Each replaces a hand-written `retrieve.go` with a `retrieve.gen.go`.

**Removed legacy Oracle plugins:** migrate, marshal, snapshot, and their associated commands and tests. These were superseded by the current plugin architecture.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR introduces a new Oracle `go/query` plugin that auto-generates `retrieve.gen.go` files from schema annotations, replacing hand-written retrieve files across all core services. The underlying `gorp` layer is significantly refactored: `WhereKeys`/`Where`/`WhereRaw` are unified under a composable `Filter` type, validators now correctly propagate through `Exec`/`Count`/`Exists`, and `MatchRaw` inside `Or`/`Not` compositions is properly handled via `evalChild`.

The only notable finding is that `canSkipExec` in `ontology/retrieve.go` is now dead code — its condition `HasFilterKeys() && !HasFilters()` is logically impossible under the new filter model — though the added test for ErrNotFound on non-existent traversals suggests the behaviour change is intentional.

Safe to merge; only P2 findings remain after prior review rounds addressed the major issues.

All P0/P1 issues raised in prior review rounds (validator skipping in Count, MatchRaw in Or/Not, ErrNotFound hiding in execKeys, RBAC performance regression) are addressed in this revision. Remaining findings are P2: dead-code optimization in canSkipExec and a minor parameter-naming shadow.

core/pkg/distribution/ontology/retrieve.go (canSkipExec is dead code)

| Filename | Overview |
|----------|----------|
| oracle/plugin/go/query/query.go | New Oracle plugin that generates gorp Retrieve query wrappers from schema annotations; well-structured with comprehensive template and filter generation. |
| x/go/gorp/filter.go | New composable filter primitives (Match, MatchRaw, MatchKeys, And, Or, Not, BoundFilter); previously-flagged MatchRaw issues inside Or/Not are now correctly handled via evalChild. |
| x/go/gorp/retrieve.go | Major refactor replacing WhereKeys/filters/rawFilters with unified Filter type; validators now run in all Count paths; ErrNotFound handling in execKeys is improved. |
| core/pkg/distribution/ontology/retrieve.go | canSkipExec optimization is permanently disabled because HasFilterKeys() && !HasFilters() is an impossible condition; the function is dead code and should be simplified or removed. |
| core/pkg/service/access/rbac/policy/retriever.go | Old Retriever replaced with generated Retrieve; WhereSubjects logic extracted to ResolveSubjects for explicit key pre-resolution before MatchKeys. |
| core/pkg/api/access/access.go | Subjects+Keys combined query now intersects (was union); this is a security improvement but is a silent API semantic change for callers combining both fields. |
| core/pkg/distribution/channel/service.go | validateChannels signature updated to match new Validator[K,E] type; now correctly propagates through generated Exec/Count/Exists pipeline via Validate(). |
| x/go/slices/slices.go | New CollectKeys helper for iter.Seq2; minor: parameter name cap shadows the built-in. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[service.NewRetrieve] --> B[Retrieve struct]
    B --> C[Where Filter]
    C --> D{Filter type}
    D -->|MatchKeys| E[filter.keys set\nHasFilterKeys = true]
    D -->|Match or MatchRaw| F[filter.eval or raw set\nHasNonKeyFilters = true]
    D -->|And or Or or Not| G[Composite Filter\nkeys via singleBoundedKeys]
    E --> H[Exec / Count / Exists]
    F --> H
    G --> H
    H --> I{HasFilterKeys?}
    I -->|yes| J[execKeys\npoint-lookup per key\nrun validators]
    I -->|no| K[execFilter\nfull scan plus raw pre-screen\nrun validators]
    J --> L[ErrNotFound if isBareKeys and missing keys]
    K --> M[ErrNotFound if no results and single-entry bound]
```

<sub>Reviews (9): Last reviewed commit: ["updated oracle plugins"](https://github.com/synnaxlabs/synnax/commit/e93ca853ad539fbeb72b0953b3bd02a63cbfa9f7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27174770)</sub>


docs/tech/rfc/0034-260404-gorp-indexes.md
```diff
@@ -0,0 +1,636 @@
...
+(< 100k entries per type), this is acceptable. If profiling reveals the shift cost is a
+problem, we can swap the backing structure to a B-tree without changing the public API.
+
+### 2.1.3 - Why Not Composite Indexes?
+
+Composite indexes (indexing on multiple fields simultaneously) add significant
+complexity: you need to define field ordering, handle partial prefix queries, and the
+number of possible composites grows combinatorially. For our access patterns, chaining a
+sorted index with a post-filter on the remaining fields is sufficient. We can revisit
+composite indexes if profiling shows a real need.
```
pjdotson, line 148: if we do add composite indexes would the API need to change at all?

docs/tech/rfc/0034-260404-gorp-indexes.md
```diff
@@ -0,0 +1,636 @@
...
+Gorp provides generic index structs that hold backing data and satisfy a common `Index`
+interface:
+
+```go
+// Index is the interface that all index types satisfy for registration on a Table.
+type Index interface {
+    populate(ctx context.Context, tx Tx) error
+    update(key any, oldEntry any, newEntry any)
+    remove(key any)
+}
```
pjdotson, line 174: Is there a reason for this not to have type parameters for the key and entries?

docs/tech/rfc/0034-260404-gorp-indexes.md
```diff
@@ -0,0 +1,636 @@
...
+    Where(channel.ByName("sensor_1", "sensor_2", "sensor_3")).
+    Entries(&results).
+    Exec(ctx, tx)
+
+// Compose with additional non-indexed filters
+table.NewRetrieve().
+    Where(channel.ByDataType(telem.Float32)).
+    Where(channel.IsVirtual()).
+    Entries(&results).
+    Exec(ctx, tx)
```
pjdotson, line 385: Is there a fundamental reason that we have to have `ByDataType`, `IsVirtual` be
functions instead of methods on the `Retrieve` type?

docs/tech/rfc/0034-260404-gorp-indexes.md
```diff
@@ -0,0 +1,636 @@
...
+### 2.4.3 - Transaction Isolation
+
+A question arises: should uncommitted writes within a transaction be visible to
+index-backed queries within that same transaction?
+
+For the initial implementation: **no.** Index queries always reflect committed state.
+This matches the current behavior where `Retrieve` reads from the underlying KV store,
+which (in Pebble) provides snapshot isolation at the transaction level. Adding
+transaction-local index overlays is complex and not needed for our current access
+patterns, where writes and reads in the same transaction are rare.
```
pjdotson: so if `Retrieve` takes in a transaction in it, it won't properly retrieve things because
the index hasn't been updated? I don't know if this really makes sense, even if its rare
it means any time we do have a read after a write in the same transaction that touches
the same thing, the read will be wrong, and is just basically a known bug we are putting
in here.

docs/tech/rfc/0034-260404-gorp-indexes.md
```diff
@@ -0,0 +1,636 @@
...
+
+`Retrieve` does not know about indexes. The intelligence lives in the `Filter`'s `Eval`
+closure, which the index's `.Filter(values...)` method produces:
+
+1. If the index's backing data is populated, the `Eval` closure looks up candidate keys
+   directly from the in-memory map, fetches those entries from Pebble by key, and
+   returns the matches. This converts the query into the existing fast `execKeys` path
+   internally.
+2. If the backing data is nil (index not registered), the `Eval` closure falls back to a
+   field comparison against each decoded entry, the same as any non-indexed filter.
```
pjdotson: what are examples where we call `Eval` and it won't hit an index?

oracle/plugin/domain/helpers.go
```diff
@@ -88,6 +88,20 @@ func HasExprFromType(t resolution.Type, domainName, exprName string) bool {
...
+
+func HasDomainFromType(t resolution.Type, domainName string) bool {
+	_, ok := t.Domains[domainName]
+	return ok
+}
+
+func HasDomainFromField(f resolution.Field, domainName string) bool {
+	_, ok := f.Domains[domainName]
+	return ok
+}
```
pjdotson, line 103: add tests for all of these

oracle/plugin/go/query/query_test.go
```diff
@@ -0,0 +1,697 @@
+// Copyright 2026 Synnax Labs, Inc.
```
pjdotson, line 1: so for tests like this we never test at all what the body of the code looks like, just
the function / method typing?

core/pkg/distribution/channel/retrieve.gen.go
```diff
@@ -0,0 +1,153 @@
...
+) Filter {
+	return func(r Retrieve) gorp.Filter[Key, Channel] {
+		return gorp.Match(func(ctx gorp.Context, e *Channel) (bool, error) {
+			return f(ctx, r, e)
+		})
+	}
+}
+
+// And returns a filter that matches when all provided filters match.
+func And(fs ...Filter) Filter {
```
pjdotson, line 51: we have to generate specific `And`, `Or`, and `Not`s for each type? so if i want to call
something that matches both names and data types I would have to call:

```go
r.Where(
    channel.And(
        channel.MatchDataTypes(telem.Float32T, telem.Float64T),
        channel.MatchNames("name1", "name2-*"),
    ),
)
```

This is a really verbose API instead of the previous method based one:

```go
r.WhereDataTypes(telem.Float32T, telem.Float64T).WhereNames("name1","name2-*").MatchAny()
```
where you have something like a `MatchAny` / `MatchAll` to give the AND / OR semantics.
It is also odd because we still have the `.WhereKeys` method as well, but everything
else needs to go through the mega `r.Where`. Finally, having them as methods instead of
functions is a bit better devX: you can get see in your IDE all of the methods on the
`Retrieve` for different filtering options directly, instead of having to call a
function in the `channel` package

core/pkg/distribution/ontology/writer_dag.go
```diff
@@ -190,27 +190,27 @@ func (d dagWriter) retrieveDescendants(ctx context.Context, id ID) (map[ID]Resou
 }
 
 func (d dagWriter) deleteIncomingRelationships(ctx context.Context, id ID) error {
-	return d.relationshipTable.NewDelete().Where(func(ctx gorp.Context, rel *Relationship) (bool, error) {
+	return d.relationshipTable.NewDelete().Where(gorp.Match(func(ctx gorp.Context, rel *Relationship) (bool, error) {
```
pjdotson, line 191: it's a little odd that we have to wrap calls like this in `gorp.Match` now. Is there a
good reason the API has to be like this instead of the previous `Where` / `WhereRaw`?
Why can't `Where` wrap the closure as a `gorp.Filter` internally?

core/pkg/distribution/channel/retrieve.go
```diff
@@ -10,163 +10,116 @@
...
-func (r Retrieve) WhereVirtual(virtual bool) Retrieve {
-	r.gorp = r.gorp.Where(func(_ gorp.Context, ch *Channel) (bool, error) {
-		isVirtual := ch.Virtual && !ch.IsCalculated()
-		return isVirtual == virtual, nil
-	}, gorp.Required())
-	return r
+// MatchVirtual returns a filter for channels that are virtual if virtual is true, or
+// are not virtual if virtual is false. Calculated channels are excluded from the
+// virtual bucket even though they are stored with Virtual=true.
+func MatchVirtual(virtual bool) Filter {
```
pjdotson, line 23: These `Match*` functions are not auto-generated in the `retrieve.gen.go` files?

pjdotson (changes requested):
Left a few comments. Generally think the API needs to be improved to make it easier to read and write (specifics in comments).

docs/tech/rfc/0034-260404-gorp-indexes.md
emilbon99: Yes, this is a design flaw. I changed it in the actual gorp indexes PR to accurately reflect transactions. Will update RFC

core/pkg/distribution/channel/retrieve.gen.go
emilbon99: 1) I don't think that it's that verbose. If you do:

`r.Where(channel.MatchDataTypes(telem.Float32, telem.Float64)).Where(channel.MatchNames("name1", "name2-*").Exec()` 

is really not that much longer than what you have. Total cost here is like 20 characters. Not that much. 

2) The key thing that your method based API ability to mix `AND` and `OR` logic. What happens if I want these two filters in AND and another filter in OR? Your API cannot express that. 

What if I want to do:

`channel.Where(
       channel.Or(
           channel.MatchNames("cat"),
           channel.And(
               channel.MatchDataTypes(yada),
               channel.MatchInternal(true)
           )
       )
))`

Your query language cannot express that cleanly.

core/pkg/distribution/ontology/writer_dag.go
emilbon99: Yes, so that matchers like `And` and `Or` can take in a mix of raw filters, prefix filters, and others. Makes complex query composition much easier.

core/pkg/distribution/channel/retrieve.go
emilbon99: Yeah they should be. Will correct.

docs/tech/rfc/0034-260404-gorp-indexes.md
emilbon99: Actual implementation has type parameters. Rewrote RFC.

docs/tech/rfc/0034-260404-gorp-indexes.md
emilbon99: Yeah we would probably need to make adjustments to support composite indexes. Not particularly concerned as the API here isn't a one way decision and can be expanded/modified as needed i the future.

docs/tech/rfc/0034-260404-gorp-indexes.md
emilbon99: Well we aren't going to have indexes for every single field on every data structure, so some filters will fall back to a full table scan. This is a standard practice and is common in things like SQL.

oracle/plugin/go/query/query_test.go
emilbon99: We're asserting on correct generation of relevant snippets as opposed to more brittle tests that assert on the whole file output. I think this is the best way to test sustainably. The nice think about generation is that almost every code generation issue will make the output not compile, so we get a lot of safety that way.

docs/tech/rfc/0034-260404-gorp-indexes.md
```diff
@@ -0,0 +1,911 @@
...
+   filter is index-backed or composes one that is), invoke it to produce the merged
+   `[]K` and the lazy-membership build function. Populate `filter.Keys` and
+   `filter.membership`. No-op for bare `Match` / `MatchRaw` filters.
+2. **OrderBy** (`r.execOrdered`). If the query has an `OrderQuery`, walk it, `GetMany`
+   the resulting keys, run any post-filters, and return.
+3. **execKeys (fast path)** (`r.execKeys`). If `WhereKeys` was called or the resolved
+   filter has `Keys != nil`, fetch only those keys via `reader.GetMany` and run the
+   remaining `Eval` / `Raw` predicates as post-checks.
+4. **execFilter (full scan)** (`r.execFilter`). Otherwise, iterate every entry in the
+   table, applying `Raw` (pre-decode) and `Eval` (post-decode) predicates.
```
pjdotson, line 657: Does this prevent queries like "channels where keys are 1,2,3 OR names are
'sens1','sens2'"? I don't know if that would ever be something we would want to do or not.

x/go/gorp/reader.go
```diff
@@ -41,42 +41,57 @@ func WrapReader[K Key, E Entry[K]](tx Tx) *Reader[K, E] {
...
+// get returns an entry with a copy of its raw bytes that outlives the
+// transaction handle.
+func (r Reader[K, E]) get(ctx context.Context, key K) (E, []byte, error) {
 	var e E
 	b, closer, err := r.tx.Get(ctx, r.keyCodec.encode(key))
 	if err != nil {
-		return e, err
+		return e, nil, err
 	}
+	raw := append([]byte(nil), b...)
```
pjdotson: couldn't you just use `bytes.Clone`?

x/go/gorp/reader.go
```diff
@@ -41,42 +41,57 @@ func WrapReader[K Key, E Entry[K]](tx Tx) *Reader[K, E] {
...
 	err = r.tx.Decode(ctx, b, &e)
-	return e, errors.Combine(err, closer.Close())
+	return e, raw, errors.Combine(err, closer.Close())
```
pjdotson: should be returning zero values when `err` is non-nil

x/go/gorp/retrieve.go
```diff
@@ -24,41 +24,40 @@ import (
 // decoded value. Returning true allows normal decode + filter processing.
 type RawFilter func(data []byte) (bool, error)
 
+// Validator is a batch check that runs on the final bound result set after
+// Exec populates entries. Returning a non-nil error causes Exec to return that
+// error. Validators are attached via Retrieve.Validate and cannot filter
+// results — if you need to filter, use Where.
+type Validator[K Key, E Entry[K]] func(ctx Context, entries []E) error
```
pjdotson, line 28: what is the point of this? don't see it used anywhere.

x/go/gorp/filter.go
```diff
@@ -0,0 +1,218 @@
...
+		Eval: func(ctx Context, e *E, _ []byte) (bool, error) {
+			return f(ctx, e)
+		},
+	}
+}
+
+// MatchRaw wraps a raw-byte predicate as a Filter. Under Where or And it
+// pre-screens before decoding; under Or or Not it is called at Eval time so
+// composition is correct, at the cost of decoding every entry.
+func MatchRaw[K Key, E Entry[K]](f func(data []byte) (bool, error)) Filter[K, E] {
```
pjdotson, line 31: Why do we have to decode under Or or Not?

core/pkg/service/label/retrieve.go
```diff
@@ -46,14 +46,19 @@ func (r Retrieve) Entries(labels *[]label.Label) Retrieve { r.gorp = r.gorp.Entr
...
 func (r Retrieve) WhereKeys(keys ...label.Key) Retrieve { r.gorp = r.gorp.WhereKeys(keys...); return r }
 
-// WhereNames filters for labels whose Name attribute matches the provided name.
-func (r Retrieve) WhereNames(names ...string) Retrieve {
-	r.gorp = r.gorp.Where(func(ctx gorp.Context, label *label.Label) (bool, error) {
-		return lo.Contains(names, label.Name), nil
-	})
+// Where applies the provided filters to the query.
+func (r Retrieve) Where(filters ...gorp.Filter[label.Key, label.Label]) Retrieve {
+	r.gorp = r.gorp.Where(filters...)
```
pjdotson: this file isn't generated?

core/pkg/distribution/channel/retrieve.go
```diff
@@ -10,163 +10,45 @@
...
-}
-
-// WhereVirtual filters the query for channels that are virtual if virtual is true, or are
-// not virtual if virtual is false.
-func (r Retrieve) WhereVirtual(virtual bool) Retrieve {
-	r.gorp = r.gorp.Where(func(_ gorp.Context, ch *Channel) (bool, error) {
+// MatchVirtual returns a filter for channels that are virtual if virtual is true, or
+// are not virtual if virtual is false. Calculated channels are excluded from the
+// virtual bucket even though they are stored with Virtual=true.
+func MatchVirtual(virtual bool) Filter {
```
pjdotson, line 23: should figure out the semantics of what this means more later. From the point of view of
a client, I don't see why I would want the virtual property of a calculated channel to
be true.

core/pkg/service/rack/rack_test.go
```diff
@@ -274,14 +276,142 @@ var _ = Describe("Rack", Ordered, func() {
...
+					Search("SearchableRackAlpha").
+					Entries(&res).
+					Exec(ctx, tx)).To(Succeed())
+			})
+			It("Should execute Count through the search path", func(ctx SpecContext) {
+				MustSucceed(svc.NewRetrieve().Search("SearchableRackAlpha").Count(ctx, tx))
+			})
+			It("Should execute Exists through the search path", func(ctx SpecContext) {
+				MustSucceed(svc.NewRetrieve().Search("SearchableRackAlpha").Exists(ctx, tx))
+			})
```
pjdotson, line 426: should assert on return values here

core/pkg/service/rack/retrieve.go
```diff
@@ -28,133 +25,39 @@ type Retrieve struct {
...
-}
-
-// WhereEmbedded filters for racks that are embedded within the Synnax server.
-func (r Retrieve) WhereEmbedded(embedded bool, opts ...gorp.FilterOption) Retrieve {
-	r.gorp = r.gorp.Where(func(ctx gorp.Context, rack *Rack) (bool, error) {
-		return rack.Embedded == embedded, nil
-	}, opts...)
-	return r
+// MatchName returns a filter that matches racks whose Name matches the provided value.
+func MatchName(name string) Filter {
```
pjdotson: why isn't this generated?

core/pkg/service/rack/retrieve.go
```diff
@@ -28,133 +25,39 @@ type Retrieve struct {
...
 
-// Count returns the number of entries matching the query.
-func (r Retrieve) Count(ctx context.Context, tx gorp.Tx) (int, error) {
-	var err error
-	r, err = r.execSearch(ctx)
-	if err != nil {
-		return 0, err
+// MatchIntegration returns a filter that matches racks that support the given
+// integration.
+func MatchIntegration(integration string) Filter {
```
pjdotson: why isn't this one generated?

pjdotson (changes requested):

x/go/gorp/retrieve.go
emilbon99: This gets used for channel validation. See channel service. It makes it so we can generate the channel retrieve query infra while keeping channel validation.

core/pkg/service/label/retrieve.go
emilbon99: Yeah, it has to do with the fact that label spans `x` and the `core`. I thought I'd leave it handwritten until we decided what we're going to do with this and the status type.

core/pkg/distribution/channel/retrieve.go
emilbon99: Copy

x/go/gorp/filter.go
emilbon99: We don't always have to decode under or or not. Fixed do omments.

emilbon99:

x/go/gorp/retrieve.go
```diff
@@ -68,19 +63,32 @@ func (r Retrieve[K, E]) HasLimit() bool { return r.limit > 0 }
...
+// key set (i.e. Where(MatchKeys(...)) was called).
+func (r Retrieve[K, E]) HasFilterKeys() bool {
+	return r.filter != nil && r.filter.keys != nil
 }
 
-// HasFilters returns true if any Where filters were added to the query.
-func (r Retrieve[K, E]) HasFilters() bool { return len(r.filters) > 0 }
+// GetFilterKeys returns the resolved filter's primary key set, or nil if the
+// query is not bounded by keys.
+func (r Retrieve[K, E]) GetFilterKeys() []K {
```
pjdotson, line 86: should these things be public methods?

pjdotson (changes requested):

pjdotson (approved):

x/go/gorp/retrieve.go
emilbon99: Yeah they need to be public for now as the ontology uses them for fast path lookup. As we get indexes in I will work on trying to reduce the footprint of the gorp public API.

PR #2188, opened 2026-04-07
SY-3198: Add Clock Skew Detection to Python and TypeScript Clients

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-3198](https://linear.app/synnaxlabs/issue/SY-3198)

## Description

Adds clock skew detection to the Python and TypeScript clients, matching the existing C++ implementation. The server already sends `node_time` in auth and connectivity responses over HTTP; the clients were not parsing it.

- Adds `ClockSkewCalculator` utility to `x/ts` and `x/py` (mirroring `x/cpp/telem/clock_skew.h`) with injectable `now()` for testability
- Adds `TimeSpan.abs()` to both TS and Python
- TypeScript client measures skew at auth time and on each connectivity poll (every 30s), exposing `clockSkew` and `clockSkewExceeded` on connection state
- Python client measures skew at auth time and emits `warnings.warn()` if threshold exceeded
- Console surfaces clock skew warnings via the existing notification adapter pattern
- Updates gRPC connectivity proto/handler to forward `node_time` for completeness
- Default threshold: 1 second (configurable via `clockSkewThreshold` / `clock_skew_threshold`)

## Basic Readiness

<h3>Greptile Summary</h3>

Adds clock skew detection to Python and TypeScript clients by introducing `ClockSkewCalculator` (midpoint synchronization algorithm) in `x/ts` and `x/py`, mirroring the existing C++ implementation. The TypeScript `Checker` uses a `checking` flag to correctly handle concurrent calls, exposes `clockSkew`/`clockSkewExceeded` on connection state, and `Provider.tsx` now surfaces skew warnings both at initial connect and on later threshold transitions — addressing the previous review concern about notifications only firing once.

- Safe to merge; all remaining findings are P2 suggestions.
- The implementation is clean and consistent across all three languages, tests cover the key scenarios, and the previous review concerns (concurrent-call accuracy in TS, UI notification on later polls) are both addressed. The only open finding is a minor thread-safety note on Python's `_skew_calc` (same pattern the TS version already fixed), which affects measurement accuracy at most and not data integrity.
- client/py/synnax/connection.py — `_skew_calc.start()`/`end()` called outside the lock; low-impact race but worth aligning with the TypeScript guard.

| Filename | Overview |
|----------|----------|
| x/ts/src/telem/clockSkew.ts | New `ClockSkewCalculator` utility using the midpoint synchronization algorithm, with injectable `now()` for testability. Clean implementation matching the C++ reference. |
| client/ts/src/connection/checker.ts | Adds `clockSkew`/`clockSkewExceeded` to connection state; uses a `checking` flag to correctly guard the `ClockSkewCalculator` against concurrent `check()` calls. `onChange` now fires on skew transitions, addressing the previous review concern. |
| client/py/synnax/connection.py | Python port of connection checker with clock skew detection; `_skew_calc.start()`/`end()` are called outside `_lock`, creating a potential race if user code and the background thread call `check()` concurrently. |
| pluto/src/synnax/Provider.tsx | Surfaces clock skew notifications both on initial connect and on later transitions via `handleChange`; addresses the previous review comment about notifications only firing at initial connect. |
| x/py/x/telem/clock_skew.py | New `ClockSkewCalculator` matching the TypeScript/C++ implementations; injectable `now()` for testability, correct midpoint algorithm. |
| client/ts/src/auth/auth.ts | Adds `nodeTime` field to `clusterInfoZ` for schema completeness; field is parsed but not yet used for skew calculation (skew is measured in the connectivity checker instead). |
| console/src/cluster/types/v2.ts | Correctly omits `clockSkewThreshold` from the persisted cluster config schema (along with `connectivityPollFrequency` and `retry`) so users don't see it as a stored connection parameter. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant App
    participant Checker
    participant ClockSkewCalc
    participant Server

    App->>Checker: new Checker(client, threshold)
    Checker->>ClockSkewCalc: new ClockSkewCalculator()
    Checker->>Checker: void this.check() [async]
    Checker->>Checker: start() [30s interval]

    rect rgb(200, 230, 255)
        Note over Checker,Server: Each check() call
        Checker->>ClockSkewCalc: start() [record localStartT]
        Checker->>Server: GET /connectivity/check
        Server-->>Checker: "{ clusterKey, nodeVersion, nodeTime }"
        Checker->>ClockSkewCalc: end(nodeTime)
        Note over ClockSkewCalc: skew = localMidpoint - nodeTime
        ClockSkewCalc-->>Checker: skew, exceeds(threshold)?
        Checker->>Checker: update state (clockSkew, clockSkewExceeded)
        alt status or skewExceeded changed
            Checker->>App: onChange(state)
        end
    end

    App->>Checker: "check() [explicit call, TS: checking=true guards skew]"
    Checker-->>App: "State { clockSkew, clockSkewExceeded }"
```

<sub>Reviews (2): Last reviewed commit: ["Address PR review feedback"](https://github.com/synnaxlabs/synnax/commit/49d8bf84ad201b9aa05fa0d659072fddb2650cd0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27615994)</sub>


client/py/synnax/synnax.py
```diff
@@ -87,6 +87,7 @@ def __init__(
         max_retries: int = 3,
         instrumentation: Instrumentation = NOOP,
         cache_channels: bool = True,
+        clock_skew_threshold: TimeSpan = TimeSpan.SECOND,
```
sy-nico, line 93: Add parameter to doc string.

x/ts/src/telem/clockSkew.ts
```diff
@@ -0,0 +1,50 @@
...
+    const halfSpan = localEndT.span(this.localStartT).valueOf() / 2n;
+    const thisMidpointT = this.localStartT.add(halfSpan);
+    const skew = thisMidpointT.valueOf() - remoteMidpointT.valueOf();
+    this.accumulatedSkew += skew;
+    this.n++;
+  }
+
+  get skew(): TimeSpan {
+    if (this.n === 0) return TimeSpan.ZERO;
+    return new TimeSpan(this.accumulatedSkew / BigInt(this.n));
```
sy-nico: I don't think we should be returning a running average of the skew. The clock could drift over time, the severity would be underreported due to more recent values having less weight. 

Since we are only polling once every 30s, I think we should just output the most recent value. If you think we need a running average for long-term smoothing, then there should be a set window.

x/py/x/telem/clock_skew.py
```diff
@@ -0,0 +1,47 @@
...
+        local_mid = start + (end - start) // 2
+        skew = local_mid - int(remote_midpoint_t)
+        self._accumulated_skew += skew
+        self._n += 1
+
+    @property
+    def skew(self) -> TimeSpan:
+        if self._n == 0:
+            return TimeSpan(0)
+        return TimeSpan(self._accumulated_skew // self._n)
```
sy-nico: Same concern as [this comment](https://github.com/synnaxlabs/synnax/pull/2188/changes#r3075007720).

client/py/synnax/auth.py
```diff
@@ -53,19 +65,37 @@ def __init__(
...
         self.authenticated = False
+        self._skew_calc = ClockSkewCalculator()
+        self._clock_skew_threshold = clock_skew_threshold
+
+    @property
+    def clock_skew(self) -> TimeSpan:
+        return self._skew_calc.skew
 
     def authenticate(self) -> None:
+        self._skew_calc.start()
```
sy-nico: I think the clock skew for the python client is only measured once during the authentication. There should be polling every 30s to match the typescript client. Without it, long-running py connections can't detect clock drift.

client/ts/src/connection/checker.ts
```diff
@@ -74,12 +89,15 @@ export class Checker {
...
     clientVersion: string,
     name?: string,
+    clockSkewThreshold: CrudeTimeSpan = TimeSpan.seconds(1),
   ) {
     this._state = { ...DEFAULT };
     this.client = client;
     this.pollFrequency = pollFreq;
     this.clientVersion = clientVersion;
     this.name = name;
+    this.skewCalc = new ClockSkewCalculator();
```
sy-nico, line 98: Both `auth.Client` and `Checker` each create their own `ClockSkewCalculator`. The auth measurement runs during login but nothing reads `auth.clockSkew` after that.  `Provider.tsx` only reads from `connectivity.clockSkew`. Is the auth-side measurement intentional, or should there be a single shared calculator?

client/ts/src/connection/checker.ts
```diff
@@ -140,6 +175,8 @@ export class Checker {
       this._state.status = "failed";
       this._state.error = err as Error;
       this._state.message = this.state.error?.message;
+    } finally {
+      this.checking = false;
     }
     if (this.onChangeHandlers.length > 0 && prevStatus !== this._state.status)
```
sy-nico: `onChange` only fires when `status` changes (e.g., connected → failed), but clock skew can cross the threshold while status stays `"connected"`.
Just verifying that we don't need to call the `onChange` when skew is detected.

pluto/src/synnax/Provider.tsx
```diff
@@ -113,6 +126,27 @@ export const Provider = ({ children, connParams }: ProviderProps): ReactElement
           variant: CONNECTION_STATE_VARIANTS[state.status],
           message: state.message ?? caseconv.capitalize(state.status),
         });
+      if (
```
sy-nico: The clock skew check here is unreachable. `handleChange` only fires on status transitions, and the initial check below (lines 209-225) already covers the `connected` + `clockSkewExceeded` case. On reconnect, `clockSkewWarned.current` would already be `true` from the first connection, so I think this is dead code that can be removed.

x/ts/src/telem/clockSkew.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  exceeds(threshold: CrudeTimeSpan): boolean {
```
sy-nico, line 41: The threshold should always be positive since we're comparing overall drift. Consider adding validation in the constructor (or at assignment in `checker.ts:100` and `auth.ts:73`) to enforce this

The `abs(threshold)` in the Python implementation (`x/py/x/telem/clock_skew.py:47`) works as a redundancy but shouldn't be needed if we validate at initialization.

x/py/x/telem/clock_skew.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def exceeds(self, threshold: TimeSpan) -> bool:
+        return abs(self.skew) > abs(threshold)
```
sy-nico: `abs(threshold)` here is fine as a redundancy, but the threshold should always be positive since we're comparing overall drift. Consider adding validation during the initialization.

x/py/x/telem/clock_skew.py
```diff
@@ -0,0 +1,47 @@
...
+        self._accumulated_skew += skew
+        self._n += 1
+
+    @property
+    def skew(self) -> TimeSpan:
+        if self._n == 0:
+            return TimeSpan(0)
+        return TimeSpan(self._accumulated_skew // self._n)
+
+    def exceeds(self, threshold: TimeSpan) -> bool:
```
sy-nico: [Mirrored comment
](https://github.com/synnaxlabs/synnax/pull/2188/changes#r3075174598)Python uses `TimeSpan`, but Typescript uses `CrudeTimeSpan`. We should standardize for language parity. Which of the two is more correct?

x/ts/src/telem/clockSkew.ts
sy-nico, line 41: [Mirrored comment ](https://github.com/synnaxlabs/synnax/pull/2188/changes#r3075168599)
Typescript uses `CrudeTimeSpan` but Python uses `TimeSpan`. We should standardize for language parity. Which of the two is more correct?

client/ts/src/auth/auth.spec.ts
sy-nico, line 1: Missing test coverage for the `Checker.check()` skew path (`checker.ts:119-138`). The `this.checking` guard and `clockSkewExceeded` state updates are untested.

client/ts/src/connection/checker.ts
```diff
@@ -95,14 +113,31 @@ export class Checker {
...
+        this._state.clockSkew = this.skewCalc.skew;
+        this._state.clockSkewExceeded = this.skewCalc.exceeds(this.clockSkewThreshold);
+        if (this._state.clockSkewExceeded && !this.clockSkewWarned) {
+          const direction = this.skewCalc.skew.valueOf() > 0n ? "ahead of" : "behind";
+          console.warn(
+            `Measured excessive clock skew between this host and ` +
+              `Synnax Core. This host is ${direction} Synnax Core ` +
+              `by approximately ${this.skewCalc.skew.abs().toString()}.`,
+          );
+          this.clockSkewWarned = true;
```
sy-nico: I don't think we should suppress repeated warnings here. The check only runs every 30s, so it won't be spammy.  Clock skew can cause real problems with clients writing to channels, and a one-time warning is easy to miss or dismiss. Same applies to `clockSkewWarned` in `Provider.tsx:148`.

Removing the suppression would also simplify the overall logic.

client/py/synnax/auth.py
sy-nico, line 1: When (if?) polling is added to the python client, don't suppress repeated warnings. The poll interval is slow enough that it won't be spammy, and clock skew can cause real problems with channel writes. 

[Mirrored typescript comment](https://github.com/synnaxlabs/synnax/pull/2188/changes#r3075238449)

client/ts/src/connection/checker.ts
sy-nico: `pollFrequency` is declared as `readonly` with a default value of `TimeSpan.seconds(30)`, but the constructor always overwrites it with the `pollFreq` parameter. The field initializer is dead code. Should the default be removed from the declaration, or should the constructor assignment be conditional?

sy-nico (changes requested):
Main concerns:
* Python client does not poll continuously for the clock skew
* The averaging logic will devalue newer measurements, so the calculator will become increasingly sluggish to react to the actual drift. I personally don't think you need to average the measurements (latest is fine), but an EMA or SMA with a sliding window also works.
* I don't think we should to suppress clock skew warnings since they are urgent and infrequent (every 30s)

sy-nico:
@emilbon99 , Does the C++ have continuous polling for the clock skew as well?

client/ts/src/connection/checker.ts
```diff
@@ -57,7 +57,7 @@ const createWarning = (
   clientIsNewer: boolean,
 ): string => {
   const toUpgrade = clientIsNewer ? "Core" : "client";
-  return `Synnax Core version ${nodeVersion != null ? `${nodeVersion} ` : ""}is too ${clientIsNewer ? "old" : "new"} for client version ${clientVersion}.
+  return `The Synnax core version ${nodeVersion != null ? `${nodeVersion} ` : ""}is too ${clientIsNewer ? "old" : "new"} for client version ${clientVersion}.
```
sy-nico: I thought the convention to follow was to capitalize `Core` since it is a noun.

core/pkg/api/connectivity/connectivity.go
```diff
@@ -43,6 +43,6 @@ func (s *Service) Check(
 		ClusterKey:  s.cluster.Key().String(),
 		NodeVersion: version.Get(),
 		NodeKey:     s.cluster.HostKey(),
-		NodeTime:    telem.Now(),
+		NodeTime:    telem.Now() - 5*telem.SecondTS,
```
sy-nico: What is the purpose of this 5s offset?

client/py/synnax/connection.py
```diff
@@ -0,0 +1,197 @@
...
+        self._lock = threading.Lock()
+        self._on_change_handlers: list[Callable[[State], None]] = []
+        self._stop_event = threading.Event()
+        self._version_warned = False
+        self.check()
+        self._thread = threading.Thread(target=self._run, daemon=True)
+        self._thread.start()
+
+    def stop(self) -> None:
+        self._stop_event.set()
```
sy-nico, line 114: Missing `self._thread.join()` after `self._stop_event.set()`

client/py/synnax/connection.py
```diff
@@ -0,0 +1,197 @@
...
+        poll_freq: CrudeTimeSpan = TimeSpan.SECOND * 30,
+        client_version: str = "",
+        name: str | None = None,
+        clock_skew_threshold: CrudeTimeSpan = TimeSpan.SECOND,
+    ) -> None:
+        self._client = client
+        self._poll_freq = TimeSpan(poll_freq)
+        self._client_version = client_version
+        self._name = name
+        self._clock_skew_threshold = TimeSpan(clock_skew_threshold)
```
sy-nico: I still think this should be validated or coerced to a positive value or at least throw a warning that a negative skew will be interpreted as a positive value.

` self._clock_skew_threshold = abs(TimeSpan(clock_skew_threshold))`

Same for Typescript (`cherker.ts:99`) and C++ (`checker.cpp:63`)

sy-nico (changes requested):
Real issues that should be resolved:
* `-5s` offset for NodeTime
* Missing `.join()` in python Checker thread

Nitpicks that won't block merge:
* Validation on the skew_threshold
* Capitalization question

core/pkg/api/connectivity/connectivity.go
emilbon99: Testing oops. Good catch

emilbon99:

sy-nico (approved):

PR #2205, opened 2026-04-09
SY-3990: Server-Side Metadata Import/Export

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-3990](https://linear.app/synnaxlabs/issue/SY-3990)

## Description

Adds server-side import/export infrastructure for metadata resources, starting with the
log service as a proof of concept. Includes RFC 0034 documenting the full design.

- New `imex` package with Envelope type, Importer/Exporter interfaces, and central
  registry service that routes by ontology resource type
- Hand-written frozen types and zyn schemas for log data v0 and v1, with migration
  function porting the existing TypeScript v0->v1 migration
- Log service ImporterExporter implementation with version dispatch, zyn validation,
  and persistence through existing Writer.Create path
- API endpoint with RBAC enforcement at `/api/v1/import` and `/api/v1/export`
- 21 tests across 3 suites with JSON testdata fixtures

The log changes are NOT complete yet, and should be considered as a proof-of-concept for migrations. When logs are strongly typed, this conversion will be considered complete.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds a new `imex` package providing server-side import/export infrastructure for metadata resources. It introduces an `Envelope` wire type, `Importer`/`Exporter` interfaces, a central registry `Service`, and uses the log service as a proof-of-concept implementation with version-dispatched migration, `zyn` schema validation, and HTTP/gRPC transport endpoints at `/api/v1/import` and `/api/v1/export`.

- **New `imex` package** with flat JSON wire format (promoted `version`/`type`/`name` fields), `UseNumber`-mode decode to preserve int64 precision, and a `parseJSONNumber` fix to let existing `zyn` number validators accept `json.Number` values from the envelope payload.
- **Log migrations** (`v0`/`v1`) with frozen data shapes, `zyn` schemas, and a `Migrate` dispatch function; the log `Service` now registers itself as an `ImportExporter` on startup and the existing `Writer.Create` path gains `ws == uuid.Nil` handling for workspace-free imports.
- **API layer** wires `/api/v1/import` (JSON-only request decoder) and `/api/v1/export` (JSON-only response encoder) with RBAC enforcement, and `grpc.go`/`http.go` aliases are cleaned up (removing the `fnoop`/`fhttp` import aliases in favour of direct package names).

The import/export endpoints are functional and well-tested, but the log export path has known deferred limitations around stored-data re-validation documented as future work.

The import/export endpoints are functional and the core happy-path tests now pass following the zyn/number.go fix. The most significant outstanding issues in log/imex.go were previously raised and the developer has explicitly deferred them until the log type is strongly typed in ORC. Migration dispatch, RBAC enforcement, and wire-format handling are correct for the current version landscape.

core/pkg/service/log/imex.go — export may produce non-importable envelopes for logs written by older Console clients. core/pkg/service/log/migrations/migrate.go — the default branch should be made explicit to harden the migration chain against future version additions.

| Filename | Overview |
|----------|----------|
| core/pkg/service/imex/imex.go | New Envelope type with custom UnmarshalJSON (UseNumber mode, promoted field stripping) and MarshalJSON; versionFromAny + legacyToNumeric (N.0.0 semver strings); Importer/Exporter/ImportExporter interfaces. Clean; null JSON input handled safely. |
| core/pkg/service/imex/service.go | Central registry with Import (wrapped in a transaction) and Export (no transaction). Well-structured; notFoundError scoped to field path. |
| core/pkg/service/log/imex.go | Import migrates envelope through version chain and persists via Writer.Create; Export calls l.Data.Unmarshal then d.ToMap(). Known limitations explicitly deferred per developer reply. |
| core/pkg/service/log/migrations/migrate.go | Version dispatch: version > LatestVersion → unsupported error; case v1.Version → parse v1; default → parse v0 + lift. Default branch should be made explicit for safety when v2 is added. |
| core/pkg/service/log/migrations/v1/data.go | ToMap() returns d.Channels as typed []ChannelEntry (not []any), which breaks zyn.Array validation on in-process round-trips (previously flagged). Fine for over-the-wire use. |
| x/go/zyn/number.go | Adds parseJSONNumber helper converting json.Number to int64/uint64/float64 before existing fast-path assertions. Fixes the XIt-skipped test issue; fallback ordering is correct. |
| core/pkg/api/imex/imex.go | API-layer thin wrapper: RBAC enforced for both Import (create) and Export (retrieve); delegates to internal imex.Service. |
| core/pkg/api/http/http.go | Removes fhttp alias (cleanup), adds JSON-only Import/Export endpoints with asymmetric codec options. |
| core/pkg/service/log/writer.go | Create gains ws == uuid.Nil guard to skip workspace ParentOf relationship on import path; named return refactors are safe mechanical cleanups. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client
    participant HTTP as HTTP Transport
    participant APII as api/imex.Service
    participant RBAC as rbac.Service
    participant SvcImEx as service/imex.Service
    participant LogImEx as log.Service
    participant Migrate as migrations.Migrate
    participant Writer as log.Writer
    participant DB as gorp.DB

    Note over Client,DB: Import Flow
    Client->>HTTP: POST /api/v1/import JSON Envelope
    HTTP->>APII: Import(ctx, Envelope)
    APII->>SvcImEx: ImporterType(env.Type)
    SvcImEx-->>APII: ontology.ResourceTypeLog
    APII->>RBAC: Enforce ActionCreate log
    RBAC-->>APII: ok
    APII->>SvcImEx: Import(ctx, Envelopes)
    SvcImEx->>DB: WithTx
    SvcImEx->>LogImEx: Import(ctx, tx, Envelope)
    LogImEx->>Migrate: Migrate(version, data)
    Migrate-->>LogImEx: v1.Data
    LogImEx->>Writer: Create(ctx, uuid.Nil, log)
    Writer->>DB: table.Create + DefineResource
    LogImEx-->>SvcImEx: key
    APII-->>HTTP: ImportResponse Key
    HTTP-->>Client: 200 JSON

    Note over Client,DB: Export Flow
    Client->>HTTP: POST /api/v1/export type+key
    HTTP->>APII: Export(ctx, ontology.ID)
    APII->>RBAC: Enforce ActionRetrieve resource
    RBAC-->>APII: ok
    APII->>SvcImEx: Export(ctx, IDs)
    SvcImEx->>LogImEx: Export(ctx, key)
    LogImEx->>DB: Retrieve log by key
    DB-->>LogImEx: Log with msgpack data
    LogImEx->>LogImEx: Unmarshal to v1.Data then ToMap
    LogImEx-->>SvcImEx: Envelope version 1
    APII-->>HTTP: flat JSON Envelope
    HTTP-->>Client: 200 JSON
```

1. `core/pkg/service/imex/service.go`, line 960 ([link](https://github.com/synnaxlabs/synnax/blob/b73429827cbc74708f3129477b408b955b3fc506/core/pkg/service/imex/service.go#L960)) 

   <a href="#">P1</a> **Export stamps server version instead of schema version, breaking future dispatch**

   `env.Version = s.exportVersion` overwrites every envelope with the server's numeric version (e.g. 54 for v0.54.x). On re-import, the dispatch in `migrateData` uses range-based comparisons like `case version >= v1.Version (5)`. This works today because no v2 exists, but once `v2.Version = 6` is introduced, any file exported from v0.54.x (version=54) will satisfy `version >= v2.Version (6)` and be routed to the v2 handler — even though its JSON was serialized in v1 format. The RFC (section 4.3.0) explicitly requires that each handler stamps its own latest schema version on export, and that the central `imex.Service` does not stamp version. The log `Export` should set `env.Version = v1.Version` before returning, so that re-imported files are always routed to the correct schema handler regardless of which server version produced the export.

2. `core/pkg/version/get.go`, line 100-117 ([link](https://github.com/synnaxlabs/synnax/blob/af29f9dcd6d170abe897d164ad55d71f65cf9e68/core/pkg/version/get.go#L100-L117)) 

   <a href="#">P1</a> `Numeric()` silently returns 0 when `Prod()` returns `"unknown"` (e.g. when the embedded VERSION file cannot be read). With `exportVersion = 0`, every exported envelope carries `Version = 0`. On re-import `0 >= v1.Version (5)` is false, so the file is routed to the v0 handler, which attempts to decode `channels` as `[]int` while the stored data is in v1 format — producing a decode error and making the export non-importable. A sentinel value that is larger than all current schema version constants would be safer, but at minimum the caller should detect and reject the zero case.
</details>

<sub>Reviews (29): Last reviewed commit: ["Refactor ImportResponse to be a struct"](https://github.com/synnaxlabs/synnax/commit/e52a6463dd60c0e2b40adb00fc75bb18e15a7bc9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27884061)</sub>


emilbon99:

core/pkg/service/log/imex.go
```diff
@@ -0,0 +1,70 @@
...
+		if err := l.Data.Unmarshal(&d); err != nil {
+			return imex.Envelope{}, err
+		}
+	}
+	return imex.Envelope{
+		Version: migrations.LatestVersion,
+		Type:    string(ontology.ResourceTypeLog),
+		Name:    l.Name,
+		Data:    d.ToMap(),
+	}, nil
```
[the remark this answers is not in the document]
pjdotson: Ignoring this. Code for log migrations will be substantially improved when they are
strongly typed on the Core, including running a migration so that all stored data is up
to the latest version.

pjdotson:
@greptile /rerun

core/pkg/api/imex/imex.go
```diff
@@ -0,0 +1,85 @@
...
+	cfg, err := xconfig.New(config.DefaultLayerConfig, cfgs...)
+	if err != nil {
+		return nil, err
+	}
+	return &Service{internal: cfg.Service.ImEx, access: cfg.Service.RBAC}, nil
+}
+
+type (
+	ImportRequest  = imex.Envelope
+	ImportResponse = string
```
emilbon99: This is the only place in the codebase where we're not returning a struct. We're ok with that?

core/pkg/service/imex/imex.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	if v, ok := m["name"]; ok {
+		s, ok := v.(string)
+		if !ok {
+			return errors.Newf("name must be a string, got %T", v)
+		}
+		e.Name = s
+		delete(m, "name")
+	}
+	if len(m) > 0 {
```
emilbon99, line 127: What happens here if you're taking in an `Envelope` with a `Data` field? In that case the data is stored in that field, not in the remainder of the map keys right? Am I missing something?

core/pkg/service/imex/imex_suite_test.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+var (
+	db  *gorp.DB
+	svc *imex.Service
+	ts  *testService
+)
+
+var _ = BeforeSuite(func(ctx SpecContext) {
+	db = gorp.Wrap(memkv.New())
```
emilbon99: There are two nice test utilities I wrote:

`DeferClose` automatically closes an `io.Closer`
`MustOpen` automatically opens an opener, asserts it succeeds, and then makes a call to `DeferClose`. 

Would be good to use these utilities where relevant in tests.

x/go/zyn/number.go
```diff
@@ -646,3 +656,31 @@ func unsignedIntegerTooLargeError() error {
...
 }
+
+// unboxJSONNumber converts an encoding/json json.Number — emitted by decoders running
+// in UseNumber mode to preserve full int64/uint64 precision — into a typed numeric
+// value the rest of NumberZ already understands. Tries int64 first to keep integral
+// precision past 2^53, falls back to uint64 for values above int64.MaxInt64, and
+// finally to float64 for fractional or otherwise non-integer numbers. A non-numeric
+// json.Number string surfaces as a conversion error. Non-json.Number inputs are
+// returned unchanged.
+func unboxJSONNumber(data any) (any, error) {
```
emilbon99: Is there a reason this is called `unbox` and not `parse`? Nitpick so address if you wish.

emilbon99 (commented):
Overall looks good. I think there are still some unanswered questions about long term future of zyn, where we could generate code, and how to make these import/export related migrations be easy to write. 

That said, one thing I would make sure you are confident about is the import and export format. Double check that you are happy with that, because it's going to be a PITA to correct later.

It doesn't show up as a problem in this PR, but you are going to eventually have to contend with potential conflicts that arise from the oracle internal way of versioning (based on core no.) and the incrementing number you have. For example, the `v2` of log for imex may live in folder `v58` or something.

core/pkg/service/imex/imex.go
pjdotson: Well in general `UnmarshalJSON` is being used on empty `Envelopes`. so this is the flow
of what happens:
- unmarshal JSON bytes into a `map[string]any` (needs the JSON to be a flat object)
- remove `version`, `type`, and `name` from that object into the Envelope struct field
- all other keys go into the `Envelope.Data` which is `map[string]any`.

emilbon99 (commented):
Approved

PR #2216, opened 2026-04-12
SY-4056: Gorp Indexes

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-####]()

## Basic Readiness

<h3>Greptile Summary</h3>

This PR introduces a full secondary-index system for gorp — `LookupIndex` (exact-match) and `SortedIndex` (ordered cursor), with per-transaction read-your-own-writes via a `deltaOverlay`, a `Table` abstraction that manages populate lifecycle and observer-driven sync, and an oracle code-generator extension that emits the index scaffolding from `@index`-annotated fields.

- **New core types**: `LookupIndex`, `SortedIndex`, `delta`, `deltaOverlay`, `Table`, `TableConfig` — together they cover populate, flush-on-commit, and concurrent read safety.
- **Filter system overhaul**: `Filter` now carries `keys`, `membership`, `eval`, `raw`, and `resolve` fields; `And`/`Or`/`Not` combinators materialize per-Exec fresh copies to eliminate the data races noted in earlier review rounds; `resolveFilter` swallows `ErrIndexInvalid` and falls back to sequential scan.
- **oracle plugin extended**: `@index` annotation now drives generation of `newIndexes()`, the `indexes` struct, and index-routed `MatchX`/`MatchXs` filter functions in `retrieve.gen.go`.

The index infrastructure is largely correct, with prior-round races and missing fallbacks now fixed, but two confirmed issues from previous reviews remain in the changed code: execOrdered ignores r.offset, and WhereRaw filters are never applied in either execOrdered or execKeys.

Both execOrdered's offset silently returning wrong pages and the matchRaw skip in execKeys/execOrdered affect query correctness on the changed code paths. They were flagged in earlier rounds and have not been addressed in this PR.

x/go/gorp/retrieve.go — execOrdered (offset handling) and execKeys (raw-byte pre-screen) need attention before the index feature is considered complete.

| Filename | Overview |
|----------|----------|
| x/go/gorp/index.go | New file: LookupIndex and SortedIndex implementations with populate lifecycle, committed-state management, and per-tx staging via deltaOverlay. SortedIndex.Get blocks on mu.RLock rather than on populateDone channel, unlike LookupIndex.Get — see comment. |
| x/go/gorp/delta.go | New file: per-tx delta state for secondary indexes. stageSet/stageDelete, merge semantics, and flush-on-commit look correct. |
| x/go/gorp/filter.go | Major rewrite of Filter, And, Or, Not. Previous data-race concerns (Or/Not mutating shared captured state) are now fixed via fresh-copy materialization. Or now sets a construction-time fallback eval before the resolver branch. evalChild correctly checks child.keys for MatchKeys-style children. |
| x/go/gorp/retrieve.go | execOrdered now calls runValidators (fixing the previous omission); match() correctly checks r.filter.keys for index-backed filters. Known unfixed issues from prior rounds: execOrdered ignores r.offset, and matchRaw is never called from execOrdered or execKeys. |
| x/go/gorp/table.go | New Table/TableConfig types with background populate goroutine, observer attachment, and lifecycle management. runPopulate cannot surface decode errors from OpenNexter (closer is io.Closer without access to Iterator.err) — issue from prior round still unfixed. |
| x/go/gorp/order_by.go | walkOrder now checks populateErrWrapped before performing binary search, fixing the prior bug where an unsorted entries slice could produce incorrect cursor bounds on populate failure. |
| oracle/plugin/go/query/query.go | Extended to emit indexes struct, newIndexes(), and index-routed filter functions for @index-annotated fields. hasVariadicFilter check correctly excludes index-routed filters. |
| x/go/gorp/entries.go | Key type changed to require comparable (needed for map keys in delta), bytes key kind removed, Entries refactored from pointer to value. Changes look consistent. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Caller
    participant Table
    participant Index
    participant deltaOverlay
    participant delta
    participant Writer

    Note over Table,Index: OpenTable — populate phase
    Table->>Index: populate() → acquires mu.Lock
    Table->>Table: attachIndexObserver (blocks on mu until finish)
    Table->>Table: sCtx.Go(runPopulate)
    Table-->>Caller: "return *Table (populate running in background)"

    Note over Caller,delta: Write tx — stageSet
    Caller->>Writer: Create/Update/Delete .Exec(ctx, tx)
    Writer->>Index: stageSet(tx, entry)
    Index->>deltaOverlay: stage(tx, key, value)
    deltaOverlay->>delta: stageSet(key, value)

    Note over Caller,delta: Commit — flush delta to committed state
    Caller->>Writer: tx.Commit(ctx)
    Writer->>delta: "onCleanup(committed=true)"
    delta->>Index: flush(d) → acquires mu.Lock

    Note over Caller,Index: Retrieve — resolve filter
    Caller->>Index: Filter(values...) → resolve closure
    Index->>Index: waitPopulated(ctx)
    Index->>Index: Get(tx, values) → mu.RLock
    Index->>deltaOverlay: resolve(tx, committed, values)
    deltaOverlay->>delta: merge(committedKeys, values)
    delta-->>Index: effectiveKeys (RYOW)
    Index-->>Caller: "resolved{keys, build, eval}"
```

1. `x/go/gorp/retrieve.go`, line 1276-1283 ([link](https://github.com/synnaxlabs/synnax/blob/22dfa532a155e8750107628ca99e3864ade92ce5/x/go/gorp/retrieve.go#L1276-L1283)) 

   <a href="#">P1</a> **`WhereRaw` filter silently ignored on the `OrderBy` path**

   `execOrdered` calls `r.match(ctx, &e, nil, nil)` with nil raw bytes. Inside `match()`, when the filter has only `Raw` set (produced by `WhereRaw`/`MatchRaw`) — `Keys` is nil and `Eval` is nil — the function falls through to `return true, nil`, so the raw predicate is never invoked. Every entry passes the filter regardless of what `WhereRaw` specifies.

   `execFilter` and `execKeys` both call `r.matchRaw(rawKey, rawValue)` first, so those paths enforce the raw filter correctly. But `execOrdered` skips `matchRaw` entirely and only calls `match()`, which doesn't re-invoke `Raw` when `Eval == nil` (it trusts the comment "Raw was already enforced by matchRaw at the pre-decode pre-screen" — an assumption that does not hold here). Any `WhereRaw(...).OrderBy(...)` query silently returns unfiltered results.

   Since `GetMany` doesn't surface raw bytes, the fix is to either run a raw-byte scan for the restricted key set inside `execOrdered` (analogous to `execKeys`' `getMany` lower-level call), or document that `WhereRaw` is unsupported with `OrderBy` and return an explicit error when both are set.

2. `x/go/gorp/retrieve.go`, line 395-400 ([link](https://github.com/synnaxlabs/synnax/blob/718972c90c4667a6b5c4c5638ad3338e94efa6ab/x/go/gorp/retrieve.go#L395-L400)) 

   <a href="#">P1</a> `execKeys` never calls `matchRaw`, so any raw-byte predicate set on the filter is silently ignored when the key-based dispatch path is taken. A caller who composes `MatchKeys` with `WhereRaw` (e.g. `r.Where(MatchKeys(k)).WhereRaw(fn)`) will receive the entry regardless of what `fn` returns, because `And(MatchKeys, MatchRaw)` has `eval=nil` (both children have no eval), so `match()` falls through to `return true, nil` after the key check passes. The raw key bytes are available as `reader.keyCodec.encode(k)` and can be passed to `matchRaw` (and also forwarded to `match` instead of the current `nil`).
</details>

<sub>Reviews (27): Last reviewed commit: ["Clean up Search test&#39;s committed views t..."](https://github.com/synnaxlabs/synnax/commit/90d91b47f13b6b3ecc2ecc6f0dbd58034836a2e4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28114652)</sub>


emilbon99:

emilbon99:

x/go/gorp/retrieve.go
```diff
@@ -201,21 +309,24 @@ func (r Retrieve[K, E]) Exists(ctx context.Context, tx Tx) (bool, error) {
...
 func (r Retrieve[K, E]) Count(ctx context.Context, tx Tx) (count int, err error) {
 	checkForNilTx("Retriever.Count", tx)
+	if err := r.resolveFilter(ctx, tx); err != nil {
+		return 0, err
+	}
 	if r.HasFilterKeys() {
-		e := make([]E, 0, len(r.filter.keys))
-		r.entries.bindMultiple(&e)
-		if _, err := r.execKeys(ctx, tx); err != nil {
+		r.entries.bindMultiple(new(make([]E, 0, len(r.filter.keys))))
```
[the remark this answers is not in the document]
emilbon99: Support for this added in go 1.26

x/go/gorp/index_storage.go
```diff
@@ -0,0 +1,484 @@
...
+//nolint:unused
+func (s *mapLookupStorage[K, V]) stageSet(tx Tx, key K, value V) {
+	if tx.txIdentity() == nil {
+		s.commitMu.Lock()
+		defer s.commitMu.Unlock()
+		s.put(key, value)
+		return
+	}
+	s.overlay.stage(tx, key, value)
+}
```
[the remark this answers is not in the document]
emilbon99: `s.commitMu` and `l.mu` are the same mutex — every storage constructor takes `&l.mu` (see `index.go:116`, `118`, `138`). The observer path (`Lookup.set`) holds `l.mu`, and `stageSet(nil-tx)` re-acquires that same lock via `commitMu`. They serialize on the same `*sync.RWMutex`.

The pattern differs from `Sorted` (which has no separate storage mutex and locks `s.mu` directly inside `s.set`), but the lock identity is the same — passed down by pointer rather than reached through the outer struct. The `lookupStorage` doc comment makes this contract explicit.

x/go/gorp/filter.go
```diff
@@ -100,105 +343,298 @@ func Or[K Key, E Entry[K]](filters ...Filter[K, E]) Filter[K, E] {
...
-		out   []K
-		count = 0
-	)
-	for _, f := range filters {
-		if f.keys == nil {
+// anyHasResolver reports whether any child filter carries a deferred
+// resolver.
+func anyHasResolver[K Key, E Entry[K]](filters []Filter[K, E]) bool {
+	for _, child := range filters {
+		if child.resolve != nil {
```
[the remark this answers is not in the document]
emilbon99: Good catch — bug confirmed and fixed.

I diverged slightly from the suggested `evalChild` swap: `And` sets `f.raw` at the top level (run pre-decode via `Retrieve.matchRaw`), and `evalChild` falls through to `f.raw` when `f.eval` is nil, so using it directly would double-run raw for any `MatchRaw` child of `And`. `Or.eval` doesn`t hit this because it only sets `f.raw` when every child is raw-only.

Targeted fix: inline a `child.keys` check at the top of `And.eval`. Handles the `MatchKeys` constraint in the fallback path without touching the raw pipeline. Regression test added under "Populate failure" that reproduces exactly the shape you described (`And(idx.Filter("alpha"), MatchKeys(1))` after a populate failure) — fails before the fix, passes after.

x/go/gorp/create.go
```diff
@@ -18,14 +18,22 @@ import (
 
 // Create is a query that creates Entries in the DB.
 type Create[K Key, E Entry[K]] struct {
-	entries   *Entries[K, E]
-	onUpdate  onUpdate[K, E]
+	// entries is the set of entries the query will write.
+	entries Entries[K, E]
+	// onUpdate is the chain of MergeExisting filters applied when an
+	// entry with a matching key already exists. Empty disables the
```
pjdotson: what is `Empty`?

x/go/gorp/delta.go
```diff
@@ -0,0 +1,205 @@
...
+type delta[SK comparable, V comparable] struct {
+	// state is the authoritative per-tx view: if a key is in state,
+	// the committed index's view of that key is ignored during resolve.
+	state map[SK]deltaEntry[V]
+	// forward is the value-to-keys inverse, mirroring non-deleted
+	// entries in state for O(1) resolve.
+	forward map[V]set.Set[SK]
+}
+
+//nolint:unused
```
pjdotson, line 39: why are there all these `nolint:unused` comments? either they should be used or they
should be deleted.

pjdotson (changes requested):
There are a bunch of `nolint:unused` comments. Can you remove those and make sure the PR is actually ready for review?

x/go/gorp/delta.go
emilbon99: There's no way to solve this, even with using an interface assertion. I don't know why.

x/go/gorp/create.go
emilbon99: Refactored this comment to work properly.

x/go/gorp/delta.go
emilbon99: You're welcome to give it a shot but I've tried a bunch of times to get around it.

x/go/gorp/table.go
```diff
@@ -15,40 +15,121 @@ import (
...
+	Indexes []Index[K, E]
+	// Instrumentation is used for migration and populate logging.
 	alamos.Instrumentation
 }
 
 // Table provides a strongly typed interface for a specific entry type within a gorp DB.
 type Table[K Key, E Entry[K]] struct {
-	DB        *DB
+	// DB is the gorp DB the Table reads and writes through.
+	DB *DB
```
pjdotson: Is there a good reason for `DB` to be public on the `Table`?

x/go/gorp/table.go
```diff
@@ -63,34 +144,147 @@ func OpenTable[K Key, E Entry[K]](
...
+	)
+	return t, nil
+}
+
+// runPopulate scans the table once and feeds every registered index
+// from the same scan. It always returns nil: populate failures are
+// recoverable (queries fall back to sequential scan) and must not be
+// treated as routine failures by signal. The terminal scan error is
+// stored on t.populateErr and passed to each index's finish closure.
+func (t *Table[K, E]) runPopulate(
```
pjdotson, line 191: if it always returns `nil` why return an `error` at all?

x/go/gorp/table.go
```diff
@@ -15,40 +15,121 @@ import (
...
+	// keyPrefix is the gorp key prefix for entry type E.
 	keyPrefix []byte
+	// indexes is the set of secondary indexes registered on the Table.
+	indexes []Index[K, E]
+	// disconnectObserver releases the index observer subscription. Nil
+	// when no observer was installed.
+	disconnectObserver func()
+	// populateCtx hosts the background populate routine. Nil when the
+	// table has no indexes.
+	populateCtx signal.Context
```
pjdotson: is there a reason to store a whole context in a struct instead of just a channel for
populate being done? is a bad practice to store `context.Context` onto structs that I
would like to move away from (https://go.dev/blog/context-and-structs).

x/go/gorp/filter.go
```diff
@@ -9,89 +9,340 @@
...
+}
+
+// linearContainsKey reports whether k is in keys via O(n) scan.
+// Handles []byte-shaped K via bytes.Equal since []byte is not ==
+// comparable; other K kinds dispatch through interface equality.
+func linearContainsKey[K Key](keys []K, k K) bool {
+	if len(keys) == 0 {
+		return false
+	}
+	if reflect.TypeOf(keys[0]).Kind() == reflect.Slice {
```
pjdotson: are there worries about using `reflect` here?

x/go/gorp/filter.go
```diff
@@ -9,89 +9,340 @@
...
+		for _, fk := range keys {
+			if bytes.Equal(reflect.ValueOf(fk).Bytes(), probe) {
+				return true
+			}
+		}
+		return false
+	}
+	probe := any(k)
+	for _, fk := range keys {
+		if any(fk) == probe {
```
pjdotson: why do you need these `any`s here? isn't `k` of type `K` and `fk` also of type `K`?

x/go/gorp/filter.go
```diff
@@ -9,89 +9,340 @@
...
 	return f
 }
 
-// Or returns a filter that matches when ANY child matches. Short-circuits on
-// the first true.
+// Or returns a filter that matches when ANY child matches.
+//
+// Composition rules:
+//   - keys are unioned only when every child has keys; a single
+//     unbounded child collapses the union to unbounded.
```
pjdotson, line 275: typo

x/go/gorp/retrieve.go
```diff
@@ -109,17 +127,31 @@ func (r Retrieve[K, E]) Offset(offset int) Retrieve[K, E] {
...
+// OrderBy walks the results in the order defined by the given OrderQuery,
+// typically obtained from SortedIndex.Ordered(dir) (optionally with .After(cursor)
+// chained for cursor-based pagination). Combine with Limit for paged walks.
+func (r Retrieve[K, E]) OrderBy(o OrderQuery[K, E]) Retrieve[K, E] {
+	r.orderBy = o
+	return r
+}
+
+// HasOrderBy returns true if OrderBy was set on the query.
+func (r Retrieve[K, E]) HasOrderBy() bool { return r.orderBy != nil }
```
pjdotson: Why does this need to be public?

x/go/gorp/retrieve_test.go
```diff
@@ -148,6 +148,50 @@ var _ = Describe("Retrieve", func() {
...
 
+		Context("Composed with another filter", func() {
+			// When MatchKeys is layered with another filter (Match, MatchRaw,
+			// WherePrefix, etc.), Exec must treat missing keys as empty slots
+			// rather than fail with ErrNotFound. The ontology Retrieve
+			// pipeline depends on this: between clauses it injects
+			// MatchKeys(nextIDs) on top of WhereTypes / WherePrefix, and a
+			// transient miss (orphaned relationship, freshly-deleted
+			// resource) must not blow up the whole traversal.
+			DescribeTable("Should NOT return ErrNotFound when MatchKeys is combined with a filter and some keys are missing",
```
pjdotson, line 159: So this is the API:
- Retrieve.Where(MatchKeys).Where(MatchNames) == Keys OR Names
which is equivalent to `Where(Or(MatchKeys,MatchNames))`
so should we allow two ways of doing the same thing?

x/go/gorp/order_by.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Direction is the iteration direction used by a SortedIndex index when it is
+// consumed via Retrieve.OrderBy.
+type Direction uint8
+
+const (
+	// Asc walks the sorted index from smallest to largest value.
+	Asc Direction = iota
+	// Desc walks the sorted index from largest to smallest value.
+	Desc
```
pjdotson: should be `DirectionAsc` and `DirectionDesc`

x/go/gorp/index.go
```diff
@@ -0,0 +1,494 @@
...
+// GetTx is the tx-aware counterpart to Get. It returns the primary keys
+// of entries whose indexed field matches any of the provided values,
+// merging committed index state with any per-tx delta staged against
+// the open transaction. When tx has no per-tx scoping or no staged
+// mutations for this index, it returns the same result as Get. Use
+// GetTx when consuming keys directly outside of Retrieve — e.g. graph
+// traversal helpers that probe the index for candidate IDs. Returns
+// ErrIndexInvalid when population failed; see Get for the failure-mode
+// contract.
+func (l *LookupIndex[K, E, V]) GetTx(tx Tx, values ...V) ([]K, error) {
```
pjdotson: why not just have a single `Get` method that allows passing `nil` in for `Tx`? generally
would like to only expose methods w/ transaction arguments across the whole Core so that
callers have to explicitly pass a `nil` transaction in case they are composing several
operations at once.

x/go/gorp/index.go
```diff
@@ -0,0 +1,494 @@
...
+		},
+		eval: func(_ Context, e *E, _, _ []byte) (bool, error) {
+			return slices.Contains(captured, s.extract(e)), nil
+		},
+	}
+}
+
+// GetTx is the tx-aware counterpart to Get. See LookupIndex.GetTx for
+// semantics.
+func (s *SortedIndex[K, E, V]) GetTx(tx Tx, values ...V) ([]K, error) {
```
pjdotson: see comment above

pjdotson (changes requested):
Overall very good. You should file an issue for the false positives on unused.

x/go/gorp/table.go
emilbon99: I removed and replaced with a channel, but I definitely disagree with your take that storing the signal context is bad here. It really depends on how you use it. 

I did a lot of research on the context field in struct and have my thoughts.

x/go/gorp/retrieve_test.go
emilbon99: Yes, there are two ways of doing the same thing, but combined where chains have `AND`, not `OR` semantics. The alternative here would be: (a) panic when Where is called multiple times (BAD) or (b) refactor this to remove the And function (bad because it hurts wire protocol for API), or (c) make it so that these structs no longer use a builder but an options pattern. (c) is a huge refactor that I just don't think is worth it.

So yes, two ways of doing the same thing. Is it ideal? No. Is it catastrophic? No. Is the refactor cost pretty big? Yes. 

Also this is what most other ORMs already do so....

emilbon99:
Refactored `ontology.Relationship` so it uses a string instead of a byte slice (format compatible so no migration needed). This allowed for removal and simplification of a lot of the gorp code.

x/go/gorp/table.go
emilbon99: Corrected

x/go/gorp/table.go
emilbon99: Made private

pjdotson (approved):

PR #2217, opened 2026-04-13
SY-4059: Switch to Length-Prefixed Format for Variable-Length Series

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4059](https://linear.app/synnaxlabs/issue/SY-4059)

## Description

Switches the encoding format for variable-length Series types (`StringT`, `JSONT`,
`BytesT`) from newline-delimited to uint32-length-prefixed across all four languages
(Go, Python, TypeScript, C++).

The previous newline-delimited encoding (`sample\nsample\n`) is broken when sample data
contains literal newlines, which is common in JSON and multi-line strings. The new format
(`[uint32 LE length][bytes][uint32 LE length][bytes]...`) is binary-safe with 4 bytes of
overhead per sample.

Changes per language:
- **Go** (`x/go/telem/`): `marshalVariable`, `unmarshalVariable`, `Len`, `Samples`,
  `At`, `Downsample`
- **Python** (`x/py/x/telem/`): `__len__`, constructors, `__variable_getitem__`,
  `__iter__variable`
- **TypeScript** (`x/ts/src/telem/`): constructors, `calculateCachedLength`,
  `atVariable`, `StringSeriesIterator`, `toStrings`, `writeVariable`
- **C++** (`x/cpp/telem/`): all string/JSON constructors, `strings()`, `at<string>()`,
  `write()`, `fill_from`, proto deserializer

The codec layer (`core/pkg/distribution/framer/codec/`) required no changes since it
transmits `Series.Data` bytes verbatim using byte-length, not sample-count.

This is Phase 1 of the variable-length DB RFC (`docs/rfcs/cesium-variable-db.md`).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR replaces newline-delimited encoding for variable-length series (`StringT`, `JSONT`, `BytesT`) with a uint32 little-endian length-prefixed format across Go, Python, TypeScript, and C++. The change is motivated by newlines appearing inside JSON and multi-line strings, which broke the old encoding. The implementation is consistent and complete across all four languages, signal encoding in `core/` and `cesium/` is updated, and tests are updated to match the new byte layout.

Safe to merge; the encoding change is consistent and correct across all four languages, and all remaining findings are P2 style/defensive-coding notes.

All P0/P1 concerns are absent. The two findings are: (1) Python __len__ counts a truncated-payload sample while Go/TS skip it — only observable on malformed data; (2) C++ helpers are named _le but use memcpy, safe on all LE target platforms. Neither affects correctness under normal operation.

x/py/x/telem/series.py (missing truncation guard) and x/cpp/telem/series.h (endian-assumption in helper names)

| Filename | Overview |
|----------|----------|
| x/go/telem/series.go | Replaces newline scanning with uint32-LE prefix scanning in Len, Samples, At, and Downsample; bounds checks are consistent and correct. |
| x/go/telem/series_factory.go | Adds MarshalVariableSample helper and rewrites marshalVariable/unmarshalVariable to use 4-byte LE length prefixes; logic is correct. |
| x/py/x/telem/series.py | Converts all encode/decode paths to uint32-LE prefix format; __len__ lacks the truncation guard that Go and TypeScript have, causing inconsistent sample count on malformed data. |
| x/ts/src/telem/series.ts | Rewrites all variable-length paths (constructors, writeVariable, calculateCachedLength, atVariable, toStrings, StringSeriesIterator) to use uint32-LE prefixes; cached-index logic is correctly updated. |
| x/cpp/telem/series.h | Replaces NEWLINE_TERMINATOR with uint32 prefix helpers; fixes write() to start at cached_byte_size; helpers use memcpy (native-endian) despite being named _le, which is safe only on LE targets. |
| cesium/control.go | Adds DecodeControlUpdate using telem.At(0) and updates EncodeControlUpdate to use MarshalVariableSample; correct. |
| core/pkg/distribution/ontology/signals/signals.go | Replaces newline-appended encoding in EncodeID/EncodeIDs and newline-scanning decode loops with MarshalVariableSample / UnmarshalSeries calls; correct. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Writer as Producer (any language)
    participant Buf as Series.Data ([]byte)
    participant Reader as Consumer (any language)

    Note over Writer,Buf: Encoding (new format)
    Writer->>Buf: write uint32-LE length (4 bytes)
    Writer->>Buf: write sample bytes (length bytes)
    Note over Writer,Buf: repeat per sample

    Note over Buf,Reader: Decoding
    Reader->>Buf: read uint32-LE length
    Reader->>Buf: read length bytes → sample
    Note over Buf,Reader: repeat until end of buffer

    Note over Writer,Reader: Go: binary.LittleEndian.Uint32<br/>Python: int.from_bytes(..., 'little')<br/>TS: DataView.getUint32(offset, true)<br/>C++: memcpy (native-LE on all targets)
```

1. `x/ts/src/telem/series.ts`, line 609-612 ([link](https://github.com/synnaxlabs/synnax/blob/b213fcb07ea85743779ff99760fd394ef65dffa0/x/ts/src/telem/series.ts#L609-L612)) 

   <a href="#">P2</a> **Stale JSDoc comment**

   The comment still says "calculated by counting newlines" but the implementation now uses length-prefixed format scanning.
</details>

<sub>Reviews (2): Last reviewed commit: ["Use telem JSON utilities in EncodeContro..."](https://github.com/synnaxlabs/synnax/commit/c992194bd9dd0e588e94fdc482f0acddf9cb94c2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28238775)</sub>

### DIFF
--- a/cesium/control.go
+++ b/cesium/control.go
@@ -149,12 +149,18 @@ func (db *DB) ControlUpdateToFrame(ctx context.Context, u ControlUpdate) Frame {
 	return telem.UnaryFrame(db.mu.digests.key, d)
 }
 
+// EncodeControlUpdate encodes a ControlUpdate into a single-sample Series. The content
+// is JSON, but the DataType is set to StringT because the control digest channel is
+// created as a StringT virtual channel in core/pkg/distribution/layer.go. Both locations
+// must be updated together if the type is ever changed to JSONT.
 func EncodeControlUpdate(_ context.Context, u ControlUpdate) (telem.Series, error) {
 	s, err := telem.NewJSONSeriesV(u)
 	s.DataType = telem.StringT
 	return s, err
 }
 
+// DecodeControlUpdate decodes a ControlUpdate from a single-sample Series. See
+// EncodeControlUpdate for why the Series has DataType StringT despite containing JSON.
 func DecodeControlUpdate(s telem.Series) (ControlUpdate, error) {
 	updates, err := telem.UnmarshalJSONSeries[ControlUpdate](s)
 	if err != nil || len(updates) == 0 {
--- a/cesium/control.go
+++ b/cesium/control.go
@@ -152,7 +152,10 @@ func (db *DB) ControlUpdateToFrame(ctx context.Context, u ControlUpdate) Frame {
 
 func EncodeControlUpdate(ctx context.Context, u ControlUpdate) (s telem.Series, err error) {
 	s.DataType = telem.StringT
-	s.Data, err = (json.Codec).Encode(ctx, u)
-	s.Data = append(s.Data, '\n')
-	return s, err
+	raw, err := (json.Codec).Encode(ctx, u)
+	if err != nil {
+		return s, err
+	}
+	s.Data = telem.MarshalVariableSample(raw)
+	return s, nil
 }
--- a/cesium/control.go
+++ b/cesium/control.go
@@ -11,6 +11,7 @@ package cesium
 
 import (
 	"context"
+	stdjson "encoding/json"
 
 	"github.com/google/uuid"
 	"github.com/synnaxlabs/cesium/internal/channel"
@@ -158,4 +159,9 @@ func EncodeControlUpdate(ctx context.Context, u ControlUpdate) (s telem.Series, 
 	}
 	s.Data = telem.MarshalVariableSample(raw)
 	return s, nil
+}
+
+func DecodeControlUpdate(s telem.Series) (ControlUpdate, error) {
+	var u ControlUpdate
+	return u, stdjson.Unmarshal(s.At(0), &u)
 }
--- a/cesium/control.go
+++ b/cesium/control.go
@@ -11,14 +11,12 @@ package cesium
 
 import (
 	"context"
-	stdjson "encoding/json"
 
 	"github.com/google/uuid"
 	"github.com/synnaxlabs/cesium/internal/channel"
 	"github.com/synnaxlabs/cesium/internal/control"
 	"github.com/synnaxlabs/x/confluence"
 	xcontrol "github.com/synnaxlabs/x/control"
-	"github.com/synnaxlabs/x/encoding/json"
 	"github.com/synnaxlabs/x/errors"
 	"github.com/synnaxlabs/x/signal"
 	"github.com/synnaxlabs/x/telem"
@@ -151,17 +149,16 @@ func (db *DB) ControlUpdateToFrame(ctx context.Context, u ControlUpdate) Frame {
 	return telem.UnaryFrame(db.mu.digests.key, d)
 }
 
-func EncodeControlUpdate(ctx context.Context, u ControlUpdate) (s telem.Series, err error) {
+func EncodeControlUpdate(_ context.Context, u ControlUpdate) (telem.Series, error) {
+	s, err := telem.NewJSONSeriesV(u)
 	s.DataType = telem.StringT
-	raw, err := (json.Codec).Encode(ctx, u)
-	if err != nil {
-		return s, err
-	}
-	s.Data = telem.MarshalVariableSample(raw)
-	return s, nil
+	return s, err
 }
 
 func DecodeControlUpdate(s telem.Series) (ControlUpdate, error) {
-	var u ControlUpdate
-	return u, stdjson.Unmarshal(s.At(0), &u)
+	updates, err := telem.UnmarshalJSONSeries[ControlUpdate](s)
+	if err != nil || len(updates) == 0 {
+		return ControlUpdate{}, err
+	}
+	return updates[0], nil
 }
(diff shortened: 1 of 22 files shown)


emilbon99:
@greptile please re-review and update PR score?

x/go/telem/series_factory.go
```diff
@@ -120,16 +120,29 @@ func NewJSONSeries[T any](data []T) (Series, error) {
...
 
-const newLine = '\n'
+// variableLengthPrefixSize is the number of bytes used for the uint32 LE length prefix
+// in variable-density series encoding.
+const variableLengthPrefixSize = 4
+
+// MarshalVariableSample wraps a single variable-length sample with a uint32 LE length
+// prefix. This is useful for code that accumulates samples into a Series.Data buffer
+// incrementally rather than using NewSeriesV.
+func MarshalVariableSample(sample []byte) []byte {
```
pjdotson, line 130: add test

x/py/x/telem/series.py
```diff
@@ -32,6 +32,8 @@
     TimeStamp,
 )
 
+VARIABLE_LENGTH_PREFIX_SIZE = 4
```
pjdotson, line 35: you need to move the tests for the series into `x/py/tests` and make sure that the
variable-length series are well-tested

x/ts/src/telem/series.ts
```diff
@@ -358,12 +358,42 @@ export class Series<T extends TelemValue = TelemValue>
...
+          offset += e.byteLength;
+        }
+        this._data = buf;
       } else if (this.dataType.equals(DataType.JSON)) {
         this.cachedLength = data_.length;
-        this._data = new TextEncoder().encode(
-          `${data_.map((d) => binary.JSON_CODEC.encodeString(d)).join("\n")}\n`,
-        ).buffer;
+        const encoded = data_.map((d) =>
+          new TextEncoder().encode(binary.JSON_CODEC.encodeString(d)),
```
pjdotson, line 380: this does the case conversions. Do we want to do that?

cesium/control.go
```diff
@@ -150,9 +149,22 @@ func (db *DB) ControlUpdateToFrame(ctx context.Context, u ControlUpdate) Frame {
...
+	s, err := telem.NewJSONSeriesV(u)
 	s.DataType = telem.StringT
-	s.Data, err = (json.Codec).Encode(ctx, u)
-	s.Data = append(s.Data, '\n')
 	return s, err
 }
+
+// DecodeControlUpdate decodes a ControlUpdate from a single-sample Series. See
+// EncodeControlUpdate for why the Series has DataType StringT despite containing JSON.
+func DecodeControlUpdate(s telem.Series) (ControlUpdate, error) {
```
pjdotson, line 164: add test

pjdotson (changes requested):
1) Is there any performance considerations for using a `uint32` length versus a varint length? Am expecting that it could be a relatively common scenario to write short strings where the varint takes up less space, and we would very rarely if ever run into a situation where the varint takes up more than 4 bytes. The other option might be to put the hard cap on string lengths at like a `uint16` if there are worries about Cesium running into performance issues with longer strings.
2) Can you go through the series tests in all 4 languages and make sure they are well tested for all 3 variable-length types?

x/ts/src/telem/series.ts
emilbon99: This is how the behavior was previously, so I think we should leave it as is for this PR. We might want to consider a different approach if necessary.

emilbon99:
I think that `uint32` is the right choice for a few reasons:

1. There's no built in `varint` implementations in TypeScript, Python, or C++. We'd have to roll our own. Not huge, but extra overhead. 
2. A lot of the performance overhead is going to come from calculations related to iteration (len(), at(), iterate) that will require us to use loops with branching every single time we need to read the length of the next sample. It's not really possible to compare the performance of this overhead when compared to the difference in sizes of moving over the network, but claude things that it will be higher. 

There are no performance implications to large length data types in cesium. 

My guess is that the typical string will be "Failed to open ox_pt_1" or something like that, with JSON values being the second most common. So if that's the most common string it's 22 characters which means we'd really save 3 bytes on that so 12% which probably won't make a difference in practice.

My take is: I don't think it's clear that `varint` will yield meaningful benefits and the logic surrounding `uint32` is simpler, so we should just go with what's simpler. 

2) done

x/ts/src/telem/series.ts
pjdotson: I'm fine to leave it as is for this PR, but IMO we shouldn't release support for persisted variable series until this gets solved.

x/py/tests/test_series.py
```diff
@@ -13,21 +13,21 @@
 import pandas as pd
 import pytest
 
-import synnax as sy
+import x.telem as sy
```
pjdotson: ??

pjdotson (changes requested):

x/ts/src/telem/series.ts
emilbon99: https://linear.app/synnax/issue/SY-4077/decide-on-whether-we-should-automatically-convert-cases-for-json

PR #2219, opened 2026-04-13
SY-4060: Support for Persisted Variable-Length Data Types in Cesium

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4060](https://linear.app/synnaxlabs/issue/SY-4060)

## Description

Adds a new `variable` per-channel DB variant in Cesium that persists variable-density
data types (`StringT`, `JSONT`, `BytesT`). Previously, these types could only be used
as virtual (non-persisted) channels.

The `variable` package (`cesium/internal/variable/`) is a sibling of `unary`, reusing
the same external dependencies (`domain.DB`, `index.Domain`, `control.Controller`) but
implementing its own writer, iterator, and delete logic with an in-memory offset cache
instead of fixed-density arithmetic.

Key changes:
- **`cesium/internal/variable/`**: New package with DB, Writer, Iterator, Delete, offset
  cache, and 48 internal tests across both memFS and osFS.
- **`cesium/internal/channel/channel.go`**: Removed the validation gate that rejected
  variable-density persisted channels.
- **`cesium/db.go`, `cesium/open.go`**: Added `variableDBs` map and routing
  (virtual -> variable -> unary fallback).
- **`cesium/channel.go`**: Retrieve, rename, rekey, validate all handle variable DBs.
- **`cesium/delete.go`**: Delete, DeleteTimeRange, GC all handle variable DBs.
- **`cesium/control.go`**: Control state iteration includes variable DBs.
- **`cesium/iterator_open.go`, `cesium/iterator_stream.go`**: Iterator opens and
  iterates both unary and variable iterators using separate arrays.
- **`cesium/writer_open.go`, `cesium/writer_stream.go`**: Variable channels sit inside
  `idxWriter.varInternal` alongside fixed channels in the same index group. Write,
  validate, commit, and close handle both.
- **Integration tests**: Variable channel tests woven into existing `writer_test.go`,
  `iterator_test.go`, and `channel_test.go` covering exclusively variable, mixed
  fixed+variable, multi-variable, embedded newlines, sub-range iteration, and
  persistence across close/reopen.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR introduces persisted variable-length channels (`StringT`, `JSONT`, `BytesT`) in Cesium by collapsing the previously planned `variable` package back into the `unary` package and adding an `offsetResolver`/`offsetTracker` abstraction that handles both fixed-density (density arithmetic, zero overhead) and variable-length (lazy per-domain offset table built by scanning length-prefixed records) channels through the same writer, iterator, and delete code paths.

All five issues flagged in prior review rounds have been resolved: the domain-reader leak in `read()` is fixed with `defer`; the `sampleCount` double-counting is corrected (alignment now stores the pre-write count); the stale offset-cache entry after multi-commit sessions is addressed by gating cache hits on `domainSize` equality; `invalidate()` is called after both delete and GC; and the stale `removeChannel` comment reflects the current two-DB structure.

- Safe to merge — all previously identified P0/P1 issues are resolved, implementation is logically correct, and test coverage is comprehensive.
- All five issues from prior review rounds (reader leak, sampleCount double-counting, stale cache after multi-commit, missing post-delete invalidation, stale comment) have been addressed with correct fixes. The only remaining finding is a P2 comment typo ("is unary" instead of "is fixed") in a test file, which has no runtime impact. Offset arithmetic is correct for both fixed-density and variable-length code paths, the cache invalidation strategy is sound, and 48 new test cases plus integration-level tests cover the new feature thoroughly.
- No files require special attention; `cesium/internal/unary/resolver.go` is the new core abstraction and looks correct.

| Filename | Overview |
|----------|----------|
| cesium/internal/unary/resolver.go | New file implementing offset resolution for both fixed-density (no-op arithmetic) and variable-length (cached per-domain scan) channels; domainSize-keyed cache invalidation correctly handles multi-commit sessions; `uint32` offset storage is documented as a ~4 GB limit matching the domain layer. |
| cesium/internal/unary/variable_test.go | New test file with comprehensive coverage of variable-length channels: write/read of StringT/JSONT/BytesT, empty strings, embedded newlines, sub-range iteration, multi-domain iteration, delete, GC, and a regression test for the multi-commit cache invalidation fix. |
| cesium/internal/unary/iterator.go | Iterator refactored to use `offsetResolver.byteOffset` instead of direct density arithmetic; `defer r.Close()` fix applied; `pickSampleOffset` helper correctly extracts the sample-index logic shared across `autoNext`, `autoPrev`, and `sliceDomain`. |
| cesium/internal/unary/writer.go | Alignment now stores the pre-write sample count (`tracker.count` before `tracker.record`), fixing the prior double-counting bug; `commitWithEnd` correctly uses `tracker.count - 1` as the last-sample index for both fixed and variable channels. |
| cesium/internal/unary/delete.go | Delete path now calls `resolveByteOffset` (opening a short-lived domain iterator) instead of density arithmetic; cache is invalidated after both delete and GC, preventing stale offset tables from causing incorrect byte boundaries. |
| cesium/writer_stream.go | sampleCount double-counting fixed (alignment holds pre-write count; `w.sampleCount = alignment.SampleIndex() + series.Len()` is now correct for variable channels); validateWrite skips unknown-density check for variable types; minor refactors to write/validateWrite logic. |
| cesium/writer_open.go | Two-pass writer-open refactored: pass 1 opens index-channel writers, pass 2 opens all non-index unary writers (fixed and variable); `idxWriter.internal` lazily initialized in both passes. |
| cesium/internal/unary/iterator_test.go | Variable rename from `unaryW` to `fixedW` is clean; one comment was accidentally changed to a nonsensical string ("this behavior is unary" instead of "this behavior is fixed") during a bulk rename. |
| cesium/db.go | DB maps restructured from flat `mu.unaryDBs`/`mu.virtualDBs` to nested `mu.dbs.unary`/`mu.dbs.virtual`; all call sites updated correctly; Metrics and Close loops unchanged in behavior. |
| cesium/internal/channel/channel.go | Removed the validation gate that rejected variable-density persisted channels, enabling `StringT`, `JSONT`, and `BytesT` channels to be stored on disk. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant C as Caller
    participant IW as idxWriter
    participant UW as unary.Writer
    participant OT as offsetTracker
    participant OR as offsetResolver
    participant OC as offsetCache
    participant D as domain.Writer/Iterator

    Note over C,D: Write path (variable-length channel)
    C->>IW: write(frame)
    IW->>UW: Write(series)
    UW->>OT: count(dw) → pre-write sampleCount
    UW->>UW: storeAlignment(DomainIdx, preCount)
    UW->>OT: record(series.Data, baseOffset)
    Note over OT: increments sampleCount & offsets[]
    UW->>D: Write(series.Data)
    UW-->>IW: "alignment (SampleIndex=preCount)"
    IW->>IW: "sampleCount = preCount + series.Len()"

    Note over C,D: Commit path
    C->>IW: Commit(ctx)
    IW->>IW: resolveCommitEnd → idx.Stamp(start, sampleCount-1)
    IW->>UW: CommitWithEnd(ctx, end)
    UW->>D: Commit(ctx, end)

    Note over C,D: Read path (variable-length channel)
    C->>OR: byteOffset(ctx, iter, sampleIdx)
    OR->>OC: get(domainIdx)
    alt cache miss or domainSize changed
        OC-->>OR: miss
        OR->>D: OpenReader(ctx)
        OR->>OR: buildOffsetTable(reader, domainSize)
        Note over OR: scans length-prefixed records
        OR->>OC: "set(domainIdx, table{domainSize, offsets[]})"
    else cache hit (domainSize matches)
        OC-->>OR: offsetTable
    end
    OR-->>C: telem.Size(offsets[sampleIdx])

    Note over C,D: Delete/GC path
    C->>UW: Delete(ctx, timeRange)
    UW->>OR: resolveByteOffset (via temp iterator)
    OR->>OC: get / rebuild table
    UW->>D: domain.Delete(calculateStart, calculateEnd)
    UW->>OR: invalidate() → clears all cached tables
```

1. `cesium/delete.go`, line 129-154 ([link](https://github.com/synnaxlabs/synnax/blob/8df572d21bd4633ec483b9c87c676ff87e34cacf/cesium/delete.go#L129-L154)) 

   <a href="#">P1</a> **`DeleteChannels` silently skips variable-density channels**

   The first-pass loop only looks up channels in `db.mu.unaryDBs`. For a variable channel, `uok` is `false`, so the condition `!uok || udb.Channel().IsIndex` is immediately `true`, and the loop hits `continue` without calling `removeChannel` or queuing the directory for removal. After `DeleteChannels` returns, the channel remains open in `db.mu.variableDBs` and its on-disk directory is untouched.

   `DeleteChannel` (singular) works correctly because it calls `removeChannel` unconditionally, and `removeChannel` does handle variable DBs. The multi-channel variant needs the same treatment:

   ```go
   // Do a pass first to remove all non-index channels
   for _, ch := range chs {
       udb, uok := db.mu.unaryDBs[ch]

       if !uok || udb.Channel().IsIndex {
           if udb.Channel().IsIndex {
               indexChannels = append(indexChannels, ch)
           }
           // Handle variable and virtual channels
           if _, varOk := db.mu.variableDBs[ch]; varOk {
               err = db.removeChannel(ch)
               if err != nil {
                   return
               }
               oldName := keyToDirName(ch)
               newName := oldName + "-DELETE-" + strconv.Itoa(rand.Int())
               err = db.fs.Rename(oldName, newName)
               if err != nil {
                   return
               }
               directoriesToRemove = append(directoriesToRemove, newName)
           }
           continue
       }
       ...
   ```

2. `cesium/delete.go`, line 128-155 ([link](https://github.com/synnaxlabs/synnax/blob/754f7831e02e91eab2f1562b3a9af5728ddc03f3/cesium/delete.go#L128-L155)) 

   <a href="#">P1</a> **`DeleteChannels` silently skips variable channels**

   When `ch` is a variable-channel key, `fok` is `false`, so the outer condition `!fok || fdb.Channel().IsIndex` is immediately `true`, and the inner `fdb.Channel().IsIndex` check (on the zero-value `Channel`) is `false` — the code simply `continue`s. Variable channels are never removed from `db.mu.variableDBs`, their directories are never renamed or deleted, and no error is returned to the caller.

   `DeleteChannel` (single-channel) correctly delegates to `removeChannel`, which handles `variableDBs`. The bulk variant needs the same coverage.
</details>

<sub>Reviews (4): Last reviewed commit: ["Invalidate unary offset table when the d..."](https://github.com/synnaxlabs/synnax/commit/3565f1e6752fccc25796201f04c434e3162cbbbf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28252106)</sub>


emilbon99:

emilbon99:

emilbon99:
@greptile re-review and update PR score

cesium/channel.go
```diff
@@ -111,27 +109,20 @@ func (db *DB) RenameChannel(ctx context.Context, key ChannelKey, newName string)
...
 // RenameChannel renames the channel with the specified key to newName.
 func (db *DB) renameChannel(ctx context.Context, key ChannelKey, newName string) error {
-	udb, uok := db.mu.unaryDBs[key]
-	if uok {
-		// There is a race condition here: one could rename a channel while it is being
-		// read or streamed from or written to. We choose to not address this since the
-		// name is purely decorative in Cesium and not used to identify channels whereas
-		// the key is the unique identifier. The same goes for the virtual database.
-		if err := udb.RenameChannelInMeta(ctx, newName); err != nil {
+	if u, ok := db.mu.dbs.unary[key]; ok {
```
pjdotson, line 116: was this race condition addressed?

cesium/delete.go
```diff
@@ -226,36 +222,31 @@ func (db *DB) DeleteTimeRange(
...
-
-		// Cannot delete an index channel that other channels rely on.
-		if uDB.Channel().IsIndex {
-			indexChannels = append(indexChannels, ch)
+		if _, ok := db.mu.dbs.virtual[ch]; ok {
 			continue
 		}
-
-		dataChannels = append(dataChannels, ch)
+		return errors.Wrapf(ErrChannelNotFound, "channel key %d not found", ch)
```
pjdotson: isn't there a function `NewNotFoundError`

cesium/internal/unary/delete.go
```diff
@@ -135,7 +160,8 @@ func (db *DB) calculateStartOffset(
 			// We stamp to sampleOffset - 1 here since if we are approximating the start
 			// sampleOffset, we want to stamp the last written sample.
 			if sampleOffset == 0 {
-				return density.Size(sampleOffset), ts, nil
+				byteOff, err := db.resolveByteOffset(ctx, domainStart, sampleOffset)
+				return byteOff, ts, err
```
pjdotson: should be returning zero-valued if err is not nil, a few other places in this file where
the same issue happens

cesium/internal/unary/iterator.go
```diff
@@ -457,39 +466,37 @@ func (i *Iterator) sliceDomain(ctx context.Context) (
...
-			// approximation is exact, we want to use the lower approximation.
-			startOffset = i.Channel.DataType.Density().Size(startApprox.Lower)
-		} else {
-			off := (startApprox.Lower + startApprox.Upper) / 2
-			startOffset = i.Channel.DataType.Density().Size(off)
-		}
+	startSample := pickSampleOffset(startApprox)
+	startOffset, err := i.resolver.byteOffset(ctx, i.internal, startSample)
+	if err != nil {
+		return 0, align, 0, err
```
pjdotson: should be returning zero instead of `align`

cesium/internal/unary/resolver.go
```diff
@@ -0,0 +1,209 @@
...
+func (c *offsetCache) set(domainIdx uint32, t *offsetTable) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.tables[domainIdx] = t
+}
+
+func (c *offsetCache) invalidateAll() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.tables = make(map[uint32]*offsetTable)
```
pjdotson: should probably just use `clear` unless you want to deallocate memory

cesium/internal/unary/variable_test.go
```diff
@@ -0,0 +1,356 @@
...
+						telem.NewSeriesSecondsTSV(20, 21),
+					)).To(Succeed())
+					jsonSeries := MustSucceed(telem.NewJSONSeriesV(
+						map[string]any{"key": "value"},
+						map[string]any{"num": 42},
+					))
+					Expect(unary.Write(ctx, jsonDB, 20*telem.SecondTS, jsonSeries)).To(Succeed())
+					frame := MustSucceed(jsonDB.Read(ctx, (20 * telem.SecondTS).Range(22*telem.SecondTS)))
+					Expect(frame.Count()).To(Equal(1))
+					Expect(frame.SeriesAt(0).Len()).To(Equal(int64(2)))
```
pjdotson: should test actual JSON values

cesium/internal/unary/variable_test.go
```diff
@@ -0,0 +1,356 @@
...
+					Expect(unary.Write(ctx, dataDB, 203*telem.SecondTS,
+						telem.NewSeriesV("w", "v"),
+					)).To(Succeed())
+
+					iter := MustSucceed(dataDB.OpenIterator(unary.IterRange(
+						(200 * telem.SecondTS).Range(205 * telem.SecondTS),
+					)))
+					Expect(iter.SeekFirst(ctx)).To(BeTrue())
+					Expect(iter.Next(ctx, telem.TimeSpanMax)).To(BeTrue())
+					Expect(iter.Value().Count()).To(BeNumerically(">=", 1))
```
pjdotson: should test against actual values

pjdotson (changes requested):
Mostly looks good. Have you done any thinking about what it would take to store boolean data types into Cesium: presumably we would only need to make changes in the `cesium/internal/unary` package?

emilbon99:
The only thing you'd need to consider about a `bool` data type is whether you want to compress the values to single bits on disk and/or within the series itself. If within the series itself `unary` would require no changes but `series` would. If not the you would need to add bit level compression to `unary`. Neither would require a new DB variant.

cesium/internal/unary/iterator.go
```diff
@@ -513,7 +520,11 @@ func (i *Iterator) approximateStart(ctx context.Context) (
 // after the end of the range, the returned value will be the number of samples in the
 // range.
 func (i *Iterator) approximateEnd(ctx context.Context) (endApprox index.DistanceApproximation, err error) {
-	endApprox.Approximation = index.Exactly(i.Channel.DataType.Density().SampleCount(telem.Size(i.internal.Size())))
+	total, err := i.resolver.domainSampleCount(ctx, i.internal)
+	if err != nil {
+		return endApprox, err
```
pjdotson: should return `0, err`

pjdotson (changes requested):

PR #2223, opened 2026-04-15
SY-4071: Arc Dot Module Syntax - Part 2

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4072](https://linear.app/synnax/issue/SY-4071/dot-module-syntax-followup)

## Description

This is the followup to #2111 (Dot Module Syntax Part 1). Part 1 added the module
syntax infrastructure (`time.now()`, `math.pow()`, `string.concat()`, etc.) and the
`CompoundFactory`/`MultiFactory` dispatch layer. This PR builds on that foundation to
qualify the remaining STL functions, add dual WASM/Flow execution support, and
consolidate scattered modules.

### Dual WASM/Flow Infrastructure

Added an `ExecContext` field to `symbol.Symbol` with three values: `ExecWASM`,
`ExecFlow`, and `ExecBoth`. This controls where a function is valid:

- The **analyzer** rejects WASM only symbols in flow statements
- The **compiler** rejects flow only symbols in `func` blocks
- The **LSP** filters completions based on the cursor's execution context

This resolves the server panic crashes where using a flow only function in a `func` block (or
vice versa) would crash via `DeriveWASMCoordinates`. Both cases now produce clear
compile errors instead.

We evaluated gating at the resolver level (the `Resolver` interface takes an
`ExecContext` param and filters internally) vs gating at the call site. Resolver gating
sounds cleaner, but the resolver can only return "not found," making a wrong context
function indistinguishable from an undefined symbol. Producing useful error messages like
"function `add` is only available in flow statements" still requires caller side context,
which makes the resolver gating redundant. Call site gating touched 19 files with +75
lines and zero interface changes. Resolver gating touched 26 files with +133/−63 lines,
changed the `Resolver` interface signature, and needed a variadic workaround on
`Scope.Resolve` to avoid ~180 additional no op callsite changes. Same outcome, more
plumbing.

`time.now` is the proof of concept: it went from `ExecWASM` to `ExecBoth`, gaining a
flow node representation that emits the current timestamp when triggered. Support for
negated literals in flow config values (e.g. `time.wait{duration=-3h}`) was also added.

### Module Qualification

Every public STL function now has a module qualified form with LSP hover docs:

| Qualified Form | Previously | Source Package |
|---|---|---|
| `authority.set()` | `set_authority()` | `control` |
| `status.set()` | `set_status()` | new `status` module |
| `selector.select()` | `select()` | `selector` |
| `stable.stable_for()` | `stable_for()` | `stable` |
| `math.avg()` | `avg()` | was `stat` |
| `math.min()` | `min()` | was `stat` |
| `math.max()` | `max()` | was `stat` |
| `math.derivative()` | `derivative()` | was `stat` |
| `math.add()` | `add()` | was `op` |
| `math.subtract()` | `subtract()` | was `op` |
| `math.multiply()` | `multiply()` | was `op` |
| `math.divide()` | `divide()` | was `op` |
| `math.mod()` | `mod()` | was `op` |
| `math.neg()` | `neg()` | was `op` |

Bare forms remain resolvable with deprecation warnings. Deprecation is driven by a
`Deprecated` field on `symbol.Symbol` (set in each module's bare resolver) with
automatic warning emission via `context.ResolveAndWarn`. The old hardcoded
`deprecatedBareFunctions` map in `text/analyze.go` has been removed.

`series.len()` and `string.len()` are now internal. Users call `len()`, which dispatches
to the correct implementation based on argument type. `SERIES` has been removed from the
parser grammar as a valid module name.

### Math Module Consolidation

Moved `avg`, `min`, `max`, `derivative` out of `stat` and `add`, `subtract`, `multiply`,
`divide`, `mod`, `neg` out of `op` into the `math` module. The `stat` package is
deleted. The `op` package retains comparison (`ge`, `gt`, `le`, `lt`, `eq`, `ne`) and
logical (`and`, `or`, `not`) operators.

Binary arithmetic ops (`add`, `subtract`, `multiply`, `divide`, `mod`) are `ExecFlow`
only. WASM has native opcodes for `+`, `-`, `*`, `/`, `%`, so host function bindings
are unnecessary; users use inline operators in func blocks. `pow` and `neg` are
`ExecBoth` since they have no native WASM opcode equivalent.

### CompoundFactory/MultiFactory Routing Fix

Fixed a latent bug where qualified names sharing the same member after prefix stripping
(e.g. `authority.set` vs `status.set` both becoming `set`) would collide. Factories now
declare a module name via the `ModuleNamer` interface so the dispatcher routes to the
correct one.

### C++ Parity

All changes are mirrored in the C++ `math` module:
- Flow nodes (`ArithmeticBinary`, `ArithmeticUnary`) for all arithmetic ops
- WASM linker bindings for `pow` and `neg` across all numeric types
- `WasmModule` and `FlowModule` merged into a single `Module` class
- `time.now` flow node

### Test Coverage

~73 Arc files changed, ~7600 lines added. Tests cover:
- ExecContext gating (analyzer, compiler, LSP)
- Symbol resolution for all qualified and bare forms
- WASM end to end (analyzer -> compiler -> runtime) for arithmetic functions
- Flow node correctness for arithmetic, `time.now`, and reduction ops
- Deprecation warnings for bare forms
- LSP hover docs and completions
- `CompoundFactory`/`MultiFactory` routing with module disambiguation
- C++ unit tests for all new flow nodes and WASM bindings

### Deferred

Section 6 from the planning doc (require module `import` statements) has been moved to
[SY-4076](https://linear.app/synnax/issue/SY-4076/make-module-import-required).

## Basic Readiness

<h3>Greptile Summary</h3>

This large PR (122 files, ~7600 lines) builds on Part 1's module syntax infrastructure to complete the Arc STL module qualification: adds `ExecContext` gating so WASM-only symbols error cleanly in flow statements and vice versa, migrates `avg/min/max/derivative` from `stat` and comparison/logic ops from `op` into the correct modules, fixes a `CompoundFactory` routing collision for same-member-name across modules (e.g. `authority.set` vs `status.set`), adds a `time.now` flow node, and adds support for negated literals in flow config values.

The previously-reported issues (nil `op` dereference for unsigned `neg`, `name` config field silently dropped in `status.set`, integer divide-by-zero in both the Go and C++ flow/WASM paths, missing `now()` offset type validation, `string.len` internal vs hover-doc conflict, and the `CompoundFactory` routing fix) all appear addressed in the current HEAD.

Safe to merge with minor follow-up; all previously-reported P0/P1 issues are addressed in the current HEAD.

All critical bugs from the prior review round are resolved. Remaining findings are P2: math.neg() is registered in WASM bindings but absent from the symbol resolver, string.concat/equal are silently removed from the user API without deprecation, and math.Module/time.Module not implementing ModuleNamer is a latent routing fragility. None block correctness of the shipped functionality.

arc/go/stl/math/math.go (neg symbol resolver gap), arc/go/stl/strings/string.go (silent API removal of concat/equal), arc/go/runtime/node/factory.go (ModuleNamer coverage)

| Filename | Overview |
|----------|----------|
| arc/go/stl/math/math.go | Migrates avg/min/max/derivative from stat; adds WASM neg bindings but omits neg from moduleMembers (PR table says math.neg() should be user-callable but it isn't, P2). |
| arc/go/runtime/node/factory.go | Adds ModuleNamer interface and module-prefix routing to CompoundFactory; math.Module and time.Module don't implement ModuleNamer, leaving them unconditionally tried for every prefixed node type (P2). |
| arc/go/stl/strings/string.go | Marks concat, equal, and len as Internal+ExecWASM; removes user-callable qualified string.concat/string.equal without deprecation warnings (P2 breaking change). |
| arc/go/stl/time/time.go | Adds Now flow node (ExecBoth) and deprecated bare resolvers for interval/wait/now; module prefix handling delegated to CompoundFactory. |
| arc/go/stl/authority/authority.go | New authority module implementing ModuleNamer; authority.set{} as qualified form with set_authority deprecated bare resolver. |
| core/pkg/service/arc/status/set.go | Adds status.set qualified form, ModuleNamer, deprecated bare resolver for set_status, and restores the name config field (previously silently dropped). |
| arc/go/compiler/expression/compiler.go | Removes compileBuiltinNow (now() handled via normal resolution + emitName=scope.Deprecated), adds ExecFlow guard in compiler, extends compileBuiltinLen to string.len/series.len dispatch. |
| arc/go/analyzer/expression/expression.go | Adds ExecFlow guard (rejects flow-only symbols in func blocks) and wires all resolve calls through ctx.Resolve() for automatic deprecation warnings. |
| arc/go/types/type.go | Adds KindSignedNumericConstant, SignedNumericConstraint, and PromoteUnsignedToSigned for neg's unsigned-to-signed implicit cast. |
| arc/go/lsp/completion.go | Adds ExecContext filtering in completions based on cursor nesting (WASM vs Flow); filters Internal symbols from completions. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph Resolution["Symbol Resolution"]
        UserCode["User Arc Code"]
        CtxResolve["ctx.Resolve() — emits deprecation warning"]
        Internal{"Internal?"}
        Deprecated{"Deprecated field set?"}
        Warn["Emit deprecation diagnostic"]
        Symbol["Return Symbol"]
    end

    subgraph ExecGating["ExecContext Gating"]
        AnalyzerFlow["Analyzer — flow node\ntext/analyze.go"]
        AnalyzerWASM["Analyzer — func block\nexpression/expression.go"]
        ExecWASM["Exec == ExecWASM\n→ reject in flow context"]
        ExecFlow["Exec == ExecFlow\n→ reject in func block"]
        ExecBoth["Exec == ExecBoth\n→ allowed everywhere"]
    end

    subgraph Factory["CompoundFactory Routing"]
        IRNode["IR Node Type e.g. status.set"]
        Strip["Strip module prefix: status + set"]
        ModuleNamer{"Implements ModuleNamer?"}
        NameMatch{"prefix == ModuleName?"}
        Skip["Skip factory"]
        Create["factory.Create(cfg)"]
    end

    UserCode --> CtxResolve --> Internal
    Internal -- yes --> Error["ErrNotFound"]
    Internal -- no --> Deprecated
    Deprecated -- yes --> Warn --> Symbol
    Deprecated -- no --> Symbol

    Symbol --> AnalyzerFlow & AnalyzerWASM
    AnalyzerFlow --> ExecWASM
    AnalyzerWASM --> ExecFlow
    Symbol --> ExecBoth

    IRNode --> Strip --> ModuleNamer
    ModuleNamer -- yes --> NameMatch
    NameMatch -- no --> Skip
    NameMatch -- yes --> Create
    ModuleNamer -- no --> Create
```

1. `arc/go/analyzer/expression/expression.go`, line 404-407 ([link](https://github.com/synnaxlabs/synnax/blob/6e28c3ae36e899a1e2127c142eb89a36b66c78a1/arc/go/analyzer/expression/expression.go#L404-L407)) 

   <a href="#">P1</a> **`string.len` missing from builtin exclusion list**

   The analyzer excludes `len` and `series.len` from `validateFunctionCall` because those symbols' stored types use WASM-level representations (i32 handles for `series.len`, ptr+len i32 pair for `string.len`) rather than the Arc-level types. The same logic applies to `string.len`, which was just added to the compiler's builtin dispatch in this PR:

   ```go
   // arc/go/compiler/expression/compiler.go
   if funcName == "len" || funcName == "series.len" || funcName == "string.len" {
       return compileBuiltinLen(ctx, funcCalls[0], funcName)
   }
   ```

   `strings.SymbolResolver` registers "len" with signature `(ptr: i32, len: i32) → u32` (the WASM host binding). When the analyzer processes `string.len(myStr)`, it finds the symbol, does not trigger the ExecFlow guard (it's `ExecWASM`), and then falls through to `validateFunctionCall` with the two-parameter WASM type. That call produces a false error: caller provides 1 argument (a `string`), but the stored type expects 2 (`i32`, `i32`). Any script using `string.len()` fails analysis even though the compiler handles it correctly.

   ```go
   if funcName != "now" && funcName != "time.now" &&
       funcName != "len" && funcName != "series.len" && funcName != "string.len" {
       validateFunctionCall(ctx, scope.Type, funcName, funcCalls[0])
   }
   ```
</details>

<sub>Reviews (18): Last reviewed commit: ["SY-4071:  Fix test"](https://github.com/synnaxlabs/synnax/commit/54995241ab8aa5aee7fb96daaa513a4d53db62a6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28470182)</sub>


core/pkg/service/arc/status/set.go
```diff
@@ -27,24 +27,47 @@ import (
...
+				Name: qualifiedMemberName,
+				Kind: symbol.KindFunction,
+				Exec: symbol.ExecFlow,
+				Type: symbolProps,
 			},
-		}),
+		},
 	}
-	SymbolResolver = symbol.MapResolver{symbolName: symbolSet}
+	SymbolResolver = symbol.CompoundResolver{bareResolver, moduleResolver}
```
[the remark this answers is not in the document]
sy-nico: Restored the schema entry, nodeConfig field, and `stat.Name` assignment

arc/go/stl/math/math.go
```diff
@@ -66,12 +167,492 @@ func NewModule(
...
+	arithmeticUnaryOps = map[string]map[telem.DataType]op.Unary{
+		negSymbolName: {
+			telem.Float64T: op.NegateF64,
+			telem.Float32T: op.NegateF32,
+			telem.Int64T:   op.NegateI64,
+			telem.Int32T:   op.NegateI32,
+			telem.Int16T:   op.NegateI16,
+			telem.Int8T:    op.NegateI8,
+		},
+	}
```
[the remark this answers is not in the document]
sy-nico: Added a `SignedNumericConstraint` that implicitly promotes unsigned types to their wider signed equivalents (u8→i16, u16→i32, u32→i64, u64→f64) during type unification. The `neg` symbol now uses this constraint, so unsigned inputs get cast before reaching the operation. Added cross-type `NegateU8/U16/U32/U64` ops for the flow node path and corresponding entries in `arithmeticUnaryOps`. Tests cover both WASM and flow paths.

sy-nico:
@greptile, concerns have been addressed and more test coverage added. Please update the review on the PR description.

sy-nico:
@greptile, added divide by 0 guard, please review again.

sy-nico:
@greptile, Implemented divide by 0 guard for c++

sy-nico:
@emilbon99 , how should we handle a divide-by-zero in Flow and WASM functions? I don't like the idea of outputting `0` as a guard. For flow nodes, we could just not output and downstream nodes don't trigger. That is elegant, but doesn't solve the problem for WASM nodes, where we can't just prevent execution of downstream functions
```
a := math.divide(some_ch, some_0_ch)
b := math.add{1,1} 
```

arc/cpp/stl/math/math.h
emilbon99: We have both a Module and FlowModule? Why?

arc/cpp/stl/time/time.h
```diff
@@ -171,6 +172,53 @@ class Wait : public runtime::node::Node {
     }
 };
 
+struct NowConfig {
+    x::telem::TimeSpan offset = x::telem::TimeSpan(0);
```
emilbon99: Why do we have an offset parameter?

arc/go/analyzer/expression/expression.go
```diff
@@ -384,12 +384,23 @@ func analyzePostfix(ctx context.Context[parser.IPostfixExpressionContext]) {
...
 			}
+			if callerFn != nil && scope.Exec == symbol.ExecFlow {
+				ctx.Diagnostics.Add(diagnostics.Errorf(
+					ctx.AST,
+					"function '%s' is only available in flow statements, not in func blocks",
+					funcName,
+				))
+				return
+			}
+			if funcName != "now" && funcName != "time.now" &&
```
emilbon99: Why should `now` by a special case handled in the analyzer? Why should `len` be a special case handled in the analyzer? I have my own opinions but want to heart your justification.

arc/go/compiler/expression/exec_context_test.go
```diff
@@ -0,0 +1,130 @@
...
+			Outputs: types.Params{{Name: ir.DefaultOutputParam, Type: types.I64()}},
+		})
+		MustSucceed(ctx.Scope.Add(ctx, symbol.Symbol{
+			Name: "avg",
+			Kind: symbol.KindFunction,
+			Type: funcType,
+			Exec: symbol.ExecFlow,
+		}))
+		expr := MustSucceed(parser.ParseExpression("avg(10)"))
+		_, err := expression.Compile(context.Child(ctx, expr))
```
emilbon99: You need to go back through and make sure your tests properly match against errors using `.Error()`

arc/cpp/stl/math/math.h
```diff
@@ -63,13 +129,689 @@ class Module : public stl::Module {
...
-                [](W base, W exp) -> W {
+                [](W a, W b) -> W {
                     return static_cast<W>(
-                        int_pow(static_cast<T>(base), static_cast<T>(exp))
+                        int_pow(static_cast<T>(a), static_cast<T>(b))
                     );
                 }
             )
             .unwrap();
+        linker
```
emilbon99: Why do these functions need to exist? Are these series arithmetic functions? WASM itself handles math operations.

arc/go/literal/literal.go
```diff
@@ -418,3 +418,25 @@ func IsExactInteger(value float64) bool {
 	// Use relative tolerance: difference should be tiny compared to the magnitude
 	return math.Abs(value-rounded)/math.Abs(rounded) < 1e-9
 }
+
+// Negate negates a parsed literal value. Supports the numeric types produced by Parse.
+func Negate(v any) any {
```
emilbon99, line 423: Why is this necessary?

arc/go/parser/ArcParser.g4
```diff
@@ -153,8 +153,13 @@ function
     | IDENTIFIER configValues
     ;
 
+// AUTHORITY and SERIES are lexer keywords but also valid module names
```
emilbon99: Why should `series.len` be a valid piece of syntax to type in an arc program? Should it just be `len`?

arc/go/runtime/node/factory.go
```diff
@@ -39,12 +40,40 @@ type Factory interface {
 	Create(ctx context.Context, cfg Config) (Node, error)
 }
 
+// ModuleNamer is an optional interface that factories can implement to declare
+// which module they belong to (e.g. "authority", "status"). When the
+// CompoundFactory encounters a qualified node type like "status.set", it skips
+// factories whose module name doesn't match the prefix. Factories that don't
+// implement this interface are always considered.
+type ModuleNamer interface {
```
emilbon99, line 48: Why should this be a separate interface? There are some factories that shouldn't implement `ModuleName`?

arc/go/stl/math/math.go
```diff
@@ -66,12 +167,514 @@ func NewModule(
 		WithFunc(func(_ context.Context, base float64, exp float64) float64 {
 			return math.Pow(base, exp)
 		}).Export("pow_f64")
+
+	builder = bindI32Binary[uint8](builder, "add", "u8", func(a, b uint8) uint8 { return a + b })
```
emilbon99: Again, don't understand why these need to be added if WASM natively supports these operators.

arc/go/stl/math/math_test.go
```diff
@@ -97,4 +214,1075 @@ var _ = Describe("Math", func() {
...
+			Node:  ir.Node{Type: t},
+			State: s.Node("math"),
+		}))
+		changed := make(set.Set[string])
+		c.Next(node.Context{Context: ctx, MarkChanged: func(o string) { changed.Add(o) }})
+		Expect(changed.Contains(ir.DefaultOutputParam)).To(BeTrue())
+		Expect(*s.Node("math").Output(0)).To(telem.MatchSeries(output))
+		Expect(*s.Node("math").OutputTime(0)).To(telem.MatchSeries(outputTime))
+	},
+		Entry("Float32 Add", "add", telem.NewSeriesV[float32](1.5, 2.5, 3.5), telem.NewSeriesSecondsTSV(1, 2, 3), telem.NewSeriesV[float32](0.5, 1.5, 2.5), telem.NewSeriesSecondsTSV(1, 2, 3), telem.NewSeriesV[float32](2.0, 4.0, 6.0), telem.NewSeriesSecondsTSV(1, 2, 3)),
```
emilbon99: Each entry should be formatted onto multiple lines.

arc/go/stl/op/op.go
emilbon99, line 1: Should this package exist?

arc/go/stl/op/op_test.go
```diff
@@ -165,281 +146,8 @@ var _ = Describe("OP", func() {
 		Entry("Uint8 NOT - all false", "not", telem.NewSeriesV[uint8](0, 0, 0), telem.NewSeriesSecondsTSV(1, 2, 3), telem.NewSeriesV[uint8](255, 255, 255), telem.NewSeriesSecondsTSV(1, 2, 3)),
 		Entry("Uint8 NOT - all true", "not", telem.NewSeriesV[uint8](1, 1, 1), telem.NewSeriesSecondsTSV(1, 2, 3), telem.NewSeriesV[uint8](254, 254, 254), telem.NewSeriesSecondsTSV(1, 2, 3)),
 		Entry("Uint8 NOT - mixed", "not", telem.NewSeriesV[uint8](0, 1, 0, 1), telem.NewSeriesSecondsTSV(1, 2, 3, 4), telem.NewSeriesV[uint8](255, 254, 255, 254), telem.NewSeriesSecondsTSV(1, 2, 3, 4)),
-		Entry("Float64 NEG - positive", "neg", telem.NewSeriesV[float64](1.5, 2.5, 3.5), telem.NewSeriesSecondsTSV(1, 2, 3), telem.NewSeriesV[float64](-1.5, -2.5, -3.5), telem.NewSeriesSecondsTSV(1, 2, 3)),
```
emilbon99, line 168: Why have these been deleted?

arc/go/stl/selector/selector.go
```diff
@@ -37,7 +38,11 @@ var (
 			},
 		}),
 	}
-	SymbolResolver = symbol.MapResolver{symbolName: symbolSelect}
+	bareResolver   = symbol.MapResolver{symbolName: symbolSelect}
+	SymbolResolver = symbol.CompoundResolver{
```
emilbon99, line 66: Why should select also be available as `selector.select`?

arc/go/stl/series/series.go
```diff
@@ -84,8 +85,25 @@ var hostSymbols = symbol.MapResolver{
 	"slice":             hostSym("slice", polyFunc(types.Params{{Name: "handle", Type: i32}, {Name: "start", Type: i32}, {Name: "end", Type: i32}}, types.Params{{Name: "result", Type: i32}})),
 }
 
+// moduleMembers contains symbols accessible via the "series" module namespace.
+// The "len" entry is a non-internal copy of the host symbol so that series.len()
+// resolves through the scope (which filters Internal symbols) while preserving
+// the concrete WASM types needed for correct import resolution.
+var moduleMembers = symbol.MapResolver{
```
emilbon99: I disagree that `series.len` should be called with a module prefix. It should just be `len`.

emilbon99: If it needs to be called internally like that it's ok but `series.len` should not be available to the user.

arc/go/stl/wasm/wasm_test.go
```diff
@@ -1101,6 +1124,363 @@ var _ = Describe("WASM", func() {
 		})
 	})
 
+	Describe("Qualified math.add()", func() {
+		DescribeTable("const, const",
+			expectOutput[int64],
+			Entry("i64 literals", "add_ii", types.I64(), `{
+				return math.add(3, 7)
```
emilbon99: In no world should we make these publicly available to callers. In as many scenarios as possible there should be one and only one way to do things. There is no reason why a caller would use `math.add` over `+`

arc/go/symbol/symbol.go
```diff
@@ -57,6 +57,34 @@ import (
...
+	// ExecFlow marks a symbol as only usable in flow statements (graph nodes).
+	ExecFlow
+	// ExecBoth marks a symbol as usable in both contexts.
+	ExecBoth
+)
+
+// Compatible reports whether a symbol with exec context e is visible in a
+// context that filters for target. An untagged symbol (e == 0) is never
+// visible. An unset filter (target == 0) shows all tagged symbols.
+func (e ExecContext) Compatible(target ExecContext) bool {
```
emilbon99, line 75: Should use bitwise operators for this. See `ChanDirection` as an example.

arc/go/text/analyze.go
```diff
@@ -29,6 +29,25 @@ import (
 	"github.com/synnaxlabs/x/diagnostics"
 )
 
+// deprecatedBareFunctions maps bare function names to their qualified replacements.
+// When the analyzer encounters one of these names, it emits a deprecation warning.
+var deprecatedBareFunctions = map[string]string{
```
emilbon99: This is a bad allocation of responsibility. It makes it so that the `text` package is aware of module specific implementation details. Why don't we make it so that the modules themselves return warnings when returning `Resolve` on a deprecated function?

integration/tests/arc/edge_cases.py
```diff
@@ -229,6 +229,45 @@
...
+    label: str
+    source: str
+    wait_substr: str
+
+
+EXEC_CONTEXT_CASES = [
+    ExecContextCase(
+        "FlowInFunc",
+        ARC_FLOW_IN_FUNC,
+        "only available in flow statements",
```
emilbon99: Are we sure these are the best messages to use? Are end users even aware of what a "flow statement" is? Might be worth thinking about how to better communicate the context in which a particular function is valid.

x/go/telem/op/negate_unsigned.go
```diff
@@ -0,0 +1,64 @@
...
+
+package op
+
+import (
+	"github.com/synnaxlabs/x/telem"
+	xunsafe "github.com/synnaxlabs/x/unsafe"
+)
+
+// NegateU8 negates uint8 values, promoting the output to int16.
+func NegateU8(input telem.Series, output *telem.Series) {
```
emilbon99, line 18: You may have already solved this but, what are the contexts in which automatic promotion to a signed integer from an unsigned integer valid? Don't want to spring unexpected conversions on end users.

emilbon99 (changes requested):
Great work overall. I think there are some key decisions we still need to iron out.

arc/cpp/stl/time/time.h
sy-nico: To be able to return the time offset. Imagine an abort occurs, and you want the last 3 hours of data

```
range.create("My_range", start = time.now(-3h), end = time.now(30min))

File: arc/cpp/stl/math/math.h
Comment (sy-nico) [reply arc/cpp/stl/math/math.h]: `Module` is for WASM modules within `func` blocks. `FlowModule` is for flow functions. I have renamed `Module` to `WasmModule` for consistency.

File: arc/go/analyzer/expression/expression.go (hunk 4 above)
Comment (sy-nico) [reply arc/go/analyzer/expression/expression.go]: ~~`len` is polymorphic, so it resolves to `string.len` or `series.len` based on argument type, which the generic validator can't handle~~

~~`now`'s offset is a `Config` param (for flow node usage), not an `Input`, so `validateFunctionCall` would reject `now(-3h)` as "expects 0 arguments." We could remove the `now` special case by also declaring the offset in `Inputs`, but that conflates the flow config and WASM argument paths.~~

~~Since `now` is `ExecBoth`, its offset needs to work in both contexts. Currently it's only declared as `Config`, and the compiler handles the WASM argument manually. If we added the offset to `Inputs` too, the same parameter would exist in both places, and we'd need to make sure the analyzer, compiler, and flow runtime all agree on which one to use in which context.~~

File: arc/cpp/stl/time/time.h (hunk 3 above)
Comment (emilbon99) [reply arc/cpp/stl/time/time.h]: `range.create("My_range", start = time.now() - 3h, end = time.now() + 30min)`

This should be the way you do that.

File: arc/go/literal/literal.go (hunk 7 above)
Comment (sy-nico) [reply arc/go/literal/literal.go:423]: `literal.Negate` is used during flow config analysis to handle negated literals in node config values (`time.now(offset=-3h)`). It operates on parsed Go values at the analyzer level.

File: arc/cpp/stl/math/math.h (hunk 6 above)
Comment (sy-nico) [reply arc/cpp/stl/math/math.h]: These functions are redundant. I've removed them and changed add/subtract/multiply/divide/mod to `ExecFlow`. Only `pow` and `neg` retain WASM host bindings.

File: arc/go/parser/ArcParser.g4 (hunk 8 above)
Comment (sy-nico) [reply arc/go/parser/ArcParser.g4]: `len()` can resolve to both `series.len() and `string.len(), so `series.len()` and `string.len()` exist as qualified forms for when the user wants to be explicit about which variant they're calling. This follows the same pattern as the rest of the module qualification work in this PR, which was to give each function to a qualified form.

File: arc/go/stl/math/math.go (hunk 10 above)
Comment (sy-nico) [reply arc/go/stl/math/math.go]: These are now `ExecFlow` only.

File: arc/go/runtime/node/factory.go (hunk 9 above)
Comment (sy-nico) [reply arc/go/runtime/node/factory.go:48]: It's a separate interface so factories can opt in. The problem it solves is when `authority.set` and `status.set` both become `set` after prefix stripping, so the first factory with a `set` handler would win even if it's the wrong module. 

`ModuleNamer` lets a factory declare "I'm `authority`" or "I'm `status`" so the dispatcher skips non-matching modules. Most factories (like `time`, `selector`, `stable`) don't have this collision problem, so making them all implement `ModuleName()` returning `""` would be unnecessary boilerplate. As a separate interface, the default behavior is "always considered."

File: arc/go/stl/op/op.go
Comment (sy-nico) [reply arc/go/stl/op/op.go:1]: Yes. It still contains the comparison (`ge`, `gt`, `le`, `lt`, `eq`, `ne`) and logical (`and`, `or`, `not`) flow operators. The arithmetic ops moved to `math`, but these stayed since they aren't math operations.

I didn't want to move these into any of the other stl libraries, so it seemed most correct to keep these under `op`. Do you think these should be moved elsewhere?

File: arc/go/stl/op/op_test.go (hunk 12 above)
Comment (sy-nico) [reply arc/go/stl/op/op_test.go:168]: `neg` moved from `op` to `math` as part of the arithmetic consolidation. These tests were moved to `math_test.go`, not deleted.

File: arc/go/stl/selector/selector.go (hunk 13 above)
Comment (sy-nico) [reply arc/go/stl/selector/selector.go:66]: Per the initial request/instruction, all functions must have qualified names.

File: arc/go/stl/series/series.go (hunk 14 above)
Comment (sy-nico) [reply arc/go/stl/series/series.go]: Which functions should explicitly have qualified names and which shouldn't? Per your instruction, the purpose of Part 2 was to ensure all functions have qualified names and are in the correct module.

The followup change would be to require module imports.

File: arc/go/stl/wasm/wasm_test.go (hunk 15 above)
Comment (sy-nico) [reply arc/go/stl/wasm/wasm_test.go]: These are now `ExecFlow` only, so they're not callable in `func` blocks. The WASM host bindings have been removed.

File: arc/go/symbol/symbol.go (hunk 16 above)
Comment (sy-nico) [reply arc/go/symbol/symbol.go:75]: Implemented

File: arc/go/text/analyze.go (hunk 17 above)
Comment (sy-nico) [reply arc/go/text/analyze.go]: Agreed on the separation of concerns. Two options:

1) Change the `Resolver` interface to return warnings alongside the symbol (`Resolve` returns `(Symbol, []string, error)`). Each module's bare resolver returns a deprecation warning. This touches 162 call sites across 39 files.

2) Add a `DeprecatedMessage` field to `Symbol`. Each module's bare resolver sets it. `text/analyze.go` checks the field after resolve. No interface change, but adds a permanent field to `Symbol` for what should be a small number of deprecated functions.

Which do you prefer?

File: integration/tests/arc/edge_cases.py (hunk 18 above)
Comment (sy-nico) [reply integration/tests/arc/edge_cases.py]: Updated to

`"function '%s' cannot be called inside a func block. Use it as a flow statement instead: %s{}"`
and
`"function '%s' cannot be used as a flow statement. Call it inside a func block instead: %s()"`

File: x/go/telem/op/negate_unsigned.go (hunk 19 above)
Comment (sy-nico) [reply x/go/telem/op/negate_unsigned.go:18]: Unsigned-to-signed promotion only occurs when the user explicitly negates an unsigned value (`-x`). Since the result can be negative, the compiler promotes to the next wider signed type (`u8`→`i16`, `u16`→`i32`, `u32`→`i64`, `u64`→`f64`). 

This will be visible to the user. If the promoted type doesn't match the destination channel, the compiler rejects it with a type mismatch error. No implicit silent conversion ever reaches the output.

File: arc/cpp/stl/time/time.h (hunk 3 above)
Comment (sy-nico) [reply arc/cpp/stl/time/time.h]: Removed parameter from time.now()

File: arc/go/analyzer/expression/expression.go (hunk 4 above)
Comment (sy-nico) [reply arc/go/analyzer/expression/expression.go]: Removing. With the offset removed, `now()` has no `Inputs` at all, so `validateFunctionCall` handles it correctly (0 expected, 0 provided).

File: arc/go/literal/literal.go (hunk 7 above)
Comment (sy-nico) [reply arc/go/literal/literal.go:423]: I removed `offset` from `time.now(). But future functions that would need this could be:

* `math.clamp{min=-10, max=100}`: clamping a sensor value to a range where the lower bound is negative (e.g., temperature)
* `math.threshold{value=-40}`:  fire when a reading crosses -40°C
* `math.scale{factor=2, offset=-273.15}`: linear transform, e.g., Kelvin to Celsius

File: arc/go/parser/ArcParser.g4 (hunk 8 above)
Comment (emilbon99) [reply arc/go/parser/ArcParser.g4]: It's fine to have them qualified, but I don't think `series.len` and `string.len` should be visible to the end user, they should be internal mechanics only.

File: arc/go/stl/op/op.go
Comment (emilbon99) [reply arc/go/stl/op/op.go:1]: These aren't math operations? Aren't they all math operations? They aren't arithmetic operators but they are definitely mathematical operations.

File: arc/go/parser/ArcParser.g4 (hunk 8 above)
Comment (sy-nico) [reply arc/go/parser/ArcParser.g4]: * `len()` is still public, even if `string.len()` is not

File: arc/go/stl/control/control.go [hunk 20]
```diff
@@ -45,6 +70,8 @@ type Module struct {
 
 func NewModule(ab *ProgramState) *Module { return &Module{auth: ab} }
 
+func (m *Module) ModuleName() string { return moduleName }
+
 func (m *Module) Resolve(ctx context.Context, name string) (symbol.Symbol, error) {
```
emilbon99: This method isn't called/used anywhere in the codebase and it doesn't seem to implement an interface. I think this is an artifact and it should be removed from this particular Module implementation and all others.

arc/go/stl/control/control.go
emilbon99: This method isn't called/used anywhere in the codebase and it doesn't seem to implement an interface. I think this is an artifact and it should be removed from this particular Module implementation and all others.

arc/go/text/analyze.go
emilbon99: Option 2 is along the lines of the right route I think. But you need to keep in mind that making every call site handle deprecation warnings individually is duplicated code. Probably needs some mechanism for the Scope.Resolve to attach and/or return relevant diagnostics so that adding deprecation warnings is not duplicated.

Also consider whether `DeprecatedMessage` is the right field. Consider an LSP that has quick fixes in order to remove the deprecation, it might need more info such as the new recommended function/type to use.

arc/cpp/stl/math/math.h
emilbon99: Why can't it be a single `Module` type that implements the signatures for both the wasm module and the node module? As far as I see the only production call sites for this are in `arc/cpp/runtime/runtime.h` where it would make sense to just have every package expose a single module type that does both workloads. It's what the `Module` contract in `/Users/emilianobonilla/Desktop/synnaxlabs/synnax/arc/cpp/stl/stl.h` already defines.

emilbon99 (changes requested):

arc/go/parser/ArcParser.g4
sy-nico: Removed series from parser grammer

arc/go/stl/control/control.go
sy-nico: `ModuleName()` implements the `ModuleNamer` interface (`arc/go/runtime/node/factory.go:48`). `CompoundFactory.Create` uses it to disambiguate which factory handles a qualified node type (e.g., `authority.set` vs `status.set` both have a member named `set`). It's only implemented on modules where the bare name collides with another module's member.

arc/go/stl/control/control.go
sy-nico: Removed dead code. `Resolve` and `Search` on the `Module` just delegate to `SymbolResolver`, which is used directly everywhere.

arc/go/text/analyze.go
sy-nico: Implemented. Added `Deprecated string` field to `Symbol` and a `ResolveAndWarn` method on `analyzer/context.Context` that wraps `Scope.Resolve` and auto-emits the warning. `Scope.Resolve` signature unchanged. Deleted the `deprecatedBareFunctions` map.

arc/cpp/stl/math/math.h
sy-nico: Merged `WasmModule` and `FlowModule` into a single `Module` class.

arc/cpp/runtime/wasm/node_test.cpp
```diff
@@ -81,13 +81,13 @@ std::vector<std::shared_ptr<stl::Module>> build_stl_modules(
     const std::shared_ptr<stl::stateful::Variables> &var_st
 ) {
     return {
-        std::make_shared<stl::channel::Module>(channel_st, str_st),
-        std::make_shared<stl::stateful::Module>(var_st, series_st, str_st),
-        std::make_shared<stl::series::Module>(series_st),
-        std::make_shared<stl::str::Module>(str_st),
+        std::make_shared<stl::channel::WasmModule>(channel_st, str_st),
```
emilbon99: All of these should just be named `Module` and not `WasmModule`.

arc/cpp/runtime/wasm/node_test.cpp
```diff
@@ -1846,7 +1845,7 @@ TEST(NodeTest, DISABLED_StringConcatWithChannelData) {
 
     const std::string source = R"(
 func concat_len(a str, b str) i64 {
-    return string.len(string.concat(a, b))
+    return len(string.concat(a, b))
```
emilbon99: Why do we have a public `string.concat` function when you could just use `a + b`? Should preferably be only one way to do things unless absolutely necessary.

arc/cpp/stl/time/time.h
```diff
@@ -171,7 +172,40 @@ class Wait : public runtime::node::Node {
...
+/// @brief Outputs the current wall-clock timestamp when triggered.
+class Now : public runtime::node::Node {
+    runtime::state::Node state;
+
+public:
+    explicit Now(const NowConfig &, runtime::state::Node &&state):
+        state(std::move(state)) {}
+
+    x::errors::Error next(runtime::node::Context &ctx) override {
+        const auto ts = x::telem::TimeStamp::now().nanoseconds();
```
emilbon99: Should be using `telem::MonoClock` for this. Also there should be no reason to call `.nanoseconds` as the series have an overload for setting a timestamp.

arc/go/analyzer/context/context.go
```diff
@@ -125,6 +125,24 @@ func (c Context[AST]) WithTypeHint(hint types.Type) Context[AST] {
 	return c
 }
 
+// ResolveAndWarn resolves a symbol by name and, if the resolved symbol is
+// deprecated, automatically adds a deprecation warning to diagnostics.
+func (c Context[AST]) ResolveAndWarn(name string) (*symbol.Scope, error) {
```
emilbon99: Why is there both a `Resolve` and a `ResolveAndWarn` function? Is there ever a case where we call `Resolve` and wouldn't want to warn?

arc/go/analyzer/expression/expression.go
```diff
@@ -384,12 +384,22 @@ func analyzePostfix(ctx context.Context[parser.IPostfixExpressionContext]) {
...
 		}
 		if scope.Kind == symbol.KindFunction {
-			validateFunctionCall(ctx, scope.Type, funcName, funcCalls[0])
 			callerFn, fnErr := ctx.Scope.ClosestAncestorOfKind(symbol.KindFunction)
 			if fnErr != nil && !errors.Is(fnErr, query.ErrNotFound) {
 				ctx.Diagnostics.Add(diagnostics.Error(fnErr, ctx.AST))
 				return
 			}
+			if callerFn != nil && scope.Exec == symbol.ExecFlow {
+				ctx.Diagnostics.Add(diagnostics.Errorf(
```
emilbon99, line 393: Is this the only place where we need to add this diagnostic?

arc/go/compiler/expression/compiler.go
```diff
@@ -412,33 +427,53 @@ func emitLiteralValue[T antlr.ParserRuleContext](
...
 		return types.I64(), nil
 	default:
-		return types.Type{}, errors.Newf("argument 1 of len: expected series or str, got %s", argType)
+		return types.Type{}, errors.Newf(
+			"argument 1 of %s: expected series or str, got %s", funcName, argType)
 	}
 }
 
-func compileBuiltinNow(ctx context.Context[parser.IPostfixExpressionContext]) (types.Type, error) {
+func compileBuiltinNow(
```
emilbon99: Why does `now` need to compile as a `builtin`? It has no special polymorphic usages and it takes no arguments. Why can't we just compile it as a standard function from the stl?

arc/go/compiler/expression/exec_context_test.go
emilbon99: This was not fixed. This assertion should be as follows:

Expect(expression.Compile(context.Child(ctx, expr))).Error().To(MatchError(ContainSubstring("YADA"))

This will: 1) assert error occurred, 2) assert the substring is right, and 3) assert that the ignored parameter has a zero value.

Make sure all of your other assertions properly do this.

arc/go/lsp/hover_test.go
```diff
@@ -112,9 +112,136 @@ var _ = Describe("Hover", func() {
...
+				},
+			}))
+
+			Expect(hover).ToNot(BeNil())
+			Expect(hover.Contents.Value).To(ContainSubstring("#### math.avg"))
+			Expect(hover.Contents.Value).To(ContainSubstring("running average"))
+		})
+
+		It("should provide hover for 'math.pow' function", func(ctx SpecContext) {
+			content := "result := math.pow(2, 3)"
```
emilbon99: We should talk about why these APIs are public. Why should a caller be allowed to call `math.pow` when they could just do `2 ^ 3`?

arc/go/lsp/hover_test.go
```diff
@@ -112,9 +112,136 @@ var _ = Describe("Hover", func() {
...
+				},
+			}))
+
+			Expect(hover).ToNot(BeNil())
+			Expect(hover.Contents.Value).To(ContainSubstring("#### selector.select"))
+			Expect(hover.Contents.Value).To(ContainSubstring("Routes input values"))
+		})
+
+		It("should provide hover for 'stable.stable_for' function", func(ctx SpecContext) {
+			content := "sensor -> stable.stable_for{duration=5s} -> output"
```
emilbon99: Should it be called `stable.for`?

arc/cpp/stl/authority/authority.h
emilbon99, line 1: If we're calling this module `authority` shouldn't the folder and namespace be called `authority` as well? Same train of logic applies to go.

arc/go/stl/op/op.go
emilbon99: This comment wasn't addressed

arc/go/stl/selector/selector.go
emilbon99: Yes, but there are some functions such as `len` and `range` that are built in keyword functions to the language. Select is a relatively primitive function, should it be a builtin?

arc/go/stl/series/series.go
```diff
@@ -84,8 +85,23 @@ var hostSymbols = symbol.MapResolver{
...
+		Type: types.Function(types.FunctionProperties{
+			Inputs:  types.Params{{Name: "handle", Type: types.I32()}},
+			Outputs: types.Params{{Name: ir.DefaultOutputParam, Type: types.I64()}},
+		}),
+	},
+}
+
 var SymbolResolver = symbol.CompoundResolver{
 	userSymbols,
+	&symbol.ModuleResolver{Name: "series", Members: moduleMembers},
```
emilbon99: Seems pretty wild that we have two symbol resolvers for the same thing her. How did you come to make that decision?

arc/go/stl/time/time.go
```diff
@@ -261,3 +262,23 @@ func (w *Wait) Reset() {
...
+
+// Now outputs the current wall-clock timestamp when triggered.
+type Now struct {
+	*node.State
+}
+
+func (n *Now) Init(_ node.Context) {}
+
+func (n *Now) Next(ctx node.Context) {
+	ts := telem.Now()
```
emilbon99: Should be using `telem.MonoClock` for this.

core/pkg/service/arc/status/set.go
```diff
@@ -63,54 +87,70 @@ func (m *Module) Search(ctx context.Context, term string) ([]symbol.Symbol, erro
...
-	return &setNode{ins: cfg.Instrumentation, stat: stat, statusSvc: m.stat}, nil
-}
+func (m *Module) ModuleName() string { return moduleName }
 
-type setNodeConfig struct {
-	StatusKey string `json:"status_key"`
-	Message   string `json:"message"`
-	Variant   string `json:"variant"`
+func (m *Module) Create(ctx context.Context, cfg node.Config) (node.Node, error) {
+	f := &statusFactory{stat: m.stat}
```
emilbon99: should just collapse the `statusFactory` and `Module` types into one `Module` type.

emilbon99 (changes requested):

arc/go/lsp/hover_test.go
sy-nico: Sure, I can update, but this isn't in scope.

arc/cpp/stl/authority/authority.h
sy-nico: Also not in scope, but I will update here.

arc/cpp/stl/time/time.h
sy-nico: The `telem::MonoClock` refactor was merged into `rc` after this PR was opened. I have back merged and implemented.

arc/go/stl/time/time.go
sy-nico: The `telem.MonoClock` refactor was merged into `rc` after this PR was opened. I have back merged and implemented.

arc/go/stl/op/op.go
sy-nico: There wasn't an action from this comment. I don't agree with lumping them with arithmetic operations. Either way, this change requires offline discussion and is out of scope for this PR.

arc/go/stl/selector/selector.go
sy-nico: Out of scope for this PR. The goal was to add module qualified names, not redesign which functions are builtins. That's a separate design call that should have its own ticket.

`len` got handled here because making `string.len`/`series.len` internal fell out of the qualification work naturally. `select` doesn't, so leaving it as `selector.select` for now.

arc/go/stl/op/op.go
sy-nico: Followup ticket: https://linear.app/synnax/issue/SY-4134/dot-module-followup

arc/go/compiler/expression/compiler.go
sy-nico: It doesn't need to be. The intent of the ticket was to explicitly enable module qualified names, not to deprecate `builtin` functions. From our offline discussion we are increasing the PR scope and will deprecate `now()` so that the function can only be called in one way.

arc/cpp/runtime/wasm/node_test.cpp
sy-nico: updated to use `+`

arc/cpp/runtime/wasm/node_test.cpp
sy-nico: fixed

arc/go/lsp/hover_test.go
sy-nico: Implemented

arc/go/stl/selector/selector.go
sy-nico: making bare select callable

arc/go/lsp/hover_test.go
sy-nico: Made internal

arc/go/analyzer/expression/expression.go
sy-nico: Yes, this is the only place. Every function call inside a func block flows through `analyzePostfix`, so this is the single chokepoint for that direction. The symmetric case (WASM-only function used as a flow statement node) is similarly handled in `text/analyze.go : analyzeFunctionNode`. There's a duplicate of this check in compiler/expression/compiler.go that's a defensive backstop.

arc/go/stl/series/series.go
sy-nico: I added `moduleMembers` so `series.len()` could resolve as user callable, since the scope filters out `Internal: true` symbols and everything in `hostSymbols` is internal. After I made `series.len` internal both resolvers only hold internal symbols, so the split is obsolete. Collapsing them.

arc/go/analyzer/context/context.go
sy-nico: Original thinking was opt-in: analyzer warns, compiler/LSP/tests skip to avoid double-warning. So `ResolveAndWarn` for the analyzer, plain `Scope.Resolve` elsewhere.

But the analyzer has ~10 other `ctx.Scope.Resolve` sites resolving user-named symbols that should also warn. Today `avg` as a function call skips the warning while `avg` as a flow node emits it. This is inconsistent.

Collapsing into a single `ctx.Resolve(name)` that always warns and migrating the analyzer sites. Compiler and tests keep `ctx.Scope.Resolve` for raw resolution.

core/pkg/service/arc/status/set.go
sy-nico: Original instinct was separation of concerns: `Module` for symbol/scope work (`Resolve`, `Search`, `ModuleName`), `statusFactory` for runtime node creation.

But they wrap the same `*status.Service` with no state or lifecycle difference, and `Module.Create` is just a thin shim that allocates a `statusFactory` and delegates. Pure indirection.

Collapsing into a single `Module` type.

emilbon99 (commented):
LGTM. Nice work

emilbon99 (approved):

PR #2243, opened 2026-04-19
SY-4084: Omit Status, Program, and Parent From Entity Wire Format

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4084](https://linear.app/synnax/issue/SY-4084)

## Description

Adds `@go marshal omit` to the derived/transient fields on `arc`, `rack`, `task`, and `device` entities so they no longer round-trip through the persisted wire format. These fields (`status` on all four, plus `program` on `arc` and `parent` on `device`) live in separate services or are computed in-process and shouldn't be embedded in the entity's storage row.

The change ships with full v54 migration support:

- Renames the existing `@go marshal skip` field directive to `@go marshal omit` for consistency, including the v54 ontology snapshot and oracle plugin tests.
- Teaches the oracle migrate plugin's schema diff (`schemasEqual` / `diffStructFields`) to detect changes in the `@go marshal` directive on a field. Without this, marshal-only changes were silently ignored — no v54 frozen codec or `migrate_auto.gen.go` was generated, leaving `MigrateX` calls referencing nonexistent symbols.
- Generates v54 frozen types + codecs + auto-migrate helpers for `arc`, `rack`, `task`, `device` (and their transitive dependencies in `arc/ir`, `arc/graph`, `x/status`, `x/label`, `x/color`, etc.).
- Wires the new `v54_drop_*` entry migrations into each service's `Migrations` slice and pins the chain ordering (`v54_drop_* -> msgpack_to_orc -> v0`) via `migrate.WithAddedDeps`. The `msgpack_to_orc` codec migrations now reference the v54 frozen types so the codec transition preserves the dropped fields long enough for `v54_drop_*` to read them.
- Skips `Migrate*` template generation for non-struct frozen types (enum/alias/distinct), which previously emitted invalid composite-literal returns like `ir.EdgeKind{}`.
- Updates the existing v0 migration tests in `rack_test.go`, `task_test.go`, and `device_test.go` to seed via the v0 frozen entry types (compatible wire format), plus a new in-package `migrate_test.go` for each of arc/rack/task/device that exercises the full gorp pipeline with both empty and populated `Status` payloads.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR strips transient/derived fields (`status`, `program`, `parent`) from the wire format of `arc`, `rack`, `task`, and `device` entities by adding `@go marshal omit` to the oracle schemas. It ships the necessary v54 frozen types, codecs, and migration helpers so existing data is safely upgraded: the `msgpack_to_orc` step is pinned to the v54 frozen types (preserving old Status bytes), then a new `v54_drop_*` entry migration re-encodes each row with the updated codec that omits the transient fields. The oracle migrate plugin is also fixed to detect marshal-directive changes in the schema diff and to skip non-struct types when emitting `MigrateX` templates.

Safe to merge; all remaining findings are P2 style/documentation issues that don't affect runtime correctness.

The migration chain logic is sound: freezing `msgpack_to_orc` to v54 types and ordering `v54_drop_*` after it via `WithAddedDeps` correctly handles both fresh and mid-migration databases. The oracle plugin fixes (struct-type guard, marshal-directive diffing) are well-tested with new unit tests. The only open question — `ir.IR.Root` being zero-valued post-migration — is flagged by the generator's own comment and is P2 pending author confirmation. All other findings are test naming style issues.

`arc/go/ir/migrations/v54/migrate.go` — confirm `ir.IR.Root` zero value is safe for existing persisted arcs.

| Filename | Overview |
|----------|----------|
| oracle/plugin/go/migrate/schema.go | Adds `@go marshal` directive comparison to both `typesEqual` and `diffStructFields` so marshal-only field changes (e.g. adding `omit`) correctly trigger schema-diff detection and migration generation. |
| oracle/plugin/go/migrate/migrate.go | Adds a struct-type guard in `renderTypeMigrateTemplate` to skip non-struct frozen types (enums, aliases, distincts) that would otherwise produce invalid composite-literal returns. |
| oracle/plugin/go/marshal/encoder.go | Renames marshal directive from `"skip"` to `"omit"` for consistency with the new schema vocabulary; simple one-line change. |
| arc/go/ir/migrations/v54/migrate.go | Manually-editable migration template for IR sub-types; oracle flagged `ir.IR.Root` as zero-valued after migration — unclear if this field's zero value is safe for existing persisted arcs. |
| core/pkg/service/arc/service.go | Upgrades migration chain: `msgpack_to_orc` now uses frozen `v54.Arc` type to preserve Status/Program in the intermediate ORC form; adds `v54_drop_program_status` entry migration with `WithAddedDeps` ordering constraint. |
| core/pkg/service/rack/service.go | Mirrors the arc pattern for rack: `msgpack_to_orc` frozen to `v54.Rack`, new `v54_drop_status` entry migration wired after it. |
| core/pkg/service/task/service.go | Same migration chain upgrade for task service; `msgpack_to_orc` uses `v54.Task`, adds `v54_drop_status` dependency. |
| core/pkg/service/device/service.go | Same migration chain upgrade for device service; `msgpack_to_orc` uses `v54.Device`, adds `v54_drop_status_parent` dependency to also strip the `parent` field. |
| core/pkg/service/arc/migrations/v54/migrate_test.go | New integration tests exercising v54→current Arc migration via gorp pipeline; covers both empty and populated Status/Program paths. Second test description is slightly ambiguous. |
| schemas/arc.oracle | Adds `@go marshal omit` to `program` and `status` fields; also renames `skip` → `omit` in ontology.oracle. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant DB as Existing DB
    participant v0 as v0 Migration
    participant msgpack as msgpack_to_orc<br/>(v54 frozen codec)
    participant drop as v54_drop_*<br/>(entry migration)
    participant current as Current Codec<br/>(omit status/program/parent)

    DB->>v0: msgpack-encoded row (legacy format)
    v0->>msgpack: decoded v0 entry
    msgpack->>drop: ORC row with Status/Program/Parent preserved
    drop->>current: AutoMigrateX() copies all fields
    current->>DB: re-encoded ORC row (Status/Program/Parent omitted)
```

<sub>Reviews (1): Last reviewed commit: ["formatting"](https://github.com/synnaxlabs/synnax/commit/9edd8ee7939dc689fb92d78edbc7134b7867357d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28912134)</sub>


x/go/status/migrations/v54/codec.gen.go
```diff
@@ -0,0 +1,106 @@
...
+	w.String(s.Description)
+	w.Int64(int64(s.Time))
+	{
+		b, err := json.Marshal(s.Details)
+		if err != nil {
+			return err
+		}
+		w.WriteWithLen(b)
+	}
+	if s.Labels != nil {
```
pjdotson, line 36: should skip encoding `Labels`

schemas/task.oracle
```diff
@@ -110,7 +110,8 @@ Task struct<
         @doc value "indicates whether to persist this task's configuration."
     }
     status   Status<StatusData>?? {
-        @doc value "is the current execution status of the task."
+        @doc value  "is the current execution status of the task."
```
pjdotson, line 113: a little odd that the formatter adds two spaces between `value` and the first `"`.

arc/go/types/migrations/v54/types.gen.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Params is a collection of named, typed parameters for function inputs, outputs, or
+// configuration.
+type Params []Param
+
+// Kind is the type category for Arc's type system, including primitives, compound
+// types, and meta-types.
+type Kind uint8
+
+//go:generate stringer -type=Kind
```
pjdotson, line 22: This go generate was not run automatically. We should add a CI check that all
`go:generate` have been run

arc/go/ir/migrations/v54/migrate.go
```diff
@@ -0,0 +1,64 @@
...
+
+package v54
+
+import (
+	"context"
+
+	ir "github.com/synnaxlabs/arc/ir"
+)
+
+func MigrateFunction(ctx context.Context, old Function) (ir.Function, error) {
```
pjdotson, line 20: Help me understand this: we only make the `v54` folder _after_ there is a new version
made? Shouldn't we have made the `v54` folder with the `v54` release? It's also odd to
me that `arc/ir/migrations/v54` depends on `arc/ir`, instead of the other way around

pjdotson (changes requested):
You should make sure that Lables on statuses + ranges and parent on ranges also are not encoded.

x/go/status/migrations/v54/codec.gen.go
emilbon99: Implemented. Ended up needing to make some changes to the oracle code generator.

schemas/task.oracle
emilbon99: It's designed to work in the same way as tags on go struct fields, where the values end up vertically aligned when they are next to each other.

arc/go/types/migrations/v54/types.gen.go
emilbon99: Want me to add this check in this issue or another one?

pjdotson: Another issue

schemas/task.oracle
pjdotson: Copy

arc/go/ir/migrations/v54/migrate.go
emilbon99: 1) The only think we take a snapshot of at release time is the schemas, then type specific migration directories are created in go lazily only if necessary. This means we reduce the amount of generated boilerplate go code if types don’t change across specific version i.e. if we don't change the `Range` struct we don't end up with `v54`, `v55`, `v56` before a change in `v57` reducing duplicated code.
2) This is a legitimate question. The answer is that directly migratable types that are stored in gorp directly e.g. `arc.Arc`, `ranger.Range` do follow the pattern that you outlined i.e. they import their migration folder the other way around. Non-direct migration patterns follow this structure. It's not a bug but it is a questionable design decision that would be worth looking into.

arc/go/types/migrations/v54/types.gen.go
emilbon99: https://linear.app/synnax/issue/SY-4094/add-go-generate-ci-check

arc/go/ir/migrations/v54/migrate.go
emilbon99: https://linear.app/synnax/issue/SY-4095/refactor-migration-folder-structure-for-dependent-data-types

core/pkg/service/task/migrations/v54/migrate_test.go
```diff
@@ -0,0 +1,122 @@
...
+					Cmd:     "start",
+				},
+				Labels: []labelv54.Label{
+					{Key: uuid.New(), Name: "active", Color: colorv54.Color{G: 200, A: 1}},
+				},
+			},
+		}
+		MustSucceed(gorp.OpenTable[v54.Key, v54.Task](
+			ctx, gorp.TableConfig[v54.Task]{DB: db},
+		))
```
pjdotson: shouldn't you be using the table here after opening it?

pjdotson (changes requested):
approved after all tests where comment applys are fixed

PR #2248, opened 2026-04-21
SY-4099: Refactor migration tests RFC

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4099](https://linear.app/synnax/issue/SY-4099/refactor-migration-tests)

## Description

Design document to refactor the migration tests. Several things that are an issue:
- the `Simulator` class and importing from `examples` in general is odd. I don't think we should export that in the Client library at all, instead that should just be example code, and any code used for integration testing should live in the `integration` folder. This is what is causing issues with the current integration tests.
- In general, simplified the logic so that the setup chain just runs simple Python scripts instead of running the full integration test cases
- was wrong to build old wheels of old versions of the Python client instead of downloading from PyPI
- test infrastructure easier to parse now.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR refactors the migration test infrastructure by replacing wheel-based client builds and a single conductor running current-repo code against old clients with per-version frozen setup scripts (flat `*_setup.py` files under `integration/tests/migration/` with `SETUP_VERSION` constants) and a new stdlib-only Python conductor (`migration_conductor.py`) that installs each version's client from PyPI into an isolated venv.

- The RFC design document (§2.0, §2.4, §4) describes a directory-per-version layout (`v0_54/setup.py`) and a shell-script orchestrator that don't match the actual flat-file + Python conductor implementation. The §4 \"Adding a New Version\" instructions point developers to the wrong path, so files added there would never be discovered.

Safe to merge after updating the RFC §4 instructions to match the actual flat-file discovery model.

The implementation itself is sound — the conductor logic, venv isolation, and version discovery are all well-designed. The one blocking concern is that the RFC §4 'Adding a New Version' section actively points contributors to the wrong directory structure, which would silently break the migration chain whenever someone follows it.

docs/tech/rfc/0036-260420-migration-test-refactor.md — §2.0 directory layout, §2.4 shell script reference, and §4 'Adding a New Version' instructions all need updating to match the actual implementation.

| Filename | Overview |
|----------|----------|
| docs/tech/rfc/0036-260420-migration-test-refactor.md | RFC describes a directory-based setup structure and shell-script orchestrator that don't match the actual flat-file + Python conductor implementation; §4 "Adding a New Version" gives instructions that would create files in the wrong location. |
| integration/migration_conductor.py | New Python conductor: manages Core binary lifecycle and orchestrates setup/verify chain; clean stdlib-only design with good error handling; PYTHONPATH not cleared before venv invocations (previously flagged). |
| .github/workflows/test.migration.yaml | Workflow simplified: replaces multi-job matrix generation with a single resolve+test pattern; version discovery via gh release list is clean; correct cleanup/upload-artifact steps preserved. |
| integration/tests/migration/channels_setup.py | Flat setup script with SETUP_VERSION constant; module-level constants are importable by verify scripts; setup logic correctly gated under __main__. |
| integration/tests/migration/setup.py | Shared helpers for setup scripts; docstring clearly explains the examples-import exception for OPC/Modbus simulators. |
| integration/tests/migration_tests.json | Updated to reference _verify suffix test cases matching the new naming convention. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[mc --from 0.55.0] --> B[discover_setup_scripts\nSETUP_VERSION <= 0.55]
    B --> C{setup groups found?}
    C -- yes --> D[For each version group]
    D --> E[install_version\ndownload Core binary]
    E --> F[start_core binary]
    F --> G[create_setup_venv\nuv pip install synnax>=X.Y,<X.Y+1]
    G --> H[run_setup_script\nfor each *_setup.py]
    H --> I[stop_core]
    I --> D
    C -- no --> J[No-op: start + stop from_version Core]
    D --> K{already_ran from_version?}
    K -- no --> J
    K -- yes --> L[start latest Core\nCI binary or go run]
    J --> L
    L --> M[uv run tc migration\nrun verify tests]
    M --> N{pass?}
    N -- yes --> O[exit 0]
    N -- no --> P[exit 1]
```

<sub>Reviews (3): Last reviewed commit: ["docstring update"](https://github.com/synnaxlabs/synnax/commit/8b7335ccfbbf5b318f7703f384c19c517e2d1dea) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29073440)</sub>


docs/tech/rfc/0036-260420-migration-test-refactor.md
```diff
@@ -0,0 +1,246 @@
...
+**Verification** (`uv run tc migration_verify`): A standard test conductor invocation
+against the latest Core (started separately). Uses the current workspace's `synnax`
+client via `uv run`. Full access to test infrastructure — Playwright, assertions, result
+reporting.
+
+## 2.1 - Version Folder Setup Scripts
+
+Each `v0_XX/setup.py` is a standalone Python script. It connects to Core at
+`localhost:9090` via `import synnax`, creates resources (channels, ranges, tasks, etc.),
+and exits. No `TestCase` inheritance, no framework imports, no assertions.
```
sy-nico: `no assertions` means that setup can silently fail. Do we have a plan for checking that the setup step worked? The intent of running the simulators was to verify that the tasks function successfully. Removing the simulators means reducing coverage.

Without this, if the `*_verify.py` cases fail, we wouldn't know if it was due to a migration issue or a silent infrastructure failure.

docs/tech/rfc/0036-260420-migration-test-refactor.md
```diff
@@ -0,0 +1,246 @@
...
+
+Setup scripts should NOT start simulators. They use the Synnax API to create and
+configure tasks directly. The verify phase handles starting simulators and validating
+that tasks actually run.
+
+When a new version is released, a snapshot of the current resource-creation logic is
+frozen into a new version folder. Once committed, it is never modified.
```
sy-nico: What if we need to add test coverage to an existing test case? Maybe add a new valve type to the schematics migration setup.

If the setup script is frozen at release and only installable from PyPI after publish, how do you test the setup script before it ships? Right now you can iterate on migration test code on a branch and run it via workflow dispatch. That development loop disappears in this proposal.

docs/tech/rfc/0036-260420-migration-test-refactor.md
```diff
@@ -0,0 +1,246 @@
...
+The fix must ensure that setup code for version X only ever runs against the version X
+client, with dependencies resolved by the package manager — not pinned manually.
+
+# 2 - Design
+
+## 2.0 - Architecture
+
+Migration testing is split into two independent concerns:
+
+**Setup chain** (`run_migration_chain.sh`): A shell script that loops through old
```
sy-nico: Moving to bash script will require an additional PS script and in general will be harder to debug, test, and extend than a python script. We'd also lose structured logging, timeout handling, and would have to completely rewrite the process management.

The pseudo code for `run_migration_chain.sh` doesn't show any of that, presumably because some of the duties would be assumed by `setup.py`? Either way, it means we would be adding unnecessary complexity just to launch the Synnax Core from a bash script instead.

docs/tech/rfc/0036-260420-migration-test-refactor.md
```diff
@@ -0,0 +1,246 @@
...
+
+When the shell script runs `uv pip install synnax==0.54.0`, the resolver automatically
+pulls the correct transitive dependencies (`pymodbus`, `asyncua`, `numpy`, etc.) that
+0.54.0 declared in its package metadata. No manual `deps.txt` or version pinning is
+needed.
+
+For the verify phase, `uv run` resolves `import synnax` to the local workspace code via
+the uv workspace configuration. No wheels, no build step.
+
+## 2.3 - Verify Tests
```
sy-nico: Is the plan still to support migration chain testing?

`v0.53.0 (Setup)` -> `v0.54.4 (Verify)` -> `v0.55.0 (Verify)` ?

docs/tech/rfc/0036-260420-migration-test-refactor.md
sy-nico: I can see the argument that we don't need to verify the tasks during `setup.py` because we already have coverage in the integration tests. But following that chain, we would need to verify against those integration tests at the previous version either:
* By looking for a release workflow and using the ref ID (adds complexity to workflow dispatch)
* Running the specific integration tests at the previous version before verifying (adds complexity and time), so we'd end up reinventing what we already have

docs/tech/rfc/0036-260420-migration-test-refactor.md
```diff
@@ -0,0 +1,246 @@
...
+
+Each `v0_XX/setup.py` is a standalone Python script. It connects to Core at
+`localhost:9090` via `import synnax`, creates resources (channels, ranges, tasks, etc.),
+and exits. No `TestCase` inheritance, no framework imports, no assertions.
+
+Setup scripts should NOT start simulators. They use the Synnax API to create and
+configure tasks directly. The verify phase handles starting simulators and validating
+that tasks actually run.
+
+When a new version is released, a snapshot of the current resource-creation logic is
```
sy-nico: `a snapshot of the current resource-creation logic is frozen into a new version folder`

I don't understand how the snapshots are supposed to function in this refactor. Is this test infrastructure or pulled from production code? If test infrastructure, what is the source of truth? Are these python scripts or config files?

sy-nico (changes requested):
This RFC solves the dependency drift problem, but introduces some other foot-guns and reduces overall coverage.

Migration tests won't come into affect until the *next* release. How do we validate during development (locally and via workflow dispatch). For example, how do we test `v0_55/setup.py` before `0.55` is on PyPI? The RFC doesn't address this. You'd need either:

* A way to install the local client into the isolated venv (which reintroduces the dependency problem)
* A "dry run" mode that uses the current workspace client
* Accept that setup scripts are only validated post-release (introduces risk)

I want to be able to run this workflow on any branch at any time. Either with a true migration (old -> build on my branch), or setup/verify both run at the current build (for baseline validation).

docs/tech/rfc/0036-260420-migration-test-refactor.md
pjdotson: I think the "chain" should be multiple setups, e.g. v0.53 (Setup) -> v0.54 (Setup) -> Current commit (Verify). It doesn't make too much senses to add migration testing for already released software, the bigger win is making sure current software supports migrations.

docs/tech/rfc/0036-260420-migration-test-refactor.md
pjdotson: I think that the `run_migration_chain.sh` should be a very minimal script for orchestrating the setup scripts. The biggest concern I have is with the dependency resolution piece. I think the best way to go about it is to have 1) the run migration chain script that is either a Python (only stdlib imports) or bash script to do the orchestration, 2) `setup.py` files for each version that only use stdlib imports and the Synnax library (the Synnax library will be scoped to whatever that version folder is).

docs/tech/rfc/0036-260420-migration-test-refactor.md
pjdotson: The setup script won't be frozen at release. We could still have the ability to write a setup script for older versions that have already been released. This language on the RFC is a little unclear.

docs/tech/rfc/0036-260420-migration-test-refactor.md
pjdotson: Maybe we should have assertions / some type of way for these to fail? I am not super sure on this point. However, I don't think it should be inheriting from the `TestCase` and other classes as it should just be a very simple script.

pjdotson:
> This RFC solves the dependency drift problem, but introduces some other foot-guns and reduces overall coverage.
> 
> Migration tests won't come into affect until the _next_ release. How do we validate during development (locally and via workflow dispatch). For example, how do we test `v0_55/setup.py` before `0.55` is on PyPI? The RFC doesn't address this. You'd need either:
> 
> * A way to install the local client into the isolated venv (which reintroduces the dependency problem)
> * A "dry run" mode that uses the current workspace client
> * Accept that setup scripts are only validated post-release (introduces risk)
> 
> I want to be able to run this workflow on any branch at any time. Either with a true migration (old -> build on my branch), or setup/verify both run at the current build (for baseline validation).

We don't need to add a script for `v0_55/setup.py` until we want to introduce new "introduction" code that starts at v55 (e.g., we add a new resource type in v0.55). So we can introduce that after we release v0.55, since there won't be any migration before then. The migration tests should only use already existing Python client / Core versions, which make sense because we only care about migration stability for software that has already been released.

.github/workflows/test.migration.yaml
sy-nico, line 1: The scripts are what needs to handle the deployment logic, not the yaml. The way this is currently set up, and with the current directory structure, means that if I test `v0.55 -> Latest`, nothing in the setup is loaded. 

The verify step will pull the latest binary and run verify on system data that doesn't exist. The scripts still need to do the heavy lifting to know that the channels, tasks, and rbac setup still needs to run. 

The other side is that as we grow, this system means we need to constantly add to the migration chain.
`v0.55 -> v0.56 -> v0.57 -> ... -> latest`.  

What this is doing is: 
`setup -> setup -> setup -> ... -> verify`

A migration chain test should be to load everything once and verify data integrity survives the updates:
`setup -> verify -> verify -> ... -> verify`

sy-nico (changes requested):
We should still go back to the model of having a script that orchestrates launching the cores and running the setup scripts. 

* We now need to have separate sets of steps for Ubuntu and WIndows, rather than one platform agnostic script
* Expanding on the logic to handle loading the various `setup.py` scripts for the revs is going to be messy.
* The logic for the setup does not allow me to select specific jump points and load in all of the relevant tests.
* logic to run verification is not functional when a corresponding setup script has not been run (tries to verify each time).

Manually triggered run: [`v0.54 -> latest`](https://github.com/synnaxlabs/synnax/actions/runs/24754195567)

.github/workflows/test.migration.yaml
pjdotson: The logic to select a specific jump point should be easily added by just allowing for an argument to the script of `from_version`, as stated in the RFC

Personally I think we should not be doing most of that logic through a YAML file, and instead be running some type of script (shell / powershell / python / whatever) for the orchestration. That way, you can test it locally.

Why would you want to test `v0.55 -> Latest`? `v0.55` isn't out yet.

I don't agree that we should be doing the verify multiple times. If we verify once at the end, that covers, unless you are worried about a scenario where we write data in v1, it gets corrupted in v2, and it gets uncorrupted in v3, but we want to go back and make a new v2.1 release when we already have v3 out. We currently don't have a support policy for old releases where we fix them for bugs, and what you are suggesting with `setup -> verify -> verify` is that we should be doing that. I don't think that makes sense at our stage.

sy-nico:
@greptile, update your review. RFQ was not followed 1:1 due to more discovery and conversation happening online. This refactor resolves the primary dependency drift issues involved with building the wheels.

------

Now, setups are run at a pre-defined version and chain up until the target "from" migration start point. This ensures we test data integrity throughout the upgrades. 
We preserve the setup/verify script mirroring for proper dev x and visual verification of coverage. By tagging the scripts, we can then build the setup chain groups by regex.

-------

We do not normally need to rewrite the setup if, say, there is a change in the client library (such as `channels.create() -> channel.create()`.  That interface will be updated/tested in the standard `test.integration` workflow. We are testing that the data survives upgrades, not the interface used to create it.

 In most situations, we can update the setup scripts without increasing the version tag (like adding coverage). In situations where there is a breaking change, we can:
A) Update the version tag. We would lose confidence in migrating from older versions for the specific topic, but it could be outweighed by the necessity to prioritize the new coverage
B) Create a new, independent test case (e.g. `ranges_new_feature_setup.py`) at a newer tag.

The options are not mutually exclusive.

-----

This refactor changes the migration to `v49_setup -> v52_setup -> v_base_no_op -> v_latest_verification`. It continues using python scripts for orchestration to simplify cross-platform usage. The new architecture allows for a simpler yaml, orchestrator, and setup script (decoupled from test_case), meaning that even accounting for Core restarts, the overall job runs faster and is easy to extend depending on needs.

integration/tests/migration/task_opc_setup.py
```diff
@@ -0,0 +1,76 @@
...
+IDX_NAME = "mig_opc_idx"
+CHANNEL_PREFIX = "mig_opc_float"
+NUM_CHANNELS = 2
+
+if __name__ == "__main__":
+    from examples.opcua.server import OPCUASim
+    from setup import kill_port, log, run
+
+    def setup(client: sy.Synnax) -> None:
+        kill_port(OPC_UA_PORT)
```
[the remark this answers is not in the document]
sy-nico: This is not an issue and one of the acceptable deviations from the RFQ. Addressed concern with doc string in `setup.py`

sy-nico:
@greptile, Update the review. Your main concern is a non-issue. This is an acceptable deviation from the RFQ that is explained in the updated doc string of `setup.py`.

integration/tests/migration/channels_calc_setup.py
```diff
@@ -0,0 +1,266 @@
...
+"""Migration setup: calculated channels with operations, expressions, and windows."""
+
+import numpy as np
+
+import synnax as sy
+
+SETUP_VERSION = "0.49"
+
+S = sy.TimeSpan.SECOND
+MS = sy.TimeSpan.MILLISECOND
```
pjdotson: pointless

integration/tests/migration/channels_calc_setup.py
```diff
@@ -0,0 +1,266 @@
...
+    t = d * WIN_DOMAIN_GAP_S
+    if seed == "cosine":
+        return float(np.cos(2 * np.pi * t / 2.5))
+    return float(t**2)
+
+
+if __name__ == "__main__":
+    from setup import log, run
+
+    def setup(client: sy.Synnax) -> None:
```
pjdotson: this `def setup` can go outside `__main__` and then you do if name is main setup()

integration/tests/migration/channels_calc_setup.py
```diff
@@ -0,0 +1,266 @@
...
+                expression=PASSTHROUGH_EXPR,
+                operations=[
+                    sy.channel.Operation(
+                        type=op_type,
+                        duration=duration,
+                        reset_channel=reset.key if uses_reset else 0,
+                    )
+                ],
+                retrieve_if_name_exists=True,
+            )
```
pjdotson, line 123: nit: should one create call for all channels

integration/tests/migration/setup.py
```diff
@@ -0,0 +1,85 @@
...
+
+import synnax as sy
+
+HOST = "localhost"
+PORT = 9090
+USERNAME = "synnax"
+PASSWORD = "seldon"
+
+S = sy.TimeSpan.SECOND
+MS = sy.TimeSpan.MILLISECOND
```
pjdotson: bad

integration/migration_conductor.py
```diff
@@ -193,166 +255,229 @@ def stop_core(proc: subprocess.Popen[bytes]) -> None:
...
+    return f"synnax>={major}.{minor},<{major}.{minor + 1}"
 
 
-def _create_client_venv(version: str) -> Path:
-    """Create a venv and install pre-built client wheels for the given version.
+def _version_to_release(version: str) -> str:
+    """Convert a version string to a release version for downloading
+    Core binaries. E.g., '0.54' -> '0.54.0'"""
+    major, minor = _parse_version(version)
+    return f"{major}.{minor}.0"
```
pjdotson: this logic should use the latest version, not version .0

pjdotson: for both the Core and PyPI

integration/migration_conductor.py
```diff
@@ -193,166 +255,229 @@ def stop_core(proc: subprocess.Popen[bytes]) -> None:
...
+
+
+def _read_setup_version(script: Path) -> str | None:
+    """Read the SETUP_VERSION constant from a setup script."""
+    m = re.search(
+        r'^SETUP_VERSION\s*=\s*["\']([^"\']+)["\']',
+        script.read_text(),
+        re.MULTILINE,
+    )
+    return m.group(1) if m else None
```
pjdotson: just use module import

.github/workflows/test.migration.yaml
```diff
@@ -7,86 +7,80 @@ on:
...
+            | sed 's/^synnax-v//' \
+            | sort -V \
+            | tail -n1
+          )
+          if [[ -z "$VERSION" ]]; then
+            echo "ERROR: Could not find any major release" >&2
+            exit 1
+          fi
+          echo "base_version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "Resolved latest major release: $VERSION"
```
pjdotson: confused

pjdotson: (this is probably in Python setup script), default should be latest versions of Synnax, NOT latest major release.

integration/migration_conductor.py
```diff
@@ -31,12 +41,12 @@
...
-import tomllib
+from datetime import datetime
 from pathlib import Path
 
-BINARY_CACHE_DIR = Path.home() / "synnax-binary-cache"
-BINARY_DIR = Path.home() / "synnax-binaries"
-DATA_DIR = Path.home() / "synnax-data"
+MIGRATION_BINARY_DIR = Path.home() / "synnax-migration-binaries"
+CI_BINARY_DIR = Path.home() / "synnax-binaries"
+DATA_DIR = Path.home() / "synnax-migration-data"
```
pjdotson, line 50: home/synnax/integration-testing

not a huge deal

PR #2260, opened 2026-04-24
SY-4112: Add Schematic Symbol to Embed Media and iFrames

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4112](https://linear.app/synnax/issue/SY-4112/create-embedcontainer-schematic-symbol)

## Description

* Add support for "Embed" symbols in the schematic:
  * `MediaEmbed` supports mjpeg, svg, png, and gifs
  * `IframeEmbed` supports iframes (e.g. Grafana dashboards) with a `sandbox`
    attribute for security. A "Block Cookies" toggle controls whether the
    embedded site can use cookies (`allow-same-origin`).
  * Stubs for `PageEmbed` to later support embedding Synnax logs, plots, and
    tables. Proof of concept was shown for page embeds, but that feature will
    require a larger refactor and is outside the scope of this feature.
* Edge resizing for symbol components via drag handles on edges and corners.
* `useStaticData` search improvements: Fuse.js keys are now collected from all
  items (not just the first), and `ignoreLocation` enables substring matching
  anywhere in a field.

image

.

image

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds `MediaEmbed` (images/MJPEG/GIF/SVG), `IframeEmbed` (with `sandbox` and a "Block Cookies" toggle), and a `PageEmbed` stub as new schematic symbols, alongside edge/corner drag-resizing for symbol components and improved `useStaticData` Fuse.js search (all-item key collection + `ignoreLocation`). Previously flagged issues — MJPEG stream abort on cleanup, `onChange` infinite-loop via ref (`useSyncedRef`), and `expect()` panics in the Tauri setup closure — are all addressed in this revision.

Safe to merge after resolving the outstanding P1 in link.ts noted in the previous review thread.

A P1 regression in console/src/log/services/link.ts (data key leaked into spread, schema validation removed) was flagged in a prior review and remains unaddressed in the current diff. All other previously flagged issues have been fixed. New code is well-tested and the remaining findings are P2 style notes.

console/src/log/services/link.ts — the { ...log.data, ...log } spread and missing anyStateZ.parse() validation.

| Filename | Overview |
|----------|----------|
| console/src-tauri/src/main.rs | Adds `create_webview_window` Tauri command and manually constructs the main window in setup to attach an `on_new_window` handler; previous `expect()` panics replaced with `?`; uses a deprecated API suppressed by `#[allow(deprecated)]` without documentation. |
| console/src/log/services/link.ts | Simplifies retrieve/spread pattern, but introduces a `data` key back into the object passed to `Log.create` and removes `anyStateZ.parse()` validation — already flagged in a previous review thread. |
| drift/src/tauri/index.ts | Extracts `buildWindowOptions` helper and replaces `WebviewWindow` constructor + event-listener dance with `invoke("create_webview_window")`; position visibility check preserved before options are built. |
| pluto/src/schematic/symbol/Symbols.tsx | Adds MediaEmbed, IframeEmbed, PageEmbed symbols and `createLabeled` resize delegation; previous feedback on MJPEG cleanup and `onChange` ref is addressed; minor redundant `hasURL` dep in useEffect. |
| pluto/src/schematic/symbol/Grid.tsx | Adds `onResize` prop wired to `NodeResizeControl` handles on edges and corners; minor `...rest` shadowing inside the map callback. |
| pluto/src/schematic/symbol/registry.ts | Registers `mediaEmbed` and `iframeEmbed` variants with default props, search terms, and a new Containers group; `withSuffix` helper neatly generates compound search term strings. |
| pluto/src/list/useStaticData.ts | Collects Fuse.js keys from all items (not just the first) and adds `ignoreLocation: true` to enable substring matching anywhere in a field. |
| pluto/src/schematic/Schematic.css | Lifts resize-control visibility from inside `.pluto-symbol-primitive` to any `.react-flow__node.selected`, enabling resize handles on embed nodes that don't use the primitive class. |
| pluto/src/schematic/symbol/Primitives.tsx | Adds `Embed` primitive with fixed-position handles and `overflow: hidden` container; removes unused `onResize` prop from `TankProps` and `CylinderProps` since resize is now centralized in Grid. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant TS as TypeScript drift
    participant Rust as Tauri Rust Command
    participant WV as WebviewWindow

    Note over TS,WV: Window creation (new flow)
    TS->>TS: buildWindowOptions(label, props)
    TS->>Rust: invoke create_webview_window options
    Rust->>WV: WebviewWindowBuilder from_config build
    WV-->>Rust: Ok
    Rust-->>TS: resolve

    Note over WV,Rust: Popup intercept
    WV->>Rust: on_new_window url
    Rust->>Rust: shell.open to external browser
    Rust-->>WV: NewWindowResponse Deny

    Note over TS: IframeEmbed render
    TS->>TS: IframeEmbedBase renders iframe with sandbox
    TS->>TS: iframeScaleStyle computes CSS transform
```

<details><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `pluto/src/schematic/symbol/Symbols.tsx`, line 901-908 ([link](https://github.com/synnaxlabs/synnax/blob/48aae05fbb8564fea1043964e3c610fc3028f438/pluto/src/schematic/symbol/Symbols.tsx#L901-L908)) 

   <a href="#">P2</a> <a href="#">security</a> **`allow-scripts` + `allow-same-origin` sandbox escape**

   When both `allow-scripts` and `allow-same-origin` are set together on an iframe sandbox, a script running inside the iframe — if it is same-origin with the Synnax app — can call `window.frameElement.removeAttribute("sandbox")` and fully escape the sandbox. MDN explicitly warns against this combination for untrusted content. For cross-origin embeds (e.g. Grafana on a different host) `window.frameElement` returns `null`, so the risk doesn't materialise in the common case, but it's worth adding a code comment here to document the known limitation alongside the existing postMessage warning.

2. `console/src/log/services/link.ts`, line 92-93 ([link](https://github.com/synnaxlabs/synnax/blob/5ddf4cff0e85da1410f3da3dd5fe4a1994746a2e/console/src/log/services/link.ts#L92-L93)) 

   <a href="#">P1</a> **`data` key leaked into spread and schema validation removed**

   The old code destructured `data` out of the retrieve result so that `log` did **not** contain a `data` key, then validated it with `anyStateZ.parse(data)`. The new code spreads `...log` directly, which re-introduces `data: log.data` into the object passed to `Log.create`. This means `Log.create` now receives both the individual flattened fields from `log.data` **and** a top-level `data` key pointing at the same nested object — a different shape than before. The silent removal of `anyStateZ.parse()` also drops the schema validation that previously guaranteed the data conformed to the expected shape before layout creation.

3. `console/src/log/services/link.ts`, line 3-5 ([link](https://github.com/synnaxlabs/synnax/blob/ac820002c0f7f35926be297d1631eb626f84ba8d/console/src/log/services/link.ts#L3-L5)) 

   <a href="#">P1</a> **`data` key re-introduced and schema validation removed**

   `{ ...log.data, ...log }` spreads `log.data`'s fields first, then overwrites with `log`'s top-level fields — which include a `data` key pointing at the same nested object. `Log.create` now receives an unexpected `data` property that the old code explicitly stripped out via destructuring. The old code also validated through `anyStateZ.parse(data)` before use; that guard is gone, so a malformed server response will propagate silently instead of throwing at the boundary.
</details>

<sub>Reviews (11): Last reviewed commit: ["SY-4112:  Hold onChange in a ref so it&#39;s..."](https://github.com/synnaxlabs/synnax/commit/d9df3ff2f0ffe862e5ded7895b238e982c92968d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29634490)</sub>


console/src/log/services/link.ts
```diff
@@ -9,9 +9,8 @@
 
 import { type Link } from "@/link";
 import { Log } from "@/log";
-import { anyStateZ } from "@/log/types";
 
 export const handleLink: Link.Handler = async ({ client, key, placeLayout }) => {
-  const { data, ...log } = await client.logs.retrieve({ key });
-  placeLayout(Log.create({ ...anyStateZ.parse(data), ...log }));
+  const log = await client.logs.retrieve({ key });
+  placeLayout(Log.create({ ...log.data, ...log }));
```
pjdotson, line 16: why aren't we parsing the log data?

console/src-tauri/src/main.rs
```diff
@@ -100,6 +120,21 @@ fn main() {
...
+            WebviewWindowBuilder::from_config(app.handle(), &main_window_config)
+                .expect("Failed to create main window builder")
+                .on_new_window(move |url, _features| {
+                    #[allow(deprecated)]
+                    let _ = tauri_plugin_shell::ShellExt::shell(&app_handle_for_main)
+                        .open(url.to_string(), None::<tauri_plugin_shell::open::Program>);
+                    NewWindowResponse::Deny
+                })
+                .build()
+                .expect("Failed to build main window");
```
pjdotson: let's talk in person about why we have to do all of this stuff. I am not really
convinced it's necessary, especially since the documentation iframe already embeds fine

pluto/src/list/useStaticData.ts
```diff
@@ -53,9 +53,11 @@ export const useStaticData = <
   }, [data, filter, sort]);
   const fuse = useMemo(() => {
     if (filteredData.length === 0) return null;
+    const keys = new Set(filteredData.flatMap(Object.keys));
     return new Fuse(filteredData, {
-      keys: Object.keys(filteredData[0]),
+      keys: [...keys],
       threshold: 0.3,
+      ignoreLocation: true,
```
pjdotson, line 60: what does `ignoreLocation` do?

pluto/src/schematic/symbol/Symbols.tsx
```diff
@@ -1347,3 +1379,98 @@ export const StateIndicatorPreview = ({
...
+}
+
+export const MediaEmbedBase = ({
+  dimensions: dims,
+  color: colorVal,
+  url,
+}: SymbolProps<MediaEmbedProps>): ReactElement => (
+  <Primitives.Embed dimensions={dims} color={colorVal} placeholder="Enter a URL">
+    {url != null && url.length > 0 ? (
+      <img src={url} style={{ width: "100%", height: "100%", objectFit: "contain" }} />
```
pjdotson: should this always be wrapped in a `img` tag? why not use a `video` tag for videos and
live streams?

pjdotson (changes requested):
Two things:

- I really don't think we should allow a general iframe embed. iframes could have potential security risks from users embedding untrusted content, all they are is just embedding a mini browser where it is likely that the user would just prefer the actual browser. It feels disconnected from our product, and I would really not want to add it unless we see a lot of evidence of user demand for it. Also, there is going to be a split between use cases for Synnax where it may primarily be used via a desktop app (like having an operator on-console during a test) and where it may primarily be used via browser (a factory setting where a lot of things are constantly running). I would be a little worried about users getting frustrated if a link that is supposed to open in a new tab doesn't work in an iframe for instance. I don't completely know what the right answer here is, but am curious to hear your thoughts. Personally, I think what might make more sense is an "image" and "video" element where you can link to an image or a video, and then we wrap that in the proper tag (we already have a `Video` element in Pluto).
- there is a lot of edited code in both the Drift and Rust code that I am not sure is necessary and also isn't tested / can cause odd issues. We should talk through this together.

pluto/src/list/useStaticData.ts
sy-nico: It tells Fuse.js to score matches equally regardless of where they appear in the string. For MediaEmbed, I wanted various searches to work:
"Media Embed, Picture Embed, Video Embed.... PNG Embed...."

The problem was that PNG Embed was so deep into the string, that searching for "PNG Embed" returned 0 results instead of finding the MediaEmbed symbol.

pluto/src/schematic/symbol/Symbols.tsx
sy-nico: Using a `video` tag prevents us from rendering PNGs, Gifs, SVGs, or other image formats. `video` is needed for HLS/DASH/MP4, which I don't see a use case for now. Maybe in the future when we have proper video stream handling.

console/src-tauri/src/main.rs
sy-nico: Without this, right-clicking a link inside an iframe and selecting "Open in New Window/Tab" silently fails (no URL opens, regardless of protocol). Tauri's default webview denies new window requests when no `on_new_window` handler is set. The change attaches a handler that opens the link in the native browser, which then let's us open `synnax://` urls that are embedded in the iframe page content.

What I tried:

* `allow-top-navigation-by-user-activation` in the iframe sandbox. Doesn't fire for custom protocols like `synnax://`
* Tauri's on_navigation handler. Fires for http/https but not custom protocols from sandboxed iframes
* `registerProtocolHandler` Browsers only support web+ prefix


Another thing I tried to do was to "intercept" clicks on a `synnax://` hyperlink to allow users to seamlessly open the page links. Since they are not valid links, when clicked from within an iframe, no navigation occurs. But I couldn't get this to work.

console/src/log/services/link.ts
sy-nico: The other handlers don't parse either, and the spread of .data already extracts the state. I was just aiming for consistency with the other `link.ts` files. Should I change this?

sy-nico:
@pjdotson , 

* The `` tag is more encompassing, and the pluto `video` tag you mentioned only handles `MP4`. It seems less invasive to use the `img` tag and not have to touch the pluto code, no?

* On security, the iframe is sandboxed by default with sandbox="allow-scripts", which blocks cookies, form submission, popups, same-origin access, and top navigation. This is more restrictive than a regular browser tab. We also provide a "Block Cookies" toggle (default: on).

* The use case is embedding live dashboards directly into schematics so operators don't have to context-switch to a browser. It's the same reason we have MediaEmbed for camera feeds. Inline visibility matters for control workflows and ops management. Of the folks I have reached out to, they were all excited at the possibility of the new feature. Supporting iFrames is a low-cost way to enable support for ideas that we could never consider, but would resolve real day-to-day pains for an operator. If Synnax can be used for anything, it would follow that it should be easy to integrate anything into Synnax.

console/src/log/services/link.ts
pjdotson: No this is fine, just wondering.

pluto/src/schematic/symbol/registry.ts
```diff
@@ -1930,6 +1943,71 @@ const customStatic: Spec<CustomStaticProps> = {
...
+    "google",
+    "iframe",
+    "webpage",
+    "website",
+    "dashboard",
+    "grafana",
+    "widget",
+    "html",
+    "web",
+  ),
```
pjdotson, line 2008: I don't think these search terms make sense. Are you going to include anything website
are users could possibly use? I would get rid of all of these `searchTerms` for both of
these elements and instead just choose better names like "Embedded Website" (users
probably don't know what an iframe is) and "Embedded Media" or just "Website" and "Media".

pluto/src/schematic/symbol/registry.ts
```diff
@@ -2188,4 +2268,10 @@ export const GROUPS: Group[] = [
       "flowStraightener",
     ],
   },
+  {
+    key: "containers",
+    name: "Containers",
+    Icon: Icon.Visualize,
+    symbols: ["mediaEmbed", "iframeEmbed"],
+  },
```
pjdotson, line 2276: Do you really want to put this into a new thing instead of just sticking in general
right now?

pjdotson: I would probably stick with general until we have more things to put in this category
that way it is easier for users to find.

pluto/src/schematic/symbol/Primitives.tsx
```diff
@@ -4861,3 +4859,45 @@ export const StateIndicator = ({
...
+      <HandleBoundary orientation="left">
+        <Handle location="left" orientation="left" left={0} top={50} id="1" />
+        <Handle location="right" orientation="left" left={100} top={50} id="2" />
+        <Handle location="top" orientation="left" left={50} top={0} id="3" />
+        <Handle location="bottom" orientation="left" left={50} top={100} id="4" />
+      </HandleBoundary>
+      <div className={CSS.BE("embed", "content")}>
+        {children ?? (
+          <Text.Text level="small" color={textColor}>
+            {placeholder}
```
pjdotson, line 4897: I don't think you should automatically do a placeholder here, instead let the preview
declare it in `Symbols.tsx`

pjdotson (changes requested):
Left a couple comments and think embedding the websites is a good idea but let's please talk sync about the Drift + Rust changes

PR #2262, opened 2026-04-25
SY-4114: Variable-Density Sample Optimizations and Fixes

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4114](https://linear.app/synnax/issue/SY-4114)

## Description

Resolves three weaknesses in cesium's variable-length offset cache and the rollover path that supports it.

**1. Tracker → cache flush on commit.** The unary writer now publishes its accumulated per-sample byte offsets into the resolver's offset cache after every successful `dw.Commit`. Previously the tracker built the offset list during writes and discarded it on close, so every reader paid a full length-prefix scan on first access — even one second after the writer finished. The publish call snaps the tracker's offsets into the cache keyed on the just-committed domain's start timestamp.

**2. Rollover correctness.** A new optional `OnRollover func(commitEnd telem.TimeStamp)` callback on `domain.WriterConfig` fires from inside `Commit` immediately after a file rollover. The unary tracker installs itself as the callback so it can publish the just-finished domain's state, reset its per-domain fields (currentStart, domainBytes, domainOffsets), and advance to the new domain. Cumulative session sample count is preserved across rollovers so alignment math and end-timestamp lookups stay correct.

**3. Cache key shape.** The offset cache is now keyed by the domain pointer's Start timestamp instead of its position in the index. Pointer Starts are unique within a DB (non-overlap invariant), immutable once set (`update` panics if Start changes), and shift-immune (positions move when other pointers insert before; Starts do not). The end-timestamp gate replaces the byte-size gate. Together these mean the cache no longer leaks domain-layer addressing details into the unary layer; both sides now coordinate via `iter.TimeRange()` and `cfg.Start`/`commitEnd`, which are already public domain vocabulary. **Zero new domain-layer surface beyond the optional `OnRollover` hook.**

**4. Buffered cold rebuild.** `buildOffsetTable` was reading 4 bytes at a time via `ReadAt` per sample — one syscall per length prefix. Now wraps the domain reader in `bufio.NewReaderSize` with a 64 KiB buffer capped at the domain's size. For an event-rate domain that's typically one buffered read covering thousands of samples, instead of thousands of 4-byte syscalls.

**5. Singleflight on cache miss.** `tableFor` routes cache misses through `golang.org/x/sync/singleflight.Group` keyed on the domain's start. Concurrent first-readers of the same cold domain collapse into a single scan. Different domains have different keys, so they don't block each other.

**6. Byte-order consolidation.** `telem.ByteOrder` is the canonical encoding for variable-length prefix decoding, but seven sites in `x/go/telem` and `cesium/internal/unary` were using `binary.LittleEndian` directly. All length-prefix encode/decode paths now go through `telem.ByteOrder` — changing the wire format requires editing exactly one line.

**7. Recording FS utility.** Added `xfs.Recorder` in `x/go/io/fs/`, a wrapper that records every `Open`, `Read`, `ReadAt`, `Write`, and `WriteAt` performed against any wrapped FS. `Sub`-derived FSes share the parent's event log so events from every branch of the tree appear in one ordered timeline. Used by the new variable-length tests to assert observable I/O patterns (e.g., \"this read produced no 4-byte ReadAts\") without spying on internal state.

## Tests

Three new tests in `cesium/internal/unary/variable_test.go`, each running under both memFS and osFS:

- **Flush on commit.** Writes 100 samples, commits, reads back, asserts zero 4-byte ReadAts on the data file (would be 100+ without the flush).
- **Rollover.** Forces a writer file rollover by setting FileSize=40 bytes, writes across both domains, reads everything back, asserts zero 4-byte ReadAts (both domains' offsets are published on rollover).
- **Cold rebuild.** Writes through one DB, closes it, reopens against the same FS so the cache is empty, reads. Asserts zero 4-byte ReadAts (the buffered scan handles the rebuild).

The recorder itself has 14 unit tests covering Open, Read, ReadAt, Write, WriteAt, Sub-derived event sharing, name-scoped queries, Reset propagation, and concurrent safety.

All cesium suites green (1000+ specs across 12 suites).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR resolves several correctness and performance weaknesses in cesium's variable-length offset cache and the writer rollover path. The tracker is moved from `Writer` to `controlledWriter` so per-sample byte offsets survive control handoffs, the cache is now keyed by domain start timestamp (immutable) rather than index position (shift-prone), cold-cache rebuilds use a 64 KiB buffered scan instead of per-sample 4-byte `ReadAt` syscalls, and concurrent cache misses for the same domain are collapsed via `singleflight`. Supporting utilities include a new `xfs.Recorder` and a shared `fstestutil.FileSystems` factory map.

The changes are extensively tested (3 new `variable_test.go` test groups covering flush-on-commit, rollover, and cold rebuild, plus 3 control-handoff scenarios).

Safe to merge; all core logic paths are well-tested and no data-corruption paths were found.

No P0/P1 findings. Two P2s in resolver.go: a phantom offset entry left in the offsets slice after truncation-path discard (harmless given current callers but fragile), and an int(length) narrowing that overflows on 32-bit for samples >2 GB. A comment/log-level mismatch is also noted. Extensive new test coverage and correct tracker-sharing across control handoffs raise confidence significantly.

cesium/internal/unary/resolver.go — truncation handling in buildOffsetTable leaves a phantom offset entry and has a 32-bit narrowing concern in the Discard call.

| Filename | Overview |
|----------|----------|
| cesium/internal/unary/resolver.go | Core cache rework: key changed to domain Start timestamp, singleflight on cache miss, buffered rebuild, new offsetTracker.publish/rollover methods. Minor: phantom offset in truncation branch; comment/log level mismatch. |
| cesium/internal/unary/writer.go | Tracker moved from Writer to controlledWriter so it is shared across control handoffs; publish(end) called after each successful Commit; baseOffset now derived from tracker.domainBytes instead of dw.Len(). |
| cesium/internal/domain/writer.go | Added OnRollover func(commitEnd telem.TimeStamp) hook to WriterConfig; fires inside Commit immediately after file rollover. Clean, minimal surface extension. |
| x/go/io/fs/recorder.go | New Recorder wraps any FS, logs every Open/Read/ReadAt/Write/WriteAt; sub-FSes share the parent's *eventLog. Concurrent-safe; well-documented. |
| cesium/internal/unary/variable_test.go | Extensive new tests: flush-on-commit, cold-rebuild, rollover, multi-commit, and three control-handoff scenarios — all verifying zero 4-byte ReadAt events via xfs.Recorder. |
| x/go/io/fs/testutil/fs.go | New shared Factory / FileSystems map; OpenOS uses ginkgo.DeferCleanup instead of manual cleanup, removing teardown boilerplate from callers. |
| cesium/internal/testutil/fs.go | Simplified to thin aliases over fstestutil.Factory and fstestutil.FileSystems, eliminating duplicated OS/mem FS factory code. |
| cesium/internal/unary/open.go | Single-line change: passes cfg.Instrumentation to newOffsetResolver so truncation warnings have context. |
| x/go/telem/series.go | Replaced binary.LittleEndian.Uint32 with telem.ByteOrder.Uint32 across all variable-length decode sites. Wire-format compatible. |
| x/go/telem/series_factory.go | Same binary.LittleEndian → telem.ByteOrder consolidation for encode paths; no semantic change. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant W as unary.Writer
    participant CW as controlledWriter (tracker)
    participant DW as domain.Writer
    participant Cache as offsetCache
    participant SF as singleflight

    note over W,Cache: Normal write + commit (no rollover)
    W->>CW: Write(series)
    CW->>CW: baseOffset = tracker.domainBytes
    CW->>CW: tracker.record(data, baseOffset)
    CW->>DW: Write(data)
    W->>CW: Commit(ctx)
    CW->>DW: Commit(ctx, end)
    DW-->>CW: ok (no rollover)
    CW->>Cache: tracker.publish(end) → cache.set(currentStart, table)

    note over W,Cache: Rollover path
    W->>CW: Commit(ctx)
    CW->>DW: Commit(ctx, end)
    DW->>CW: OnRollover(commitEnd) callback
    CW->>Cache: tracker.rollover → publish(commitEnd), reset domainOffsets/domainBytes
    DW-->>CW: ok (rollover done)
    CW->>Cache: tracker.publish(end) [no-op: domainOffsets=nil]

    note over W,Cache: Read path (cache miss → singleflight rebuild)
    W->>Cache: tableFor(iter) → cache miss
    Cache->>SF: rebuilds.Do(start:end, func)
    SF->>DW: iter.OpenReader(ctx)
    SF->>SF: buildOffsetTable (bufio 64KiB scan)
    SF->>Cache: cache.set(start, table)
    SF-->>W: offsetTable
```

<sub>Reviews (4): Last reviewed commit: ["updated cesium tests"](https://github.com/synnaxlabs/synnax/commit/bae7b375f13d74068ddc7dbc1009eb30041992c0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29723253)</sub>


x/go/io/fs/recorder_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	xfs "github.com/synnaxlabs/x/io/fs"
+	. "github.com/synnaxlabs/x/testutil"
+)
+
+var _ = Describe("Recorder", func() {
+	var rec *xfs.Recorder
+	BeforeEach(func() { rec = xfs.NewRecorder(xfs.NewMem()) })
```
pjdotson: would be good to also have test here with real FS

cesium/internal/unary/resolver.go
```diff
@@ -64,20 +78,34 @@ func (c *offsetCache) invalidateAll() {
...
+		)
+		lenBuf = make([]byte, 4)
+		pos    int64
+	)
 	for pos+4 <= int64(domainSize) {
-		t.offsets = append(t.offsets, uint32(pos))
-		n, err := r.ReadAt(buf, pos)
-		if err != nil && err != io.EOF {
+		if _, err := io.ReadFull(br, lenBuf); err != nil {
+			if errors.IsAny(err, io.EOF, io.ErrUnexpectedEOF) {
```
pjdotson, line 109: why should we skip on `ErrUnexpectedEOF`?

cesium/internal/unary/resolver.go
```diff
@@ -145,65 +178,151 @@ func (r *offsetResolver) invalidate() {
...
-	// cache hit on the domain size matching what the table was built from.
-	if cached, ok := r.cache.get(domainIdx); ok && cached.domainSize == size {
+func (r *offsetResolver) tableFor(ctx context.Context, iter *domain.Iterator) (*offsetTable, error) {
+	tr := iter.TimeRange()
+	// A pointer's Start is immutable but its End advances as the writer commits more
+	// samples to the same domain. Gate the cache hit on the cached End matching the
+	// pointer's current End so a cached table from before a later commit cannot
+	// answer queries against the extended range.
+	if cached, ok := r.cache.get(tr.Start); ok && cached.end == tr.End {
 		return cached, nil
```
pjdotson, line 232: should you use `==` or `<=` of `>=`

pjdotson (changes requested):

cesium/internal/unary/resolver.go
emilbon99: My take is that we should try to rebuilt the offset cache to the best of our ability regardless of whether the file is corrupted, which means that we should gracefully handle io.ErrUnexpectedEOF. That said, I added logging when this case occurs to ensure that we aware of the issue.

cesium/internal/unary/resolver.go
emilbon99: Updated a comment explaining why == is the right approach.

pjdotson (approved):

PR #2266, opened 2026-04-26
SY-4118: Improve Mosaic Tab Rename Synchronization

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-####]()

## Basic Readiness

<h3>Greptile Summary</h3>

This PR refactors how mosaic tab names are synchronized with remote resources. It introduces a `UseName` hook pattern attached to each layout `Renderer`, a new `CustomTabName` component in the Mosaic that calls `retrieve()` via `useEffect` and `onRename` via a stable callback, and a `createUseFluxName` factory that wires up server-side name observables.

- **`createUseFluxName`** now defaults `isEnabled` to `true` when no `useEnabled` guard is provided, fixing Arc, Range Overview, and similar "always-remote" renderers that previously silently no-op'd on retrieve/rename.
- **`CustomTabName`** replaces per-renderer `useSyncName` calls with a centralized component that calls `retrieve()` in a `useEffect` dependency array (fixing the unstable-ref-callback warning from previous iterations) and correctly discards the unused first argument from `DefaultName.onRename`.
- **Task `wrapForm`** drops `Layout.useSyncName` and instead attaches `useName` directly to the wrapped component, with `onRename` gated on `taskKey != null`.

Safe to merge; the name-sync refactor is structurally sound and all previously identified defects have been corrected.

The two most impactful bugs from earlier iterations — `createUseFluxName` silently no-op-ing when `useEnabled` is omitted, and the unstable ref-callback triggering repeated network fetches — are both addressed. The `handleRename` argument order is now correct. The only new finding is a cosmetic duplicate in a `useCallback` dep array that does not affect behavior.

The duplicate dependency entries in `handleFileDrop` inside `console/src/layouts/Mosaic.tsx` are worth a quick cleanup, but nothing else needs special attention.

| Filename | Overview |
|----------|----------|
| console/src/layout/useFluxName.ts | New factory `createUseFluxName` correctly defaults `isEnabled=true` when `useEnabled` is omitted, fixing the silent no-op regression from the prior `useRemoteCreated?.(key) === true` pattern. |
| console/src/layouts/Mosaic.tsx | Adds `CustomTabName` and `TabName` components for per-renderer name hooks; `handleRename` correctly uses `(_: string, name: string)` signature; minor: `store` and `fluxStore` are duplicated in the `handleFileDrop` `useCallback` dep array. |
| console/src/layout/useFluxName.spec.ts | Comprehensive new tests covering enabled/disabled/undefined guards, callback stability across re-renders, and key-change targeting. |
| console/src/hardware/common/task/Form.tsx | Replaces `Layout.useSyncName` with a `useName` attached to `Wrapper`; `onRename` is correctly gated on `taskKey != null` and `retrieve` similarly guards the base call. |
| console/src/arc/editor/Editor.tsx | Attaches `useName` via `createUseFluxName` without a `useEnabled` guard, so Arc is treated as always-remote — intentional given Arc resources always live on the server. |
| console/src/range/overview/Overview.tsx | Same always-remote pattern as Arc; `useName` wired via `createUseFluxName` with no guard, which is correct for Range resources. |
| pluto/src/tabs/types.ts | Adds `NameProps` interface used by the new `Name` render-prop slot, cleanly separating the name rendering contract from the selector internals. |
| pluto/src/tabs/Selector.tsx | Threads the `Name` prop through `SelectorButton` to allow custom tab name renderers; default `Name=DefaultName` preserves backward compatibility. |
| pluto/src/text/Editable.tsx | Minor change; `outline` prop added with a default, straightforward addition. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant U as User
    participant DN as DefaultName (Pluto)
    participant CTN as CustomTabName
    participant UF as createUseFluxName
    participant Store as Redux Store
    participant Server as Server

    Note over CTN,UF: Mount — retrieve name from server
    CTN->>UF: useName(tabKey, handleLayoutRename)
    UF->>Server: "useRetrieve({ onChange: handleLayoutRename })"
    CTN->>UF: useEffect → retrieve()
    UF->>Server: "baseRetrieve({ key: tabKey })"
    Server-->>UF: observable fires name
    UF-->>CTN: onChange(name) → handleLayoutRename(name)
    CTN->>Store: "dispatch(Layout.rename({ key, name }))"

    Note over U,Server: User renames tab in UI
    U->>DN: double-click → type new name
    DN->>CTN: onRename(tabKey, newText)
    CTN->>Store: handleLayoutRename(name) → dispatch(Layout.rename)
    CTN->>Server: "onRename(name) → update({ key, name })"
    Server-->>UF: observable confirms new name
    UF-->>CTN: onChange(name) → handleLayoutRename(name)
    CTN->>Store: dispatch(Layout.rename) [idempotent]
```

<details><summary><h3>Comments Outside Diff (4)</h3></summary>

1. `console/src/arc/editor/Editor.tsx`, line 11-12 ([link](https://github.com/synnaxlabs/synnax/blob/62aaefa5dbb6313119ad01d96c06f638d9148d13/console/src/arc/editor/Editor.tsx#L11-L12)) 

   <a href="#">P2</a> **Single-letter variable names reduce readability**

   `arc` was renamed to `a`, and the same pattern appears in `Log.tsx` (`log` → `l`) and `Table.tsx` (`table` → `t`). These one-letter names make the code harder to scan — the original descriptive names conveyed the type at a glance.

   

2. `console/src/layout/useSyncName.spec.tsx` 

   <a href="#">P2</a> **Tests deleted with no replacement**

   `useSyncName.spec.tsx` is removed along with the hook it tested, but no tests for the new `NameHook` / `CustomTabName` / `useName` pattern are added. The prior tests covered the two-way sync (remote → layout slice and layout state → remote). These paths are now handled by each renderer's `useName` hook and by `CustomTabName`'s `handleNameChange` / `handleRename`, neither of which has coverage.

3. `console/src/hardware/common/task/Form.tsx`, line 83-85 ([link](https://github.com/synnaxlabs/synnax/blob/62aaefa5dbb6313119ad01d96c06f638d9148d13/console/src/hardware/common/task/Form.tsx#L83-L85)) 

   <a href="#">P2</a> **`retrieve` silently no-ops when `taskKey` is null**

   When the task hasn't been persisted yet (`taskKey == null`), `retrieve()` returns without fetching, so `onChange` is never called and the tab name stays at whatever value the layout slice already has. This is probably intentional, but the asymmetry with other `useName` hooks (which always call `baseRetrieve`) could cause the tab name to be stale if the task key becomes available after the component mounts — the observable won't start until the next retrieve.

4. `console/src/lineplot/LinePlot.tsx`, line 449-453 ([link](https://github.com/synnaxlabs/synnax/blob/5f7249e68b367e0e728ecc6a9feda5e8b5088419/console/src/lineplot/LinePlot.tsx#L449-L453)) 

   <a href="#">P1</a> **`handleTitleChange` drops workspace sync**

   `syncDispatch` was replaced with plain `dispatch`. The `syncDispatch` variant propagates `Layout.rename` across windows (multi-window workspace sync); plain `dispatch` is local-only. When the plot's inline title editor is used to rename the layout the workspace will no longer be notified, causing the name to diverge between open windows.

   Previously `syncDispatch(Layout.rename(...))` handled both local update and workspace propagation in one call. Now only `dispatch` (local) and `renameRemote` (server resource) are called — the workspace sync path is missing for this code path.
</details>

<sub>Reviews (15): Last reviewed commit: ["Merge branch &#39;rc&#39; of https://github.com/..."](https://github.com/synnaxlabs/synnax/commit/c962baddccbf30ed4102f9d6ed74e6ec7f2502e3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29743804)</sub>


console/src/layouts/Mosaic.tsx
```diff
@@ -171,6 +178,53 @@ const ModalContent = ({ node, tabKey }: ModalContentProps): ReactElement => {
...
+    [hookOnRename, dispatch, tabKey],
+  );
+  return (
+    <Tabs.DefaultName
+      tabKey={tabKey}
+      name={name}
+      level={level}
+      editable={editable}
+      onRename={hookOnRename != null ? handleRename : fallbackOnRename}
+    />
```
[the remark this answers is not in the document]
emilbon99: Wrong

console/src/hardware/common/task/Form.tsx
```diff
@@ -178,17 +178,6 @@ export const wrapForm = <S extends task.Schemas = task.Schemas>({
       onChange: (d) => form.set("rackKey", d.data?.rack),
```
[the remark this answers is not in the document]
emilbon99: Not concerned about this

pluto/src/text/Editable.tsx
```diff
@@ -183,7 +184,7 @@ export const Editable = ({
     el.addEventListener(START_EDITING_EVENT_NAME, () => setEditable(true));
   }, []);
 
-  const combinedRef = useCombinedRefs(ref, refCallback);
+  const combinedRef = useCombinedRefs(propsRef, ref, refCallback);
```
pjdotson, line 187: can you add a test for this?

pluto/src/tabs/Selector.tsx
```diff
@@ -311,25 +315,23 @@ export interface SelectorButtonProps extends Spec {
...
-  onRename?: (key: string, name: string) => void;
-  name: string;
-  tabKey: string;
-  editable?: boolean;
-}
+export interface DefaultNameProps
+  extends NameProps, Omit<Text.TextProps, "level" | "onChange"> {}
 
-const Name = ({
+export const DefaultName = ({
```
pjdotson, line 324: Why do we need the ability to pass an entirely new name component here? why can't we get
this functionality just by exposing the `name` and `onRename` props?

pluto/src/tabs/Selector.tsx
```diff
@@ -9,6 +9,7 @@
 
 import { box, type location, scale, xy } from "@synnaxlabs/x";
 import {
+  type ComponentType,
```
pjdotson, line 12: changes in this module should have tests

console/src/table/selectors.ts
```diff
@@ -33,6 +33,14 @@ export const selectOptional = select as (
 export const useSelect = (key: string): State =>
   useMemoSelect((state: StoreState) => select(state, key), [key]);
 
+export const selectIsRemoteCreated = (
+  state: StoreState,
+  key: string,
+): boolean | undefined => selectOptional(state, key)?.remoteCreated;
+
+export const useSelectIsRemoteCreated = (key: string): boolean | undefined =>
```
pjdotson, line 41: add tests

console/src/schematic/selectors.ts
```diff
@@ -195,6 +195,14 @@ export const selectIsSnapshot = (state: StoreState, key: string): boolean | unde
...
   useMemoSelect((state: StoreState) => selectIsSnapshot(state, key), [key]);
 
+export const selectIsRemoteCreated = (
+  state: StoreState,
+  key: string,
+): boolean | undefined => selectOptional(state, key)?.remoteCreated;
+
+export const useSelectIsRemoteCreated = (key: string): boolean | undefined =>
+  useMemoSelect((state: StoreState) => selectIsRemoteCreated(state, key), [key]);
+
```
pjdotson, line 205: add tests

pluto/src/tabs/Selector.tsx
pjdotson: It looks like everywhere this is used, the component just uses `DefaultName` as a final argument anyways.

pjdotson (changes requested):

pluto/src/tabs/Selector.tsx
emilbon99: We need somewhere for the tab specific `useName` hook to live. You have three options:

1. Pass `useName` as a hook argument, which I think is a bad pattern in react.
2. Keep the `Name` component.
3. Take a `renderName` render prop. 

I think no. 2 or no. 3 are the right option. Trying to get the hook lifecycle to live somewhere else is just hacky.

emilbon99: If you have a better solution please lmk.

pjdotson: Right, but the only thing "different" for each tab is just `name` and `onRename`, so why
can't they be passed in? instead of passing a whole component

emilbon99: Because the `useName` hook needs somewhere to live.

emilbon99: One `useName` hook needs to be mounted and called per tab. If you pass just the `name` and `onRename` where do you mount that hook? You can't call them all dynamically in the mosaic. You need a custom tab component and/or render prop scoped per tab so the hook lifecycle is valid.

pjdotson: So the tab provider context values both has the tab names and the `onRename`
functionality in the context value. Now the custom name component needs to make sure it
modifies the context value `name` as well. Why is this hook not just the first line in
the layout renderers, that a) when the name is updated in flux, triggers the rename in
the tabs provider, and b) updates a per-tab custom `onRename` in the context value, so
renaming that tab now renames through Flux?

emilbon99: Because that approach requires nasty effect based mounting and synchronization, and is a contrived way of solving the problem IMO. Let's talk verbally.

PR #2274, opened 2026-04-28
SY-4127: Divide F64 Typecast - Trade and RFC

sy-nico:
# Pull Request

[SY-4127](https://linear.app/synnax/issue/SY-4127/refactor-arc-divide-by-0-to-not-panic-on-a-crash)

<h3>Greptile Summary</h3>

This PR introduces two new RFC documents for SY-4127: a trade study and a formal RFC specifying that `/` and `**` in Arc will always return `f64` (with IEEE 754 divide-by-zero semantics), replacing the current behavior of returning the input type and panicking on integer divide-by-zero. The trade study thoroughly covers type theory, precision, reliability, safety, and language identity; the RFC details the implementation plan across the analyzer, compiler, WASM runtime, flow nodes, and series operations.

Safe to merge — documentation-only PR with no production code changes.

Both files are new RFC/trade-study markdown documents with no code changes. The two findings are P2 documentation clarity issues that do not affect correctness of the spec's conclusions.

docs/tech/rfc/0038-260425-division-power-f64-typecast.md — minor wording clarifications in §3.2 and §4.3.

| Filename | Overview |
|----------|----------|
| docs/tech/rfc/0038-260425-division-power-f64-typecast.md | New RFC specifying that `/` and `**` always return `f64` with IEEE 754 divide-by-zero behavior; two minor documentation clarity issues found: identical code examples in §3.2 and ambiguous "first operator" wording in §4.3. |
| docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md | New trade study evaluating Option A (always f64) vs Option B (preserve input type); comprehensive analysis across type theory, precision, safety, and reliability criteria; no new issues beyond those already addressed in prior review threads. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["Arc Expression: a / b or a ** b"] --> B{Analyzer: InferMultiplicative / InferPower}
    B --> C["Force result type to f64"]
    C --> D{Compiler: binary.go}
    D --> E["Strip caller type hint (§4.3)"]
    E --> F["Emit host function call\nmath.divide(T,T) → f64 / math.pow(f64,f64) → f64"]
    F --> G{WASM Runtime}
    G --> H["Cast operands to float64"]
    H --> I{Division by zero?}
    I -- "n/0 (n≠0)" --> J["f64(+Inf) or f64(-Inf)"]
    I -- "0/0" --> K["f64(NaN)"]
    I -- "normal" --> L["f64 result"]
    J --> M{isSeriesTruthy}
    K --> M
    L --> M
    M -- "+Inf / -Inf" --> N["truthy → control fires"]
    M -- "NaN" --> O["non-truthy → control blocked"]
    M -- "normal f64" --> P["truthy if non-zero"]
```

<sub>Reviews (4): Last reviewed commit: ["SY-4127: Reviewer feedback and RFC updat..."](https://github.com/synnaxlabs/synnax/commit/99368dddcb784f8182ee519fbd84777e1caed1e3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29959343)</sub>

### DIFF
--- a/docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
+++ b/docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
@@ -41,11 +41,18 @@ outputs must match inputs.**
 
 ### Type System Consistency
 
+**What "consistency" means here.** Option A's consistency is that each operator returns
+a type matching **what the operation actually represents**, not that every operator
+returns its input type. `+`, `-`, `*`, `%` stay in the integer domain because they
+produce exact integer results from integer inputs. `/` and `**` do not, so they return
+`f64`. The rule is in the math (like derivative), not in matching the input datatype,
+which ignores what the operators actually do.
+
 **"Divide and pow behave differently from add/subtract/multiply."**
 
 LabVIEW, numpy, and scipy all have separate divide (float) and quotient (integer)
 operations. This is standard. Both `divide` and `pow` can produce non-integer results
-from integer inputs (`1/2 = 0.5`, `2^-1 = 0.5`), unlike add, subtract, and multiply,
+from integer inputs (`1/2 = 0.5`, `2**-1 = 0.5`), unlike add, subtract, and multiply,
 which always produce exact integers from integer inputs.
 
 Arc already has precedent: `math.derivative` returns `f64` regardless of input type
@@ -53,6 +60,10 @@ because rate of change is inherently continuous. Division and power have the sam
 property.
 
 **"A user might expect `i32 / i32 → i32`."**
+
+Arc's users come from Python, LabVIEW, and MATLAB. In all of these environments, `/`
+produces a result in the continuous domain. Customer research confirms this expectation.
+A C-style integer truncation would surprise the majority of Arc's actual users.
 
 Two integer division variants can be added later if requested: `math.floor_division()`
 (round toward negative infinity, Python `//` semantics: `-7 // 2 = -4`) and
@@ -67,13 +78,32 @@ In Python, `int(1) / int(3)` returns `float(0.333...)`. In numpy,
 returns `float64(inf)`. The Synnax Python client's `Series.to_numpy()` returns
 `numpy.ndarray`, so when a user queries two integer series from Synnax and divides them,
 numpy casts to `float64` and produces `+/-Inf`/`NaN` on zero denominators. Option A
-makes Arc consistent with behavior that already exists in our own client library.
+makes Arc consistent with behavior that already exists in our own client library. The
+Python client can already write `+Inf`, `-Inf`, and `NaN` directly to `f64` channels:
+
+```python
+import synnax as sy
+import numpy as np
+
+client = sy.Synnax()
+
+idx = client.channels.retrieve("inf_nan_time")
+ch = client.channels.retrieve("inf_nan_f64")
+
+VALUES = [np.inf, -np.inf, np.nan]
+
+now = sy.TimeStamp.now()
+timestamps = [int(now + i * sy.TimeSpan.SECOND) for i in range(len(VALUES))]
+
+with client.open_writer(start=now, channels=[idx.key, ch.key]) as w:
+    w.write({idx.key: timestamps, ch.key: VALUES})
+```
 
 For power, numpy's `np.power(int, negative_int)` raises `ValueError` entirely rather
 than produce a wrong integer answer. `np.float_power` always returns `float64`. Option A
 uses `float_power` semantics for both `divide` and `pow`.
 
-### Precision
+### Precision Errors
 
 | Scenario       | Option A (`f64` output)  | Option B (integer output)   |
 | -------------- | ------------------------ | --------------------------- |
@@ -81,8 +111,9 @@ uses `float_power` semantics for both `divide` and `pow`.
 | `7 / 3`        | `2.333...`, correct      | `2`, 14% error (truncation) |
 | `(2^53+1) / 1` | loses 1 bit of precision | exact                       |
 
-Option B loses precision on every non-exact division. Option A only loses precision for
-integers exceeding 2^53 (~9 quadrillion), which is outside the range of any real sensor.
+Option B has increased error due to not having the ability to represent precision for a
+solution in the continuous domain. Option A only loses precision for integers exceeding
+2^53 (~9 quadrillion), which is outside the range of any real sensor.
 
 For `f32` inputs, promoting to `f64` is a free precision gain: `f32` has ~7 decimal
 digits of precision, while `f64` has ~15. `f32(1.0) / f32(3.0)` computed in `f64`
@@ -130,17 +161,21 @@ Every mainstream language treats these differently. Python 3: `int / int = float
 `int * int = int`. Rust, Go, C: multiplication wraps, division truncates. Nobody
 auto-promotes multiplication.
 
+Arc integer arithmetic wraps on overflow (two's complement, WASM semantics), the same as
+C and Go.
+
 **Overflow is a width problem. Division is a domain problem. Different problems,
-different solutions.**
+different solutions.** Overflow behavior is out of scope for this RFC and is an
+independent design decision that does not affect the case for Option A.
 
 ### Divide by Zero
 
-| Scenario             | Option A (`f64`)                                                     | Option B (integer)                               |
-| -------------------- | -------------------------------------------------------------------- | ------------------------------------------------ |
-| `n/0`, `-n/0`, `0/0` | `+Inf`, `-Inf`, `NaN`. Deterministic per IEEE 754.                   | Panic (current), or requires new error handling. |
-| Downstream impact    | `+Inf`, `-Inf`, and `NaN` are all non-truthy. Won't trigger control. | No output, possible cascade failure.             |
-| Debugging            | `+Inf`, `-Inf`, `NaN` visible in the Log view, traceable.            | Depends on resolution.                           |
-| Recovery             | System keeps running.                                                | Depends on resolution.                           |
+| Scenario             | Option A (`f64`)                                                                           | Option B (integer)                               |
+| -------------------- | ------------------------------------------------------------------------------------------ | ------------------------------------------------ |
+| `n/0`, `-n/0`, `0/0` | `+Inf`, `-Inf`, `NaN`. Deterministic per IEEE 754.                                         | Panic (current), or requires new error handling. |
+| Downstream impact    | `NaN` is non-truthy (won't trigger control). `+Inf`/`-Inf` are truthy, matching Python/JS. | No output, possible cascade failure.             |
+| Debugging            | `+Inf`, `-Inf`, `NaN` visible in the Log view, traceable.                                  | Depends on resolution.                           |
+| Recovery             | System keeps running.                                                                      | Depends on resolution.                           |
 
 #### What Option B actually requires
 
@@ -192,7 +227,7 @@ What does "reliable" mean for a telemetry language?
 
 The operations that preserve input type (add, subtract, multiply, min, max) do so
 because integer inputs always produce exact integer results. Division and power do not
-have that property: `1 / 2` is `0.5`, `2^-1` is `0.5`. Returning `f64` is the
+have that property: `1 / 2` is `0.5`, `2**-1` is `0.5`. Returning `f64` is the
 mathematically correct choice, consistent with how `math.derivative` already works.
 
 The counterargument is that low-level hardware engineers expect C-style integer
@@ -203,8 +238,10 @@ MATLAB, where division returns floats.
 
 `+Inf`, `-Inf`, and `NaN` are safe in Arc's control model:
 
-- **Non-truthy**: `isSeriesTruthy` treats `+Inf`, `-Inf`, and `NaN` as non-truthy. They
-  will not trigger control logic.
+- **Predictable truthiness**: `isSeriesTruthy` treats `NaN` as non-truthy (it is
+  genuinely undefined, since `0/0` has no magnitude). `+Inf` and `-Inf` are truthy,
+  consistent with Python and JavaScript. Programs that need to guard against
+  divide-by-zero triggering control should check the denominator explicitly.
 - **Visible**: `∞`/`Infinity`, `-∞`/`-Infinity`, and `NaN` already render in the UI.
   They are not silent.
 - **Consistent**: Both flow nodes and WASM functions produce the same values.
@@ -212,29 +249,28 @@ MATLAB, where division returns floats.
 Under Option B, either the system outputs incorrect data (sentinel value of 0), silently
 drops the output (no signal to the user), or behaves differently depending on whether
 the division is in a flow node or a `func` block. All three are worse for safety than a
-visible, non-truthy IEEE 754 value.
+visible IEEE 754 value with well-defined truthiness semantics.
 
 ---
 
 ## Trade Summary
 
-| Criterion            | Option A | Option B                        |
-| -------------------- | -------- | ------------------------------- |
-| Type strictness      | Wins     |                                 |
-| Type consistency     | Wins     |                                 |
-| Precision            | Wins     |                                 |
-| Timestamp precision  |          | Wins (not practically relevant) |
-| Overflow consistency | Wins     |                                 |
-| Divide by zero       | Wins     |                                 |
-| Language identity    | Wins     |                                 |
-| Reliability          | Wins     |                                 |
-| Safety               | Wins     |                                 |
+| Criterion            | Option A | Option B |
+| -------------------- | -------- | -------- |
+| Type strictness      | ✓        | ✓        |
+| Type consistency     | ✓        |          |
+| Precision            | ✓        |          |
+| Overflow consistency | ✓        |          |
+| Divide by zero       | ✓        |          |
+| Language identity    | ✓        |          |
+| Reliability          | ✓        |          |
+| Safety               | ✓        |          |
 
 ## Recommendation
 
 Option A is the recommended option. It wins on every criterion evaluated in this study
-except timestamp precision, where Option B preserves exact nanosecond values. Under
-Option A, the precision loss is ~2-4 nanoseconds on year-long time deltas.
+except the timestamp precision sub-topic, where Option B preserves exact nanosecond
+values. Under Option A, the precision loss is ~2-4 nanoseconds on year-long time deltas.
 
 The intuition behind Option B is reasonable. "The output type should match the input
 type" is a clean mental model, and integer division is familiar to anyone who has
--- a/docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
+++ b/docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
@@ -1,0 +1,252 @@
+# Division and Power Output Type: Trade Study
+
+## Overview
+
+This trade study evaluates two options for the output type of `divide` and `pow`
+operations in Arc, and the behavior of division by zero.
+
+## Options
+
+- **Option A**: `divide` and `pow` always return `f64`. Division by zero produces IEEE
+  754 values (`+Inf`, `-Inf`, `NaN`). This is what LabVIEW, numpy, and scipy do.
+- **Option B**: `divide` and `pow` return the input type (no typecast to `f64`).
+  Division by zero must be handled separately, either by erroring, trapping, or
+  producing a sentinel value. This is what C does.
+
+---
+
+## Trade Criteria
+
+### Type Strictness and Strong Typing
+
+These terms overlap in casual use but mean different things:
+
+- **Strictly typed**: Every value has a definite type known at compile time, and the
+  compiler enforces type rules at every boundary.
+- **Strongly typed**: No implicit coercion between unrelated types. The programmer must
+  explicitly convert (e.g., JavaScript's `"5" + 3 = "53"` is weak typing).
+
+Option A satisfies both. The `/` operator has a **concrete, compile-time return type:
+`f64`**. Always. There is no ambiguity, no runtime coercion, no type that depends on
+context. If you try to assign the result to an `i32` channel, the compiler rejects it
+with a clear error. That is the opposite of "fuzzy."
+
+The concern behind Option B is often: "the output type doesn't match the input type, and
+that feels surprising." That is a fair UX reaction, but it is a preference, not a
+type-system property. LabVIEW is considered strongly typed: every wire has an explicit
+type, and the divide node explicitly returns `DBL` (double).
+
+**"Type-strict" means the compiler enforces types at every boundary. It does not mean
+outputs must match inputs.**
+
+### Type System Consistency
+
+**"Divide and pow behave differently from add/subtract/multiply."**
+
+LabVIEW, numpy, and scipy all have separate divide (float) and quotient (integer)
+operations. This is standard. Both `divide` and `pow` can produce non-integer results
+from integer inputs (`1/2 = 0.5`, `2^-1 = 0.5`), unlike add, subtract, and multiply,
+which always produce exact integers from integer inputs.
+
+Arc already has precedent: `math.derivative` returns `f64` regardless of input type
+because rate of change is inherently continuous. Division and power have the same
+property.
+
+**"A user might expect `i32 / i32 → i32`."**
+
+Two integer division variants can be added later if requested: `math.floor_division()`
+(round toward negative infinity, Python `//` semantics: `-7 // 2 = -4`) and
+`math.quotient()` (truncate toward zero, C semantics: `-7 / 2 = -3`). These differ only
+on negative inputs. YAGNI applies to both. For now, `i32(math.divide(a, b))` works for
+anyone who needs integer output.
+
+**Existing behavior in the Synnax ecosystem:**
+
+In Python, `int(1) / int(3)` returns `float(0.333...)`. In numpy,
+`np.int64(1) / np.int64(3)` returns `float64(0.333...)` and `np.int64(1) / np.int64(0)`
+returns `float64(inf)`. The Synnax Python client's `Series.to_numpy()` returns
+`numpy.ndarray`, so when a user queries two integer series from Synnax and divides them,
+numpy casts to `float64` and produces `+/-Inf`/`NaN` on zero denominators. Option A
+makes Arc consistent with behavior that already exists in our own client library.
+
+For power, numpy's `np.power(int, negative_int)` raises `ValueError` entirely rather
+than produce a wrong integer answer. `np.float_power` always returns `float64`. Option A
+uses `float_power` semantics for both `divide` and `pow`.
+
+### Precision
+
+| Scenario       | Option A (`f64` output)  | Option B (integer output)   |
+| -------------- | ------------------------ | --------------------------- |
+| `1 / 2`        | `0.5`, correct           | `0`, 100% error             |
+| `7 / 3`        | `2.333...`, correct      | `2`, 14% error (truncation) |
+| `(2^53+1) / 1` | loses 1 bit of precision | exact                       |
+
+Option B loses precision on every non-exact division. Option A only loses precision for
+integers exceeding 2^53 (~9 quadrillion), which is outside the range of any real sensor.
+
+For `f32` inputs, promoting to `f64` is a free precision gain: `f32` has ~7 decimal
+digits of precision, while `f64` has ~15. `f32(1.0) / f32(3.0)` computed in `f64`
+produces a more accurate result than computing in `f32`. Option B would return the less
+precise `f32` answer.
+
+#### Timestamps
+
+Synnax timestamps are int64 nanoseconds since epoch (~1.7 x 10^18), which exceeds f64's
+2^53 exact integer range. So `timestamp / n` would lose nanoseconds.
+
+However, dividing a raw timestamp (`now / 3`) is not a meaningful operation. What is
+meaningful is dividing time _deltas_ (`(now - yesterday) / 3`). A one-day delta is ~8.6
+x 10^13 nanoseconds (well within 2^53). Even a one-year delta (~3.15 x 10^16) only loses
+~2-4 nanoseconds of precision in the f64 representation, which is negligible for any
+real sensor application.
+
+More importantly, under integer division, `timedelta / n` where `timedelta < n` returns
+**zero**. That is not a precision loss; that is a total loss. The f64 path gives a
+meaningful sub-nanosecond answer; the integer path gives nothing.
+
+### Overflow and Consistency with Other Operations
+
+**"If division auto-promotes to f64, shouldn't multiplication also auto-promote to
+prevent overflow? Otherwise the type system is inconsistent."**
+
+No. Division and overflow are fundamentally different problems:
+
+- Division and power produce results that **leave the integers entirely**. `1 / 2 = 0.5`
+  is not representable as any integer type. The correct result lives in a different
+  numeric domain.
+- Overflow produces a result that **is still an integer**, just one that doesn't fit in
+  the container. `u8(255) * u8(255) = 65025` is a perfectly valid integer; the problem
+  is the container width, not the numeric domain.
+- For division, there is a single well-defined promotion target: `f64`. For overflow,
+  there is no single right answer: promote to `u16`? `u32`? `u64`? It depends on the
+  values at runtime, which would require arbitrary-precision integers or
+  runtime-dependent types, both of which actually _would_ make the type system fuzzy.
+
+If multiplication promotes to prevent overflow, so must addition and subtraction (they
+also overflow). Then every arithmetic operation returns a wider type, and having
+different integer widths becomes pointless.
+
+Every mainstream language treats these differently. Python 3: `int / int = float`, but
+`int * int = int`. Rust, Go, C: multiplication wraps, division truncates. Nobody
+auto-promotes multiplication.
+
+**Overflow is a width problem. Division is a domain problem. Different problems,
+different solutions.**
+
+### Divide by Zero
+
+| Scenario             | Option A (`f64`)                                                     | Option B (integer)                               |
+| -------------------- | -------------------------------------------------------------------- | ------------------------------------------------ |
+| `n/0`, `-n/0`, `0/0` | `+Inf`, `-Inf`, `NaN`. Deterministic per IEEE 754.                   | Panic (current), or requires new error handling. |
+| Downstream impact    | `+Inf`, `-Inf`, and `NaN` are all non-truthy. Won't trigger control. | No output, possible cascade failure.             |
+| Debugging            | `+Inf`, `-Inf`, `NaN` visible in the Log view, traceable.            | Depends on resolution.                           |
+| Recovery             | System keeps running.                                                | Depends on resolution.                           |
+
+#### What Option B actually requires
+
+The current behavior before this change is a panic that crashes the flow scheduler. That
+is a bug, not a design choice. If we pursued Option B properly, we would need to resolve
+it. The options are different depending on context:
+
+**Flow nodes** (e.g. `divide{}` in flow statements):
+
+1. **Don't output** (skip the tick). Downstream nodes never fire. The user gets no
+   signal that something went wrong. Silent failure.
+2. **Report an error**. `node.Context` has a `report_error` callback, but it is for
+   logging, not stopping execution. We would need to define what "error" means: stop the
+   whole program? Just this node? Just this tick?
+3. **Output a sentinel value** (e.g. 0). This is lying about the result, which is the
+   original problem.
+
+**WASM functions** (division inside `func` blocks):
+
+WASM already traps on integer divide-by-zero (controlled termination). But division
+backed by host functions returns values, not errors. We would need an error ABI (return
+a tuple, or set a global error flag) to surface the failure.
+
+**The inconsistency problem**: under Option B, a `divide{}` flow node behaves
+differently from division in a `func` block. The user writes the same operation in two
+contexts and gets two different failure modes.
+
+Option A avoids all of this. Both paths produce the same IEEE 754 result. No error
+handling infrastructure needed. No inconsistency between flow and WASM.
+
+### Language Identity
+
+Arc's actual users convert Python scripts and LabVIEW diagrams. The C user is
+hypothetical.
+
+| Context                            | Option A                                               | Option B  |
+| ---------------------------------- | ------------------------------------------------------ | --------- |
+| Arc users converting Python        | Matches Python `/` behavior                            |           |
+| LabVIEW precedent (decades, FPGAs) | Exact same behavior (divide returns DBL)               |           |
+| C behavior expectation             | Addressed by future `//` operator if needed            | Matches C |
+| FPGA deployment concern            | LabVIEW proves float divide works predictably on FPGAs |           |
+
+Designing for a hypothetical C user at the expense of current Python and LabVIEW users
+violates YAGNI.
+
+### Reliability
+
+What does "reliable" mean for a telemetry language?
+
+The operations that preserve input type (add, subtract, multiply, min, max) do so
+because integer inputs always produce exact integer results. Division and power do not
+have that property: `1 / 2` is `0.5`, `2^-1` is `0.5`. Returning `f64` is the
+mathematically correct choice, consistent with how `math.derivative` already works.
+
+The counterargument is that low-level hardware engineers expect C-style integer
+division. Every customer we have spoken to disagrees. They come from Python, LabVIEW, or
+MATLAB, where division returns floats.
+
+### Safety
+
+`+Inf`, `-Inf`, and `NaN` are safe in Arc's control model:
+
+- **Non-truthy**: `isSeriesTruthy` treats `+Inf`, `-Inf`, and `NaN` as non-truthy. They
+  will not trigger control logic.
+- **Visible**: `∞`/`Infinity`, `-∞`/`-Infinity`, and `NaN` already render in the UI.
+  They are not silent.
+- **Consistent**: Both flow nodes and WASM functions produce the same values.
+
+Under Option B, either the system outputs incorrect data (sentinel value of 0), silently
+drops the output (no signal to the user), or behaves differently depending on whether
+the division is in a flow node or a `func` block. All three are worse for safety than a
+visible, non-truthy IEEE 754 value.
+
+---
+
+## Trade Summary
+
+| Criterion            | Option A | Option B                        |
+| -------------------- | -------- | ------------------------------- |
+| Type strictness      | Wins     |                                 |
+| Type consistency     | Wins     |                                 |
+| Precision            | Wins     |                                 |
+| Timestamp precision  |          | Wins (not practically relevant) |
+| Overflow consistency | Wins     |                                 |
+| Divide by zero       | Wins     |                                 |
+| Language identity    | Wins     |                                 |
+| Reliability          | Wins     |                                 |
+| Safety               | Wins     |                                 |
+
+## Recommendation
+
+Option A is the recommended option. It wins on every criterion evaluated in this study
+except timestamp precision, where Option B preserves exact nanosecond values. Under
+Option A, the precision loss is ~2-4 nanoseconds on year-long time deltas.
+
+The intuition behind Option B is reasonable. "The output type should match the input
+type" is a clean mental model, and integer division is familiar to anyone who has
+written C. But that intuition breaks down when applied to operations that are
+mathematically continuous: `1 / 2` is `0.5`, not `0`. Every language that our users
+actually come from (Python, LabVIEW, MATLAB, numpy) resolved this the same way. Option B
+is not more type-safe or strict than Option A. The `/` operator under Option A has a
+single, concrete, compile-time return type (`f64`). That is the definition of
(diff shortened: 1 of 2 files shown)


docs/tech/rfc/0038-260425-division-power-f64-typecast.md
```diff
@@ -0,0 +1,251 @@
...
+
+**Consistency.** Both execution contexts (flow nodes and WASM `func` blocks) should
+produce identical results for the same operation. Under the current behavior, a
+`divide{}` flow node and division inside a `func` block have different failure modes on
+divide-by-zero.
+
+# 2 - Non-Goals
+
+- **Integer division operators.** `//`, `math.quotient()`, and `math.floor_division()`
+  are future work if requested. This RFC only changes `/` and `^`.
```
pjdotson: we may want to consider editing `^` to `**` in the future, as that is also a symbol for
bitwise XOR (which we don't use right now, but may end up wanting to support in the future)

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,252 @@
...
+real sensor application.
+
+More importantly, under integer division, `timedelta / n` where `timedelta < n` returns
+**zero**. That is not a precision loss; that is a total loss. The f64 path gives a
+meaningful sub-nanosecond answer; the integer path gives nothing.
+
+### Overflow and Consistency with Other Operations
+
+**"If division auto-promotes to f64, shouldn't multiplication also auto-promote to
+prevent overflow? Otherwise the type system is inconsistent."**
```
pjdotson, line 173: there isn't any discussion anywhere about integer overflows and how this fits into that piece

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,252 @@
...
+is a bug, not a design choice. If we pursued Option B properly, we would need to resolve
+it. The options are different depending on context:
+
+**Flow nodes** (e.g. `divide{}` in flow statements):
+
+1. **Don't output** (skip the tick). Downstream nodes never fire. The user gets no
+   signal that something went wrong. Silent failure.
+2. **Report an error**. `node.Context` has a `report_error` callback, but it is for
+   logging, not stopping execution. We would need to define what "error" means: stop the
+   whole program? Just this node? Just this tick?
```
pjdotson, line 224: so will we never allow flow nodes to return an error or conditionally do stuff based on errors?

docs/tech/rfc/0038-260425-division-power-f64-typecast.md
```diff
@@ -0,0 +1,251 @@
...
+- **Series operation tests**: `telem/op` generated operation tests
+- **C++ runtime tests**: `math.h` host function tests
+- **Integration tests**: `stl_math.py` end-to-end arc program tests
+
+New test scenarios required by this change:
+
+- **Division by zero**: `n/0` produces `+Inf`, `-n/0` produces `-Inf`, `0/0` produces
+  `NaN` (instead of panicking)
+- **`isSeriesTruthy` on special values**: `+Inf`, `-Inf`, and `NaN` all return
+  non-truthy
```
pjdotson: `+Inf` and `-Inf` should evaluate to truthy

docs/tech/rfc/0038-260425-division-power-f64-typecast.md
```diff
@@ -0,0 +1,251 @@
...
+- **Integer division output type**: `i32 / i32` returns `f64`, `i64 / i64` returns `f64`
+- **Series integer division output type**: `series<i64> / scalar` returns `series<f64>`
+- **Chained expressions**: `a / b * c` produces `f64` throughout when inputs are integer
+- **f32 precision gain**: `f32 / f32` computed in `f64` produces a more precise result
```
pjdotson: the result will be the same precision, as the inputs, you can't magically create
precision, you can only lose it.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,252 @@
...
+than produce a wrong integer answer. `np.float_power` always returns `float64`. Option A
+uses `float_power` semantics for both `divide` and `pow`.
+
+### Precision
+
+| Scenario       | Option A (`f64` output)  | Option B (integer output)   |
+| -------------- | ------------------------ | --------------------------- |
+| `1 / 2`        | `0.5`, correct           | `0`, 100% error             |
+| `7 / 3`        | `2.333...`, correct      | `2`, 14% error (truncation) |
+| `(2^53+1) / 1` | loses 1 bit of precision | exact                       |
```
pjdotson, line 144: this discussion of precision doesn't make sense - if we define `"/"` as integer
division, the `7 / 3` equals `2` exactly and there is no error.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,252 @@
...
+
+Option B loses precision on every non-exact division. Option A only loses precision for
+integers exceeding 2^53 (~9 quadrillion), which is outside the range of any real sensor.
+
+For `f32` inputs, promoting to `f64` is a free precision gain: `f32` has ~7 decimal
+digits of precision, while `f64` has ~15. `f32(1.0) / f32(3.0)` computed in `f64`
```
pjdotson, line 151: this isn't a gain - sure we have more decimal places, but the last 8 aren't meaningful
at all. We just changed the type to a different one.

docs/tech/rfc/0038-260425-division-power-f64-typecast.md
```diff
@@ -0,0 +1,251 @@
...
+- **WASM integration tests**: End-to-end execution through wazero
+- **Series operation tests**: `telem/op` generated operation tests
+- **C++ runtime tests**: `math.h` host function tests
+- **Integration tests**: `stl_math.py` end-to-end arc program tests
+
+New test scenarios required by this change:
+
+- **Division by zero**: `n/0` produces `+Inf`, `-n/0` produces `-Inf`, `0/0` produces
+  `NaN` (instead of panicking)
+- **`isSeriesTruthy` on special values**: `+Inf`, `-Inf`, and `NaN` all return
```
pjdotson: should we have constants for +-Inf (and support for +-0) and `NaN` that users can write?

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,252 @@
...
+### Type System Consistency
+
+**"Divide and pow behave differently from add/subtract/multiply."**
+
+LabVIEW, numpy, and scipy all have separate divide (float) and quotient (integer)
+operations. This is standard. Both `divide` and `pow` can produce non-integer results
+from integer inputs (`1/2 = 0.5`, `2^-1 = 0.5`), unlike add, subtract, and multiply,
+which always produce exact integers from integer inputs.
+
+Arc already has precedent: `math.derivative` returns `f64` regardless of input type
```
pjdotson: might be worth noting that `math.derivative` can "feel" different because this is
written as an explicit function call where you can hover over derivative and see the
docs on what type it returns, versus these other functions which have built-in symbols
in the language.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
pjdotson: So the only actual precision loss happens when we have large integers. What I think you
should be getting at is "user expectation": what does a user expect `1 /2` and `1 / 2.0`
and `1.0/2.0` and `1.0/2` to return? should be a separate trade

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,252 @@
...
+prevent overflow? Otherwise the type system is inconsistent."**
+
+No. Division and overflow are fundamentally different problems:
+
+- Division and power produce results that **leave the integers entirely**. `1 / 2 = 0.5`
+  is not representable as any integer type. The correct result lives in a different
+  numeric domain.
+- Overflow produces a result that **is still an integer**, just one that doesn't fit in
+  the container. `u8(255) * u8(255) = 65025` is a perfectly valid integer; the problem
+  is the container width, not the numeric domain.
```
pjdotson, line 182: this argument really doesn't make sense to me. Division only leaves the integers if we
define division as leaving the integers, which is the whole point of this discussion.
One big drawback that isn't mentioned in option A explicitly is that we lose the
consistency on the point "do primitive math operations change the type of the inputs" -
currently they all keep the input types the same. I also think you should try to put
together a consistent story here of why integer overflow is okay but integer division is
not, why it makes sense to have specifically sized types like `i8`, `u32` instead of
just general `number` or `int`,`float` or `int`,`uint`,`float` types

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,252 @@
...
+
+| Context                            | Option A                                               | Option B  |
+| ---------------------------------- | ------------------------------------------------------ | --------- |
+| Arc users converting Python        | Matches Python `/` behavior                            |           |
+| LabVIEW precedent (decades, FPGAs) | Exact same behavior (divide returns DBL)               |           |
+| C behavior expectation             | Addressed by future `//` operator if needed            | Matches C |
+| FPGA deployment concern            | LabVIEW proves float divide works predictably on FPGAs |           |
+
+Designing for a hypothetical C user at the expense of current Python and LabVIEW users
+violates YAGNI.
```
pjdotson, line 285: the point of this is that we try to think ahead to these hypothetical users so that we
don't make breaking changes in the Arc language design later to move the language
towards something that makes sense for them. I don't think YAGNI makes that much sense here

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,252 @@
...
+| -------------------- | -------- | ------------------------------- |
+| Type strictness      | Wins     |                                 |
+| Type consistency     | Wins     |                                 |
+| Precision            | Wins     |                                 |
+| Timestamp precision  |          | Wins (not practically relevant) |
+| Overflow consistency | Wins     |                                 |
+| Divide by zero       | Wins     |                                 |
+| Language identity    | Wins     |                                 |
+| Reliability          | Wins     |                                 |
+| Safety               | Wins     |                                 |
```
pjdotson: you might want to rethink this table. A and B are the same on strictness so I don't get
why that is marked as a win for A. I also don't get why A is a win for type consistency
or precision (see comment in precision section). Why is overflow consistency marked as a
win for A either?

pjdotson (commented):
I left comments here, but we should be mostly having discussions about this in-person, I just wanted to put extra things to think about. I also think that the error / exception handling is something that needs to be discussed as well: what does returning errors for functions look like in both the flow and func contexts (this is something that also is encountered with ranges + statuses RFC and it really does not feel like we have a consistent story on that, and instead are just doing things slapshot as they arise).

docs/tech/rfc/0038-260425-division-power-f64-typecast.md
sy-nico: Why? It's a divide by `0` operation.

Not arguing for or against. I think I'm actually in the middle on this decision. My gut was "any divide by 0 op should be invalid". `+/- Inf` truthiness means "only a numerator of 0 is non-truthy"

sy-nico: Actually, just looked up how other languages handle truthiness and `+/- Inf` are truthy for Python, JavaScript, Ruby. Updating Doc

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: There is nothing inherently wrong with returning an error on a node. The main issue with not typecasting is having different behaviors across contexts. Even if both flow nodes and WASM func blocks throw errors, what's the mechanism? Do they behave the same? Does one block execution while the other skips a tick? Do we need a new error ABI for WASM? Designing consistent error semantics across both execution contexts gets complicated fast.

Deterministic float semantics sidestep all of this. The result is the same in both contexts, no new infrastructure is needed, and IEEE 754 gives engineers a shared, well-understood language for reasoning about the special values.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: The whole section is a discussion on overflow. What do you want to see added to the discussion? What are your concerns?

If the concern is f64 precision loss on large integers when casting, that is covered in the Precision section (`(2^53+1) / 1` loses 1 bit). If the concern is what Arc does on integer overflow for add/subtract/multiply, I've added a sentence clarifying that Arc wraps (two's complement, WASM semantics), unchanged by this RFC. Let me know if there's something else specific you're looking for.

docs/tech/rfc/0038-260425-division-power-f64-typecast.md
sy-nico: My wording could be better, but that's not quite the point I'm trying to make, which is that performing the division in `f64` reduces rounding error in the operation itself. The result is closer to the true mathematical value than if the division were computed in `f32`. This really applies to all non-64bit data types, so I'll make sure to update the docs.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: This section should more aptly be named "Precision Errors"

The division symbol `/` is not defined as integer division. It is defined as mathematical division. Section title has been updated. 

The user expects a solution in the "continuous domain" (validated by customer research). If you still disagree with this then I can reach out to more external folks over slack. If you have heard otherwise, let me know by whom.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: Agreed that the extra bits don't recover information that wasn't in the original inputs. The point is that the division operation itself introduces less rounding error when computed in `f64`. 

Whether those extra bits are meaningful depends on the use case, but having them is strictly better than not having them. The alternative is performing the division in the input type and introducing avoidable rounding error on top of what the inputs already carry. 

Framed correctly: computing in `f64` is a gain (or no gain if you still don't agree). Not computing in `f64` is an objective loss.

docs/tech/rfc/0038-260425-division-power-f64-typecast.md
sy-nico: Yes... I did not consider this.

The python already allows you to write `+/- Inf` & `NaN` to a channel since it is already using numpy under-the-hood. I tested this but don't remember the semantics. Updating RFC.

sy-nico:
```python

import synnax as sy
import numpy as np

client = sy.Synnax()

idx = client.channels.retrieve("inf_nan_time")
ch = client.channels.retrieve("inf_nan_f64")

VALUES = [np.inf, -np.inf, np.nan]

now = sy.TimeStamp.now()
timestamps = [int(now + i * sy.TimeSpan.SECOND) for i in range(len(VALUES))]

with client.open_writer(start=now, channels=[idx.key, ch.key]) as w:
    w.write({idx.key: timestamps, ch.key: VALUES})

```

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: Agreed this is worth surfacing. I had a gap in the RFC, so I updated the LSP section.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: We've covered these points, but consolidating here (and also updating docs to be more explicit)

1. `We lose consistency on 'primitive math ops don't change input type'.`
 
I'd frame it the opposite way. Option A increases consistency with what each operation actually represents: `+`, `-`, `*`, `%` stay in the integer domain because they produce exact integer results from integer inputs. `/` and `^` don't, so they return `f64`. The consistency is in the math, not in a uniform "input type in, input type out" rule that doesn't reflect what these operators do.


2. `Why integer overflow is okay but integer division is not.`

Integer overflow isn't "okay", it's out of scope. Overflow is a container-width problem (`u8(255) * u8(255)` is a valid integer, just doesn't fit in 8 bits). Divide-by-zero and integer truncation are numeric-domain problems. Different category, different RFC.

3. `Why it makes sense to have specifically sized types like `i8`, `u32` instead of general `number` or `int`, `float`.` 

Also out of scope. Arc targets hardware telemetry, where channel types match the underlying sensor/protocol width (a 12-bit ADC stored as `u16`, a tachometer count as `u32`, an `i16` thermocouple reading, etc.). Sized types in Arc exist so a channel and its consuming flow share the same width without coercion. That is a separate language design decision, not something this RFC touches

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: The trade study already covers this. `math.quotient()` and `math.floor_division()` can be added later without changing `/` . The argument about catering to C developers is really irrelevant to the big picture. 

We aren't developing ourselves into a corner and we can easily implement future operations when needed.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: - Strictness: I'll mark it as a tie. The section really exists to address the earlier concern that A was not strict, not to argue A is stricter than B.
- Consistency: A still wins. Each operator returns a type that matches what the math produces (continuous results -> `f64`, integer results -> input type). B applies a uniform input-shape rule that ignores what the operators actually do.
- Overflow consistency: A applies one rule across domain and width problems. B doesn't actually handle the domain side of `/`,  which introduces more footguns.

docs/tech/rfc/0038-260425-division-power-f64-typecast.md
sy-nico: I will make this change here. It's the opportune time given that `math.pow()` hasn't been released into the public yet.

sy-nico:
@pjdotson , replying to [your top-level comment](https://github.com/synnaxlabs/synnax/pull/2274#pullrequestreview-4191256767).

- Agreed, but it was time to put the concerns on paper since the discussion was starting to go in circles. I needed your thoughts/concerns/objections in a way that I could address point by point. Something that gets hard to do in-person the longer the conversation drags on.
- Also agreed that error/exception handling needs to be discussed, but not here. Typecasting division to `f64` cleanly sidesteps this conversation to the contexts where errors/statuses actually make sense (metadata CRUD, etc.), rather than math operations. 
- If we went with Option B, error handling would absolutely need to be part of this RFC. That's another reason Option A is the cleaner choice: it removes the error-handling question from the math layer entirely and lets us tackle it in the contexts where it actually belongs, rather than shoe-horn error handling and create more avenues for inconsistency, which backs us into a corner.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,288 @@
...
+## Trade Criteria
+
+### Type Strictness and Strong Typing
+
+These terms overlap in casual use but mean different things:
+
+- **Strictly typed**: Every value has a definite type known at compile time, and the
+  compiler enforces type rules at every boundary.
+- **Strongly typed**: No implicit coercion between unrelated types. The programmer must
+  explicitly convert (e.g., JavaScript's `"5" + 3 = "53"` is weak typing).
```
pjdotson: I don't see where you are getting these definitions of strongly vs strictly. From what
I've read before about typing in languages, there is a general strong typing vs weak
typing axis, along with static typing vs dynamic typing and nominal vs structual. But I
haven't heard and can't really seem to find a difference between strict and strong

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
pjdotson: That isn't true. The division symbol is defined as whatever the language defines the
division symbol as. We are discussing what the division symbol should be defined as. My
point is that in a lot of these categories you say option A wins for the same exact
point (users would initially expect `/` to be mathematical division). I don't disagree
with that point, but you should just put that in one section by itself instead of
mangling with the definition of "precision"

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,288 @@
...
+outputs must match inputs.**
+
+### Type System Consistency
+
+**What "consistency" means here.** Option A's consistency is that each operator returns
+a type matching **what the operation actually represents**, not that every operator
+returns its input type. `+`, `-`, `*`, `%` stay in the integer domain because they
+produce exact integer results from integer inputs. `/` and `**` do not, so they return
+`f64`. The rule is in the math (like derivative), not in matching the input datatype,
+which ignores what the operators actually do.
```
pjdotson: See my comment below, but it feels like that the sections: type system consistency,
precision errors, and consistency with other operations are mostly reiterating the same
point. There should be one section called user expectation that goes for option A
because most current users expect it to return a float, and another section for
mathematical operator type consistency that goes with option B because when we perform
these primitive math operators (`/`, `%`, `+`, etc.) that are binary operators on two
objects of the same type, the output is the same type as the inputs.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,288 @@
...
+What does "reliable" mean for a telemetry language?
+
+The operations that preserve input type (add, subtract, multiply, min, max) do so
+because integer inputs always produce exact integer results. Division and power do not
+have that property: `1 / 2` is `0.5`, `2**-1` is `0.5`. Returning `f64` is the
+mathematically correct choice, consistent with how `math.derivative` already works.
+
+The counterargument is that low-level hardware engineers expect C-style integer
+division. Every customer we have spoken to disagrees. They come from Python, LabVIEW, or
+MATLAB, where division returns floats.
```
pjdotson: seems to be the same argument as some of the above sections as well

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
pjdotson: so if you do want a section called precision it should go to option B because right now
you can't precisely divide very large integers (really it should get merged in with the
following timestamps section)

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,288 @@
...
+| Criterion            | Option A | Option B |
+| -------------------- | -------- | -------- |
+| Type strictness      | ✓        | ✓        |
+| Type consistency     | ✓        |          |
+| Precision            | ✓        |          |
+| Overflow consistency | ✓        |          |
+| Divide by zero       | ✓        |          |
+| Language identity    | ✓        |          |
+| Reliability          | ✓        |          |
+| Safety               | ✓        |          |
```
pjdotson: should add a section explicitly stating that error / exception behavior would need to be
added here as well for option B.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: Moved content to the `Reliability and User Expectations` section.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: Consolidated the user expectation thread into `Reliability and User Expectations` and trimmed overlap across type consistency, precision, and overflow.

Option B doesn't win on type consistency. "Binary op on `(T, T)` returns `T`" is a C convention, not a type system invariant. Division isn't closed over integers (`1/2 = 0.5` has no integer representation), so the rule discards the correct result for `/`. Stronger type systems than C reject it: Python (`int/int -> float`), Haskell typeclass numerics, LabVIEW (`divide -> DBL`).

Also, option A's `(T, T) -> f64` doesn't break the rule. The operator promotes both operands to `f64` and operates on two values of the same type, like Python's `int/int -> float`. Promotion is part of the signature, not implicit coercion.

`Type System Consistency` captures this: output matches what the operation represents (continuous for `/` and `**`, integer for `+`,`-`, `*`, `%`), not that every operator returns its input type.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: "Option B wins precision because `7/3 = 2` exactly" is circular: it scores precision using the very definition the trade is deciding. Under the mathematical semantic the RFC adopts for `/`, `2` is a 14% error, not zero.
Option B's only genuine precision win is integers above ~9 quadrillion (outside any sensor range); `f32` inputs flip it since Option A computes in `f64`'s ~15 digits vs `f32`'s ~7, and the logic extends to any input narrower than `f64`.

User expectation now lives in its own section, `Reliability and User Expectations`, per your earlier comment. Added a sentence at the top of `Precision Errors` naming the framing explicitly so the criterion isn't ambiguous.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: This is already covered in `Divide by Zero` under `What Option B actually requires`. That subsection walks through what each context needs: flow nodes pick between skipping the tick, a new error ABI, or a sentinel value, and WASM funcs need an error ABI for host-function division (WASM already traps on integer divide by zero).

The whole value of Option A on this front is that none of that infrastructure is required. IEEE 754 (`+Inf`, `-Inf`, `NaN`) gives both contexts the same deterministic result. The subsection's last line: "Option A avoids all of this. Both paths produce the same IEEE 754 result. No error handling infrastructure needed."

I added an `Error handling required` row to the Trade Summary if a glance-level callout helps, but I don't think a separate section is needed since the content already lives under `Divide by Zero`.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: Strict vs strong isn't a canonical axis. I updated the section to use static vs strong, which match the standard programming-language type axes you mentioned (static/dynamic, strong/weak). Sources are also cited in `Static and Strong Typing` section.

Substantive claim is unchanged: under Option A, `/` is statically typed (concrete compile-time return type `f64`) and strongly typed (no implicit coercion; explicit cast required to assign to non-`f64`).

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: Actually, I am going to write a whole new section for the Error handling so that the content for this topic isn't scattered.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: I added more to this section and included a different criteria for `Same-type rule uniformity`, which is the topic that I think you were trying to get at and you clarified over a call.

docs/tech/rfc/0038-260425-division-power-f64-typecast.md
emilbon99: The `^` operator has been released to the production, so changing `^` for `**` is a breaking change. We agreed during the early design of arc not to support bitwise operations. Has that sentiment changed? This is also out of scope for this RFC.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
```diff
@@ -0,0 +1,448 @@
...
+There is no silent bridging between unrelated types.
+
+The internal promotion of integer operands to `f64` inside the divide operation is part
+of the operator's defined signature `(T, T) -> f64`, not implicit coercion across the
+type system. Every strongly typed language with numeric operators does this: Python
+(`int / int -> float`), Haskell (typeclass-driven numerics), and LabVIEW (divide node
+returns `DBL` from integer inputs) are all considered strongly typed.
+
+The intuition that "outputs should match inputs" is a reasonable UX preference, but it
+is not what static or strong typing means. LabVIEW is the closest precedent in the
```
emilbon99, line 64: How did you come to the conclusion that LabVIEW is the closest precedent? What about structured text? Structured text does not coerce `INT / INT` division, and it is arguably more similar to arc than LabVIEW is.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
emilbon99: I agree with @pjdotson here, there isn't an explanation of what should happen when I do `uint8(255) + uint8(255)`. You say 'closed over the integer domain', but that's the crux of what you're missing. Sure if we only had an `int` type like python, then `+` is really closed over the integer domain. 

But if you have `uint8` the `uint8` domain is `[0, 255]`, this means that that `+` is NOT closed on that domain `uint8(255) + uint8(255)` is NOT closed on the `uint8` domain. as `255 + 255` is `510` which is clearly not contained by `[0,255`.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
emilbon99: #2 and #3 may be out of scope for the implementation of this RFC, but I strongly feel that they are _essential_ to have a proposal for if we choose to implement type casting on division.

You say that overflow is a container width problem, but in many ways so is division by zero. The fundamental reason why we are doing a cast from integer to float on division is because we _cannot_ represent `NaN` using a standard integer type in computing i.e. the integer _container_ does _not_ contain a value in it's domain that can be used to represent `NaN`. If the domain of `uint8` per say was not [0, 255) but it also contained `NaN` then we wouldn't have the problem here.

You're arguing we need float division because the integer container is not wide enough to represent the output domain that we need for safety (i.e. the integer domain does not contain `NaN`).

Integer overflow is a fundamentally similar problem i.e. without coercing the output type the output domain does not contain sufficient values to represent `uint8(255) + uint8(255)` in a safe manner.

There NEEDS to be a plan (not an implementation, a plan) for how we are going to handle integer overflows AND whether we need sized types before I am comfortable merging a floating point auto-coerce implementation.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
emilbon99: The fundamental problem here is that parts of our language have already catered to `C` developers. Notably, the fact that every type is `sized` isn't something that test engineers need to know about or give a shit about. 

We've already created a language that blends a shitty line between catering to test engineers and `C` developers.

Again, before I'm comfortable merging an implementation that coerces on floating point division, we need to have plans and make hard decisions about how the wider type system should work.

emilbon99 (changes requested):
If we really decide we want to auto-cast on division, I will concede. What I will not concede is making that decision _without_ considering what the long term direction of the larger type system will look like.

We are on a train to a language that is a shitty hybrid between one optimized for software engineers and test engineers, and we're making micro decisions about the language without considering the macro design.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: LabVIEW is the closest precedent because the domain and user base match. Synnax is a data acquisition and automation platform; LabVIEW is a data acquisition and automation language. Our selling point is that our customers can migrate from LabVIEW workflows to Synnax. That is the concrete user story driving the product.

Structured Text is a PLC programming language. Its domain is discrete control logic on programmable controllers, not data acquisition, streaming telemetry, or even triggering statuses and range creation. The overlap with Arc is that both are industrial; that is where it ends.

If the counter is that our users do not write LabVIEW themselves but use systems built in it, that cuts ST and C equally. Our users do not write Structured Text or C either. The question is which language's divide behavior matches the mental model our actual users carry, and that is LabVIEW. Citing ST/C while disputing LabVIEW is choosing which precedent counts, not analyzing them.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: Sized types are not a concession to C developers, they are how telemetry languages model hardware. Sensor ADCs are i16 or i32, IMUs produce i32 or f32, IEEE 754 single and double precision are f32 and f64. Synnax channels are typed in those widths because the data on the wire is. LabVIEW exposes I8, I16, I32, I64, U8, U16, U32, U64, SGL, and DBL. Test engineers absolutely do work with sized types: they pick channel data types, choose storage formats, and read datasheets that specify bit widths. Saying they don't "give a shit" about sized types contradicts what the LabVIEW user base actually does every day.

The "C developer vs. test engineer" framing is also a false binary. NumPy users explicitly select np.int32, np.int64, np.float32, np.float64. MATLAB users work with int8, uint16, single, double. Researchers, scientists, and students in those ecosystems work with sized types daily and are not C developers. Sized types are the standard interface to numeric data across scientific and engineering computing, not a tell that a language is catering to systems programmers.

The framing that the existing language has to be resolved before this RFC can land is the same process gate pointed out in another comment. Sized types shipped without this macro-design conversation. If they need revisiting, that is a separate conversation, and it is not a precondition for fixing a scheduler-crashing panic.

If there is a specific commitment about the wider type system that would change the answer here, name it. Otherwise this is the same one-sided standard: holding one in-flight change to a bar the rest of the language was never held to.

sy-nico: A few: issues with "parts of our language have already catered to C developers":

- Strong typing is not catering to C. C has weak typing in multiple dimensions (implicit numeric promotion, void* casts). The languages most associated with strong typing (Haskell, Rust, OCaml) are not C-developer-oriented. "The language has sized types" is not evidence of "the language caters to C."

- If the question is who the language should cater to, it is Synnax users. They come from Python, MATLAB, LabVIEW, and the Synnax client libraries. None of those are C. The framing does not match who is actually using the platform.

- Making incremental progress toward one identity, alongside an urgent fix, does not invalidate everything that has been built. Re-litigating prior decisions as a precondition for an urgent fix is process used to defer a known bug indefinitely.

- Even granting your premise: those decisions were yours, made without the macro consideration now being demanded of this RFC. And it is not as though I have never raised objections about the language drifting toward catering to C users.

The language is young, and we have barely reached a crossroads. Many of the fixes in flight are for stability, not identity. We are not so deeply committed to a C-developer identity that a narrow fix is going to turn the language on its head. It is not too late to keep aligning the language with how our users actually think and write code, and this RFC is part of doing that.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: Gating this merge on a plan for overflow and sized types is indefinite deferral by another name. The current behavior, a scheduler-crashing panic on integer divide-by-zero, was not held to that bar.

sy-nico: `You're arguing we need float division because the integer container is not wide enough to represent the output domain that we need for safety (i.e. the integer domain does not contain NaN).`

No. The argument is consistency, not crashing, and industry standards. The other benefits in the RFC are direct rebuttals to concerns you raised (not seeing the big picture). Reframing those rebuttals as my premise is a bad faith argument.

`because we cannot represent NaN using a standard integer type in computing`

Not the point. `f64` inherits IEEE 754, so we do not have to design for `NaN` ourselves. `math.derivative` already auto-casts to `f64`. Its input is an integer divided by an integer timestamp differential. That call was made unilaterally without a plan for sized types or overflow. The principle in this RFC is the one already shipped in the language and Python client.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: `+` IS closed on `uint8` in Arc. `uint8(255) + uint8(255)` wraps to `uint8(254)` (WASM two's complement, §Overflow). The result is a `uint8`.

`1 / 2 = 0.5` cannot wrap to any integer width. `0.5` is not an integer. That is the asymmetry.

NumPy has fixed-width integer types, wraps on overflow, and promotes `np.int32 / np.int32` to `np.float64`. Sized types and division-cast already coexist.

sy-nico:
@emilbon99 , Relying to [this comment](https://github.com/synnaxlabs/synnax/pull/2274#pullrequestreview-4224158205).

The premise that this decision is being made without considering the long term is not accurate. There are two other Arc RFCs open right now doing exactly that work, in writing, in public, with the team. These are the macro design conversations and they are happening.

The divide by 0 bug shipped without anyone gating it on the macro design. Picking up the pieces of those decisions and being told the fix has to clear a bar the current implementation across all of the libraries and APIs was not held to is an inconsistent standard.

This RFC was not written in a vacuum. The trade study weighs multiple languages and libraries (Python, NumPy, LabVIEW, C), Arc's existing patterns (`math.derivative` as the precedent for continuous-domain promotion), constraints the language already imposes as you shipped it, and behaviors of our already-released client libraries (the Python client already type casts with division and natively works with `Inf`/`NaN`). I am working with what we have. If you do not like what you wrote, one RFC cannot retroactively address every concern across the codebase.

If there is a specific type-system direction that would change the answer here, name it. Until then, "consider the macro" is a process veto. This RFC is not going to expand to a +20k line change resolving floor-division, integer promotion, and error ABIs. None of those were previously considered. None of those are blockers for a narrow, sound change.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
pjdotson: This doesn't make sense to me. You are saying that (`u8`) 255 + 255 = 254 is okay, but
(`u8`) 1 / 2 = 0 is problematic? Mathematical addition is not closed on `u8` because (without overflow
semantics) 255 + 255 = 510 which is out of bounds of `u8`. Mathematical division is also not closed
because 1 / 2 = 0.5 which isn't representable as a `u8`. So why is one okay but the
other not? If doing integer division is bad, why is doing overflow semantics okay: they are both deviations from how these operations are defined in a math
textbook to deal with the reality of computing.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
pjdotson: The argument is not indefinite deferral, it is that it doesn't make sense to talk about
this behavior in isolation of the sized types, integer overflow, casting mismatch (e.g.
casting a `NaN` `f32` to a `uint8`) that makes sense for the language overall. You are
correct that the current implementation wasn't held to this bar, but that is because we
didn't put a lot of thought into some parts of the current implementation. Since we have
already spent hours thinking through this and related issues, we should actually solve
the overall language architecture problem as it relates to numeric types instead of just
solving this one issue.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
pjdotson: If you want to have a quick hotfix for schedular-crashing panics, then we should just do
the minimal changes (dividing by 0 results in zero) to make sure panics are avoided,
instead of doing this larger change of adding type coercions. I think this RFC needs to
actually consider the whole story: why have ten different numeric types, why have
separate float and integer types, why allow integer overflows, what does casting between
types look like for casts that don't have worldwide agreed-upon behavior, should we /
shouldn't we cast to a floating point type on division or power operations.

pjdotson (changes requested):
See comments, this RFC should really address the whole picture of type casting for division, how many different numeric types, type casting floats to ints for things like out-of-range values, `NaN`, and `+/-Inf`.

If you are worried about running into panics upon getting 0s in telemetry sources, you should ship a minimal fix that will avoid those panics, even if it has behavior we won't want long run like making integer division by zero result in a zero.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: Dividing by 0 resulting in 0 is a bad hotfix. It introduces wrong behavior just to fix a wrong behavior. You keep saying "whole story," but it's just an argument to deny a claim and jump to the next topic. 

The integer overflow conversation is a different conversation. i'm not saying it's not important. But it's absolutely not required for this. It's a separate topic, as has been routinely stated.

sy-nico: It already does consider "why have different numeric types". Or at the very least it's been answered. Because Synnax is a hardware and data acquisition software and it interfaces with various hardware. So it needs to be able to manage data of different types as set by external manufacturers or users.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: So which do you want to resolve first? In order, what specific issues and RFCs do you want to see? What **is** the big picture? What needs to be buttoned up before you are comfortable implementing a sound architectural decision?
Keep in mind the elevated level of criticism that is being applied to this topic, as opposed to something like, the `->`/`=>`, the various original stl implementations, the lack of divide-by-0 consideration, the behavior of derivative, or the units.

sy-nico: I will say, following the same standards, that a hot-fix to return 0 on a divide by 0 is unacceptable.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
sy-nico: I'm not saying "(u8) 255 + 255 = 254" is okay. These topics are orthogonal. Do you want an RFC and a change implemented to tackle the width/wrapping topic? What do you want to see about it? I am happy to implement in parallel or sequentially, but these will be fundamentally independent.

They are included in the RFC and this conversation because it has been routinely brought up as a reason to reject a well thought out and scoped change.

docs/tech/rfc/0038-260425-division-power-f64-typecast.md
sy-nico: I am indifferent. I can add support for `**` and add a deprecation warning for `^`. I will update the RFC to reflect what we decide, but this is a minor "bonus" change.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
pjdotson: The elevated level of criticism is because this is a major change away from the original
vision of the language, and the lack of consideration we have given to those other
things has bit us in the ass.

The point of the hotfix is if you think that it is completely unacceptable to have
panicking behavior and we need to do something, we should do the lowest-footprint change
and then come back later with the more thought-out change.

docs/tech/rfc/0038-260425-divide-typecast-f64-trade-study.md
pjdotson: I don't think these topics are orthogonal. I think all of these topics need to be
thought of together (not that changes for them need to be implemented in the same PR):
- addition,subtraction,multiplication,exponentiation underflow/overflow
- division,modulo by zero
- exponentiation domain errors
- out-of-range float -> int casts
- invalid (NaN) float -> int casts
- how many numeric types to have:
    1) single `number` type
    2) two types, `int` and `float`
    3) three types, `uint`, `int`, and `float`
    4) ten types (various densities of `uint`, `int`, and `float`)

We shouldn't change this until we have the whole vision put together because they are related, so we should
think of all of these together and make sure there is a coherent vision of what we want
with Arc. Otherwise, we are making a major change to the language by adding type
coercions to primitive mathematical operators without thinking through what we actually
want out of numbers in Arc.

The question I want answered is "why is integer overflow okay but integer division bad"?
If the ultimate answer to that is "integer overflow is bad as well", then that has implications
on the number of numeric types in Arc. If you say "integer division is bad because it
can panic and we don't want that to happen", that has implications on all of the other
types of potential panics from numeric operations.

The ultimate reason not to approve this is because I honestly think this is a change
that makes the language worse, unless we fully commit to a different type of language
that consolidates down to one or two numeric types.

pjdotson (changes requested):

PR #2275, opened 2026-04-28
SY-4122: Arc Statuses RFC

sy-nico:
# Pull Request

[SY-4122](https://linear.app/synnax/issue/SY-4122/arc-statuses)

<h3>Greptile Summary</h3>

This PR introduces RFC 0037 and its companion trade study, specifying an Arc `status` module with `status.set` (upsert by name or key) and `status.delete`, covering WASM host functions, Flow node implementation, type-system prerequisites (`Optional` and `AllowedLiterals` on `types.Param`), and two new `Writer[D]` service methods (`Update` and `UpsertByName`).

- **`status.set`**: UUID-vs-name dispatch with preserve-on-omit semantics; concurrent same-node creates are serialized via a `UpsertByName` transaction; by-name multi-match and by-key miss both surface as error-level task statuses with handle 0.
- **`status.delete`**: Single `key_or_name` dispatch; multi-match by name deletes all rows and emits an info-level count; zero-match emits a warning.
- **Trade study**: Evaluates upsert API shape and update expression syntax alternatives, recommending Shape 1 + Option A (shared payload, polymorphism) as the lowest-cost path given Arc's current type system.

Two defects require author attention before implementation: a pseudocode method call that would not compile, and a naming conflict across the two companion documents.

The `UpsertByName` pseudocode calls `scoped.NewRetrieve()` on a writer type that does not expose it, which would produce a compilation error if transcribed directly. The trade study uses `identifier` throughout all Arc examples while the RFC uses `key_or_name`, creating an authoritative naming conflict that a developer must resolve before coding.

Both files need attention: `0037-260427-arc-status-updates.md` for the `UpsertByName` retrieve-through-writer pattern, and `0037-trade-study.md` to align all Arc examples to the `key_or_name` parameter name.

| Filename | Overview |
|----------|----------|
| docs/tech/rfc/0037-260427-arc-status-updates.md | Main RFC for Arc status module: adds `status.set` and `status.delete`, defines WASM host functions, type-system prerequisites (Optional/AllowedLiterals on Param), service methods (Update, UpsertByName), and name-resolution logic. One substantive pseudocode defect: `UpsertByName` calls `scoped.NewRetrieve()` on a Writer type that does not expose retrieve methods. |
| docs/tech/rfc/0037-trade-study.md | Companion trade study evaluating Axis 1 (upsert API shape) and Axis 2 (update expression syntax). All Arc code examples still use `identifier` as the config parameter name, conflicting with the RFC's canonical `key_or_name`; a developer reading both would see contradictory parameter names for the same field. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Arc as Arc Program
    participant Host as WASM Host (set)
    participant Svc as status.Service
    participant DB as gorp/DB

    Arc->>Host: set(keyOrNameHandle, messageHandle, variantHandle)
    Host->>Host: uuid.Parse(keyOrName)

    alt key is UUID
        Host->>Svc: Writer.Update(ctx, key, overlay)
        Svc->>DB: gorp.NewUpdate.WhereKeys(key).ChangeErr(...).Exec(ctx, tx)
        alt found
            DB-->>Svc: updated row
            Svc-->>Host: nil
            Host-->>Arc: strings.Create(key)
        else query.ErrNotFound
            DB-->>Svc: ErrNotFound
            Svc-->>Host: ErrNotFound
            Host-->>Arc: 0 (error sentinel)
        end
    else key is name string
        Host->>Svc: Writer.UpsertByName(ctx, name, overlay)
        Svc->>DB: WithTx then WhereNames(name)
        alt exactly one match
            DB-->>Svc: one Status row
            Svc->>DB: Update(ctx, match.Key, overlay)
            DB-->>Svc: nil
            Svc-->>Host: (match.Key, nil)
            Host-->>Arc: strings.Create(key)
        else zero matches
            DB-->>Svc: empty result
            Svc->>DB: Set(ctx, newStatus)
            DB-->>Svc: nil
            Svc-->>Host: (newStatus.Key, nil)
            Host-->>Arc: strings.Create(newStatus.Key)
        else multiple matches
            DB-->>Svc: multiple Status rows
            Svc-->>Host: errMultipleMatches
            Host-->>Arc: 0 (error sentinel)
        end
    end
```

<sub>Reviews (11): Last reviewed commit: ["SY-4112: More updates"](https://github.com/synnaxlabs/synnax/commit/f0aeec27419dd529e9880a976e01e20ef4889371) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29960639)</sub>

### DIFF
--- a/docs/tech/rfc/0037-260427-arc-status-updates.md
+++ b/docs/tech/rfc/0037-260427-arc-status-updates.md
@@ -3,8 +3,7 @@
 **Feature Name**: Arc Status Module Updates <br /> **Status**: Draft <br /> **Start
 Date**: 2026-04-27 <br /> **Authors**: Nico Alba <br />
 
-**Related:** [RFC 0030 - Arc Module System](./0030-260221-arc-modules.md),
-[RFC 0036 - Arc Ranges Module](./0036-260427-arc-ranges.md)
+**Related:** [RFC 0030 - Arc Module System](./0030-260221-arc-modules.md)
 
 # 0 - Summary
 
@@ -21,8 +20,14 @@ interfaces adapted from the Python and TypeScript Synnax clients.
   state across the platform.
 - **Variant**: The severity level of a status. One of `success`, `info`, `warning`,
   `error`, `loading`, or `disabled`.
-- **WASM Form**, **Flow Form**, **`ExecBoth`**, **String Handle**: See
-  [RFC 0036, Section 1](./0036-260427-arc-ranges.md#1---vocabulary).
+- **WASM Form**: An Arc symbol invoked inside a `func` block; compiled to WebAssembly
+  and called as a host function with positional arguments.
+- **Flow Form**: An Arc symbol invoked as a graph node with named curly-brace config
+  (e.g., `trigger -> status.set{identifier=..., ...}`).
+- **`ExecBoth`**: A symbol exec mode where the same function is exposed in both WASM and
+  Flow form, sharing one `Inputs` parameter list.
+- **String Handle**: A `u32` handle returned by host functions that allocate strings on
+  the WASM side. Handle 0 is the error sentinel.
 
 # 2 - Motivation
 
@@ -44,14 +49,25 @@ module in line with other Arc modules and the Python/TypeScript client interface
 - **Test campaign teardown**: Automated test sequences often create per-run statuses
   that should be cleaned up when the campaign completes.
 
+## 2.1 - Breaking Change
+
+This RFC replaces the existing `status.set` Flow config shape wholesale. The current
+`status_key`/`name`/`variant`/`message` fields collapse into the new `identifier`/
+`variant`/`message` shape, and the deprecated bare `set_status` symbol is removed.
+Existing Arc programs that use `status.set` will fail to compile until updated.
+
+This is acceptable because `status.set` is a recent addition with a small, controllable
+user population. We don't have customers depending on the old shape, so the cost of a
+migration tool or a compatibility shim outweighs the benefit. Carrying two shapes
+forward (the muddled `status_key`/`name` split and the new unified `identifier`) would
+entrench the inconsistency this RFC exists to remove.
+
 # 3 - Prerequisite: String Handle Error Sentinel
 
 `status.set` returns the status key as a string handle so the caller can reference the
 status later for updates or deletion by key (avoiding name-resolution overhead on
-repeated calls). The same handle 0 error sentinel convention from
-[RFC 0036, Section 3](./0036-260427-arc-ranges.md#3---prerequisite-string-handle-error-sentinel)
-applies: handle 0 is falsy in WASM conditional expressions, `Get(0)` returns
-`("", false)`, and host functions return 0 on failure.
+repeated calls). Handle 0 is the error sentinel: handle 0 is falsy in WASM conditional
+expressions, `Get(0)` returns `("", false)`, and host functions return 0 on failure.
 
 ```go
 key := status.set("Pressure Check", "success", "All nominal")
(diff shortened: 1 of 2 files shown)


docs/tech/rfc/0037-260427-arc-status-updates.md
```diff
@@ -0,0 +1,484 @@
...
+        return
+    }
+    for _, s := range results {
+        if err := statusSvc.NewWriter(nil).Delete(ctx, s.Key); err != nil {
+            reportError(ctx, err)
+        }
+    }
+    if len(results) > 1 {
+        reportInfo(ctx, "Deleted %d statuses named '%s'", len(results), identifier)
+    }
```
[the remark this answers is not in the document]
sy-nico: Partial deletion of duplicate-named statuses" is not an inconsistent state. It's progress toward the goal. A single-writer transaction would be marginally cleaner, but calling this P1 is wrong. I'd leave it as-is

integration/tests/arc/stl_math.py
```diff
@@ -272,9 +272,10 @@ def _setup_pow_channels(self) -> None:
 
     def _setup_op_channels(self) -> None:
         for c in OP_BINARY_CASES:
-            create_virtual_channel(self.client, c.a_ch, sy.DataType.FLOAT64)
-            create_virtual_channel(self.client, c.b_ch, sy.DataType.FLOAT64)
-            create_virtual_channel(self.client, c.out_ch, sy.DataType.FLOAT64)
+            dtype = sy.DataType.INT64 if c.label == "mod" else sy.DataType.FLOAT64
+            create_virtual_channel(self.client, c.a_ch, dtype)
+            create_virtual_channel(self.client, c.b_ch, dtype)
+            create_virtual_channel(self.client, c.out_ch, dtype)
```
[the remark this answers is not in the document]
sy-nico: This test was incorrectly added to the PR. This has been removed.

docs/tech/rfc/0037-260427-arc-status-updates.md
```diff
@@ -0,0 +1,484 @@
...
+status.set(identifier: string, variant: string, message: string, name: string) -> string
+```
+
+| Param        | Type     | Required | Default | Description                                                        |
+| ------------ | -------- | -------- | ------- | ------------------------------------------------------------------ |
+| `identifier` | `string` | yes      |         | Status key or name (auto-detected)                                 |
+| `variant`    | `string` | yes      |         | `success`, `info`, `warning`, `error`, `loading`, or `disabled`    |
+| `message`    | `string` | yes      |         | Status message text                                                |
+| `name`       | `string` | no       | `""`    | Human-readable name to set on the status                           |
+| `status_key` | `string` | no       |         | Explicit key (Flow-only, alternative to name-based auto-detection) |
```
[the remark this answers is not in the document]
sy-nico: Unified into a single table with `status_key` as the shared identifier field across both WASM and Flow forms.

docs/tech/rfc/0037-260427-arc-status-updates.md
```diff
@@ -0,0 +1,454 @@
...
+
+**Returns:** Status key string. In WASM form, returned as a string handle (handle 0 on
+failure; see Section 3).
+
+`status_key` is only available in Flow form as an explicit alternative to `name`. In
+WASM form, the `identifier` argument handles both key and name via auto-detection.
+
+**Backward compatibility**: The existing `status_key` Flow config field is preserved.
+Programs using `status.set{status_key="...", ...}` continue to work unchanged. The
+`name` config field is a new alternative for Flow form, not a replacement. WASM form
```
pjdotson: I don't understand this: your signature above does not have a `name` argument

docs/tech/rfc/0037-260427-arc-status-updates.md
```diff
@@ -0,0 +1,454 @@
...
+
+# 6 - Implementation Plan
+
+## 6.0 - Modified Files
+
+| File                                    | Change                                                                                                                       |
+| --------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| `core/pkg/service/arc/status/set.go`    | Change `set` to `ExecBoth`, add WASM host function bindings, update symbol types, add name-based identification to Flow node |
+| `core/pkg/service/arc/status/delete.go` | New file: `delete` symbol, WASM host function, Flow node                                                                     |
+| `core/pkg/service/arc/runtime/task.go`  | Register status WASM host functions in WASM builder, pass `*status.Service` to host function closures                        |
```
pjdotson: should update C++ too

pjdotson (changes requested):
The biggest thing to note here is that I don't think we should have separate function signatures (arguments and returns) for WASM and Flow. This is going to confuse users because something with the same name (`status.set` or `status.delete`) will have two different ways of calling it and return two different things.

docs/tech/rfc/0037-260427-arc-status-updates.md
sy-nico: I got myself confused when working on the ranges rfc in parallel. My aim/suggestion is to have a unified "identifer" field that accepts the key or name streing for the status. I'm still exploring this idea. 

I think it would make sense to have a `status.create(name="") function, where the behavior would be:
- create if `not_found`
- If one is found, throw a notification and return.
- If multiple are found, throw a warning and return the first.
  - Or we could decide to throw an error and return `None`.

We just need a clean way for operators to initialize/define the statuses and make sure they are in the system before their sequences run and they call the standard `status.set(id, variant, msg)` with low friction.

docs/tech/rfc/0037-260427-arc-status-updates.md
```diff
@@ -0,0 +1,853 @@
...
+        }
+        if variantHandle != 0 {
+            stat.Variant = status.Variant(strings.Get(variantHandle))
+        }
+        stat.Time = telem.Now()
+    }
+
+    var stat status.Status[any]
+    if _, err := uuid.Parse(identifier); err == nil {
+        err := statusSvc.NewRetrieve().WhereKeys(identifier).Entry(&stat).Exec(ctx, nil)
```
emilbon99: We already have built-in update capabilities in gorp. There is no need to invent your own retrieve/write mechanisms here. If you need an update mechanism, then you should update the status service to support it.

emilbon99: Do you want me to leave syntax correction comments like this in an RFC?

emilbon99:
1) We don't have a separation between `set` and `create` in our existing client libraries. This means we're intentionally deviating from the existing pattern we've laid out over the last two years. Do you feel your justification for having separate methods is warranted, and, if so, should we be changing our client libraries?  **I highly recommend doing research into how the best ORMs and client libraries in the world solve this problem. Are there novel ways of working across updates/creates that you haven't discovered yet? Examples are prisma, django, etc.** These companies have spent years thinking about these problems. How much of what you wrote in your RFC is based off of what industry says we should do? If so, cite examples.

2) From what I've read it doesn't seem like you have a proper solution to the reactive vs. imperative context versions of `status.set` i.e. these are two separate function signatures. You say that the configuration parameters map directly to the function input parameters? How and why? This is another new standard you're introducing. These two things are fundamentally different in arc. 

```
func set{
    cat string
}() {

}
```

vs.

```
func set(cat string) {

}
```

One gets called like this `set{cat="dog"}` and one takes its input parameter as `string_chan -> set{}` which then receives `cat` via string chan. If you're proposing we merge the two somehow, how do we do it? does this apply only to statuses? If not, how does it positively or negatively affect other functions. This doesn't seem well thought out.

emilbon99 (changes requested):
See my long comment attached to the RFC

docs/tech/rfc/0037-260427-arc-status-updates.md
sy-nico: Updated the RFC to add a `Writer.Update` method on the status service that wraps `gorp.NewUpdate`, and reworked the host function to compose service methods instead of opening retrieve/write directly.

sy-nico:
@emilbon99 , replying to [this top-level comment](https://github.com/synnaxlabs/synnax/pull/2275#issuecomment-4353916762)

1. The `set`/`create` split is gone. `create` was removed and `set` is now a true upsert, matching the Python and TS clients.

2. Vocabulary slip on my part. Should have declared the params as `Config`, not `Inputs`. The model is the same as `trigger -> time.wait{duration=3s} -> next`: wire triggers, `Config` carries the values. WASM fills the same slots positionally. Fixed in the type declarations and the Vocabulary entry.

docs/tech/rfc/0037-260427-arc-status-updates.md
```diff
@@ -0,0 +1,850 @@
...
+
+```
+status.set(identifier: string, message?: string, variant?: string) -> string
+```
+
+| Param        | Type     | Required | Default                                | Description                                                                                |
+| ------------ | -------- | -------- | -------------------------------------- | ------------------------------------------------------------------------------------------ |
+| `identifier` | `string` | yes      | n/a                                    | Status key (UUID) or name                                                                  |
+| `message`    | `string` | no       | preserve existing / `""` on create     | If supplied, overwrites; if omitted, preserves the existing value (or `""` on create).     |
+| `variant`    | `string` | no       | preserve existing / `"info"` on create | If supplied, overwrites; if omitted, preserves the existing value (or `"info"` on create). |
```
pjdotson: should very strongly consider having some type of string enum type for statuses so users
can only pass in allowed statuses (`"success"`, `"info"`, etc.). Probably for the v1
implementation it is fine to keep it as a string type, but in the future this is
definitely something that users would want to know at compile-time.

docs/tech/rfc/0037-260427-arc-status-updates.md
```diff
@@ -0,0 +1,850 @@
...
+
+**Resolution logic:**
+
+1. `uuid.Parse(identifier)`.
+   - Parseable: attempt key lookup via `WhereKeys(identifier)`.
+     - On success: apply the update (see below) to that row and return its key.
+     - On `query.ErrNotFound`: emit an error-level task status (UUIDs are
+       server-assigned and cannot be created by the caller), return handle 0.
+     - On any other error: emit an error-level task status, return handle 0.
+   - Not parseable: continue to step 2.
```
pjdotson, line 239: this doesn't make sense. Some of our statuses right now have non-UUID keys (like task
statuses). Shouldn't we always assume it is a key first?

docs/tech/rfc/0037-260427-arc-status-updates.md
```diff
@@ -0,0 +1,850 @@
...
+
+**Signature:**
+
+```
+status.set(identifier: string, message?: string, variant?: string) -> string
+```
+
+| Param        | Type     | Required | Default                                | Description                                                                                |
+| ------------ | -------- | -------- | -------------------------------------- | ------------------------------------------------------------------------------------------ |
+| `identifier` | `string` | yes      | n/a                                    | Status key (UUID) or name                                                                  |
```
pjdotson: why did you decide on having an `identifier` shape for both keys and names instead of a
separate key and name Param, where only one is required?

pjdotson (changes requested):
Two things:

1) I would really like to see longer code examples (like ~100 line) of the Arc code that thinks about having a long running automation and making sure your alerts all trigger the same statuses.
2) I don't think you should be _afraid_ of saying we need to add new changes to Arc for things like different ways of passing parameters, some type of record type, etc. These obviously will take work, but the first step should be thinking "what is the ideal API for working with statuses in Arc", and not worrying about implementation difficulty.

PR #2283, opened 2026-04-30
SY-3833: Strongly Type Schematics on the Core

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

This is the next split in the SY-3833 schematic strong-typing series. It moves the schematic from an opaque `data: record` blob into a fully typed shape on the server (`Legend`, `Node`, `Edge` with nested `Handle`, `EdgeProps`, structured `props`/`legend`/`authority`), and brings server-side migration support to parity with what the console already has.

Highlights:

- **`schemas/schematic.oracle`** rewritten as the strongly-typed shape; client TS, Go, and protobuf types regenerate from it. Console v6 state types now alias the oracle-generated `nodeZ`/`edgeZ`/`legendZ`/`edgePropsZ`/`handleZ`/`segmentZ` so the wire format and console state can never drift.
- **One-time legacy unwrap chain** at `core/pkg/service/schematic/migrations/legacy/v0..v5/`. Each version owns its own `Lift` step that decodes the blob if its own version matches, otherwise recurses into the predecessor. The top-level `MigrateSchematic` is a normal typed-to-typed step (`AutoMigrateSchematic` for gorp-entry fields + a single inline projection from `v5.Data` to the typed body).
- **`v55_lift_typed_schematic`** wired into the gorp `Migrations` slice (sequenced after `msgpack_to_orc`) so existing clusters lift their stored schematics on first read after the upgrade.
- **Generator extensions** in `oracle/plugin/go/`:
  - Codec generator emits `make(map[string]msgpack.EncodedJSON, n)` for `map<string, record>` fields (was emitting `interface{}`).
  - Protobuf translator handles non-primitive map values: `record` via `structpb.NewStruct`/`AsMap`, struct values via per-element `XToPB`/`XFromPB`, with a new `MapValueConversion.{Forward,Backward}HasError` template branch for error-returning conversions.
  - Added `int16`/`uint16` cases to the primitive-conversion table so `int16?` fields round-trip through proto's `int32`.
  - Test fixture generator emits `msgpack.EncodedJSON{...}` literals for `record` so generated codec tests compile.
- **`SetData(ctx, key, Schematic)`** now takes a typed body, preserves `Key` and `Name` from the existing entry, and refuses snapshot edits. The TS client (`SetDataBody`) and console call sites (`Schematic.tsx`, `services/link.ts`, `services/ontology.tsx`, `range/overview/Snapshots.tsx`, `workspace/services/ontology.tsx`, `export.ts`) project console state into the typed shape directly.
- **Slice fix**: `applyEdgeChanges` now initializes `props[edge.key] = {}` on every add so newly-drawn edges always have a parseable EdgeProps entry. Selectors (`selectNodeProps`, `selectElementProps`, `selectEdgeProps`) safe-parse against the loose props record and return `undefined` on absence/mismatch instead of throwing.
- **Schema tweaks vs. sy-3833 inspiration**:
  - `Node.measured` dropped (pluto re-measures at render time; storing it server-side was dead weight).
  - `Node.z_index` is `int16?` (right-sized, optional to match v0 wire format).
  - `EdgeProps.variant` uses `@validate default "pipe"`, `EdgeProps.segments` defaults to `[]`, so `edgePropsZ.parse({})` produces valid defaults — used both at edge-add time and during the v5→v6 lift.

Test coverage: 25 schematic specs + 2 v55 gorp e2e specs + 26 symbol specs all pass; per-version legacy migration tests exercise the v0→v5 chain through realistic blob fixtures.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR strongly types the schematic on the server side, moving from an opaque `data: record` blob to a fully-typed `Schematic` (`Legend`, `Node`, `Edge`, `Handle`, structured `Configs`). Proto, Go codec, and TypeScript types all regenerate from the new oracle schema, and a one-time legacy migration chain (`v0`\u2192`v5` via `legacy/` packages, then `v55_lift_typed_schematic` in gorp) brings existing clusters forward on first read after upgrade.

- **Server-side migration ladder** (`migrations/legacy/v0\u2026v5`, `MigrateSchematic`): each version frozen, `legacy.MigrateData` dispatches on version string and walks the chain to `v5.Data`; `MigrateSchematic` then projects into the typed body, preserving `Key`/`Name`/`Snapshot` from the gorp entry.
- **Generator updates** (`oracle/plugin/go/pb`, `marshal`): `int16`/`uint16` now emit casts in the proto translator, `map<string,record>` fields emit typed `EncodedJSON` maps, and the test-fixture generator emits correct `msgpack.EncodedJSON` literals.
- **Console wiring** (`slice.ts fromRemote`, `Schematic.tsx`, all service call sites): `fromRemote` replaces every previous `{ ...schematic.data, key, name }` spread, consistently setting `remoteCreated: true` and projecting the wire `Legend`/`Node[]`/`Edge[]`/`configs` into the Redux state type.

Safe to merge. The migration chain is correct end-to-end and all previously-identified defects have been resolved.

All previously-flagged defects are addressed: int16/int32 casts are explicit in the proto translator, ZERO_LEGEND carries correct pixel units, migrateLegend unconditionally initializes Colors, Configs is never left nil after MigrateSchematic, and fromRemote consistently sets remoteCreated: true across all call sites. The two remaining observations are defensive hygiene items that do not represent present bugs.

core/pkg/service/schematic/codec.gen.go — Legend.Colors stays nil when the ORC presence bit is false, relying on the TypeScript nullishToEmpty guard rather than the Go layer being self-consistent with migrate.go.

| Filename | Overview |
|----------|----------|
| core/pkg/service/schematic/migrate.go | New file: MigrateSchematic + helpers. Colors always initialized via make(); Configs nil-guard after migrateProps; edge data lifted into Configs correctly. |
| core/pkg/service/schematic/codec.gen.go | Rewrites ORC codec for typed Schematic. Legend.Colors and Schematic.Configs use presence-bit pattern (nil when not present), mitigated by TypeScript nullishToEmpty guard. |
| core/pkg/service/schematic/pb/translator.gen.go | Complete rewrite of proto translator. ZIndex int16↔int32 casts now correct. nil Configs guard present. |
| console/src/schematic/slice.ts | Introduces fromRemote() helper setting remoteCreated: true; applyEdgeChanges seeds configs[edge.key] on every add. |
| client/ts/src/schematic/types.gen.ts | Wire-format types fully generated from oracle. legendZ.colors uses record.nullishToEmpty; nodes/edges use array.nullishToEmpty. |
| console/src/schematic/types/v6.ts | v6 aliases oracle nodeZ/edgeZ/legendZ from @synnaxlabs/client; single migrateLegendColors declaration; stateMigration correctly converts props→configs. |
| core/pkg/service/schematic/migrations/legacy/legacy.go | New entry-point package. Version-string dispatch handles empty string (→ v0) through v5. Orphan edges filtered uniformly after chain completes. |
| client/ts/src/schematic/client.ts | setData now accepts typed SetDataBody. ZERO_LEGEND exported with correct units: { x: 'px', y: 'px' }. |
| core/pkg/service/schematic/service.go | Wires v55_lift_typed_schematic migration after msgpack_to_orc with correct dependency ordering. |
| x/ts/src/record/record.ts | nullishToEmpty overloaded with typed key/value schema variant; null/undefined → {} coercion preserved in both arities. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Stored msgpack blob] -->|CodecMigration| B[v55.Schematic]
    B -->|v55_lift_typed_schematic| C[schematic.Schematic Typed]

    subgraph legacy[legacy.MigrateData]
        D{version}
        D -->|empty/v0| E[v0 chain]
        D -->|1.0.0| F[v1 chain]
        D -->|2.0.0| G[v2 chain]
        D -->|3.0.0| H[v3 chain]
        D -->|4.0.0| I[v4 chain]
        D -->|5.0.0| J[v5 direct]
        E & F & G & H & I & J --> K[v5.Data]
    end

    B --> D
    K -->|migrateNode/Edge/Legend/Props| C
    C -->|ORC| L[HTTP JSON]
    C -->|PB| M[gRPC]
    L & M -->|schematicZ.parse| N[TS Schematic]
    N -->|fromRemote| O[Redux State remoteCreated=true]
```

1. `console/src/schematic/types/v6.ts`, line 157-167 ([link](https://github.com/synnaxlabs/synnax/blob/7d8b91731bb6111a6e43fe200644b292ed972458/console/src/schematic/types/v6.ts#L157-L167)) 

   <a href="#">P0</a> **Duplicate `migrateLegendColors` declaration — compile error**

   `migrateLegendColors` is declared twice as a `const` in the same module scope (lines 145 and 157). TypeScript will reject this with `Cannot redeclare block-scoped variable 'migrateLegendColors'`, preventing the console from building. The second declaration (returning `LegendState["colors"]`) is the old version that should have been removed when the type was updated to use `Legend["colors"]`.

2. `client/ts/src/schematic/client.ts`, line 156-160 ([link](https://github.com/synnaxlabs/synnax/blob/9ec21e9fa13ee163d992f751af2b527e6d6e1f24/client/ts/src/schematic/client.ts#L156-L160)) 

   <a href="#">P1</a> **`ZERO_LEGEND` position renders at `5000%` when units absent**

   `ZERO_LEGEND.position` is `{ x: 50, y: 50 }` with no `units`. The sticky-XY renderer at `x/ts/src/spatial/sticky/sticky.ts` interprets missing units as **decimal** fractions: `${pos.x * 100}%` → `5000%`. Any schematic created via `ZERO_NEW` (which embeds this legend) will have its legend rendered far outside the viewport.

   The v6.ts internal `ZERO_LEGEND` correctly uses `units: { x: "px", y: "px" }` to anchor the legend at 50 px. The same units must be set here so the two defaults remain consistent.
</details>

<sub>Reviews (17): Last reviewed commit: ["addressed pull request comments"](https://github.com/synnaxlabs/synnax/commit/79b3aa1abade32912be84a5a8952589958dbadfb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30352499)</sub>


core/pkg/service/schematic/pb/translator.gen.go
```diff
@@ -44,9 +438,27 @@ func SchematicFromPB(pb *Schematic) (schematic.Schematic, error) {
...
+	}
 	r.Name = pb.Name
 	r.Snapshot = pb.Snapshot
+	r.Authority = control.Authority(pb.Authority)
+	if pb.Props != nil {
+		r.Props = make(map[string]msgpack.EncodedJSON, len(pb.Props))
+		for k, v := range pb.Props {
+			r.Props[k] = msgpack.EncodedJSON(v.AsMap())
+		}
+	}
```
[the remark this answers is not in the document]
emilbon99: `(*structpb.Struct).AsMap()` is nil-safe — `GetFields()` returns `nil` on a nil receiver, so `AsMap()` on a nil `*structpb.Struct` returns an empty map rather than panicking. Skipping the entry would also be a behavior change (silently dropping the key vs. preserving it as an empty bag), so leaving the loop as-is.

schemas/schematic.oracle
```diff
@@ -15,26 +19,100 @@ Key = uuid {
...
+        @doc value "is the unique node identifier within the schematic."
+    }
+    position spatial.XY {
+        @doc value "is the top-left position of the node."
+    }
+    z_index  int16?     {
+        @doc value """
+            is the stacking order of the node within the schematic. Higher
+            values render above lower values. Set by the user via
+            send-to-back / bring-to-front actions.
```
sy-nico, line 47: Is this necessary? We don't have a concept of moving symbols forwards or backwards on a schematic.

console/src/schematic/slice.ts
```diff
@@ -55,6 +56,15 @@ export const purgeSliceState = (state: RootState): RootState => {
 
 export const PERSIST_EXCLUDE = [purgeSliceState];
 
+export const fromRemote = (s: schematic.Schematic): State => ({
+  ...ZERO_STATE,
+  ...s,
+  // TODO: remove this assertion when schematic element ocnfigs are strongly typed on
```
sy-nico: type: `ocnfigs`

oracle/plugin/go/pb/pb.go
sy-nico, line 1: Add tests for the new error-returning branches in `buildMapValueConversion`: `map<K, record>` (sets `ForwardHasError`) and `map<K, StructType>` (sets both). 

The existing `map<uint32, uint8>` test only hits the non-error primitive path.

sy-nico (changes requested):
Only request is to add tests. Non-blocking is to address a typo and a question I have regarding the node's z-index.

schemas/schematic.oracle
emilbon99: Yeah, it is necessary because even though we don't allow user customizable z-indexes (yet) we do specify them statically and those need to be passed to react flow dynamically.

emilbon99: There may be a hacky workaround so we doing need this field, but I figure our users are going to be able to edit z-indexes manually so we might as well add it.

sy-nico (approved):

PR #2287, opened 2026-05-01
SY-4145: Fix Schematic Keyboard Triggering

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4145](https://linear.app/synnax/issue/SY-4145/fix-schematic-keyboard-triggering)

## Description

Suppress keyboard triggers when target is outside the trigger region.
Fixed at the trigger level rather than checking modal state, so it applies universally to any overlay (modals, dialogs, command palette, etc.).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes keyboard shortcut triggers firing inside schematic when a modal/dialog/overlay is open, by suppressing triggers whose `target` is in a sibling DOM subtree (not an ancestor or descendant of the region). It also closes the previously noted gap where `"end"` events were emitted even when the matching `"start"` had been blocked, by routing `removed` triggers through the same `filterInRegion` guard.

Safe to merge — logic is correct, tests are updated, and the only findings are P2 style suggestions.

No P0 or P1 issues found. The ancestor-or-descendant containment check correctly blocks sibling-subtree events while preserving the existing document.body-targeted keyboard shortcut path. Both start and end filtering are now symmetric.

| Filename | Overview |
|----------|----------|
| pluto/src/triggers/hooks.ts | Applies `filterInRegion` to `removed` triggers (fixing spurious `"end"` events) and extends the non-mouseEvent branch to require `target` and `region.current` share an ancestor-descendant relationship, blocking sibling-subtree events. Logic is correct; minor style nit with `region.current!` inside the filter. |
| pluto/src/triggers/Triggers.spec.tsx | New test covers sibling-subtree suppression; two existing tests updated to reflect that `filteredRemoved` now also gates `"end"` events. Missing explicit `"end"`-count assertion in the new test. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Trigger event fires] --> B{added or removed?}
    B --> C[filterInRegion for added]
    B --> D[filterInRegion for removed]
    C --> E{region provided?}
    D --> E
    E -- no --> F[pass through]
    E -- yes --> G{region.current null?}
    G -- yes --> H[block all]
    G -- no --> I{regionMustBeElement or Mouse trigger?}
    I -- yes --> J{cursor in box AND target === region.current?}
    I -- no --> K{cursor in box AND region contains target OR target contains region?}
    J -- yes --> L[allow trigger]
    J -- no --> M[block trigger]
    K -- yes --> L
    K -- no --> M
    L --> N{added.length > 0?}
    L --> O{filteredRemoved.length > 0?}
    N -- yes --> P[fire stage: start]
    O -- yes --> Q[fire stage: end]
```

1. `pluto/src/triggers/hooks.ts`, line 69-77 ([link](https://github.com/synnaxlabs/synnax/blob/456423a088fce21a2af60c7edaa585951fd25400/pluto/src/triggers/hooks.ts#L69-L77)) 

   <a href="#">P2</a> **Spurious `"end"` events without a preceding `"start"`**

   With this change, a keyboard trigger's `"start"` is now suppressed when the event target is in a sibling subtree (correct), but the `removed` diff is **never** passed through `filterInRegion`. That means when the keys are released after a blocked start, the callback is still called with `stage: "end"` even though `stage: "start"` was never fired.

   For callers like `useHeldRef`, a spurious `"end"` is benign (`purge` on an empty list is a no-op). But any caller that treats `"start"` and `"end"` as a matched pair (e.g. setting a loading state to true on start and false on end) could enter an inconsistent state here.

   Consider applying the same target-containment guard to `removed`:
   ```typescript
   const filteredRemoved = filterInRegion(e.target, e.cursor, removed, region, regionMustBeElement);
   if (added.length > 0)
     f?.({ ...base, stage: "start", triggers: added, prevTriggers: e.prev });
   if (filteredRemoved.length > 0)
     f?.({ ...base, stage: "end", triggers: filteredRemoved, prevTriggers: e.prev });
   ```
</details>

<sub>Reviews (2): Last reviewed commit: ["SY-4145: Filter removed triggers through..."](https://github.com/synnaxlabs/synnax/commit/0aa793136a9df611a65a9bf6d67e9da316637e94) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30484715)</sub>

### DIFF
--- a/pluto/src/triggers/hooks.ts
+++ b/pluto/src/triggers/hooks.ts
@@ -92,7 +92,10 @@ const filterInRegion = (
   return added.filter((t) => {
     const rg = regionMustBeElement ?? t.some((v) => v.includes("Mouse"));
     if (rg) return box.contains(b, cursor) && target === region.current;
-    return box.contains(b, cursor);
+    return (
+      box.contains(b, cursor) &&
+      (region.current!.contains(target) || target.contains(region.current))
+    );
   });
 };
 
--- a/pluto/src/triggers/hooks.ts
+++ b/pluto/src/triggers/hooks.ts
@@ -70,11 +70,18 @@ export const use = ({
       const removed = res[1];
       if (added.length === 0 && removed.length === 0) return;
       added = filterInRegion(e.target, e.cursor, added, region, regionMustBeElement);
+      const filteredRemoved = filterInRegion(
+        e.target,
+        e.cursor,
+        removed,
+        region,
+        regionMustBeElement,
+      );
       const base = { target: e.target, cursor: e.cursor };
       if (added.length > 0)
         f?.({ ...base, stage: "start", triggers: added, prevTriggers: e.prev });
-      if (removed.length > 0)
-        f?.({ ...base, stage: "end", triggers: removed, prevTriggers: e.prev });
+      if (filteredRemoved.length > 0)
+        f?.({ ...base, stage: "end", triggers: filteredRemoved, prevTriggers: e.prev });
     });
   }, [f, memoTriggers, listen, loose, region, double, regionMustBeElement]);
 };
--- a/pluto/src/triggers/Triggers.spec.tsx
+++ b/pluto/src/triggers/Triggers.spec.tsx
@@ -439,6 +439,53 @@ describe("Triggers", () => {
       fireEvent.keyUp(document.body, { code: "KeyA" });
     });
 
+    it("should not trigger when target is in a sibling subtree of the region", async () => {
+      Element.prototype.getBoundingClientRect = mockBoundingClientRect(0, 0, 100, 100);
+      const callback = vi.fn();
+      const regionRef = { current: document.createElement("div") };
+      const C = () => {
+        Triggers.use({
+          callback,
+          triggers: [["Control", "V"]],
+          region: regionRef,
+          loose: true,
+        });
+        return <div ref={regionRef}>Target Region</div>;
+      };
+      const { container } = render(
+        <Triggers.Provider>
+          <C />
+          <input data-testid="modal-input" />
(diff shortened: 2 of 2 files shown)


pluto/src/triggers/hooks.ts
emilbon99, line 1: We already have a `regionMustBeElement` option? Is that option still necessary? I remember getting the triggers to work correctly in all scenarios was pretty tricky and it took me a while to get right, so I'm a bit concerned that this will unexpectedly break triggers in places where we don't expect. 

Just double check and make sure you're confident that this fix is the right one and is sustainable, and if `regionMustBeElement` is necessary.

emilbon99 (changes requested):

pluto/src/triggers/hooks.ts
sy-nico: Yes, `regionMustBeElement` is still necessary. They handle different cases:

- `regionMustBeElement` (auto-enabled for mouse trigger and stricter): `target === region.current`. Exact match, right for clicks.
- New containment check (keyboard triggers): `region.contains(target) || target.contains(region)`. Looser, but blocks sibling subtrees like modals. Keyboard events target `document.body`, so `regionMustBeElement` would be too strict and block every keyboard shortcut.

The one scenario where the containment check could be an issue: a detached floating toolbar (sibling in the DOM, not a child of the canvas region) where a user focuses an input and presses a shortcut meant to affect the canvas. The containment check would block it since the toolbar is a sibling subtree. In practice, no caller uses this pattern today. All callers that pass `region` (viewport, diagram, table, log) set it to a container ref that wraps all relevant interactive elements.

emilbon99 (approved):

PR #2290, opened 2026-05-02
SY-3833-2: Add Schematic Action Codec and Dispatch Endpoint

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Lays the server-side foundation for granular schematic mutations. Introduces an oracle-generated discriminated-union action codec, a hand-written reducer, and a new `Writer.Dispatch` endpoint that applies an action sequence atomically. The existing `SetData` endpoint is **not** removed; both coexist so console and pluto continue to work unchanged through the next split.

### Oracle codegen

- New `action` grammar production inside struct bodies, alongside fields, domains, and field omissions. Lexer and parser regenerated via ANTLR (most of the line-count delta is parser regen).
- `resolution.Action` and `analyzer.collectAction` lift action declarations into the resolution table. `resolveTypeRefs` now walks action payload field types so cross-references inside action bodies (e.g. `node Node`) qualify correctly to `schematic.Node`.
- New `oracle/plugin/go/actions` plugin generates an `Action` envelope, payload structs, `Reduce`, `ReduceAll`, and `NewXxxAction` constructors for any struct that declares actions. Optional fields use the same pointer-prefix logic as `gotypes` (skip slice / map / `msgpack.EncodedJSON`) — v1's actions plugin had this wrong and would have produced `*msgpack.EncodedJSON` for `record?` fields.
- Formatter learned to emit `actionDef` blocks so `oracle fmt` / `oracle sync` preserves them.

### Schematic schema and runtime

- `schemas/schematic.oracle` declares six actions: `SetNodePosition`, `AddNode`, `RemoveNode`, `SetEdge`, `RemoveEdge`, `SetProps`.
- Hand-written `Handle` methods live in `core/pkg/service/schematic/actions.go` alongside the `ScopedAction{Key, SessionKey, Actions}` envelope used for cluster broadcast and client self-dedup.
- `Writer.Dispatch(ctx, key, sessionKey, actions)` applies a sequence inside a single gorp transaction, rejects snapshots with `validate.ErrValidation`, and notifies `actionObserver` on success.
- `Service` wires `actionObserver` to a signals translator that publishes `ScopedAction` to `sy_schematic_set` (JSON) / `sy_schematic_delete` (UUID) when `cfg.Signals` is non-nil.
- API exposes `Dispatch` as a new freighter endpoint at `/api/v1/schematic/dispatch`, registered alongside `SetData` in the layer table, gRPC noop registry, and HTTP router.

### Tests

- 5 new specs for the actions plugin (`oracle/plugin/go/actions/actions_test.go`) covering: no actions emits no files, full discriminated-union codec, snake_case naming, no pointer prefix on slices/maps/records.
- 4 new ginkgo specs for `Writer.Dispatch` covering: single SetNodePosition, multi-action atomic apply, snapshot rejection, no-op on missing keys.

### Notes

- TS-side codegen (a TypeScript actions sub-plugin) is intentionally deferred to the next split, where the pluto consumer will actually call `client.schematics.dispatch`.
- v1 also defined a `SetNodeDimensions` action that targeted a `measured spatial.Dimensions?` field. The v2 schema still has `z_index` instead, so `SetNodeDimensions` is not included here. It can be added in a follow-up if the pluto consumer needs it.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR introduces a server-side schematic action codec for granular schematic mutations: an oracle-generated discriminated-union `Action` type, hand-written `Handle` methods, a new `Writer.Dispatch` endpoint that applies an action sequence atomically, and a corresponding TS codec generated by a new oracle TypeScript actions plugin. The existing `SetData` path is unchanged.

- **Go side**: `actions.gen.go` emits an `Action` envelope with nine payload variants, nil-checked dispatch in `Reduce`, and `ReduceAll`; hand-written `Handle` methods in `actions.go`; `Writer.Dispatch` applies the sequence inside a gorp transaction and notifies the `actionObserver`; `service.go` wires the observer into the cluster signals pipeline (`sy_schematic_set` for JSON action sequences, `sy_schematic_delete` for UUID deletions).
- **TS side**: `actions.gen.ts` emits Zod payload schemas, a discriminated `actionZ` union, action constructors, a `Handlers` interface, and `createReducer`/`createReduceAll` factories using immer; `client.ts` adds a `dispatch` method to the schematic client.
- **Oracle codegen**: New `action` grammar production, `go/actions` and `ts/actions` plugins, formatter support for `actionDef` blocks, and updated analyzer/resolution for action field type cross-references.

Not safe to merge — multiple issues identified in prior review rounds remain unaddressed and are still present in the code.

The prior review threads identified a cluster of real defects that are confirmed still present in the current diff. The TS actionZ discriminated-union schema uses camelCase payload keys (e.g. setNodePosition) while Go serializes them as snake_case (set_node_position), meaning every dispatched action and every broadcast action will silently carry the wrong key; clients can neither send nor receive actions correctly. scopedActionZ.sessionKey has the same mismatch against Go's session_key, so parsed broadcast messages will never have a valid session identifier. Writer.Dispatch notifies actionObserver inside the WithTx callback before the outer commit, so a failed commit still emits a ScopedAction that subscribers apply to local state against a change that was never persisted. And gorp.Update.Exec returns nil when MatchKeys finds no rows, so Dispatch silently succeeds on missing keys — the test that asserts query.ErrNotFound for a missing schematic will fail at runtime.

oracle/plugin/ts/actions/actions.go (template emits wrong payload key casing), client/ts/src/schematic/actions.gen.ts (carries the key mismatch into the generated file), client/ts/src/schematic/client.ts (scopedActionZ session key field name), core/pkg/service/schematic/writer.go (observer-before-commit and missing-key silent success)

| Filename | Overview |
|----------|----------|
| core/pkg/service/schematic/actions.gen.go | Generated Action envelope with nine variants, nil-guarded Reduce switch, ReduceAll, and constructor helpers. Nil checks for nil payload are now present (addressed from prior review). Actions declared in oracle schema match generated variants. |
| core/pkg/service/schematic/actions.go | Hand-written Handle methods for all nine action payload types including SetNodeMeasured (compilation issue from prior review resolved). RemoveNode.Handle uses the append-delete slice pattern which mutates the backing array shared with the caller's copy, but ReduceAll always reassigns state so this is safe in practice. |
| core/pkg/service/schematic/writer.go | Dispatch applies actions inside a gorp ChangeErr callback then notifies actionObserver. Two unresolved issues from prior threads: actionObserver fires before the outer WithTx commits, and gorp.Update.Exec returns nil for missing keys so Dispatch silently succeeds rather than returning query.ErrNotFound. |
| core/pkg/service/schematic/service.go | Wires actionObserver to a signals translator that publishes ScopedAction JSON to sy_schematic_set; a separate delete-only gorp publisher covers sy_schematic_delete. Design correctly separates action-dispatch signals from CRUD deletion signals. |
| oracle/plugin/ts/actions/actions.go | Template emits camelCase (e.g. setNodePosition) as the payload object key inside the discriminated union, constructors, and createReducer accessor. Go serializes the same field as set_node_position (snake_case), creating a persistent TS/Go wire-format mismatch flagged in prior threads. |
| client/ts/src/schematic/actions.gen.ts | Generated file carries the camelCase payload key mismatch from the TS actions template (e.g. setNodePosition vs Go's set_node_position). Zod parsing of server-sent actions and client-side dispatch will fail due to key mismatch (prior thread). |
| client/ts/src/schematic/client.ts | dispatch() method and dispatchReqZ correctly use snake_case session_key. scopedActionZ uses camelCase sessionKey while Go marshals the field as session_key, breaking Zod validation of broadcast messages (prior thread). |
| core/pkg/service/schematic/writer_test.go | New Dispatch specs are well-structured. The missing-key test expects query.ErrNotFound but will receive nil because gorp.Update.Exec does not invoke ChangeErr when no rows match — confirmed in prior thread. Tests for observer notification are accurate. |
| core/pkg/distribution/signals/publisher.go | Refactored to support optional set/delete channels (leave Name empty to disable). Validation, Override, and the publish loop all correctly guard on channel name. The empty-frame guard prevents spurious writes. |
| core/pkg/distribution/signals/gorp.go | Added DisableSet/DisableDelete flags with correct Validate, Override, and filtering logic. Returns len(out) > 0 from the translate function to avoid empty-change notifications. |
| oracle/plugin/go/actions/actions.go | Go actions plugin correctly emits snake_case JSON tags and the nil-guarded Reduce switch. processField mirrors gotypes optionality handling for pointer-prefix logic, skipping slices/maps/msgpack.EncodedJSON. |
| core/pkg/api/schematic/schematic.go | Dispatch API handler enforces access control then delegates to Writer.Dispatch inside WithTx. Observer fires before WithTx commits (prior thread). |
| x/ts/src/caseconv/caseconv.ts | getSchemaShape now merges all discriminated-union variants instead of returning on first match. Merged shape allows case-conversion lookup to find a field name declared in any union variant. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client as TS Client
    participant API as API Layer (Dispatch)
    participant Writer as schematic.Writer
    participant DB as gorp.DB (tx)
    participant Observer as actionObserver
    participant Signals as Signals Publisher
    participant Channel as sy_schematic_set

    Client->>API: dispatch(key, sessionKey, actions[])
    API->>API: access.Enforce (update on schematic)
    API->>DB: WithTx(ctx, fn)
    DB->>Writer: Dispatch(ctx, key, sessionKey, actions)
    Writer->>DB: table.NewUpdate().Where(MatchKeys).ChangeErr(...)
    DB->>Writer: ChangeErr callback(schematic)
    Writer->>Writer: check Snapshot → ErrValidation if true
    Writer->>Writer: ReduceAll(state, actions)
    Writer-->>DB: updated schematic
    DB-->>Writer: Exec returns nil (or err)
    Note over Writer,Observer: Observer fires here — before WithTx commits
    Writer->>Observer: Notify(ScopedAction{key, sessionKey, actions})
    Observer->>Signals: Translate(ScopedAction → JSON bytes)
    Signals->>Channel: write sy_schematic_set
    DB-->>API: Commit (may fail after Observer already fired)
    API-->>Client: empty response / error
```

<sub>Reviews (19): Last reviewed commit: ["fixed signal publishers"](https://github.com/synnaxlabs/synnax/commit/003363cd4b629e4e6f587859384b855d4e941139) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30539271)</sub>


client/ts/src/schematic/actions.gen.ts
```diff
@@ -0,0 +1,190 @@
...
+  type: "set_authority",
+  setAuthority: payload,
+});
+
+export const setLegend = (payload: SetLegendPayload): Action => ({
+  type: "set_legend",
+  setLegend: payload,
+});
+
+export const reduce = (state: Schematic, action: Action): Schematic => {
```
sy-nico: `TS` reduce returns `Schematic` (no error channel); Go returns `(Schematic, error)`. The TS handlers can never signal failure even though the Go reducers can. 

Is this asymmetric API contract purposeful? Or something that is planned to be corrected later. If it's the latter, write a ticket and leave an inline comment with the `sy-####`

client/ts/src/schematic/actions.gen.ts
```diff
@@ -0,0 +1,190 @@
...
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+// Code generated by Oracle. DO NOT EDIT.
+import { caseconv, control, record, spatial, zod } from "@synnaxlabs/x";
+import { produce } from "immer";
+import { z } from "zod";
+
+import {
```
sy-nico: Circular Imports:

`actions.gen.ts` imports the `handle*` functions from the hand-written `actions.ts`, while `actions.ts` imports `*Payload` types from this file. Module hoisting works today, but generated code depending on hand-written code via a cycle is fragile. Consider moving the handlers to a third file both can import from, or have the generator accept handlers as a parameter.

core/pkg/service/schematic/actions.go
```diff
@@ -0,0 +1,107 @@
...
+			break
+		}
+	}
+	return state, nil
+}
+
+// Handle appends the node and, if Config is non-nil, seeds the configs map
+// under the node's key.
+func (a AddNode) Handle(state Schematic) (Schematic, error) {
+	state.Nodes = append(state.Nodes, a.Node)
```
sy-nico: There's no duplicate-key guard. Calling `AddNode` twice with the same key leaves the schematic with two nodes sharing a key, and `RemoveNode` will only delete the first. Either upsert (mirror `SetEdge`'s in-place replace) or return a validation error. Same applies to the TS handler in `actions.ts`.

core/pkg/service/schematic/writer.go
sy-nico: Docstrings throughout this file say "log" instead of "schematic" (here, `Create`, `Rename`, `Copy`, `Delete`). This is a pre-existing, but is an easy fix.

sy-nico (changes requested):

client/ts/src/schematic/actions.gen.ts
emilbon99: It's not really asymmetric it's just the two error handling models. Go returns errors as values, where typescript throws exceptions.  To signal an error state in type script you would just throw the error. 

Only exception is `freighter/ts` which should probably be refactored around #2.

client/ts/src/schematic/actions.gen.ts
emilbon99: Very good point. Refactored to use a factory pattern.

core/pkg/service/schematic/actions.go
emilbon99: I refactored this so the action is `SetNode` which updates existing nodes. Decided it was good to follow the same pattern as in `Edge`

core/pkg/service/schematic/writer.go
sy-nico: This line still says `logs` instead of `schematics`

core/pkg/service/schematic/actions_test.go
sy-nico, line 1: Missing coverage for unknown `Type`, nil payload pointer (will panic in `Reduce`), or multiple payload pointers set.

sy-nico (changes requested):
Github is having trouble adding comments. I'll just add publish the two that succeeded

PR #2291, opened 2026-05-02
SY-3833-3: Migrate Schematic Component to Pluto-Owned State

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Third PR in the SY-3833 schematic split series. Wires the schematic action codec from #2290 through to the consumer side: Pluto now owns schematic graph state via flux + dispatch, and Console renders the new self-contained Schematic component.

**Stacked on [#2290](https://github.com/synnaxlabs/synnax/pull/2290).** Targets `sy-3833-2`. Flatten in order.

### What changes

**Oracle codegen — TS actions plugin**

- New `oracle/plugin/ts/types/actions.go` mirrors the Go actions plugin. Generates `actions.gen.ts` per schema-with-actions: payload zod schemas, discriminated union `actionZ`, action constructors, `reduce` / `reduceAll` (the latter using `structuredClone`, no immer dependency).
- Wired into `(*Plugin).Generate`.

**Client SDK**

- `client/ts/src/schematic/actions.gen.ts` (generated) — payload schemas + constructors for the six actions.
- `actions.ts` — hand-written `handle*` mutators called by the generated reducer; semantics mirror the Go `Handle` methods byte-for-byte.
- `scoped.ts` — `scopedActionZ` for decoding `sy_schematic_set` signal frames.
- `client.schematics.dispatch(key, sessionKey, actions)` — calls the `/schematic/dispatch` endpoint added in Part 2.

**Pluto**

- `queries.ts` gains: `useDispatch` (optimistic + rollback + augment edge segments on node moves via `connector.updateSegmentsForPositionChanges`), `useSelectProps`, `useSelectEdge`, `useSelectElementDigests`, `useSelectElementsInfo`, `useSelectElementNames`, `useSelectSnapshot`, `useSelectAuthority`, `useAddNode`, plus a `sy_schematic_set` listener that re-applies remote action batches and self-dedups via session key.
- `Schematic.tsx` becomes self-contained: drops the `create(hooks)` factory. Pulls graph state with `useRetrieve`, dispatches actions for node / edge changes and drops. Stubs `onUndo` / `onRedo`.
- New `node/Node.tsx` extracted from the old factory; reads per-node props via `useSelectProps`, dispatches `setProps` on changes.
- `edge/Edge.tsx` reads via `useSelectProps` and dispatches directly. `EdgeProps` extension dropped.

**Console**

- `console/src/schematic/Schematic.tsx` shrinks 521 → 100 lines. Drops `Base.create(hooks)`, `useUndoableDispatch`, `useSyncComponent`, `useLoadRemote`, `useAddSymbol`, and the in-component graph-state effect handlers. Renders `<Base.Schematic resourceKey={layoutKey}>` directly. Re-exports `HAUL_TYPE` from Pluto for the channel ontology drag-and-drop wiring.

### Out of scope (follow-ups)

- Console schematic slice still carries graph fields (nodes/edges/props/legend.colors/etc.). Those are unread now since the canvas reads through Pluto, but slice gut + v6 simplification + selectors trim + toolbar migration are scoped to a follow-up to keep this PR reviewable.
- Undo / redo for action history (stubbed).
- Off-page-reference cross-schematic navigation (was deleted in v1; will be ported back via `Schematic.useRetrieve` in the toolbar follow-up).
- `pendingUpload` deferred-import path (depends on imex which isn't in v2 Pluto).
- Removing the `SetData` endpoint (kept alive alongside `Dispatch` until console toolbar is fully migrated).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR migrates schematic graph state from the Console Redux slice to a Pluto-owned flux store, completing the third part of the SY-3833 schematic split series. Console now renders a self-contained `Base.Schematic` component with a `resourceKey` prop, dispatching server-side actions for all node/edge mutations.

- **Pluto `queries.ts`**: Adds `useDispatch` with optimistic updates and rollback, `useSelectAllNodes/Edges/Configs`, `useSelectSnapshot`, and an `ACTION_LISTENER` that streams incremental `sy_schematic_set` frames into the flux store with session-key deduplication.
- **Console `Schematic.tsx`**: Shrinks from ~521 to ~100 lines by removing `useUndoableDispatch`, `useSyncComponent`, `useLoadRemote`, and in-component graph-state reducers; a new `useAutoUpload` hook handles the v5\u2192v6 migration upload path.
- **Client SDK `actions.ts`**: Adds hand-written mutators mirroring Go `Handle` methods; `client.ts` routes `rename` through the dispatch endpoint.

<h3>Confidence Score: 2/5</h3>

Multiple active defects across the dispatch/retrieval layer and Console rendering path produce wrong behaviour or crashes in production.

Schematics opened from ontology/link always start in edit mode; useSelect(key) destructure throws TypeError during creation gap; useAutoUpload fires create without awaiting so failures are silent; rename echoes back due to empty session token; Form.use runs before the null guard in IndividualConfig; Node.resolveSpec crashes on unregistered variants.

pluto/src/schematic/queries.ts, pluto/src/schematic/Schematic.tsx, console/src/schematic/Schematic.tsx, console/src/schematic/useUpload.ts, console/src/schematic/services/ontology.tsx, client/ts/src/schematic/client.ts

| Filename | Overview |
|----------|----------|
| console/src/schematic/Schematic.tsx | Major simplification — drops graph state management, delegates to Pluto. Several regressions remain: `useSelect(key)` can return undefined causing a destructure TypeError; double-click fires twice; control status display is broken. |
| pluto/src/schematic/queries.ts | New flux selectors and `useDispatch` with optimistic update/rollback. Silent data loss when `current` is null at dispatch time; `useSelectConfigs` key-vs-index misalignment affects all multi-element callers. |
| pluto/src/schematic/Schematic.tsx | Refactored to self-contained component owning graph state via flux. Stale `handleClearSelection` closure; `onDoubleClick` double-fires; keyboard shortcuts active during modal dialogs. |
| console/src/schematic/useUpload.ts | New hook to handle v5→v6 migration uploads. `create(...)` is not awaited so failed uploads are silent and `clearPendingUpload` fires regardless; `name` dep can trigger duplicate concurrent creates. |
| console/src/schematic/useCreate.ts | New hook wrapping Pluto's `useCreate`. `activeWorkspace`, `update`, `maybeChangeWorkspace`, and `placeLayout` are all absent from the `useCallback` dependency array, causing stale closures on workspace switch. |
| console/src/schematic/services/ontology.tsx | `loadSchematic` and `handleMosaicDrop` both call `Schematic.create({ key, name })` without `editable: false`, so every schematic opened via ontology panel or mosaic drag starts in edit mode. |
| pluto/src/schematic/clipboard.ts | New clipboard hook with copy/paste support. Paste correctly remaps node and edge keys. Pasted edge keys are excluded from the `onPaste` selection callback, so edges are not selected after paste. |
| console/src/schematic/toolbar/Properties.tsx | Migrated to Pluto-owned dispatch. `Form.use` is called before the `config == null` guard, risking a crash when element is absent from flux store. `ELEMENT_REGISTRY` lookup is unguarded. |
| console/src/schematic/types/v6.ts | v6 state removes all graph fields; `pendingUpload` added for migration. Migration correctly guards `remoteCreated` — `pendingUpload` is only set when `!state.remoteCreated`. |
| client/ts/src/schematic/client.ts | `rename` now routes through `dispatch` with a hardcoded empty string session token. The `ACTION_LISTENER` dedup compares against `client.key`, so every rename echoes back to the originating client and is applied twice. |

</details>

<details><summary><h3>Comments Outside Diff (8)</h3></summary>

1. `console/src/schematic/Schematic.tsx`, line 496 ([link](https://github.com/synnaxlabs/synnax/blob/a90c53fe584fa4302f5071fa581e3dc53499ad05/console/src/schematic/Schematic.tsx#L496)) 

   <a href="#">P1</a> **Viewport hard-coded to origin on every mount — saved pan/zoom position lost**

   `viewport={{ position: { x: 0, y: 0 }, zoom: 1 }}` is hard-coded, so every time the schematic tab is opened or the component remounts, the view resets to origin at 1× zoom regardless of where the user had panned/zoomed. Changes are still persisted to Redux via `handleViewportChange`, but the initial value is never restored. `useSelectViewport` already exists in `selectors.ts` and returns the stored viewport; it just isn't called here.

2. `pluto/src/schematic/queries.ts`, line 1098-1116 ([link](https://github.com/synnaxlabs/synnax/blob/8f97ce78049555cd01e789adc453cc091d40ff2c/pluto/src/schematic/queries.ts#L1098-L1116)) 

   <a href="#">P1</a> **Silent data loss when optimistic dispatch races with missing store entry**

   When `current` is `null` (schematic not yet loaded in the flux store), the optimistic `reduceAll` update is skipped and no rollback is pushed, but the network call still fires. After the server writes the action, the `ACTION_LISTENER` self-deduplicates via `changed.sessionKey === client.key` and also returns early — so the local store is never updated. The canvas will silently show stale state until the next full `useRetrieve` cycle. This can happen if `useDispatch` is called (e.g. from a toolbar interaction) before the initial retrieve has resolved.

3. `pluto/src/schematic/Schematic.tsx`, line 771 ([link](https://github.com/synnaxlabs/synnax/blob/8f97ce78049555cd01e789adc453cc091d40ff2c/pluto/src/schematic/Schematic.tsx#L771)) 

   <a href="#">P1</a> **`handleClearSelection` missing `onSelectionChange` in its deps array**

   `handleClearSelection` closes over `onSelectionChange` but the `useCallback` dependency array is empty (`[]`). If the parent re-renders with a new `onSelectionChange` reference (e.g. because a prop changed), the stale callback will be kept — Ctrl+A deselect will call the old handler. Add `onSelectionChange` to the dependency array.

4. `console/src/schematic/Controller.tsx`, line 31-37 ([link](https://github.com/synnaxlabs/synnax/blob/f30b3ab664ba68d61bb027a335e9f6d1de0ad194/console/src/schematic/Controller.tsx#L31-L37)) 

   <a href="#">P1</a> **Edit mode not cleared when control is acquired**

   The old `setControlStatus` reducer set `schematic.editable = false` when `control === "acquired"`, so that any user who was in edit mode was automatically kicked out when another user acquired control. The new reducer only updates `controlStatus` — it no longer touches `editable`. As a result, if User A has the schematic open in edit mode and User B acquires control, User A's diagram stays editable (`canEdit = hasUpdatePermission && editable` remains `true`). The `ToggleEdit` button becomes visually disabled, but the underlying diagram is still accepting node/edge changes, which are then dispatched to the server under User B's control session.

5. `console/src/schematic/slice.ts`, line 248 ([link](https://github.com/synnaxlabs/synnax/blob/5dc9500415f0e0bc290636726187cbcf62e46e0f/console/src/schematic/slice.ts#L248)) 

   <a href="#">P1</a> **Stale `controlStatus` survives persistence — acquired state restored on reload**

   `PERSIST_EXCLUDE` was previously `[purgeSliceState]`, which called `purgeState` to reset `control` to `"released"` before the Redux store was serialised to localStorage. The new export is `export const PERSIST_EXCLUDE: never[] = []`, so the persistence middleware no longer runs any purge. If a user closes the app while `controlStatus === "acquired"` (and `editable === false`, since `setControlStatus` couples them), those values are rehydrated on next launch. The schematic appears locked out of editing and shows "acquired" even though the server no longer holds a control lease — the user would need to know to manually toggle the edit button to recover.

6. `pluto/src/schematic/queries.ts`, line 274-285 ([link](https://github.com/synnaxlabs/synnax/blob/1f50fad977d75126792e162abcd18ff652b7e442/pluto/src/schematic/queries.ts#L274-L285)) 

   <a href="#">P1</a> **Flux store never updated with actual server response after `useCreate` succeeds**

   The optimistic entry is written as `store.schematics.set(data.key, data as schematic.Schematic)`, but on success the actual server response `s` is only returned from `update` — it is never written back to the store. The old code explicitly called `store.schematics.set(s.key, s)` after the server round-trip. As a result, `useSelectAllNodes`/`useSelectAllEdges` will serve the raw optimistic object rather than the canonically-stored schematic for the lifetime of the session. If the server normalises any field, the diagram renders stale data until the next full `useRetrieve` cycle.

7. `pluto/src/schematic/queries.ts`, line 284-287 ([link](https://github.com/synnaxlabs/synnax/blob/5d77e9d82651e70f215515f0436db6760d06d340/pluto/src/schematic/queries.ts#L284-L287)) 

   <a href="#">P1</a> **Server response never written back to flux store after `useCreate` succeeds**

   The optimistic entry (`data as schematic.Schematic`) is placed in the store before the network call, but the actual server response `s` is only surfaced via `afterSuccess` — it is never written back with `store.schematics.set(s.key, s)`. For the rest of the session, `useSelectAllNodes`, `useSelectAllEdges`, and `useSelectConfigs` will serve the raw optimistic object. Any field the server normalises will be silently stale until the next full `useRetrieve` cycle.

8. `console/src/schematic/slice.ts`, line 248 ([link](https://github.com/synnaxlabs/synnax/blob/5d77e9d82651e70f215515f0436db6760d06d340/console/src/schematic/slice.ts#L248)) 

   <a href="#">P1</a> **Stale `controlStatus` and `editable` survive app restart**

   The previous `PERSIST_EXCLUDE = [purgeSliceState]` ran `purgeState` before serialisation, resetting `control` to `"released"` and `selected` to `[]`. The new empty array removes all purging. If a user closes the app while `controlStatus === "acquired"` (which also sets `editable = false`), those values are rehydrated on next launch — the schematic appears locked with no automatic recovery.
</details>

<sub>Reviews (44): Last reviewed commit: ["touch ups"](https://github.com/synnaxlabs/synnax/commit/7c6d140eddfeda9444c7156264d57a5fbd0aa717) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30540616)</sub>


pjdotson (changes requested):
Can you make sure that if this is using changes from another branch, the PR targets that branch instead of `rc`?

core/pkg/service/schematic/pb/schematic.proto
```diff
@@ -14,21 +14,10 @@ syntax = "proto3";
 package service.schematic.pb;
 
 import "google/protobuf/struct.proto";
```
pjdotson, line 16: is there even a reason to worry about generating schematic protos and stuff if it
doesn't use Protobufs yet?

core/pkg/service/schematic/migrations/testdata/v3_shop_flow.migrated.json
```diff
@@ -2,19 +2,6 @@
   "key": "00000000-0000-0000-0000-000000000001",
```
pjdotson: remove this schematic.

core/pkg/service/schematic/migrations/testdata/v5_hardware_workspace.migrated.json
```diff
@@ -2,19 +2,6 @@
   "key": "00000000-0000-0000-0000-000000000001",
```
pjdotson, line 2: why do we need to remove the authority and legend keys from the tests? this is what is
out there on the live JSON files.

core/pkg/service/schematic/actions.gen.go
```diff
@@ -67,42 +70,40 @@ type RemoveEdge struct {
...
 	RemoveEdge      *RemoveEdge      `json:"remove_edge,omitempty" msgpack:"remove_edge,omitempty"`
 	SetConfig       *SetConfig       `json:"set_config,omitempty" msgpack:"set_config,omitempty"`
-	SetAuthority    *SetAuthority    `json:"set_authority,omitempty" msgpack:"set_authority,omitempty"`
-	SetLegend       *SetLegend       `json:"set_legend,omitempty" msgpack:"set_legend,omitempty"`
 }
 
 // Reduce applies the action to the given state by dispatching on Type to the
 // matching payload's Handle method. Unknown action types and actions with a nil
 // payload pointer for the named Type return state unchanged.
 func (a Action) Reduce(state Schematic) (Schematic, error) {
```
pjdotson: is there a reason you chose to make `Reduce` a method on the `Action`? IG it doesn't
really matter but conceptually it is weird to think about it, we don't do anything on
the action and a signature like

```go
func Reduce(Schematic, ...Action) (Schematic, error) {}
```

makes a little more sense to me, and then you wouldn't have separate `Reduce` and
`ReduceAll` as well.

core/pkg/service/schematic/actions.gen.go
```diff
@@ -166,6 +157,11 @@ func ReduceAll(state Schematic, actions []Action) (Schematic, error) {
 	return state, nil
 }
 
+// NewRenameAction wraps a Rename payload in an Action envelope.
+func NewRenameAction(p Rename) Action {
```
pjdotson: personally I think calling these types something like `RenamePayload` instead of just
`Rename` makes a little more sense to me, otherwise we have a struct type which is a
verb which is not something we generally do.

core/pkg/service/schematic/actions.go
```diff
@@ -107,23 +112,47 @@ func (a RemoveEdge) Handle(state Schematic) (Schematic, error) {
...
-	return state, nil
-}
-
-// Handle replaces the schematic's legend configuration with the new value.
-func (a SetLegend) Handle(state Schematic) (Schematic, error) {
-	state.Legend = a.Legend
+	if existing := state.Configs[a.Key]; existing != nil {
+		merged := make(msgpack.EncodedJSON, len(existing)+len(a.Config))
+		for k, v := range existing {
+			merged[k] = v
```
pjdotson: can use `maps.Copy` for a lot of these.

core/pkg/service/schematic/codec.gen.go
```diff
@@ -14,8 +14,6 @@ package schematic
 import (
 	"encoding/json"
 
-	"github.com/synnaxlabs/x/color"
-	"github.com/synnaxlabs/x/control"
 	"github.com/synnaxlabs/x/encoding/msgpack"
```
pjdotson, line 17: doesn't this need to be a migration or the Orc decoding won't work?

client/ts/src/schematic/client.ts
```diff
@@ -100,13 +97,7 @@ export class Client {
   }
 
   async rename(key: Key, name: string): Promise<void> {
```
pjdotson, line 99: should this method be removed?

client/ts/src/schematic/actions.gen.ts
```diff
@@ -70,41 +77,34 @@ export const removeEdgePayloadZ = z.object({
...
-export type SetAuthorityPayload = z.infer<typeof setAuthorityPayloadZ>;
-
-/** SetLegend replaces the schematic's control-legend overlay configuration. */
-export const setLegendPayloadZ = z.object({
-  legend: legendZ,
-});
-
-export type SetLegendPayload = z.infer<typeof setLegendPayloadZ>;
-
 export const ACTION_TYPES = {
```
pjdotson: what is the point of this?

integration/console/case.py
```diff
@@ -58,6 +58,20 @@ def setup(self) -> None:
         )
         self.page = self.context.new_page()
 
+        # Forward browser console messages to the test log for debugging.
```
pjdotson: remove

pjdotson: or properly type

pluto/src/log/aether/telem/sources.ts
```diff
@@ -145,16 +148,24 @@ export class StreamMultiChannelLog
         let pushed = 0;
         for (const [key, chMeta] of this.channelMeta) {
           const allocated = res.get(key);
-          const isJSON = chMeta.dataType.equals(DataType.JSON);
+          const isString = chMeta.dataType.equals(DataType.STRING);
```
pjdotson, line 151: why are these changes in this PR?

pjdotson (changes requested):
There are a bunch of changes from other PRs in this one. Why are they there, it makes this hard to review? Left some comments on Go code but the TS code needs to have those changes taken out of before I will review again.

core/pkg/service/schematic/migrations/testdata/v5_hardware_workspace.migrated.json
emilbon99: This represents the migrated state. We agreed to keep the legend and authority out of it for now and keep it as session state. Do you want it back?

core/pkg/service/schematic/pb/schematic.proto
emilbon99: I mean oracle generates these things by default. I think it's good that it does. Ensures that everything is in sync.

core/pkg/service/schematic/actions.gen.go
emilbon99: Refactored

core/pkg/service/schematic/codec.gen.go
emilbon99: Yeah, the migration was already merged in a separate PR, we just modified the `migrate.go` file. All is good here, and will get tested more when everything is in.

client/ts/src/schematic/client.ts
emilbon99: I mean I wanted to keep client library API consistency until we make a longer term decision on how we want it to look.

client/ts/src/schematic/actions.gen.ts
emilbon99: Removed from the generator template. Good catch.

client/ts/src/schematic/actions.gen.ts
```diff
@@ -70,41 +77,23 @@ export const removeEdgePayloadZ = z.object({
...
-  remove_node: "remove_node",
-  set_edge: "set_edge",
-  remove_edge: "remove_edge",
-  set_config: "set_config",
-  set_authority: "set_authority",
-  set_legend: "set_legend",
-} as const;
-
 export const actionZ = z.discriminatedUnion("type", [
+  z.object({ type: z.literal("rename"), rename: renamePayloadZ }),
```
pjdotson, line 96: just curious, is there a reason for the action type to look like this:

```json
{
    "type": "rename",
    "rename": {"name": "new name"}
}
```

instead of this:

```json
{
    "type": "rename",
    "payload": {"name": "new name"}
}
```

integration/console/case.py
```diff
@@ -58,6 +60,19 @@ def setup(self) -> None:
         )
         self.page = self.context.new_page()
 
+        def _forward_console(msg: ConsoleMessage) -> None:
+            try:
+                text = msg.text
+            except Exception:
```
pjdotson: why would this ever run into an `Exception`?

integration/console/case.py
```diff
@@ -58,6 +60,19 @@ def setup(self) -> None:
...
+                text = "<unprintable>"
+            self.log(f"[browser:{msg.type}] {text}")
+
+        def _forward_page_error(exc: Error) -> None:
+            self.log(f"[browser:pageerror] {exc}")
+
+        self.page.on("console", _forward_console)
+        self.page.on("pageerror", _forward_page_error)
```
pjdotson: should we be failing tests if these happen?

core/pkg/service/schematic/actions.gen.go
```diff
@@ -12,201 +12,183 @@
...
+// reduced state.
+func Reduce(state Schematic, actions ...Action) (Schematic, error) {
 	var err error
 	for _, a := range actions {
-		state, err = a.Reduce(state)
+		switch a.Type {
+		case ActionTypeRename:
+			if a.Rename == nil {
+				continue
+			}
```
pjdotson, line 110: shouldn't these be errors? if you call rename but don't pass the proper rename payload?

core/pkg/service/schematic/actions.go
```diff
@@ -23,8 +25,14 @@ type ScopedAction struct {
...
 
+// Handle replaces the schematic's name.
+func (a RenamePayload) Handle(state Schematic) (Schematic, error) {
+	state.Name = a.Name
+	return state, nil
+}
+
 // Handle moves the named node to the given position. No-op if no node matches.
-func (a SetNodePosition) Handle(state Schematic) (Schematic, error) {
+func (a SetNodePositionPayload) Handle(state Schematic) (Schematic, error) {
```
pjdotson: nit: receiver should be `p` and not `a`

pluto/src/triggers/Provider.tsx
```diff
@@ -98,17 +98,26 @@ export const Provider = ({
...
 
   // The current trigger.
   const [, setCurr] = useStateRef<RefState>({ ...ZERO_REF_STATE });
 
   const updateListeners = useCallback((state: RefState, target: HTMLElement): void => {
     const next = state.next.length > 0 ? [state.next] : [];
     const prev = state.prev.length > 0 ? [state.prev] : [];
-    const event = { target, next, prev, cursor: cursor.current };
-    registry.current.forEach((_, f) => f(event));
+    let minPriority = Number.NEGATIVE_INFINITY;
```
pjdotson: use `-Infinity` instead

pluto/src/schematic/clipboard.spec.tsx
```diff
@@ -0,0 +1,259 @@
...
+  type PropsWithChildren,
+  type ReactElement,
+} from "react";
+import { describe, expect, it, vi } from "vitest";
+
+import { Errors } from "@/errors";
+import { Schematic } from "@/schematic";
+import { createAsyncSynnaxWrapper } from "@/testutil/Synnax";
+
+/* eslint-disable @typescript-eslint/unbound-method */
```
pjdotson, line 27: why is this okay? we should be commenting when we disable linters or type checkers the
reason why.

pluto/src/schematic/Schematic.tsx
```diff
@@ -9,97 +9,149 @@
...
 }
+const AUTO_RENDER_INTERVAL = TimeSpan.seconds(1).milliseconds;
+const DRAG_HANDLE_SELECTOR = `.${Node.DRAG_HANDLE_CLASS}`;
 
-export const elementConfigZ = z.discriminatedUnion("variant", [
-  ...Node.configZ.options,
-  ...Edge.configZ.options,
-]);
-export type ElementConfig = z.infer<typeof elementConfigZ>;
+export const Schematic = ({
```
pjdotson, line 51: is there not any value we can get from adding a `Schematic.spec.tsx` here? even if some
of the React flow stuff doesn't work properly.

console/src/schematic/services/palette.tsx
```diff
@@ -15,10 +15,8 @@ import { Palette } from "@/palette";
 import { Schematic } from "@/schematic";
 
 const CreateCommand: Palette.Command = ({ placeLayout, ...listProps }) => {
-  const handleSelect = useCallback(
-    () => placeLayout(Schematic.create()),
-    [placeLayout],
-  );
+  const create = Schematic.useCreate({});
+  const handleSelect = useCallback(() => create(), [create]);
```
pjdotson, line 19: useless function definition, cant you just pass `create` to `onSelect`?

console/src/schematic/Selectable.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { useCallback } from "react";
+
+import { LAYOUT_TYPE } from "@/schematic/layout";
+import { useCreate } from "@/schematic/useCreate";
+import { Selector } from "@/selector";
+
+export const Selectable: Selector.Selectable = ({ layoutKey: key, onPlace }) => {
+  const hasCreatePermission = Access.useCreateGranted(schematic.TYPE_ONTOLOGY_ID);
+  const create = useCreate({});
+  const handleClick = useCallback(() => create({ key }), [onPlace, key]);
```
pjdotson: why is `onPlace` a dependency, shouldn't `create` be a dependency?

pjdotson (changes requested):

integration/console/case.py
emilbon99: I"m just going to remove this.

integration/console/case.py
emilbon99: I'm just going to remove this.

client/ts/src/schematic/actions.gen.ts
emilbon99: Yes, because then you couldn't strongly type the payload in go. It's a tagged union, a lot like the `types.Type` in arc. 

This is really the only way to do typed struct unions in go. I guess we could add a translation layer in typescript, but that doesn't seem worth the pain.

console/src/schematic/services/palette.tsx
emilbon99: Because `onSelect` passes arguments that don't match the shape of the `create` arguments.

core/pkg/service/schematic/actions.gen.go
emilbon99: I made it return an error.

pluto/src/schematic/Schematic.tsx
emilbon99: I think there are probably tests you can write that can thoroughly exercise most code paths. I would like to do it in a separate issue: https://linear.app/synnax/issue/SY-4186/write-improved-schematic-unit-tests

pjdotson (approved):

PR #2295, opened 2026-05-06
SY-4150: Enable writing time.now() to `timestamp` and `int64` - Flow and WASM Pairity

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4150](https://linear.app/synnax/issue/SY-4150/enable-writing-i64-ns-to-timestamp-channels)

## Description

`time.now()` and `time.now{}` now write into either an `i64` or `timestamp` channel in any combination. Previously only the unit-matching pair worked.

Two layers were rejecting the cross-unit case:

- **WASM**: `DeriveTypeSuffix` produced `channel.write_i64 ns` for timestamp channels; host bindings only register `write_i64`. Strip the unit before suffix derivation, and add the `Int64T` <-> `TimeStampT` equivalence fix to the Go core codec and C++ client codec so the resulting `i64` series encodes for a `timestamp` channel.
- **Analyzer**: The IR-edge type check rejected `i64 ns -> i64` because `types.Equal` is unit-strict. Add `types.UnitsAssignable(a, b *Unit)`, a small predicate that treats a missing unit as a wildcard but still rejects two non-nil mismatched units (e.g. `f32 psi` vs `f32 bar`), and call it  from the one `Check` site that validates flow-edge compatibility, alongside an inline `Kind` equality guard. Constraint-solver callsites stay strict.

All three layers (analyzer, codec, writer validator) now agree that `i64` and `timestamp` are wire-compatible, while genuine unit mismatches still surface a `type mismatch` diagnostic.

image

image

## Basic Readiness

<h3>Greptile Summary</h3>

This PR makes `time.now()` and `time.now{}` writable to both `timestamp` and `int64` channels interchangeably, fixing a mismatch between how the Arc analyzer, WASM codegen, and the frame codecs each validated the `i64 ns ↔ i64` type pair.

- **Analyzer (`check.go`)**: Flow-edge type check relaxed from `types.Equal` to a Kind + `UnitsAssignable` guard, where a nil unit on either side acts as a wildcard — preserving `f32 psi → f32 bar` rejection while allowing `timestamp (i64 ns) ↔ i64`.
- **WASM codegen (`derive.go`)**: `suffixForType` strips units before producing the binding name, so a timestamp-typed channel correctly resolves to `write_i64` rather than the non-existent `write_i64_ns`.
- **Server Go codec + C++ client codec**: An `isEquivalent` guard treats `Int64T` and `TimeStampT` as wire-compatible, keeping the encoder from rejecting cross-type series.

All three enforcement layers (analyzer, WASM codegen, frame codec) are updated symmetrically, genuine unit mismatches (e.g., `f32 psi → f32 bar`) still surface errors, and the change is guarded by unit, integration, and end-to-end tests.

The three code paths that previously disagreed on `i64 ns ↔ i64` compatibility are now aligned. The nil-unit wildcard semantics introduced by `UnitsAssignable` are intentional and documented, non-nil unit mismatches are still rejected, and the regression suite explicitly covers the boundary cases. No unguarded data-loss or silent encoding path was found.

| Filename | Overview |
|----------|----------|
| arc/go/analyzer/types/check.go | Flow-edge compatibility check replaced from strict `types.Equal` to `t1.Kind != t2.Kind || !types.UnitsAssignable(...)`, relaxing nil-unit handling; constraint-solver callsites are left strict. |
| arc/go/types/type.go | Adds `UnitsAssignable(a, b *Unit) bool` — nil on either side acts as a wildcard (matching WASM/Cesium wire-type parity), while two non-nil mismatched units still return false. |
| arc/go/compiler/resolve/derive.go | Introduces `suffixForType` which strips the unit before calling `String()`, ensuring `timestamp` (i64 ns) resolves to the `write_i64` WASM host binding rather than a non-existent `write_i64_ns`. |
| core/pkg/distribution/framer/codec/codec.go | Adds `isEquivalent` guard so that `Int64T` and `TimeStampT` are treated as wire-compatible in the server-side frame encoder, mirroring the same change in the C++ client. |
| client/cpp/framer/codec.cpp | Symmetric `is_equivalent` check for `INT64_T` / `TIMESTAMP_T` interop added to the C++ client encoder; correct and consistent with the Go-side change. |
| integration/tests/arc/stl_time.py | Expands the `time.now` integration test to cover both WASM and flow modes writing to both `TIMESTAMP` and `INT64` channels; drift validation retained for the timestamp case. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["Arc source\ntime.now() → int_out / ts_out"] --> B["Analyzer\narc/go/analyzer/types/check.go"]
    B -- "KindI64 match\nUnitsAssignable(ns, nil) = true" --> C["Analysis passes ✓"]
    A --> D["Compiler\narc/go/compiler/resolve/derive.go"]
    D -- "suffixForType strips unit\ntimestamp → i64 suffix" --> E["write_i64 WASM binding ✓"]
    C --> F["Runtime write"]
    E --> F
    F --> G["Go codec\ncore/.../framer/codec/codec.go"]
    G -- "isEquivalent:\nInt64T ↔ TimeStampT" --> H["Encode succeeds ✓"]
    F --> I["C++ codec\nclient/cpp/framer/codec.cpp"]
    I -- "is_equivalent:\nINT64_T ↔ TIMESTAMP_T" --> J["Encode succeeds ✓"]
    H --> K[("Channel store\nINT64 or TIMESTAMP")]
    J --> K
```

<sub>Reviews (3): Last reviewed commit: ["SY-4150: add unit mismach tests"](https://github.com/synnaxlabs/synnax/commit/c4307631cb9f5ba818e037a3233270ad64837325) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30942967)</sub>

### DIFF
--- a/arc/go/analyzer/flow/flow_test.go
+++ b/arc/go/analyzer/flow/flow_test.go
@@ -328,6 +328,20 @@ producer{} -> consumer{}`))
 				Expect((*ctx.Diagnostics)[0].Message).To(ContainSubstring("is not equal to argument type"))
 			})
 
+			It("Should detect unit mismatch between func output and next func input", func(bCtx SpecContext) {
+				ast := MustSucceed(parser.Parse(`
+func producer() f32 psi {
+    return 5psi
+}
+func consumer(v f32 bar) {}
+producer{} -> consumer{}`))
+				ctx := context.CreateRoot(bCtx, ast, resolver)
+				analyzer.AnalyzeProgram(ctx)
+				Expect(ctx.Diagnostics.Ok()).To(BeFalse())
+				Expect((*ctx.Diagnostics)[0].Message).To(ContainSubstring("return type"))
+				Expect((*ctx.Diagnostics)[0].Message).To(ContainSubstring("is not equal to argument type"))
+			})
+
 			It("Should detect when func with multiple params is used without routing table", func(bCtx SpecContext) {
 				ast := MustSucceed(parser.Parse(`
 func source() u8 {
--- a/arc/go/analyzer/types/check.go
+++ b/arc/go/analyzer/types/check.go
@@ -69,7 +69,7 @@ func Check(
 		return Check(cs, t1.Unwrap(), t2.Unwrap(), source, reason+" (element types)")
 	}
 
-	if !types.Equal(t1, t2) {
+	if !types.Equivalent(t1, t2) {
 		return newTypeMismatchError(t1, t2, reason)
 	}
 	return nil
--- a/arc/go/analyzer/types/check.go
+++ b/arc/go/analyzer/types/check.go
@@ -69,7 +69,7 @@ func Check(
 		return Check(cs, t1.Unwrap(), t2.Unwrap(), source, reason+" (element types)")
 	}
 
-	if !types.Equivalent(t1, t2) {
+	if t1.Kind != t2.Kind || !types.UnitsAssignable(t1.Unit, t2.Unit) {
 		return newTypeMismatchError(t1, t2, reason)
 	}
 	return nil
--- a/arc/go/analyzer/types/check_test.go
+++ b/arc/go/analyzer/types/check_test.go
@@ -132,11 +132,12 @@ var _ = Describe("Check", func() {
 	})
 
 	Describe("Unit Compatibility", func() {
-		It("should succeed for same kind with different units", func() {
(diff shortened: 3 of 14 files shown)


emilbon99 (approved):
This PR LGTM, although I think we should more carefully consider how units actually work in Arc relatively soon.

PR #2298, opened 2026-05-06
SY-3880: Arc String Conversion Support

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-3880](https://linear.app/synnax/issue/SY-3880)

## Description

Adds support for converting numeric primitives (i8-i64, u8-u64, f32, f64) to strings via Arc's existing typecast syntax (e.g., `str(42)`, `str(3.14)`). Floats use shortest round-trippable formatting matching Go's `strconv.FormatFloat(_, 'g', -1, bitSize)`; the reverse direction (str to numeric) remains rejected.

`str()` also works as a flow stage (`channel -> str -> log`). Both Go and C++ WASM runtimes now materialize string-typed outputs into a fresh series at the end of each tick rather than attempting to resize variable-density buffers in place, which previously caused a runtime trap.

Conversion routing lives in a single `Resolver.EmitNumericToString` dispatcher so the planned f-string feature can reuse it without duplicating the source-type to host-fn switch.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds `str()` typecast support that converts numeric primitives (i8–i64, u8–u64, f32, f64) to their decimal string representation, and fixes a runtime trap that occurred when attempting to resize variable-density string buffers in place during WASM tick evaluation.

- **Conversion routing** is centralised in `Resolver.EmitNumericToString`, dispatching to one of six new `string.from_*` host functions (backed by `strconv.FormatFloat` with the correct `bitSize` on the Go side and `std::to_chars(…, std::chars_format::general)` on the C++ side). Float formatting matches Go's `strconv.FormatFloat(_, 'g', -1, bitSize)` contract.
- **Both runtimes (Go WASM and C++ WASM)** now accumulate string outputs into a temporary slice per tick and materialise a fresh series at the end, instead of calling `Resize` on a variable-density buffer (which would panic or trap). The `str()` cast also works as an inline flow stage (`channel -> str -> log`).

The change is safe to merge. The numeric-to-string conversion path is well-isolated, the variable-density buffer fix correctly avoids the Resize trap in both runtimes, and test coverage spans unit, integration, and end-to-end layers across all numeric source types.

The core logic — accumulating string handles into a temporary slice per tick and materialising a fresh series at the end — is symmetric between Go and C++ runtimes and verified by multi-sample tests. The Go Data-only assignment is correct because telem.Series.Len() for variable-density types always recomputes from Data (Resize panics on variable-density, so cachedLength is never set persistently for string series). Float formatting uses the right bitSize (32 vs 64) to match Go's strconv contract. No logic errors or unsafe paths were found.

| Filename | Overview |
|----------|----------|
| arc/go/compiler/resolve/emit.go | New EmitNumericToString and EmitFixedCall functions; type-switch routing from all 10 numeric kinds to the correct from_* host fn looks correct. |
| arc/go/stl/strings/string.go | Six new host functions added; FormatFloat bitSize is correctly 32 for f32 and 64 for f64. SymbolResolver entries match host function signatures. |
| arc/go/stl/wasm/node.go | String outputs now accumulate into a []string slice; only out.Data is replaced at tick end. Verified safe: telem.Series.Len() for variable-density types recomputes directly from Data (cachedLength is never set persistently via Resize since Resize panics on variable-density). DPanicf on unregistered handle miss is a good defensive addition. |
| arc/cpp/stl/str/str.h | format_float correctly handles NaN/±Inf and now checks the to_chars error code (returning empty string on failure). Six new from_* linker bindings added. |
| arc/cpp/runtime/wasm/node.h | String outputs accumulate into a lazy-allocated vector-of-vectors and are materialised as a full Series assignment at tick end, correctly bypassing the Resize-on-variable-density trap. |
| arc/go/compiler/expression/cast.go | Hint suppression for str() targets prevents numeric literals from inheriting the string type; extractType extended to recognise STR primitive; EmitCast short-circuits to EmitNumericToString for string targets. |
| arc/go/str_cast_test.go | Comprehensive end-to-end tests covering literals, channel reads, flow-stage usage, and concat patterns across all numeric types. Ghost-precision f32 cases included. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["str(x) in Arc source"] --> B{Analyzer isValidCast}
    B -->|"numeric source"| C["Compiler: compileTypeCast\nhint suppressed for str target"]
    B -->|"str source"| D["Rejected: cannot cast"]
    C --> E["EmitCast → EmitNumericToString"]
    E --> F{source kind}
    F -->|"i8/i16/i32"| G["string.from_i32"]
    F -->|"u8/u16/u32"| H["string.from_u32"]
    F -->|"i64"| I["string.from_i64"]
    F -->|"u64"| J["string.from_u64"]
    F -->|"f32"| K["string.from_f32"]
    F -->|"f64"| L["string.from_f64"]
    G & H & I & J --> M["Host fn: FormatInt/Uint → string handle uint32"]
    K & L --> N["Host fn: FormatFloat 'g' -1 bitSize → string handle uint32"]
    M & N --> O["WASM stack: i32 handle"]
    O --> P{Runtime: string output?}
    P -->|yes| Q["Accumulate into stringResults per tick"]
    P -->|no| R["setValueAt numeric output"]
    Q --> S["End of tick: materialise fresh Series from accumulated strings"]
    R --> T["Resize to sample count"]
```

<sub>Reviews (4): Last reviewed commit: ["SY-3380: Implement feedback"](https://github.com/synnaxlabs/synnax/commit/f739706dfbc9ebaddfc2fb38921df7ff0c82f779) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31089983)</sub>

### DIFF
--- a/arc/cpp/runtime/wasm/node.h
+++ b/arc/cpp/runtime/wasm/node.h
@@ -31,6 +31,7 @@ class Node : public node::Node {
     std::vector<Module::Function::Result> results;
     std::vector<int> offsets;
     std::vector<bool> string_inputs;
+    std::vector<bool> string_outputs;
     std::shared_ptr<stl::str::State> str_state;
     bool initialized = false;
     bool is_entry_node = false;
@@ -58,6 +59,9 @@ public:
         this->string_inputs.resize(node.inputs.size());
         for (size_t i = 0; i < node.inputs.size(); i++)
             this->string_inputs[i] = node.inputs[i].type.kind == types::Kind::String;
+        this->string_outputs.resize(node.outputs.size());
+        for (size_t i = 0; i < node.outputs.size(); i++)
+            this->string_outputs[i] = node.outputs[i].type.kind == types::Kind::String;
     }
 
     x::errors::Error next(node::Context &ctx) override {
@@ -84,8 +88,19 @@ public:
         for (auto &offset: this->offsets)
             offset = 0;
 
+        // String outputs are variable-density and cannot be resize'd. Their
+        // data buffer is built once at the end of the loop from accumulated
+        // strings. Numeric outputs are pre-sized here so set() can do
+        // fixed-stride writes per sample.
+        std::vector<std::vector<std::string>> string_results;
         for (size_t i = 0; i < this->ir.outputs.size(); i++) {
-            this->state.output(i)->resize(max_length);
+            if (this->string_outputs[i]) {
+                if (string_results.empty())
+                    string_results.resize(this->ir.outputs.size());
+                string_results[i].reserve(max_length);
+            } else {
+                this->state.output(i)->resize(max_length);
+            }
             this->state.output_time(i)->resize(max_length);
         }
 
@@ -131,7 +146,15 @@ public:
             for (size_t j = 0; j < results.size(); j++) {
                 auto [value, changed] = results[j];
                 if (!changed) continue;
-                this->state.output(j)->set(this->offsets[j], value);
+                if (this->string_outputs[j]) {
+                    // WASM returned an i32 string handle; materialize it
+                    // to its actual string value, mirroring the input-side
+                    // conversion above.
+                    const auto handle = static_cast<uint32_t>(std::get<int32_t>(value));
+                    string_results[j].push_back(this->str_state->get(handle));
+                } else {
+                    this->state.output(j)->set(this->offsets[j], value);
+                }
                 this->state.output_time(j)->set(this->offsets[j], ts);
                 this->offsets[j]++;
             }
@@ -139,7 +162,13 @@ public:
 
         for (size_t j = 0; j < this->ir.outputs.size(); j++) {
             const auto off = this->offsets[j];
-            this->state.output(j)->resize(off);
+            if (this->string_outputs[j]) {
+                *this->state.output(
+                    j
+                ) = x::telem::Series(string_results[j], x::telem::STRING_T);
+            } else {
+                this->state.output(j)->resize(off);
+            }
             this->state.output_time(j)->resize(off);
             if (off > 0) ctx.mark_changed(j);
         }
--- a/arc/cpp/runtime/wasm/node.h
+++ b/arc/cpp/runtime/wasm/node.h
@@ -162,13 +162,11 @@ public:
 
         for (size_t j = 0; j < this->ir.outputs.size(); j++) {
             const auto off = this->offsets[j];
-            if (this->string_outputs[j]) {
-                *this->state.output(
-                    j
-                ) = x::telem::Series(string_results[j], x::telem::STRING_T);
-            } else {
-                this->state.output(j)->resize(off);
-            }
+            auto &out = this->state.output(j);
+            if (this->string_outputs[j])
+                *out = x::telem::Series(string_results[j], x::telem::STRING_T);
+            else
+                out->resize(off);
             this->state.output_time(j)->resize(off);
             if (off > 0) ctx.mark_changed(j);
         }
--- a/arc/go/compiler/resolve/emit.go
+++ b/arc/go/compiler/resolve/emit.go
@@ -10,6 +10,8 @@
 package resolve
 
 import (
+	"context"
+
 	"github.com/synnaxlabs/arc/compiler/wasm"
 	"github.com/synnaxlabs/arc/types"
 	"github.com/synnaxlabs/x/errors"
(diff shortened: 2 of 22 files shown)


arc/cpp/runtime/wasm/node.h
```diff
@@ -131,15 +146,29 @@ class Node : public node::Node {
...
                 this->state.output_time(j)->set(this->offsets[j], ts);
                 this->offsets[j]++;
             }
         }
 
         for (size_t j = 0; j < this->ir.outputs.size(); j++) {
             const auto off = this->offsets[j];
-            this->state.output(j)->resize(off);
+            if (this->string_outputs[j]) {
+                *this->state.output(
```
emilbon99: The formatting here is weird. See if you can coerce the formatter to be nicer. Also you can get rid of the `{}` since there is a single expression.

arc/go/compiler/resolve/emit.go
```diff
@@ -301,6 +301,83 @@ func (r *Resolver) EmitStringLen(w *wasm.Writer, wID int) {
 	r.EmitCall(w, wID, "string.len", ct)
 }
 
+// EmitStringFromI32 emits a call to string.from_i32.
+func (r *Resolver) EmitStringFromI32(w *wasm.Writer, wID int) {
```
emilbon99: Do all of these sub-functions need to be public? Is there anywhere we call anything other than `EmitNumericToString`?

arc/go/compiler/resolve/emit.go
```diff
@@ -301,6 +301,83 @@ func (r *Resolver) EmitStringLen(w *wasm.Writer, wID int) {
...
+	ct := types.Function(types.FunctionProperties{
+		Inputs:  types.Params{{Type: types.I32()}},
+		Outputs: types.Params{{Type: types.I32()}},
+	})
+	r.EmitCall(w, wID, "string.from_i32", ct)
+}
+
+// EmitStringFromU32 emits a call to string.from_u32.
+func (r *Resolver) EmitStringFromU32(w *wasm.Writer, wID int) {
```
emilbon99: Seems like a lot of duplicated code across these functions. output type is always the same. ONly thing that varies is method name and input type. Could always pass the input type as a parameter then do `fmt.Sprintf("string.from_%s", t)`. Would cut out vast majority of changes in this file.

arc/go/stl/strings/string.go
```diff
@@ -62,6 +63,66 @@ var SymbolResolver = &symbol.ModuleResolver{
 				Outputs: types.Params{{Name: ir.DefaultOutputParam, Type: types.I64()}},
 			}),
 		},
+		"from_i32": {
```
emilbon99, line 66: Kind of interesting that the type signatures are duplicated here and in the wasm writer. Nothing super wrong with it but I did notice it.

arc/go/stl/wasm/node.go
```diff
@@ -122,8 +123,20 @@ func (n *nodeImpl) Next(ctx node.Context) {
 	for j := range n.offsets {
 		n.offsets[j] = 0
 	}
+	// String outputs are variable-density and cannot be Resize'd . Their
```
emilbon99: `Resize'd` -> `resized`.

arc/go/stl/wasm/node.go
```diff
@@ -186,14 +199,38 @@ func (n *nodeImpl) Next(ctx node.Context) {
...
 		for j, value := range res {
 			if value.Changed {
-				setValueAt(*n.Output(j), n.offsets[j], value.Value)
+				if n.stringOutputs[j] {
+					// WASM returned an i32 string handle; materialize it
+					// to its actual string value, mirroring the input-side
+					// conversion above.
+					s, ok := n.strings.Get(uint32(value.Value))
+					if !ok {
+						ctx.ReportError(errors.Newf(
```
emilbon99: Is this a user error or an implementation error in the arc language itself? If it's a programming error we should probably `DPanic`.

emilbon99 (changes requested):
Looks pretty good overall. Comments are mostly nitpicks.

arc/cpp/runtime/wasm/node.h
sy-nico: Made an attempt

arc/go/compiler/resolve/emit.go
sy-nico: Collapsed them all into `EmitNumericToString` since nothing else called them.

arc/go/compiler/resolve/emit.go
sy-nico: Folded into a single `EmitNumericToString` with a switch picking the input type and suffix, then `"string.from_" + suffix` since it's marginally faster than `fmt.Sprintf()`

arc/go/stl/strings/string.go
sy-nico: Added `EmitFixedCall` that looks up the signature from the `SymbolResolver` by name, so `string.go` is the sole source of truth now.

arc/go/stl/wasm/node.go
sy-nico: Changed to panic

emilbon99 (approved):

PR #2299, opened 2026-05-07
SY-4153: Fix Float32 Channel Value Display Precision

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4153](https://linear.app/synnax/issue/SY-4153/f32-values-display-with-extraneous-precision)

## Description

Float32 channel values rendered with f64-widening artifacts (e.g. `1.2339999675750732` instead of `1.234`) because JavaScript widens every numeric read to f64 before stringification. Adds a `stringifyFloat32` helper that returns the shortest decimal round-tripping through f32, and applies it at the surfaces that consume raw channel samples: the log view, `StreamChannelValue` (schematic value/gauge, `Value` component, table cells), and `findByXValue` (line plot tooltip, measure tool).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes float32 channel values displaying with f64-widening artifacts (e.g. `1.2339999675750732` instead of `1.234`) by introducing a `stringifyFloat32` helper in `series.ts` and routing `DataType.FLOAT32` through it in `asString`. Because `sources.ts` already uses `buf.asString(i, true)` to produce log entries, the fix propagates uniformly to all display surfaces without touching each call site.

- **`stringifyFloat32`** probes decimal precision 1–9 until a round-trip through `Math.fround` succeeds, returning `parsed.toString()` — a JS analogue of Go's `strconv.FormatFloat(v, 'g', -1, 32)`. IEEE 754 guarantees a match at ≤9 significant digits for any finite float32.
- **Tests** cover typical values, integer-valued floats, subnormals, large values, non-finite values, idempotency, and integration through `StreamMultiChannelLog`.

Safe to merge — change is narrowly scoped to DataType.FLOAT32 in asString, all other types are unaffected, and the new helper is deterministic with no side effects.

The algorithm is mathematically sound (IEEE 754 guarantees convergence within 9 significant digits for any finite float32), the fallback is unreachable dead code rather than a misfire, all display surfaces use asString as the single exit point, and the test suite covers normal, edge, and non-finite cases as well as idempotency.

| Filename | Overview |
|----------|----------|
| x/ts/src/telem/series.ts | Adds stringifyFloat32 helper and routes DataType.FLOAT32 through it in asString. Algorithm is correct: IEEE 754 guarantees toPrecision(9) always round-trips for any finite float32. |
| x/ts/src/telem/series.spec.ts | Adds six new test cases covering typical values, integer-valued floats, subnormals, large values, non-finite values, and idempotency under f32 round-trip. |
| pluto/src/log/aether/telem/sources.spec.ts | Adds mock channels for INT64 and JSON, plus three integration tests exercising the f32 fix and existing INT64/JSON stringification through the log view pipeline. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["series.asString(index)"] --> B{dataType.isVariable?}
    B -- yes --> C["atVariable() JSON/string"]
    B -- no --> D{DataType.UUID?}
    D -- yes --> E["atUUID()"]
    D -- no --> F["v = this.at(index)"]
    F --> G{DataType.FLOAT32?}
    G -- yes --> H["stringifyFloat32(v)"]
    G -- no --> I["String(v)"]
    H --> J{Number.isFinite?}
    J -- no --> K["f32.toString() NaN / Infinity"]
    J -- yes --> L["Loop p = 1..9 via toPrecision"]
    L --> M{fround roundtrip matches?}
    M -- yes --> N["return parsed.toString()"]
    M -- no --> L
```

<sub>Reviews (9): Last reviewed commit: ["SY-4153: Make stringifyFloat32 module-pr..."](https://github.com/synnaxlabs/synnax/commit/b092d23bc37404e1796ed7f050280638df75c9de) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31095805)</sub>


pluto/src/log/aether/telem/sources.ts
```diff
@@ -146,13 +147,18 @@ export class StreamMultiChannelLog
...
             for (let i = start; i < buf.length; i++) {
               const raw = buf.at(i, true);
               this.entries.push({
                 channelKey: chMeta.key,
                 timestamp: now,
-                value: isJSON ? JSON.stringify(raw) : String(raw),
+                value: isJSON
+                  ? JSON.stringify(raw)
+                  : isF32
+                    ? stringifyFloat32(raw as number)
```
pjdotson: A better pattern would be to add an iterator to the series class that allows you to
iterate through it as strings or get a particular value as a string, rather than callers
having to worry about things like `JSON.stringify` or `stringifyFloat32`

pluto/src/vis/line/aether/line.ts
```diff
@@ -363,7 +364,8 @@ export class Line extends aether.Leaf<typeof stateZ, InternalState> {
     if (ySeries == null) return result;
 
     const alignmentDiff = Number(ySeries.alignment - xSeries.alignment);
-    result.value.y = Number(ySeries.at(index - alignmentDiff));
+    const rawY = Number(ySeries.at(index - alignmentDiff));
+    result.value.y = cleanFloat32(rawY, ySeries.dataType);
```
pjdotson: again, instead of doing something like `cleanFloat32`, it makes more sense for `at` to
return the lower-precision number.

x/ts/src/telem/telem.ts
```diff
@@ -2129,6 +2129,22 @@ export class DataType
...
+    const parsed = parseFloat(f32.toPrecision(p));
+    if (Math.fround(parsed) === f32) return parsed.toString();
+  }
+  return f32.toString();
+};
+
+/** If dataType is FLOAT32, returns value rounded through the shortest
+ *  f32-roundtrippable decimal. Otherwise returns value unchanged. */
+export const cleanFloat32 = (value: number, dataType: DataType): number =>
+  dataType.equals(DataType.FLOAT32) ? parseFloat(stringifyFloat32(value)) : value;
```
pjdotson: note that there may be performance issues with converting a number to a string and back every
time this is called. There should probably just be a function (either builtin or one we
write) that directly converts a `number` to a floating-point with only 32-bit precision.

pjdotson (changes requested):
See comments. Should probably just make this so 'stringifying' / 'cleaning' are handled directly through series iterators.

pluto/src/log/aether/telem/sources.ts
sy-nico: Already implemented `Series.asString` in  PR #2300. This callsite will use it once that merges.

pluto/src/vis/line/aether/line.ts
sy-nico: The value from a `Float32Array` is already f64 widened by the time JS sees it. `Math.fround` is a no op. The shortest decimal string loop is the only way to recover 1.234 from 1.2339999675750732. Pushing it into `at()` would add that cost to every read, including rendering hot paths that don't need it.

x/ts/src/telem/telem.ts
sy-nico: There isn't a builtin for this. 

`Math.fround` is a no op on values already from a `Float32Array`. The shortest decimal loop (max 9 iterations of `toPrecision` + `parseFloat`) is the only way to recover `1.234` from `1.2339999675750732`. A fixed precision like `toPrecision(9)` doesn't work because it produces longer results (`1.23399997` instead of `1.234`).

pluto/src/log/aether/telem/sources.ts
pjdotson: in the future can you merge that branch into this one and have the PR target that
branch? doesn't really make sense to be reviewing the code otherwise.

pjdotson (changes requested):
see comment and will rereview once right branch is targeted

pluto/src/log/aether/telem/sources.ts
sy-nico: These were independent fixes. The merge wasn't clean either way, so it was worth keeping these separate and tackling one at a time, rather than trying to change the base of either branch.

Merge conflict was resolved by taking `origin/rc`'s version, removing the unused `stringifyFloat32` import from `sources.ts`, and moving the FLOAT32 cleaning into `Series.asString` (with a matching test in `series.spec.ts`). I also trimmed the now-redundant INT64/JSON tests and fixtures from `sources.spec.ts` since origin/rc already covers them.

sy-nico:
@pjdotson , merge was resolved cleanly and easily without having to intermingle the independent changes. Ready for your review.

x/ts/src/telem/telem.ts
```diff
@@ -2129,6 +2129,22 @@ export class DataType
...
+  for (let p = 1; p <= 9; p++) {
+    const parsed = parseFloat(f32.toPrecision(p));
+    if (Math.fround(parsed) === f32) return parsed.toString();
+  }
+  return f32.toString();
+};
+
+/** If dataType is FLOAT32, returns value rounded through the shortest
+ *  f32-roundtrippable decimal. Otherwise returns value unchanged. */
+export const cleanFloat32 = (value: number, dataType: DataType): number =>
```
pjdotson: Let's talk about this sync. I think the `stringifyFloat32` stuff makes sense, but
`cleanFloat32` does not. IMO we should only do the shortest round trip conversion
when we convert from number -> string (e.g. the Stringifier pipeline for the value on a
schematic). so it

pjdotson (changes requested):
Let's talk in sync about this

x/ts/src/telem/telem.ts
sy-nico: I'm good to drop it. I originally implemented it because I thought it was needed for other components. Each has its own justification for why it wasn't needed.

Precision display is configurable:
- Schematic Value
- Schematic Gauge
- Table cells

Handled by `smartRound`:
- Line plot tooltip
- Line plot measure tool

--------

Note: `smartRound`33 cleans the widening incidentally, not by design (it rounds by y-axis span, with no FLOAT32 awareness). For typical charts the span is much larger than f32 noise so widening gets swallowed, but a narrow y-range (e.g. data oscillating between 1.234 and 1.235)

#### Case 1: Narrow y-range
Y-axis range is [1.234, 1.235] & measurement between [1.234, 1.235]. `smartRound` sees span = 0.001 → auto-adjusts to ~6–7 decimal places to preserve resolution. f32 widening artifacts live at ~digit 7, so
they poke through the rounding. A reading of 1.234 would display as `1.2339999` instead of `1.234`.
This is not an issue for f64 data. If users need high precision data, they would/should represent it with an f64.

#### Case 2: Wide y-range, tiny delta
Y-axis range is [0, 100] & ΔY between two close points is 0.001. `smartRound` sees span = 100 → picks ~2 decimal places. The 0.001 delta rounds to `0.00`. Widening is masked but measurement resolution
is lost. This happens regardless of `cleanFloat32`.  Not a widening regression but also not an issue given the context.

x/ts/src/telem/telem.ts
```diff
@@ -2129,6 +2129,17 @@ export class DataType
   ]);
 }
 
+/** Shortest decimal string that round-trips through f32 — JS analogue of Go's strconv.FormatFloat(_, 'g', -1, 32). */
+export const stringifyFloat32 = (value: number): string => {
```
pjdotson: since this isn't needed outside of this package, you should not `export` it, and instead
test it through the `asString` tests.

pjdotson (changes requested):

x/ts/src/telem/telem.ts
sy-nico: Moved `stringifyFloat32` into `series.ts` as a module-private const so it's no longer part of the package's public API. Removed the standalone test suite and migrated every case (typical decimals, integer-valued floats, signed zero, subnormals, max f32, non-finite, idempotency) into the `asString` tests.

PR #2300, opened 2026-05-07
SY-4155: Render Line Breaks in Logs

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4155](https://linear.app/synnax/issue/SY-4155/render-line-breaks-in-logs)

## Description

`\n` in a log channel's string value did not produce a visible line break in the Console.

Canvas2D `fillText` renders `\n` as an unknown glyph. Fixed in the log telem source by splitting each string sample on `\n` and pushing one `LogEntry` per piece, leaving the renderer's "one entry = one line" invariant intact. Continuation entries drop the `timestamp`/`[name]` prefix but keep its width as whitespace so values stay column-aligned. Numeric and JSON channels take a single-push fast path with no split.

JSON log entries now display raw stored keys (`user_id`) instead of the previous JS-transformed camelCase (`userId`)

image

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes `
` characters in log channel string values not rendering as visible line breaks in the Console. The approach splits each STRING sample on `/
?
/` at the telem source layer, pushing one `LogEntry` per segment with continuation entries carrying a whitespace-only prefix to preserve column alignment.

- **`asString()` on `Series`** — new method returning raw stored bytes without camelCase conversion, giving JSON log entries their original `snake_case` keys.
- **GC orphan fix in `gcEntries`** — MAX_ENTRIES eviction now advances past leading continuation entries before splicing so the log head is always a non-orphan line.

Safe to merge — all three previously flagged concerns are addressed with tests and the atVariable refactor preserves at() behavior.

The split logic lives entirely in the telem source layer, the renderer needs only a one-line guard, and time-based GC cannot produce orphan continuations because all segments of a multi-line sample share the same timestamp.

| Filename | Overview |
|----------|----------|
| pluto/src/log/aether/telem/sources.ts | Core fix: STRING samples split on /\r?\n/ pushed as continuation entries; MAX_ENTRIES GC advances past orphaned continuations before splicing. |
| x/ts/src/telem/series.ts | Adds asString() returning raw bytes; refactors atVariable() to string | undefined with type branching moved to at(). |
| pluto/src/log/aether/Log.ts | Continuation entries get whitespace-only prefix matching the normal prefix length for column alignment. |
| pluto/src/log/aether/telem/types.ts | Adds optional continuation?: boolean to LogEntry; backward-compatible. |
| pluto/src/log/aether/telem/sources.spec.ts | Covers \n, \r\n, consecutive newlines, no-split fast paths, and MAX_ENTRIES orphan eviction edge case. |
| x/ts/src/telem/series.spec.ts | asString suite covering STRING, JSON raw keys, BYTES, numeric, UUID, and out-of-bounds cases. |
| pluto/src/log/aether/Log.spec.ts | Six renderer tests for continuation alignment across prefix combinations, multi-group, and interleaved channels. |

</details>

1. `pluto/src/log/aether/telem/sources.ts`, line 225-228 ([link](https://github.com/synnaxlabs/synnax/blob/203a7f3894cfe9a7f7d75782e02ed619950d6851/pluto/src/log/aether/telem/sources.ts#L225-L228)) 

   <a href="#">P1</a> **MAX_ENTRIES GC can leave orphaned continuation entries**

   `gcEntries` evicts entries with a blind `splice(0, excess)` when the array exceeds `MAX_ENTRIES` (100 000). That cut is agnostic to the `continuation` flag, so if the eviction boundary falls mid-group (e.g. the first non-continuation entry of a multi-line sample is removed but its continuation siblings are not), the surviving continuation entries become the new head of `this.entries`. They render with a whitespace-only prefix and no label or timestamp, making the top of the log appear to start in the middle of a sentence with no attribution.
</details>

<sub>Reviews (7): Last reviewed commit: ["SY-4155: tests"](https://github.com/synnaxlabs/synnax/commit/5f5912411ad4c82c62dcda30dc9f6396b170e994) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31135293)</sub>

### DIFF
--- a/pluto/src/log/aether/Log.spec.ts
+++ b/pluto/src/log/aether/Log.spec.ts
@@ -719,6 +719,54 @@ describe("log/aether/Log", () => {
       // Should contain scientific notation (uses unicode ᴇ)
       expect(log.state.selectedText).toMatch(/[eEᴇ]/);
     });
+
+    it("should render continuation entries without name/timestamp but aligned to the value column", () => {
+      const entries: LogEntry[] = [
+        { channelKey: 1, timestamp: TimeStamp.milliseconds(1000), value: "hello" },
+        {
+          channelKey: 1,
+          timestamp: TimeStamp.milliseconds(1000),
+          value: "world",
+          continuation: true,
+        },
+        { channelKey: 1, timestamp: TimeStamp.milliseconds(2000), value: "again" },
+      ];
+      const { log } = setupWithContext(entries, REGION_500, {
+        showChannelNames: true,
+        showReceiptTimestamp: false,
+        channels: [{ channel: 1 }],
+        channelNames: { "1": "log" },
+        selectionStart: 0,
+        selectionEnd: 2,
+      });
+      log.render();
+      const lines = log.state.selectedLines.map((l) => l.text);
+      expect(lines[0]).toContain("[log]");
+      expect(lines[0]).toContain("hello");
+      // Continuation: no [log], but leading whitespace preserves value alignment.
+      expect(lines[1]).not.toContain("[log]");
+      expect(lines[1]).toMatch(/^\s+world$/);
+      expect(lines[1].indexOf("world")).toBe(lines[0].indexOf("hello"));
+      expect(lines[2]).toContain("[log]");
+      expect(lines[2]).toContain("again");
+    });
+
+    it("should render an entry with empty value but no continuation flag with its prefix intact", () => {
+      const entries: LogEntry[] = [
+        { channelKey: 1, timestamp: TimeStamp.milliseconds(1000), value: "" },
+      ];
+      const { log } = setupWithContext(entries, REGION_500, {
+        showChannelNames: true,
+        showReceiptTimestamp: false,
+        channels: [{ channel: 1 }],
+        channelNames: { "1": "log" },
+        selectionStart: 0,
+        selectionEnd: 0,
+      });
+      log.render();
+      const [line] = log.state.selectedLines.map((l) => l.text);
+      expect(line).toContain("[log]");
+    });
   });
 
   describe("selectedText and selectedLines", () => {
--- a/pluto/src/log/aether/Log.spec.ts
+++ b/pluto/src/log/aether/Log.spec.ts
@@ -767,6 +767,128 @@ describe("log/aether/Log", () => {
       const [line] = log.state.selectedLines.map((l) => l.text);
       expect(line).toContain("[log]");
     });
+
+    it("should render a three-line continuation group with consistent alignment", () => {
+      const entries: LogEntry[] = [
+        { channelKey: 1, timestamp: TimeStamp.milliseconds(1000), value: "first" },
+        {
+          channelKey: 1,
+          timestamp: TimeStamp.milliseconds(1000),
+          value: "second",
+          continuation: true,
+        },
+        {
+          channelKey: 1,
+          timestamp: TimeStamp.milliseconds(1000),
+          value: "third",
+          continuation: true,
+        },
+      ];
+      const { log } = setupWithContext(entries, REGION_500, {
+        showChannelNames: true,
+        showReceiptTimestamp: true,
+        channels: [{ channel: 1 }],
+        channelNames: { "1": "sensor" },
+        selectionStart: 0,
+        selectionEnd: 2,
+      });
+      log.render();
+      const lines = log.state.selectedLines.map((l) => l.text);
+      expect(lines[0]).toContain("[sensor]");
+      expect(lines[0]).toContain("first");
+      const valueCol = lines[0].indexOf("first");
+      expect(lines[1]).not.toContain("[sensor]");
+      expect(lines[1].indexOf("second")).toBe(valueCol);
+      expect(lines[2]).not.toContain("[sensor]");
+      expect(lines[2].indexOf("third")).toBe(valueCol);
+    });
+
+    it("should render continuation with both timestamp and channel name hidden", () => {
+      const entries: LogEntry[] = [
+        { channelKey: 1, timestamp: TimeStamp.milliseconds(1000), value: "line1" },
+        {
+          channelKey: 1,
+          timestamp: TimeStamp.milliseconds(1000),
+          value: "line2",
+          continuation: true,
+        },
+      ];
+      const { log } = setupWithContext(entries, REGION_500, {
+        showChannelNames: false,
+        showReceiptTimestamp: false,
+        channels: [{ channel: 1 }],
+        selectionStart: 0,
+        selectionEnd: 1,
+      });
+      log.render();
+      const lines = log.state.selectedLines.map((l) => l.text);
+      expect(lines[0]).toBe("line1");
+      expect(lines[1]).toBe("line2");
+    });
+
+    it("should render continuation entries interleaved between different channels", () => {
+      const entries: LogEntry[] = [
+        { channelKey: 1, timestamp: TimeStamp.milliseconds(1000), value: "a1" },
+        {
+          channelKey: 1,
+          timestamp: TimeStamp.milliseconds(1000),
+          value: "a2",
+          continuation: true,
+        },
+        { channelKey: 2, timestamp: TimeStamp.milliseconds(1000), value: "b1" },
+        {
+          channelKey: 2,
+          timestamp: TimeStamp.milliseconds(1000),
+          value: "b2",
+          continuation: true,
+        },
+      ];
+      const { log } = setupWithContext(entries, REGION_500, {
+        showChannelNames: true,
+        showReceiptTimestamp: false,
+        channels: [{ channel: 1 }, { channel: 2 }],
+        channelNames: { "1": "chA", "2": "chB" },
+        selectionStart: 0,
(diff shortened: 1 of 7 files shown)


emilbon99:
@sy-nico removing myself as a reviewer. don't think two are necessary here.

pluto/src/log/aether/telem/types.ts
```diff
@@ -16,6 +16,8 @@ export interface LogEntry {
   channelKey: channel.Key;
   timestamp: TimeStamp;
   value: string;
+  // True when this entry was produced by splitting a sample on \n;
+  continuation?: boolean;
```
pjdotson, line 20: maybe instead of having a continuation boolean, value should be an `[]string`.

pluto/src/log/aether/Log.spec.ts
```diff
@@ -719,6 +719,54 @@ describe("log/aether/Log", () => {
...
+        showReceiptTimestamp: false,
+        channels: [{ channel: 1 }],
+        channelNames: { "1": "log" },
+        selectionStart: 0,
+        selectionEnd: 0,
+      });
+      log.render();
+      const [line] = log.state.selectedLines.map((l) => l.text);
+      expect(line).toContain("[log]");
+    });
```
pjdotson, line 769: doesn't look like any of these test a multi-string?

pluto/src/log/aether/telem/sources.ts
```diff
@@ -146,15 +149,26 @@ export class StreamMultiChannelLog
...
-              const raw = buf.at(i, true);
-              this.entries.push({
-                channelKey: chMeta.key,
-                timestamp: now,
-                value: isJSON ? JSON.stringify(raw) : String(raw),
-              });
-              pushed++;
+              const value = isJSON
+                ? JSON.stringify(buf.at(i, true))
+                : String(buf.at(i, true));
```
pjdotson: see comment on other PR, but it would be good to just automatically get a string version
at a certain index from calling `at`

pluto/src/log/aether/telem/sources.ts
```diff
@@ -146,15 +149,26 @@ export class StreamMultiChannelLog
...
+              if (!isString || !value.includes("\n")) {
+                this.entries.push({ channelKey: chMeta.key, timestamp: now, value });
+                pushed++;
+                continue;
+              }
+              const lines = value.split(/\r?\n/);
```
pjdotson, line 160: should consider different rendering on Windows vs Unix browsers based off of the `\r\n`.

pjdotson (changes requested):

pluto/src/log/aether/telem/types.ts
sy-nico: I tried this out, and while the source side gets simpler, the renderer assumes one entry = one visual line throughout scroll, selection, GC adjustment, and rendering. Switching to `[]string` requires a `lineCount` field, a `sliceByLine` helper, and touching ~11 call sites in Log.ts. Net complexity is higher. Keeping the flat model with `continuation` confines the change to the source + a one-liner in the renderer.

pluto/src/log/aether/Log.spec.ts
sy-nico: Added more tests

pluto/src/log/aether/telem/sources.ts
sy-nico: The regex `/\r?\n/` already handles both. The `?` makes `\r`optional, so it splits correctly on `\n` (Unix) and `\r\n` (Windows) line endings. The endings come from the source device data, not the browser, and both forms produce identical results after the split.

pluto/src/log/aether/telem/sources.ts
sy-nico: Implemented `Series.asString` so callers get the string directly without a round-trip through parse and stringify.

pluto/src/log/aether/telem/sources.ts
```diff
@@ -145,16 +148,24 @@ export class StreamMultiChannelLog
...
           const pushSamples = (buf: Series, start: number): void => {
             for (let i = start; i < buf.length; i++) {
-              const raw = buf.at(i, true);
-              this.entries.push({
-                channelKey: chMeta.key,
-                timestamp: now,
-                value: isJSON ? JSON.stringify(raw) : String(raw),
-              });
-              pushed++;
+              const value = buf.asString(i, true);
```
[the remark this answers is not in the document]
sy-nico: Intentional. The old code ran JSON values through `snakeToCamel` before display, so log entries showed `userId` instead of the actual stored `user_id`. `asString` returns the raw stored bytes, which is what a log viewer should display.

sy-nico:
@greptile, Regarding this P1 finding (https://github.com/synnaxlabs/synnax/pull/2300#discussion_r3205394164). It is a non-issue. See justification:

Intentional. The old code ran JSON values through `snakeToCamel` before display, so log entries showed `userId` instead of the actual stored `user_id`. `asString` returns the raw stored bytes, which is what a log viewer should display.

Please update review.

x/ts/src/telem/series.ts
```diff
@@ -845,9 +862,9 @@ export class Series<T extends TelemValue = TelemValue>
       }
     }
     const slice = new Uint8Array(buf, start, len);
-    if (this.dataType.equals(DataType.STRING))
-      return new TextDecoder().decode(slice) as T;
-    return caseconv.snakeToCamel(JSON.parse(new TextDecoder().decode(slice))) as T;
+    const str = new TextDecoder().decode(slice);
+    if (asString || this.dataType.equals(DataType.STRING)) return str as T;
+    return caseconv.snakeToCamel(JSON.parse(str)) as T;
```
pjdotson: nit but it is a little odd to have an option passed into this function just to determine
what happens here. I would make this return a type string, and then return the result
`new TextDecoder().decode(slice)`, and remove the `asString` argument. Then at the call
site in `asString` you just return the result and the other callsite in `at` you can
deal with the JSON parsing

x/ts/src/telem/series.spec.ts
```diff
@@ -369,6 +369,56 @@ describe("Series", () => {
     });
   });
 
+  describe("asString", () => {
```
pjdotson, line 372: should have tests 1) that JSON case converts, 2) for a Bytes data type, including with
invalid UTF-8 data

pluto/src/log/aether/telem/sources.spec.ts
```diff
@@ -370,6 +370,157 @@ describe("StreamMultiChannelLog", () => {
...
+      c.streamHandler?.(new Map([[c.channelA.key, new MultiSeries([series])]]));
+      const entries = log.value();
+      expect(entries).toHaveLength(3);
+      expect(entries.map((e) => e.value)).toEqual(["line1", "line2", ""]);
+      expect(entries.map((e) => e.continuation === true)).toEqual([false, true, true]);
+      expect(entries.every((e) => e.channelKey === c.channelA.key)).toBe(true);
+      const ts = entries[0].timestamp;
+      expect(entries.every((e) => e.timestamp === ts)).toBe(true);
+    });
+
```
pjdotson, line 395: can you add a test for when the sample value is an empty string? and when data type is bytes?

pluto/src/log/aether/telem/types.ts
pjdotson: The change might be a bigger refactor, but you also want to consider duplicating state
and making it impossible to represent illegal state. If i have a `[]LogEntry` that is a
continuation, then the `channelKey` and `timestamp` MUST be the same for both otherwise
we have made illegal state. Now code editing this in the future has to make sure that
illegal state never gets represented. It is much better to just refactor to support the
state data types that make since, otherwise the code becomes more brittle, there can be
odd intersecting edge cases, and it is general harder to understand and easier to write
bugs when changing this code.

x/ts/src/telem/series.spec.ts
pjdotson: For invalid UTF-8 in byte slices, you will want to use �.

pjdotson (changes requested):

pluto/src/log/aether/telem/sources.spec.ts
sy-nico: added

x/ts/src/telem/series.ts
sy-nico: `atVariable` now returns `string`, `at` handles the JSON parse + `snakeToCamel` at the call site, and the `asString` flag is gone.

x/ts/src/telem/series.spec.ts
sy-nico: Added. New tests cover snake_case vs camelCase contrast between `asString` and `at`, a BYTES happy path, and a BYTES invalid UTF-8 case that asserts the `�` replacement character.

pluto/src/log/aether/telem/types.ts
sy-nico: Spinoff issue: https://linear.app/synnax/issue/SY-4177/refactor-logentryvalue-to-string-remove-continuation

PR #2306, opened 2026-05-08
SY-3038: Action-Based Undo/Redo with Server-Synchronized Dispatch

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-3038](https://linear.app/synnax/issue/SY-3038)

## Description

Adds an action-based dispatch substrate with per-document undo/redo support, plus the schematic integration. The substrate is split into two reusable pieces in `flux/base/undoable.ts` and the existing `flux/dispatch.ts`:

- **`UndoableUnaryStore`** — a UnaryStore extension that holds doc state alongside per-doc undo/redo stacks. Owns pure transitions: `replay`, `recordEntry` (with same-kind/same-targets/same-window coalescing and stale-entry walking), `prepareUndo`, `prepareRedo`, `beginTransaction`, `applyRemote`, and cascade-delete (deleting a doc drops its undo stack).
- **`createUndoableStore`** — produces the `UnaryStoreConfig` (factory + remote listener); registered once in the Flux Provider.
- **`createDispatch`** — produces `useDispatch`, `useUndo`, `useRedo` hooks bound to a store key + send function.

The schematic side gains an action-based reducer where each handler returns `{ inverse, targets }`. The oracle `ts/actions` plugin generates the dispatcher boilerplate (`Handlers`, `HandlerResult`, `ReduceAllResult`, `createReduceAll`) so `client/ts/src/schematic/actions.ts` is just the per-action handler bodies.

A real bug surfaced and was fixed during testing: `prepareUndo` was swapping forward/inverse on the redo entry, then `prepareRedo` returned `entry.forward` (the original inverse). Redo applied the inverse a second time instead of restoring state. Fixed by shuttling entries between stacks unchanged. State-restoration assertions added to dispatch tests so this can't regress.

## Test coverage

- `pluto/src/flux/base/undoable.spec.ts` — 44 cases covering every public method of the store (replay, recordEntry, coalescing edges, stale-entry walking, transaction commit/abort/send-fail, applyRemote, cascade-delete variants, listener semantics).
- `pluto/src/flux/dispatch.spec.tsx` — 21 cases covering the React hook layer, including multi-action dispatch, undefined-doc guard, redo state restoration, multi-key isolation, and empty-stack no-ops. Uses `waitFor` instead of fixed sleeps.
- `client/ts/src/schematic/actions.spec.ts` — 57 inverse-computation cases preserved.
- `oracle/plugin/ts/actions/actions_test.go` — 12 cases against the new template.
- `x/ts/src/deep/copy.spec.ts` — added a `Proxy.revocable` test pinning the detached-snapshot guarantee the schematic relies on.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR introduces an action-based undo/redo substrate (`UndoableUnaryStore` + `createDispatch`) with per-document stacks, coalescing, stale-entry walking, and server-seq-based echo dedup, then wires it into the schematic editor so Ctrl+Z/Ctrl+Y work end-to-end.

- **`pluto/src/flux/base/undoable.ts` + `dispatch.ts`**: New reusable substrate; `prepareUndo`/`prepareRedo` shuttle entries between stacks unchanged (fixing the previously documented forward/inverse swap bug); 65 test cases cover all public methods.
- **`client/ts/src/schematic/actions.ts` + `actions.gen.ts`**: Handlers now return `{ inverse, targets }` instead of `void`; the oracle code-gen template is updated to emit `HandlerResult`/`ReduceAllResult` and a single `createReduceAll` factory.
- **`core/pkg/service/schematic/writer.go`**: Stamps a monotonically increasing `seq` (`atomic.Uint64`, per service node) on every broadcast so clients can drop stale echoes; server restart resets the counter but clients reload state on reconnect.

Safe to merge with awareness of open issues from prior rounds: remote rename actions don't propagate to the ontology sidebar, and setNode undo leaves a phantom config entry when a config is newly added to an existing node.

The core undo/redo machinery and Go seq-stamping changes are solid and well-tested (122+ cases). The open issues from prior review rounds affect specific edge cases rather than the primary dispatch loop, but remain unresolved. New findings in this pass are minor.

pluto/src/schematic/queries.ts (remote rename path missing Ontology.renameFluxResource) and client/ts/src/schematic/actions.ts (setNode inverse doesn't clean up newly added config on undo) warrant follow-up before shipping.

| Filename | Overview |
|----------|----------|
| pluto/src/flux/base/undoable.ts | New UndoableStore with per-doc undo/redo stacks, coalescing, stale-entry walking, and seq watermarks; own-echo path re-applies idempotent actions and triggers spurious notifications due to missing equal function. |
| pluto/src/flux/dispatch.ts | New createDispatch factory producing useDispatch/useUndo/useRedo hooks; commit ordering correctly prevents the undo-stack race; undo/redo correctly skip preprocessing. |
| client/ts/src/schematic/actions.ts | Hand-written handlers now return inverse+targets; setNode existing-node path leaves a phantom config when oldConfig is null but payload.config is not. |
| client/ts/src/schematic/actions.gen.ts | Generated createReduceAll accumulates inverse/targets; inner reduce switch has no default case — unknown action type past zod parsing would return undefined and throw on result.inverse.length. |
| pluto/src/schematic/queries.ts | Replaces old per-action listener with createUndoableStore+createDispatch; remote applyRemote path no longer calls Ontology.renameFluxResource, so peer renames don't update the sidebar tree name. |
| console/src/schematic/toolbar/Properties.tsx | Renamed to Properties, fetches configByKey at the top level; displayName still says PropertiesControls; configByKey.size===0 guard shows misleading prompt when only config-less elements are selected. |
| core/pkg/service/schematic/writer.go | Adds atomic.Uint64 seq stamping on Dispatch broadcasts; increment is race-free and per-node reset-on-restart limitation is documented. |
| oracle/plugin/ts/actions/actions.go | Template updated to emit HandlerResult/ReduceAllResult and createReduceAll with inverse accumulation; createReducer removed; tests updated. |
| pluto/src/flux/base/store.ts | Adds UnscopedStore interface and optional factory field to UnaryStoreConfig; get overloads added for type safety. |
| console/src/schematic/Schematic.tsx | Wires useUndo/useRedo into useTriggers; enableTriggers now correctly guards keyboard shortcuts behind hasUpdatePermission and selectEditable. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant UI as React UI
    participant D as useDispatch
    participant Store as UndoableStore
    participant Server as Synnax Server
    participant Peer as Other Clients

    UI->>D: dispatch(key, actions)
    D->>Store: replay optimistic update
    Store-->>D: processed, inverse, targets, rollback
    D->>Store: recordEntry push undo stack
    D->>Server: schematics.dispatch(processed)
    Server-->>D: echo seq own sessionKey
    D->>Store: applyRemote isOwnEcho true reducer runs remoteTouched skipped
    Server-->>Peer: broadcast seq other sessionKey
    Peer->>Store: applyRemote isOwnEcho false state updated targets marked

    UI->>D: undo()
    D->>Store: prepareUndo pop entry return inverse
    D->>Store: replay inverse skipPreprocess
    D->>Store: commit move entry undo to redo stack
    D->>Server: schematics.dispatch(inverse)
```

1. `pluto/src/schematic/queries.ts`, line 193-196 ([link](https://github.com/synnaxlabs/synnax/blob/8f8b1ec7c3cc12da3360ade50e5430d867c16a46/pluto/src/schematic/queries.ts#L193-L196)) 

   <a href="#">P2</a> **Local store delete is ordered after the server call — rollbacks inconsistent on server failure**

   `store.schematics.delete(keys)` is invoked after `await client.schematics.delete(data)`. This is the correct order for ensuring the server delete is authoritative. However, its rollback destructor is pushed after success — if any subsequent failure inside `createUpdate` unexpectedly triggers the rollback chain, it would attempt to restore a schematic that no longer exists on the server. Worth verifying that the enclosing `createUpdate` framework only runs rollbacks on failures that precede the `return data` line. Does `createUpdate` guarantee it only calls rollbacks for operations that ran before the exception? If so, this is safe since `store.schematics.delete` is the final operation before `return data`.
</details>

<sub>Reviews (16): Last reviewed commit: ["Lowercase &#39;schematic&#39; in properties plac..."](https://github.com/synnaxlabs/synnax/commit/6dd5b90d1314205e4d8eb6704865ad7f1225ced6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31372779)</sub>


pluto/src/flux/base/undoable.ts
```diff
@@ -0,0 +1,497 @@
...
+    const entry = state.undo[idx];
+    return {
+      actions: entry.inverse,
+      commit: () =>
+        this.updateUndo(scope, key, (s) => ({
+          ...s,
+          undo: s.undo.slice(0, idx),
+          redo: [...s.redo, entry],
+        })),
+    };
```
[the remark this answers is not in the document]
emilbon99: The race described here doesn't exist. `prepared.commit()` is invoked synchronously by `apply` in `dispatch.ts` before the `await send(...)`:

```ts
const r = store[storeKey].replay(...);     // sync
const stackRollback = record(r);            // sync — calls prepared.commit()
try {
  await send(...);                          // network round-trip starts here
} catch { ... }
```

The `record` closure passed by `useUndo` is `() => prepared.commit()`, so the full `prepareUndo → commit` sequence runs inside a single JS task with no async boundary. Nothing else (no `recordEntry`, no `applyRemote`, no other dispatch) can mutate `s.undo` between `idx` being captured and `commit` consuming it, so `s.undo.slice(0, idx)` is safe. Closing as not a bug.

docs/tech/rfc/0037-260508-action-based-undo-redo.md
```diff
@@ -0,0 +1,489 @@
...
+
+## 3.1 - Undoability
+
+Per-document-type code passes an `isUndoable(action) => bool` callback into
+`createUndoableStore`. The substrate filters non-undoable actions out of the recorded
+`forward` before pushing a stack entry; if every action in a transaction is
+non-undoable, nothing is pushed. For the schematic, `set_node_measured` is the only
+non-undoable action — it carries layout-derived dimensions from the renderer, not user
+intent. Its reducer also returns empty targets so a remote `set_node_measured` cannot
+invalidate undoables targeting the same node.
```
pjdotson, line 161: should `setNodeMeasured` even be stored in the server-side then? That way all actions
should theoretically be undoable?

docs/tech/rfc/0037-260508-action-based-undo-redo.md
```diff
@@ -0,0 +1,489 @@
...
+
+A schema-level `undoable: bool` annotation on `.oracle` action declarations was
+considered and dropped. A per-store callback captures the same intent in one line and
+avoids a generator change; promoting it to a generated `UndoableActions` set is a
+follow-up if a second document type accumulates more than a handful of exclusions.
```
pjdotson, line 166: how would you deal with wanting your ctrl+z to undo session state but not server state?

pjdotson (changes requested):
This sorks okay
This implementation ~works (but still has bugs and things that don't work). However, I am a little concerned that the current approach to the concurrent editing feels like you didn't do your homework on different frameworks for implementing concurrent editing (OT vs CRDT) and making sure that we implement one of those properly here, otherwise I think we will run into odd bugs and edge cases. For instance a lot of the data structures for implementing those will have a revision counter (OT) or client-revision vector (CRDT) to make sure actions are applied in a way that makes sense - e.g for a table, if Alice inserts row 18, edits a cell, and then deletes row 18, and Bob inserts his row 18 and edits a cell with a different value, Bob cell should still stay after Alices delete.

A more well thought-out version doesn't need to be done for this PR, which can be thought of more as restoring the schematic undo functionality, but should be done before we add concurrent editing + undo/redo to all data types.

Notes:
- Undo didn't work for changing a label of a symbol
- Undo didn't work after changing symbol scale via dragging
- add back integration tests that were commented out

emilbon99:
I did quite a bit of research on `CRDT` vs `OT` and made an explicit decision not to over-engineer the initial implementation of this tooling. Pretty much everything works as is. I added a counter based conflict resolution mechanism, which addresses issues you mentioned. 

Basically it's at a point where state is guaranteed consistent, but it is fundamentally a last write wins model. So your table example might not work correctly, but I felt like this was a non-issue for the initial implementation.

docs/tech/rfc/0037-260508-action-based-undo-redo.md
emilbon99: Not implemented and haven't really considered it yet. How important is it that session state is undoable? Not very right? I guess if you wanted to undo moving a legend.

We should consider this once we run into it.

docs/tech/rfc/0037-260508-action-based-undo-redo.md
emilbon99: Discussed verbally.

console/src/schematic/toolbar/Properties.tsx
```diff
@@ -47,21 +47,21 @@ export interface PropertiesProps {
...
-  },
-);
-PropertiesControls.displayName = "PropertiesControls";
+export const Properties = memo(({ layoutKey }: PropertiesProps): ReactElement => {
+  const selected = useSelectSelected(layoutKey);
+  const configByKey = Schematic.useSelectConfigs({ key: layoutKey, keys: selected });
+  if (selected.length === 0 || configByKey.size === 0)
+    return (
+      <Text.Text status="disabled" center>
+        Select a Schematic element to configure its properties.
```
pjdotson: should be lowercase schematic

docs/tech/rfc/0037-260508-action-based-undo-redo.md
pjdotson: sg

pjdotson (changes requested):
Approved after comment addressed and CI passes.

Will definitely want to improve our framework for this later but is fine as a v1.

PR #2310, opened 2026-05-09
SY-4159: Arc String Formatting

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4159](https://linear.app/synnax/issue/SY-4159)

## Description

Arc string literals come in eight forms, crossing two delimiters with optional `r` (raw) and `f` (format) prefixes. Backslash escapes and placeholder parsing are fully independent passes: the `r` prefix toggles only the standard escape table, and the `f` prefix toggles only `{expr}` placeholders with `{{` / `}}` as the literal-brace escape.

### Syntax

| Delimiter   | Plain       | `r` (raw)   | `f` (format) | `rf` / `fr` |
|-------------|-------------|-------------|--------------|-------------|
| `"..."`     | single-line | no escapes  | placeholders | both        |
| `` `...` `` | multi-line  | no escapes  | placeholders | both        |

```go
x := "hello"                              // single-line
y := `line 1
line 2`                                   // multi-line
z := f"pi: {3.14:.2f}"                    // format
w := f`report:
  v={sensor}
  t={t:.2f}`                              // multi-line format
p := r"C:\Users\path"                     // raw (no escape processing)
q := r`
verbatim
backslash: \n stays as two chars
`                                         // raw multi-line
m := rf"C:\logs\{name}.txt"               // raw plus format
n := rf`path: {dir}\subdir
name: {file}`                             // raw plus format multi-line
```

### Escape semantics

Two passes operate independently:

1. **Backslash escapes** (gated by the `r` prefix): `\b`, `\t`, `\n`, `\f`, `\r`, `\\`, `\uXXXX`, and the delimiter escape (`\"` only inside `"..."`, `` \` `` only inside `` `...` ``; the non-delimiter character is a literal everywhere). Raw mode skips this pass entirely; unrecognized escapes pass through verbatim.
2. **Placeholder parsing** (gated by the `f` prefix): `{expr}` and `{expr:spec}` interpolate. `{{` escapes to a literal `{` and `}}` escapes to a literal `}`. A single `}` outside a placeholder is plain text.

Because the two passes never cross, backslashes have no special meaning to the placeholder parser. `rf"C:\logs\{name}.txt"` keeps every backslash literal and still interpolates `{name}`.

### Implementation

The grammar uses two string tokens, `STR_LITERAL` (double-quoted, single-line) and `STR_LITERAL_MULTI` (backtick, multi-line), each accepting an optional `r` / `f` / `rf` / `fr` prefix. Structural validation happens in the analyzer rather than the lexer, which keeps the grammar small and centralizes prefix logic in the `literal` package.

- **Parser**: `STR_LITERAL`, `STR_LITERAL_MULTI`, and a `STR_PREFIX` fragment. The `literal` rule alternates `STR_LITERAL | STR_LITERAL_MULTI`.
- **Literal package**: `StripQuotes(text)` peels the prefix and delimiter, returning the body plus a `StringFlags{Raw, Format, Multi}` struct. `UnescapeString(body, multi)` applies the standard escape table with the `multi`
  flag selecting which delimiter escape is interpreted. `FmtStrParse` walks the body looking for `{...}` placeholders, treating `{{` and `}}` as literal-brace escapes. For format-prefixed strings without placeholders, `ParseString` runs `FmtStrParse` and concatenates the text segments so the constant value has `{{` / `}}` collapsed before reaching the IR.
- **Analyzer**: `AnalyzeFmtStrLiteral` accepts either string terminal; body offset is prefix length plus one (both delimiters are single characters).
- **Compiler**: `compileStringLiteral` runs `StripQuotes` and dispatches on flags. Non-format strings emit a single literal segment; format strings parse via`FmtStrParse` and lower through `EmitFmtSegments`.
- **LSP**: `expandStringToken` handles both terminal types; format-string placeholders are colorized per segment.
- **TextMate grammar**: a single pattern uses a backreference to match either delimiter symmetrically. One scope name, `string.quoted.arc`.
- **Console placeholder suggest**: a permissive single-branch regex with lookbehind / lookahead that rejects the halves of a `{{` literal-brace escape.
- **C++ runtime**: `format_float_value` dispatches on the verb with hardcoded per-verb `snprintf` format strings (passing precision via `%.*X` form), which satisfies MSVC C4774 on Windows without a pragma.

### Test coverage

All 52 `arc/go` test suites pass. End-to-end integration assertions in `integration/tests/arc/stl_string.py` cover:

- `f"{{a}}"` -> `{a}` (function context, no-placeholder doubled braces)
- `f"{{{42}}}"` -> `{42}` (function context, doubled braces around placeholder)
- `rf"C:\logs\{{abc}}.txt"` -> `C:\logs\{abc}.txt` (function context)
- `rf"C:\logs\{{{42}}}.txt"` -> `C:\logs\{42}.txt` (function context)
- `trig -> rf"C:\logs\{{abc}}.txt" -> out` -> `C:\logs\{abc}.txt` (flow context)
- `trig -> rf"C:\logs\{42}.txt" -> out` -> `C:\logs\42.txt` (flow context)

Multi-line and raw-format permutations are also exercised in `literal_test.go`, `fmt_test.go`, `parser_test.go`, `compiler/expression/{literal,fmt_str}_test.go`, `analyzer/expression/fmt_str_test.go`, `formatter_test.go`, `lsp/semantic_test.go`, `text/text_test.go`, and `parser/ast_test.go`.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR decouples raw-string semantics from format-string placeholders in Arc, introducing eight orthogonal string forms by crossing two delimiters (`\"...\"` single-line, backtick multi-line) with optional `r`/`f`/`rf`/`fr` prefixes. The previous `STR_LITERAL_RAW` token is replaced by `STR_LITERAL_MULTI` with a `STR_PREFIX` ANTLR fragment.

- **`literal` package**: new `StripQuotes`/`UnescapeString`/`FmtStrParse`/`splitSpec` implement all eight forms; `splitSpec` correctly tracks `[]` depth fixing prior slice-colon false splits; spec validation is now a whitelist `verbAllowedKinds` map.
- **Compiler, analyzer, LSP**: all callers use `parser.StringTerminal`; `expandStringToken` tokenizes format-string placeholders inline with correct multiline column tracking.
- **Console**: `PLACEHOLDER_RE` regex updated with lookbehind/lookahead to reject `{{` escapes while correctly triggering on `\\{` adjacency.

Safe to merge; implementation is internally consistent and all 52 test suites pass.

The single discrepancy found — AnalyzeFmtStrSegments operating on the raw body while the compiler unescapes first — only matters for the exotic \u007B-encoded brace pattern, unlikely in real Arc programs, and does not cause silent wrong output.

arc/go/analyzer/expression/fmt_str.go — raw vs. unescaped body discrepancy with FmtStrParse warrants follow-up if \uXXXX escape coverage in format strings becomes important.

| Filename | Overview |
|----------|----------|
| arc/go/literal/fmt_str.go | New core file: FmtStrParse, splitSpec (correct bracket-depth tracking), StripQuotes, UnescapeString. Uses {{/}} for literal brace escaping as documented. |
| arc/go/analyzer/expression/fmt_str.go | New analyzer; passes raw body to FmtStrParse while compiler unescapes first — discrepancy for \uXXXX-encoded brace characters. |
| arc/go/lsp/semantic.go | expandStringToken splits format-string tokens into semantic tokens with correct per-line splitting for multiline strings. Offset arithmetic verified correct. |
| arc/go/compiler/expression/literal.go | compileStringLiteral refactored; correctly unescapes non-raw strings before FmtStrParse. |
| arc/go/stl/strings/string.go | New format_* WASM host functions; formatWithSpec uses fmt.Sprintf with validated spec. Memory pointer captured correctly via *Module. |
| console/src/arc/editor/text/placeholderSuggest.ts | PLACEHOLDER_RE correctly rejects {{ and handles \{ adjacency via regex backtracking. |

</details>

1. `arc/go/fmtstring/fmtstring.go`, line 161-175 ([link](https://github.com/synnaxlabs/synnax/blob/5d6806502a8c3ca01909a48ba3f4b251717c4054/arc/go/fmtstring/fmtstring.go#L161-L175)) 

   <a href="#">P2</a> **`ValidateNumericSpec` accepts non-numeric verbs like `%T` and `%v`**

   `ValidateNumericSpec` validates by calling `fmt.Sprintf` and checking for `%!` in the output. However, verbs like `%T` (type name) and `%v` (generic) also pass this check for integer/float types — `fmt.Sprintf("%T", int64(0))` returns `"int64"`, not `"%!T(int64=0)"`. A user writing `{x%T}` would get the runtime type string `"int32"` instead of the numeric value, which is almost certainly unintended. The validator should be restricted to explicitly allow only numeric verbs (e.g., `d`, `f`, `e`, `g`, `x`, `o`, `b`, `s` on strings) rather than relying on the absence of `%!`.
</details>

<sub>Reviews (18): Last reviewed commit: ["SY-4159: Switch literal-brace escape to ..."](https://github.com/synnaxlabs/synnax/commit/60788d7390223b720ec4ecd5bef837bf692076ab) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31433347)</sub>


emilbon99:
@pjdotson is a better reviewer for this pr :)

console/src/code/Editor.tsx
```diff
@@ -177,11 +192,16 @@ const use = ({
         target: container,
       }),
     );
+    const placeholderSuggestDispose =
+      language === "arc"
```
pjdotson: If this is Arc-specific, it belongs in `console/src/Arc`, not `console/src/Code`.

console/src/code/placeholderSuggest.ts
```diff
@@ -0,0 +1,19 @@
...
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+// monaco-vscode-api ignores `tokenTypes` in language config, so the parent
+// `string.quoted.raw.arc` scope suppresses the popup inside `{...}`. The only
+// alternative is `quickSuggestions.strings: "on"`, which fires in every string.
+export const PLACEHOLDER_RE = /`[^`]*(?<!\\)\{[^}]*$/;
```
pjdotson: shouldn't be exported

docs/site/src/pages/reference/control/arc/how-to/data-processing.mdx
```diff
@@ -390,6 +390,43 @@ output is in units per second (e.g., psi/second if pressure is in psi).
...
+Backtick format strings build human-readable status messages from live channel values.
+Placeholders inside `{...}` accept any expression, and a `%spec` after the expression
+controls formatting:
+
+```arc
+func status{sensor str} (value f64) str {
+    return `{sensor}: {value%.2f} psi`
+}
+
+pressure -> status{sensor="ox_pt_1"} -> pressure_status
```
pjdotson: bad example function name as it conflicts with `status` module

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,73 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+| ------------- | ------------------------------------------------------ |
+| `{expr}`      | Insert `expr` using its default string form            |
+| `{expr%spec}` | Format `expr` using a printf-style spec (e.g., `%.2f`) |
+
+<Details>
+<Text.Text slot="summary" level="p"><Icon.Reference /> Format spec reference</Text.Text>
+
+| Spec    | Types    | Use               | Example                 |
+| ------- | -------- | ----------------- | ----------------------- |
+| `%d`    | int      | Decimal           | `42` → `42`             |
```
pjdotson: wht about unsigned decimal?

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,73 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+| `%b`    | int      | Binary            | `5` → `101`             |
+| `%#b`   | int      | Binary, prefix    | `5` → `0b101`           |
+| `%o`    | int      | Octal             | `8` → `10`              |
+| `%#o`   | int      | Octal, prefix     | `8` → `010`             |
+| `%x`    | int, str | Hex               | `255` → `ff`            |
+| `%#x`   | int      | Hex, prefix       | `255` → `0xff`          |
+| `%X`    | int, str | Hex, uppercase    | `255` → `FF`            |
+| `%c`    | int      | Unicode character | `65` → `A`              |
+| `%U`    | int      | Unicode escape    | `65` → `U+0041`         |
```
pjdotson: doesn't seem like other languages give a built-in one for Unicode escape, my take is we
shouldn't support it because it could mess with the intuition for a lowercase `"u"` for
unsigned integers.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,73 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+| `%f`    | float    | Decimal           | `3.14` → `3.140000`     |
+| `%e`    | float    | Scientific        | `3.14` → `3.140000e+00` |
+| `%E`    | float    | Scientific, upper | `3.14` → `3.140000E+00` |
+| `%g`    | float    | Compact           | `3.14` → `3.14`         |
+| `%G`    | float    | Compact, upper    | `3.14` → `3.14`         |
+| `%s`    | str      | String            | `"ok"` → `ok`           |
+| `%q`    | str      | Quoted            | `"ok"` → `"ok"`         |
+| `%v`    | any      | Default form      | `3.14` → `3.14`         |
```
pjdotson: Isn't this just a no-op? why can't this just be what happens when the user passes in the
variable anyways?

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,73 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+| `%5s`   | str      | Width 5           | `"ok"` → `␣␣␣ok`        |
+| `%-5s`  | str      | Width 5, left     | `"ok"` → `ok␣␣␣`        |
+
+</Details>
+
+| Escape   | Meaning          |
+| -------- | ---------------- |
+| `` \` `` | Literal backtick |
+| `\{`     | Literal `{`      |
+| `\}`     | Literal `}`      |
```
pjdotson: there isn't any reason to support escaping the right curly brace. If you don't like the
asymmetry of having to escape one but not the other you could do something like
JavaScript and make the variable sign `This is my ${variable}` instead of `This is my {variable}`.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,73 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+| `%b`    | int      | Binary            | `5` → `101`             |
+| `%#b`   | int      | Binary, prefix    | `5` → `0b101`           |
+| `%o`    | int      | Octal             | `8` → `10`              |
+| `%#o`   | int      | Octal, prefix     | `8` → `010`             |
+| `%x`    | int, str | Hex               | `255` → `ff`            |
+| `%#x`   | int      | Hex, prefix       | `255` → `0xff`          |
+| `%X`    | int, str | Hex, uppercase    | `255` → `FF`            |
+| `%c`    | int      | Unicode character | `65` → `A`              |
+| `%U`    | int      | Unicode escape    | `65` → `U+0041`         |
+| `%f`    | float    | Decimal           | `3.14` → `3.140000`     |
```
pjdotson: why can't i format integers with `%f`?

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,73 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+```
+
+Each placeholder is a regular Arc expression. Numeric values are converted to their
+default string form unless a format spec is given. Format specs follow printf-style
+conventions and are validated against the placeholder type at compile time.
+
+| Form          | Meaning                                                |
+| ------------- | ------------------------------------------------------ |
+| `{expr}`      | Insert `expr` using its default string form            |
+| `{expr%spec}` | Format `expr` using a printf-style spec (e.g., `%.2f`) |
```
pjdotson: Are you sure you want to do `%` instead of `:` for formatting? In Python, you can do
something like this:

```python
f"0.1234:.2%"
```

which gives you:

```
12.34%
```

So personally I think we should do `:`.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,73 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+
+<Details>
+<Text.Text slot="summary" level="p"><Icon.Reference /> Format spec reference</Text.Text>
+
+| Spec    | Types    | Use               | Example                 |
```
pjdotson: I would split this table up into a few different tables. First, you have a general table
that just gives all of the lowercase letters, then you say something like "using an
uppercase can print differently" (list where that matters), then you say something about
the `#`, then something about precision / padding, etc.

console/src/code/placeholderSuggest.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    ["double-quoted string is not a backtick raw string", '"{x', "y", false],
+    ["empty buffer never matches", "", "a", false],
+    ["typed empty string never fires", "`hello {x", "", false],
+    ["multi-char typed input never fires (paste)", "`hello {x", "ab", false],
+    ["space is not a word char", "`hello {x", " ", false],
+    ["caret on second line inside placeholder", "x = `hi\n{na", "m", true],
+    ["closed placeholder followed by new open one", "`{a} {b", "c", true],
+    ["escaped open brace does not trigger", "`hello \\{va", "l", false],
+    ["real placeholder after escaped brace still triggers", "`\\{ {va", "l", true],
+  ])("%s", (_label, buffer, typed, expected) => {
```
pjdotson: silly to have to use an empty variable here. Should do something instead like in
`x/ts/src/notation/notation.spec.ts`, it looks a little better and you avoid this unused variable.

pjdotson (changes requested):
Mostly looks good. I haven't looked through the Go code yet, wanted to get general functionality stuff back to you incase you have to change a lot. A couple issues:


1)

LSP color formatting is a little weird. If i mistype something like this, `.2` gets colored in red. Shouldn't this instead be a compiler error?

```
time.wait {
    0s
} -> `Price is ${float+float2%f.2}` -> arc_string_test
```

2)

See https://linear.app/synnax/issue/SY-4171/fix-arc-node-update-bug, I don't know if this was a preexisting issue or an issue introduced here.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
sy-nico: `%d` works for both. The `int` label in the Types column covers signed (`i8` to `i64`) and unsigned (`u8` to `u64`). Go's `fmt` formats both as plain decimal under `%d`, so there's no separate `%u` verb.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
sy-nico: Format specs pass straight through to Go's `fmt.Sprintf`, so the supported verbs match Go's exactly. I am intentionally not curating which verbs are allowed because each exception is a rule to maintain and a way for the docs to drift from runtime behavior. 

On the unsigned-ambiguity concern: Go's `fmt` has no `%u` verb at all, so there's nothing for `%U` to clash with. `%d` already handles both signed and unsigned ints.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
sy-nico: Same reason as above: format specs pass straight through to Go's `fmt.Sprintf`, and Go's `fmt` only accepts `%f` for floats. Calling `fmt.Sprintf("%f", int64(0))` returns `"%!f(int64=0)"`, which our validator catches at compile time.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
sy-nico: The example you provided is incorrect.

```python
>>> print(f"0.1234:.2%")
0.1234:.2%
```

The correct format is:
```python
>>> print(f"{0.1234:.2%}")
12.34%
```
Which is another language using the curly braces to enclose the expression.


The go equivalent is similar.
```go
>>> fmt.Sprintf(`{(0.1234* 100.0)%.2f}%`)
// "12.34%"

Comment (sy-nico) [reply docs/site/src/pages/reference/control/arc/reference/syntax.mdx]: The `%` separator follows Go's `fmt` convention. The format verbs (`%d`, `%05d`, `%.2f`) are Go's, unchanged, and `%` is the prefix Go uses for them. Switching to `:` would mean either translating verbs at the parser layer (so `:05d` in source becomes `%05d` at runtime, a translation rule we'd have to maintain) or living with `{expr:%05d}` (double-separator).

Same principle as the `%U` and `%f` threads above: leaning on Go's `fmt` directly avoids reinventing the wheel and keeps us from adding rules or exceptions of our own.

File: docs/site/src/pages/reference/control/arc/reference/syntax.mdx (hunk 6 above)
Comment (sy-nico) [reply docs/site/src/pages/reference/control/arc/reference/syntax.mdx]: This row is noise, so I'm going to remove it.


Explanation: `%v` is Go's "default format" verb. It exists to give composite types (structs, slices, maps, channels, pointers) a canonical representation without you having to remember a type-specific verb, e.g., `fmt.Sprintf("%v", []int{1,2,3})` returns `"[1 2 3]"`.

That use case doesn't apply here. Arc only allows scalar types (`i*`, `u*`, `f32`, `f64`, `str`) in placeholders, and for each of those the bare `{x}` form already produces byte-identical output to `{x%v}`.

Comment (sy-nico) [reply docs/site/src/pages/reference/control/arc/reference/syntax.mdx]: Added a blacklist for `v` and `T`

File: docs/site/src/pages/reference/control/arc/how-to/data-processing.mdx (hunk 3 above)
Comment (sy-nico) [reply docs/site/src/pages/reference/control/arc/how-to/data-processing.mdx]: fixed

File: console/src/code/Editor.tsx (hunk 1 above)
Comment (sy-nico) [reply console/src/code/Editor.tsx]: Moved to `console/src/arc/editor/` and refactored `Code.Editor` to accept a generic `extensions` prop, so Arc wires the trigger from its own call sites.

File: console/src/code/placeholderSuggest.spec.ts (hunk 11 above)
Comment (sy-nico) [reply console/src/code/placeholderSuggest.spec.ts]: Refactored to pattern with a typed `TEST_CASES` array.

File: docs/site/src/pages/reference/control/arc/reference/syntax.mdx (hunk 10 above)
Comment (sy-nico) [reply docs/site/src/pages/reference/control/arc/reference/syntax.mdx]: Revamped the entire section.

File: docs/site/src/pages/reference/control/arc/reference/syntax.mdx (hunk 7 above)
Comment (sy-nico) [reply docs/site/src/pages/reference/control/arc/reference/syntax.mdx]: Bare `{...}` matches the rich-format-spec family (Python f-strings, Rust `format!`, .NET) where placeholders look like `{x:.2f}`. 

`${...}` belongs to the no-format-spec family (JS template literals, shell). Arc has format specs.

Escape semantics are a wash either way. A literal `${` would sitll need `\$` or `\${`.

**Comment by sy-nico** [comment]
@pjdotson ,

1) Implemented a regex spec "shape" is valid.

2) This is pre-existing (happens with numerics too). This was intentional in the original design, though, I agree it's something that we should change. I left a comment in the ticket explaining more.

File: docs/site/src/pages/reference/control/arc/reference/syntax.mdx (hunk 7 above)
Comment (pjdotson) [reply docs/site/src/pages/reference/control/arc/reference/syntax.mdx]: should only need two escapes

File: console/src/arc/editor/text/placeholderSuggest.ts [hunk 12]
```diff
@@ -0,0 +1,33 @@
...
+// alternative is `quickSuggestions.strings: "on"`, which fires in every string.
+const PLACEHOLDER_RE = /`[^`]*(?<!\\)\{[^}]*$/;
+const WORD_CHAR_RE = /^\w$/;
+
+export const shouldTriggerSuggestion = (
+  bufferBeforeCaret: string,
+  typedChar: string,
+): boolean => WORD_CHAR_RE.test(typedChar) && PLACEHOLDER_RE.test(bufferBeforeCaret);
+
+export const triggerSuggestInPlaceholders: Code.EditorExtension = (editor) =>
```
pjdotson, line 24: Fix "no suggestions" popup

docs/site/src/pages/reference/control/arc/how-to/data-processing.mdx
```diff
@@ -390,6 +390,43 @@ output is in units per second (e.g., psi/second if pressure is in psi).
...
+    }
+    return `pressure nominal: {value%.1f} psi`
+}
+
+pressure -> alert{} -> log
+```
+
+See
+[Format String Literals](/reference/control/arc/reference/syntax#format-string-literals)
+for the full syntax and supported format specs.
```
pjdotson: writing tip: you should probably move the link to the first sentence in this section,
e.g. `[Backtick format strings](link)`

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,119 @@ Strings are enclosed in double quotes. Escape sequences are supported.
 | `\\`   | backslash       |
 | `\"`   | double quote    |
 
+### Format String Literals
```
pjdotson: just to be clear on docs and how we refer to this, there are two types whose names are:
- "string literals"
- "format string literals"
and all of our documentation should refer to them as such, not as "backtick string
literals" or "quoted string literals" and they should not be capitalized. Make sure this
is consistent in the docs

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,119 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+
+Format strings are enclosed in backticks. They embed expressions inside `{...}`
```
pjdotson: so this sentence should start "Format string literals" instead of format strings. You
may want to consider changing the names, so we could have "string literals" (quoted
strings) and "format strings" (backtick strings).

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
pjdotson: switch to colon as per in-person discussion.

console/src/code/Editor.tsx
```diff
@@ -177,11 +184,14 @@ const use = ({
...
+      extensions?.map((ext) => ext(editorRef.current!)) ?? [];
 
     return () => {
       contentDispose.dispose();
       triggerDispose.dispose();
       contextMenuDispose.dispose();
+      extensionDisposables.forEach((d) => d.dispose());
       editorRef.current?.dispose();
       model?.dispose();
     };
```
pjdotson, line 196: this `useEffect` should have `extensions` as a dependency.

console/src/code/Editor.tsx
```diff
@@ -177,11 +184,14 @@ const use = ({
         target: container,
       }),
     );
+    const extensionDisposables =
+      extensions?.map((ext) => ext(editorRef.current!)) ?? [];
```
pjdotson: avoid type assertions.

console/src/channel/Calculated.tsx
```diff
@@ -39,6 +39,8 @@ const NAME_INPUT_PROPS: Partial<Input.TextProps> = {
   placeholder: "Name",
 };
 
+const EDITOR_EXTENSIONS = [Arc.Editor.triggerSuggestInPlaceholders];
```
pjdotson: there should just be a constant `Arc.Editor.EXTENSIONS` in the arc/editor module that
you then pass in below.

console/src/arc/editor/text/placeholderSuggest.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { type Code } from "@/code";
+
+// monaco-vscode-api ignores `tokenTypes` in language config, so the parent
+// `string.quoted.raw.arc` scope suppresses the popup inside `{...}`. The only
+// alternative is `quickSuggestions.strings: "on"`, which fires in every string.
+const PLACEHOLDER_RE = /`[^`]*(?<!\\)\{[^}]*$/;
+const WORD_CHAR_RE = /^\w$/;
+
+export const shouldTriggerSuggestion = (
```
pjdotson, line 19: module placement comment: this `shouldTriggerSuggestion` should live in
`arc/editor/text` as it is only used by the text editor. You probably want to call it from the calculated channel as `Arc.Editor.Text.EXTENSIONS`.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
sy-nico: Removed the third row. A bare `}` now outputs as `}` and there are no more errors for `"Unmatched {"`

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
sy-nico: Switched from `%` to `:`

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
sy-nico: `U` was added to the verb blacklist.

sy-nico:
@greptile, 

Regarding the unresolved "comment outside of diff; A few clarifications make this a non-issue:

1. The function is `ValidateSpec`, not `ValidateNumericSpec`. It also handles strings, not just numeric types.

2. `T`, `v`, and `U` are already rejected by the blacklist. `ValidateSpec` checks `strings.ContainsAny(spec, blacklistedVerbs)` *before* the `%!` probe, so `T`, `v`, and `U` are rejected for every numeric and string type without relying on `fmt.Sprintf`'s output. See `blacklistedVerbs = "vTU"` in `fmtstring.go`.

3. The delimiter is `:`, not `%`. The example `{x%T}` would already be a parse error under the current syntax: `%` is no longer the spec separator. The current form is `{x:T}`, which is then caught by the blacklist as above.

The newly-added test matrix in `fmtstring_test.go` explicitly pins that `v`, `T`, and `U` are rejected against every supported type (33 entries: 8 int, 2 float, and 1 string, multiplied by 3 verbs), so a future change that weakens the blacklist will fail tests rather than slip through. If we ever do want to allow one of these verbs, the rejection is a one-line edit to `blacklistedVerbs` and the failing tests make the scope of the change obvious.

Please update your overall review on the PR description.

console/src/arc/editor/placeholderSuggest.ts
```diff
@@ -0,0 +1,33 @@
...
+export const triggerSuggestInPlaceholders: Code.EditorExtension = (editor) =>
+  editor.onDidChangeModelContent((e) => {
+    const ch = e.changes[0];
+    if (e.changes.length !== 1 || ch.rangeLength !== 0) return;
+    const model = editor.getModel();
+    const pos = editor.getPosition();
+    if (model == null || pos == null) return;
+    const buffer = model.getValue().slice(0, model.getOffsetAt(pos));
+    if (shouldTriggerSuggestion(buffer, ch.text))
+      editor.trigger("synnax", "editor.action.triggerSuggest", { auto: true });
```
pjdotson: should the source be `"synnax"` or `"Arc LSP"` or something else?

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,129 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+| `{expr}`      | Insert `expr` using its default string form           |
+| `{expr:spec}` | Format `expr` using a printf-style spec (e.g., `.2f`) |
+
+<Details>
+<Text.Text slot="summary" level="p"><Icon.Reference /> Format spec reference</Text.Text>
+
+A format spec has the form:
+
+```
+[flags][width].[precision][verb]
```
pjdotson: the `.` is optional too. should probably put the optional part in different type of brackets

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,129 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+```
+
+Only the verb is required. Each bracketed part is described below.
+
+#### Verb
+
+The verb selects the conversion. Each verb accepts specific placeholder types.
+
+| Verb | Types | Use               | Example                 |
```
pjdotson, line 191: make sure you have very comprehensive tests for this table, including testing things we
said are not supported don't do anything.

arc/go/ir/string_fmt.go
```diff
@@ -0,0 +1,14 @@
...
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package ir
+
+// StringFmtSyntheticPrefix tags Function entries synthesized for backtick
+// literals with placeholders, gating synthetic emission in the compiler.
+const StringFmtSyntheticPrefix = "fmt$"
```
pjdotson: this constant probably doesn't belong here if it not used in the ir package.

arc/go/graph/analyze.go
```diff
@@ -85,7 +85,7 @@ func Analyze(
 		g.Functions[i] = fn
 	}
 
-	// Step 3 & 4: Create Fresh Types and IR Nodes
+	// Step 3 & 4A: Create Fresh Types and IR Nodes
```
pjdotson: 4A?

arc/go/fmtstring/fmtstring.go
```diff
@@ -0,0 +1,204 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package fmtstring
```
pjdotson: Why should there be a whole new `fmtstring` package at the top level of the Arc
directory? Does this make sense as module design instead of putting it into `literal`?
If you look at the other top-level packages—`analyzer`, `compiler`, `types`, etc.—does
it make sense for `fmtstring` to belong here?

arc/go/fmtstring/fmtstring.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// StripDelimiters returns the inner body of a `...` raw string token, or
+// ok=false if text isn't well-formed. \` escapes are left verbatim.
+func StripDelimiters(text string) (body string, ok bool) {
+	if len(text) < 2 || text[0] != '`' || text[len(text)-1] != '`' {
+		return "", false
+	}
+	return text[1 : len(text)-1], true
+}
```
pjdotson: If `text` ends with an unescaped backtick, won't this incorrectly return true for okay?

arc/go/fmtstring/fmtstring.go
```diff
@@ -0,0 +1,204 @@
...
+
+// HasPlaceholder reports whether any segment is a placeholder.
+func HasPlaceholder(segs []Segment) bool {
+	return slices.ContainsFunc(segs, func(s Segment) bool { return s.IsPlaceholder })
+}
+
+// StripDelimiters returns the inner body of a `...` raw string token, or
+// ok=false if text isn't well-formed. \` escapes are left verbatim.
+func StripDelimiters(text string) (body string, ok bool) {
```
pjdotson: nit: don't use names in the type signatures of the returns (`body`, `ok`) if they don't
add value.

arc/go/fmtstring/fmtstring.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		b.WriteByte(body[i])
+	}
+	return len(body), b.String()
+}
+
+func findPlaceholderClose(body string, open int) (int, error) {
+	for i := open + 1; i < len(body); i++ {
+		switch body[i] {
+		case '{':
```
pjdotson: so if i have a function call with a config bracket `{` inside the formatting string,
this returns an error?

arc/go/fmtstring/fmtstring.go
```diff
@@ -0,0 +1,204 @@
...
+			return 0, errors.New("unmatched '{'")
+		case '}':
+			return i, nil
+		}
+	}
+	return 0, errors.New("unmatched '{'")
+}
+
+// SplitSpec splits a placeholder body on the last `:`, yielding (expr, spec).
+func SplitSpec(body string) (expr, spec string, err error) {
```
pjdotson: this isn't used by an outside package, should be private

arc/go/fmtstring/fmtstring.go
```diff
@@ -0,0 +1,204 @@
...
+// offsets (End exclusive). SpecOffset is the body offset of `%`, or -1 if
+// no spec.
+type Segment struct {
+	Text          string
+	Spec          string
+	IsPlaceholder bool
+	Start         int
+	End           int
+	SpecOffset    int
+}
```
pjdotson: should add comments explain what each property is, e.g. what is `Spec`

arc/go/fmtstring/fmtstring.go
```diff
@@ -0,0 +1,204 @@
...
+const intBlockedVerbs = "q"
+
+// specShape enforces the canonical anatomy [flags][width][.precision][verb].
+// Anything trailing the verb (e.g., "f.2") fails this check; Go's fmt would
+// otherwise treat it as literal text and silently accept the malformed spec.
+var specShape = regexp.MustCompile(`^[#+\- 0]*\d*(\.\d+)?[a-zA-Z]$`)
+
+// ValidateSpec probes fmt.Sprintf with a typed dummy and reports an error if
+// the spec is not a valid Go fmt verb for the given type.
+func ValidateSpec(spec string, t types.Type) error {
```
pjdotson: I think that you are going about this a little wrong for validating the spec. What you
should do is two things:

1) make a regex that strictly compiles the spec, so you can only use the specific
   characters that we support, for instance.
2) Figure out the verb from the spec, e.g. `x` correspond to which allowed types via a
   map[char][]types.Type and then return if it is not allowed.

There isn't any reason to call gofmt here and the logic is difficult to follow in case
we wanted to add verbs not supported through Go.

arc/go/fmtstring/fmtstring_test.go
```diff
@@ -0,0 +1,423 @@
...
+
+	It("validates against the constraint of a constrained type variable", func() {
+		intConstraint := types.IntegerConstraint()
+		Expect(fmtstring.ValidateSpec("d", types.Variable("T", &intConstraint))).To(Succeed())
+	})
+
+	It("rejects a spec invalid for the variable's constraint", func() {
+		stringConstraint := types.String()
+		err := fmtstring.ValidateSpec(".2f", types.Variable("T", &stringConstraint))
+		Expect(err).To(MatchError(ContainSubstring("invalid format spec")))
```
pjdotson: No need to declare the intermediate `err` variable for these specs, should just call `Expect(fmtstring.Validatespec())`

arc/go/stl/strings/string.go
```diff
@@ -63,69 +64,87 @@ var SymbolResolver = &symbol.ModuleResolver{
...
+	coerce func(T) any,
+) wazero.HostModuleBuilder {
+	return builder.NewFunctionBuilder().
+		WithFunc(func(_ context.Context, v T, ptr, length uint32) uint32 {
+			return m.strings.Create(formatWithSpec(m.memory, ptr, length, coerce(v)))
+		}).Export(name)
+}
+
+func formatWithSpec(memory api.Memory, ptr, length uint32, value any) string {
+	if memory == nil {
```
pjdotson: why do we need this guard?

arc/go/stl/strings/string.go
```diff
@@ -180,30 +199,23 @@ func NewModule(
...
-	builder = builder.NewFunctionBuilder().
-		WithFunc(func(_ context.Context, v float32) uint32 {
-			return s.Create(strconv.FormatFloat(float64(v), 'g', -1, 32))
-		}).Export("from_f32")
-	builder = builder.NewFunctionBuilder().
-		WithFunc(func(_ context.Context, v float64) uint32 {
-			return s.Create(strconv.FormatFloat(v, 'g', -1, 64))
-		}).Export("from_f64")
+		WithFunc(func(_ context.Context, handle, ptr, length uint32) uint32 {
+			str, _ := s.Get(handle)
```
pjdotson: we don't do anything different if `ok` is `false`?

arc/go/analyzer/flow/expression.go
```diff
@@ -13,23 +13,49 @@ import (
...
 	if parser.IsLiteral(ctx.AST) {
+		if lit := parser.GetLiteral(ctx.AST); lit != nil {
+			if rawStr := lit.STR_LITERAL_RAW(); rawStr != nil {
+				body, ok := fmtstring.StripDelimiters(rawStr.GetText())
+				if !ok {
+					ctx.Diagnostics.Add(diagnostics.Errorf(ctx.AST,
+						"invalid raw string literal: %s", rawStr.GetText()))
+					return
+				}
+				segs, perr := fmtstring.Parse(body)
```
pjdotson: nit: name this `err`

arc/go/analyzer/flow/expression_test.go
```diff
@@ -202,6 +202,89 @@ var _ = Describe("AnalyzeSingleExpression", func() {
 		})
 	})
 
+	Describe("Raw String Format Literals", func() {
+		It("should create KindConstant for raw string without placeholders", func(bCtx SpecContext) {
```
pjdotson, line 206: add more tests for errors

arc/go/analyzer/flow/flow_upstream_trigger_test.go
```diff
@@ -0,0 +1,218 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package flow_test
```
pjdotson: this should not be a separate file, should just go in `flow_test.go`.

arc/go/analyzer/expression/string_fmt_test.go
```diff
@@ -0,0 +1,355 @@
...
+			Entry("f64 channel :e", "`{chF64:e}`"),
+			Entry("u8 channel :d", "`{chU8:d}`"),
+			Entry("string channel :s", "`{chStr:s}`"),
+			Entry("string channel :q", "`{chStr:q}`"),
+			Entry("integer literal :d", "`{123:d}`"),
+			Entry("float literal :.2f", "`{3.14:.2f}`"),
+		)
+
+		// Go's fmt accepts %x and %b on float64; pin so a future validator change cannot regress.
+		DescribeTable("accepts Go-fmt-valid integer-shaped specs on floats",
```
pjdotson: This looks like it contradicts your table in the docs site

arc/go/fmt_str_test.go
```diff
@@ -0,0 +1,481 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package arc_test
```
pjdotson, line 10: is this package the right place for comprehensive tests of the string formatting?

arc/go/compiler/expression/literal.go
```diff
@@ -27,14 +28,30 @@ func compileLiteral(
...
 
+func compileRawStringLiteral(
+	ctx context.Context[parser.ILiteralContext],
+	text string,
+) (types.Type, error) {
+	parsed, err := literal.ParseRawString(text, types.String())
+	if err != nil {
+		return types.Type{}, err
+	}
+	body, _ := parsed.Value.(string)
```
pjdotson: should check if there was a conversion error

arc/go/compiler/string_fmt.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package compiler
```
pjdotson: If this function is only used in `compiler.go`, there is no need to put it inside
an entirely new module

arc/go/text/analyze.go
```diff
@@ -498,6 +500,40 @@ func analyzeExpression(
...
+	if sym.Kind == symbol.KindFunction && parser.IsLiteral(ctx.AST) {
+		if literalCtx := parser.GetLiteral(ctx.AST); literalCtx != nil {
+			if rawStr := literalCtx.STR_LITERAL_RAW(); rawStr != nil {
+				outputType := ctx.Constraints.ApplySubstitutions(sym.Type.Outputs[0].Type)
+				parsedValue, err := literal.Parse(literalCtx, outputType)
+				if err != nil {
+					ctx.Diagnostics.Add(diagnostics.Error(err, ctx.AST))
+					return nodeResult{}, false
+				}
+				body, _ := parsedValue.Value.(string)
```
pjdotson: need to check the type cast

arc/go/text/analyze.go
```diff
@@ -498,6 +500,40 @@ func analyzeExpression(
...
+				segments, perr := fmtstring.Parse(body)
```
pjdotson: need to log or do something with the error here, and don't call it `perr`

arc/go/text/text_test.go
```diff
@@ -2453,6 +2453,72 @@ var _ = Describe("Text", func() {
...
+			parsedText := MustSucceed(text.Parse(text.Text{Raw: source}))
+			inter, diagnostics := text.Analyze(ctx, parsedText, resolver)
+			Expect(diagnostics.Ok()).To(BeTrue(), diagnostics.String())
+
+			for _, f := range inter.Functions {
+				Expect(strings.HasPrefix(f.Key, ir.StringFmtSyntheticPrefix)).To(BeFalse(),
+					"unexpected fmt$ synthetic %q for placeholder-free literal", f.Key)
+			}
+		})
+	})
```
pjdotson, line 2615: what about tests for when there is a bad spec or wrong type or something?

arc/go/lsp/semantic_test.go
```diff
@@ -23,8 +25,13 @@ import (
 // Tests in this file pin both the legend ordering and the token-type routing
 // for STR_LITERAL vs STR_LITERAL_RAW to those ids.
 const (
-	tokenTypeString    = uint32(4)
-	tokenTypeStringRaw = uint32(21)
+	tokenTypeVariable          = uint32(3)
```
pjdotson: unused

pjdotson (changes requested):
Issues:
- LSP: why are brackets orange if there is no space in between them? that makes it look like part of the string.
- the format commant on the LSP should have some opinionated formatting.
- colors: 1) why have string literals be the same color as numeric constants like `34`? 2) Why is the string text color in a format string different than the text color in a string literal?

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
sy-nico: Fully agree. Fixing.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
sy-nico: Added fuzz testing and explicit tests for each example.

arc/go/graph/analyze.go
sy-nico: From when I introduced a `4B` step, which has since been removed. Correcting this.

console/src/code/Editor.tsx
sy-nico: Updated to capture monaco.editor.create's non-null return into a local const, avoiding the type assertion in the extensions callback.

console/src/code/Editor.tsx
sy-nico: Added

console/src/channel/Calculated.tsx
sy-nico: Implemented

console/src/arc/editor/text/placeholderSuggest.ts
sy-nico: Moved `placeholderSuggest` into text editor module.

console/src/arc/editor/placeholderSuggest.ts
sy-nico: I went with `Arc LSP`, but I also considered `placeholderSuggest`, `Arc Autosuggest`, and `Arc Autocomplete`. 
Let me know if any of those speak to you.

arc/go/ir/string_fmt.go
sy-nico: Deleted this file and moved to `arc/go/compiler/string_fmt.go`.

arc/go/fmtstring/fmtstring.go
sy-nico: This was originally placed in `stl/strings/` when `string.fmt()` was a function, but I decided to move it up and could not justify a better home.

Per discussion, I tried moving to `analyzer/`, but that would require importing it into 3 `compiler/` files, which breaks the pattern where no compiler files import anything from analyzer.

Merged into `literal/`. I also standardized names so that naming convention is `fmt` first, then `str`/`string` second.

arc/go/fmtstring/fmtstring.go
sy-nico: Yes, this will incorrectly return true within a function block (although it properly errors in the flow context). Fixed by counting the trailing backslash run before the closer.

arc/go/fmtstring/fmtstring.go
sy-nico: removed

arc/go/fmtstring/fmtstring.go
sy-nico: Yes. No current use case for a { inside a placeholder, but easy to lift later by tracking brace depth. The error ("unmatched '{'") at least points at the right thing if someone hits it.

arc/go/fmtstring/fmtstring.go
sy-nico: Originally chose to probe using `fmt.Sprintf` so I wouldn't have to worry about what types supported which verbs. Once we started adding blacklisted verbs and combos, then it the complexity grew to the point that it was no longer efficient. 

`fmt.Sprintf` is now removed. Verbs and their allowed kinds now live in an explicit `verbAllowedKinds` table. Adding a new verb is one map entry

arc/go/stl/strings/string.go
sy-nico: Removed

arc/go/stl/strings/string.go
sy-nico: update to `if !ok { return 0 }`

arc/go/analyzer/expression/string_fmt_test.go
sy-nico: fixed

arc/go/fmt_str_test.go
sy-nico: Removed the "Compile-time diagnostics" block since every entry was duplicated by `analyzer/expression/fmt_str_test.go`. 

What remains in `arc_test` is genuine E2E: runtime harness, frame output, flow-form synthesis, and the docs-pinned examples table, all of which need the full pipeline.

arc/go/compiler/expression/literal.go
sy-nico: `ParseRawString` always returns a string `Value` when err is nil, so the assertion can't fail. Dropped the `, _` to match `compileStringLiteral`

arc/go/compiler/string_fmt.go
sy-nico: Folded into `compiler.go`.

arc/go/text/analyze.go
sy-nico: Guarded by `STR_LITERAL_RAW()` above, so `Value` is guaranteed to be a string when err is nil. Dropped the `, _`.

arc/go/text/analyze.go
sy-nico: Added a diagnostic and return `nodeResult{}, false` instead of being silently swallowed.

arc/go/lsp/semantic_test.go
sy-nico: Removed

arc/go/text/text_test.go
sy-nico: Bad-spec / wrong-type cases are covered at the analyzer layer in `analyzer/expression/fmt_str_test.go`. `text.Analyze` calls `analyzer.AnalyzeProgram` at `analyze.go`, so those diagnostics propagate through this path. Tests here intentionally focus on IR-construction shape. 

I added a smoke test at this ayer anyway

sy-nico:
@pjdotson ,

> Issues:
> 
> - LSP: why are brackets orange if there is no space in between them? that makes it look like part of the string.
> - the format commant on the LSP should have some opinionated formatting.
> - colors: 1) why have string literals be the same color as numeric constants like 34? 2) Why is the string text color in a format string different than the text color in a string literal?

- Fixed
- Agreed this would be nice. Tracking as a follow-up since the formatter currently treats raw strings as opaque tokens, and wiring expression formatting into placeholder bodies needs a new `FormatExpression` entry point that doesn't exist today. [SY-4187](https://linear.app/synnax/issue/SY-4187/apply-expression-formatting-inside-format-string-placeholders)
- 1. I did not change the colors. String literals and numerics have always been green.
- 2. Specifically to differentiate between them for better UX. Raw/format strings have distinct semantics (placeholders, multi-line, backtick-delimited), so a different base color makes it instantly clear which flavor you're looking at.

console/src/arc/editor/placeholderSuggest.ts
pjdotson: What is the equivalent of this in normal VSCode extensions like the Go or Python one?
that should probably tell us what to do.

arc/go/fmtstring/fmtstring.go
pjdotson: then this is a known bad algorithm and should be fixed.

docs/site/src/pages/reference/control/arc/reference/syntax.mdx
```diff
@@ -122,6 +123,131 @@ Strings are enclosed in double quotes. Escape sequences are supported.
...
+<Details>
+<Text.Text slot="summary" level="p"><Icon.Reference /> Format spec reference</Text.Text>
+
+A format spec has the form:
+
+```
+[flags][width][.precision][verb]
+```
+
+Only the verb is required. Each bracketed part is optional. If `precision` is given, the
```
pjdotson, line 184: >> Only the verb is required. Each bracketed part is optional.

verb is bracketed above

arc/go/literal/fmt_str.go
```diff
@@ -0,0 +1,209 @@
...
+	'G': floatKinds,
+}
+
+// FmtStrValidateSpec reports an error if spec is not a supported verb for t,
+// or if t is not a formattable kind.
+func FmtStrValidateSpec(spec string, t types.Type) error {
+	if spec == "" {
+		return nil
+	}
+	if t.Kind == types.KindVariable {
```
pjdotson, line 305: so you couldn't format `KindU8`, `KindNumericConstant`? LMK if I don't understand this properly.

pjdotson (changes requested):
Two things:
1) I think the string literals and format strings should have the same color in the LSP, it matches what I see in VSCode for most languages and that is probably a good guide for what we should be doing.
2) You should make sure to add one or two integration tests leveraging format strings.

arc/go/fmtstring/fmtstring.go
sy-nico: Fixed to track brace depth

arc/go/literal/fmt_str.go
sy-nico: Both are formattable. `KindU8` lives in `intKinds` and `KindNumericConstant` lives in `floatKinds` ([fmt_str.go:151-164](arc/go/literal/fmt_str.go#L151-L164)); both end up in `formattableKinds`, which gets checked a few lines below the snippet you highlighted. The `spec == ""` and `KindVariable` cases are just early exits. `KindU8` formats with `d`/`b`/`c`/`o`/`x`/etc.; `KindNumericConstant` formats with `e`/`f`/`g`/etc.

arc/go/literal/fmt_str.go
```diff
@@ -0,0 +1,236 @@
...
+	if backslashes%2 == 1 {
+		return "", false
+	}
+	return text[1 : len(text)-1], true
+}
+
+// FmtStrParse splits a format-string body into ordered segments at `{...}`
+// placeholders. `\{` escapes to a literal `{`; a bare `}` outside a placeholder
+// is plain text.
+func FmtStrParse(body string) ([]FmtStrSegment, error) {
```
pjdotson, line 141: I think the grammar here needs to be changed:

https://chatgpt.com/share/e/6a0b7b1d-b3c4-8001-ba5e-a2a4fd97c059.

pjdotson (changes requested):
see comment and let's huddle and talk through about:
1) escaping
2) "template string" might be a better name for this than "format string"

PR #2311, opened 2026-05-12
SY-4172: Improve Aether Rendering Stability and Performance

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4172](https://linear.app/synnax/issue/SY-4172)

## Description

Rewrites the Aether main-thread layer to fix a long-standing class of React 19 warnings and to harden the worker connection across StrictMode's pseudo-unmount/remount cycle. Along the way, removes the `RoutedWorker` multiplexing layer from `@synnaxlabs/x` — Aether was its only consumer and never multiplexed beyond a single hard-coded `"vis"` route, so the abstraction was paying a configuration tax for no benefit.

**Stability**
- Replaces the previous `useState`-driven Aether store with a standalone `Store` class consumed via `useSyncExternalStore`. Eliminates the React 19 "Can't perform a React state update on a component that hasn't mounted yet" warning, the "Cannot update a component while rendering a different component" warning, and the lingering `act()` warning in the EchoOnUpdate regression pin.
- Listeners are now keyed by string and persist across the register-unregister cycle React StrictMode introduces, so subscriptions wired during a pseudo-remount stay live once the real register fires.
- `Store.dispose()` parks the worker as a `NOOP_WORKER` sentinel; subsequent sends lazily re-attach via `ensureAttached`, which is what lets the same `useRef`-held Store survive StrictMode's reused-fiber pseudo-cycle.

**Configuration as a constructor-time concern**
- `new Store({})` (or a `<Provider>` with no worker source and `workerEnabled` defaulted to `true`) now throws `ValidationError` immediately. The only legitimate no-worker mode is the explicit `workerEnabled: false` kill-switch.
- Drops the runtime "no worker" reject path inside `invokeMethodAsync` — that branch was a misplaced config check, not an exception. With construction validating eagerly, the `Store.worker` field is always non-null, so the scattered null guards in `send`, `dispose`, and the invoke path are gone.

**Type safety**
- `Entry<S>` is generic over its schema, and `getEntry<S>` / `setEntry<S>` are the single seam where storage erasure lives. Reduces main-thread type assertions from 7 to 3 (one read-side, one write-side, one schema-claim site in `buildMethods`).
- `InvokeTracker.track` and `invokeMethodAsync` no longer pretend to be generic over the response type — the wire layer stays honest with `unknown`, and the schema claim is asserted exactly once at the `CallersFromSchema<M>` boundary.

**RoutedWorker removal**
- `@synnaxlabs/x/worker` is deleted. `SenderHandler`, `wrapWorker`, and `createMockPair` live next to the Aether message types in `pluto/src/aether/message.ts`. Aether's `aether` namespace re-exports `createMockPair` and the message types so test helpers in Console and elsewhere keep working without reaching into `@synnaxlabs/x` for worker plumbing.
- Drops the `workerKey` prop from `Aether.Provider` (and the `"vis"` literal from `Pluto.Provider`) since there is no router to key into.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR refactors Aether's React-to-WebWorker state synchronization architecture, replacing `useState`-driven state management with a standalone `Store` class using `useSyncExternalStore`, and hardening the system against React 19 StrictMode pseudo-unmount cycles.

- **`Store` class** (`pluto/src/aether/store.ts`): New standalone class owns worker comms, entries map, listener map, snapshot cache (`snapshotsByKey`), and `InvokeTracker`; `getSnapshot` falls back to cached snapshots during the unregister→subscribe gap.
- **`RoutedWorker` removal**: `x/ts/src/worker/worker.ts` deleted; `SenderHandler`, `wrapWorker`, `createMockPair` moved to `pluto/src/aether/aether/message.ts`.
- **Error identity preservation**: `errors.Payload` extended with optional `name`/`stack` fields; `withNative`/`applyNative` helpers carry native error identity across worker/network boundaries.

Safe to merge; only P2 suggestions with no blocking defects found.

The core Store refactor is architecturally sound, StrictMode hardening is correct, the test suite is comprehensive, and both findings are non-blocking quality improvements rather than current defects.

pluto/src/aether/store.ts (currentError not cleared on dispose) and pluto/src/aether/aether/message.ts (wrapWorker missing onmessageerror)

| Filename | Overview |
|----------|----------|
| pluto/src/aether/store.ts | New Store class replaces useState-driven state; snapshot caching, InvokeTracker, and ensureAttached are well-structured. Minor: currentError not cleared on dispose. |
| pluto/src/aether/main.tsx | Provider/use/useLifecycle refactored to useSyncExternalStore; render-phase registration is intentional for parent-before-child ordering. |
| pluto/src/aether/aether/message.ts | Wire types and wrapWorker/createMockPair extracted here; wrapWorker missing onmessageerror handler for externally-supplied workers. |
| x/ts/src/errors/errors.ts | payloadZ extended with optional name/stack; withNative/applyNative helpers correctly preserve native error identity across boundaries. |
| pluto/src/aether/main.spec.tsx | Comprehensive test suite covering invoke lifecycle, StrictMode survival, error surfacing, and unidirectional state. |
| pluto/src/aether/aether/aether.ts | Worker-side tree unchanged in structure; error-variant dispatch and invoke method routing look correct. |
| pluto/src/memo/useMemoCompare.ts | Added useMemoArray (element-wise Object.is) and useMemoDeepEqual; straightforward and well-tested. |
| pluto/src/pluto/Pluto.tsx | Worker.Provider wrapper removed; workerURL/workerEnabled passed directly, reducing provider nesting. |
| x/ts/src/errors/errors.spec.ts | New tests cover name/stack round-trip for generic and typed errors; edge cases well-covered. |

</details>

<sub>Reviews (10): Last reviewed commit: ["Accept CrudeTimeSpan for ProviderProps.i..."](https://github.com/synnaxlabs/synnax/commit/5716eccaf3db00a98bbee86c43ee11f4c2daa34c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31691887)</sub>


pluto/src/aether/aether/message.ts
```diff
@@ -0,0 +1,134 @@
...
+  handle: (handler: (value: WorkerMessage) => void) => void;
+}
+
+/** Sentinel used when `workerEnabled: false`. Lets the store stay non-null
+ * without runtime null checks; any other missing-worker configuration is a
+ * constructor-time error. */
+export const NOOP_MAIN_COMMS: MainComms = { send: () => {}, handle: () => {} };
+
+/** Adapts a `Worker` to {@link MainComms} for main-thread use. */
+export const wrapWorker = (worker: Worker): MainComms => ({
```
pjdotson, line 94: Can you add explicit tests for all of the functions in this file in `aether.spec.ts`?

pluto/src/memo/useMemoCompare.ts
```diff
@@ -32,3 +32,17 @@ export const useMemoDeepEqual = <T>(value: T): T => {
   else if (!deep.equal(ref.current, value)) ref.current = value;
   return ref.current;
 };
+
+/**
+ * Returns a referentially stable array as long as its contents are element-wise equal
+ * to the previous render's contents. Faster than {@link useMemoDeepEqual} for primitive
+ * arrays since it short-circuits on length and uses `===` per element.
+ */
+export const useMemoPrimitiveArray = <T extends primitive.Value>(
```
pjdotson: is there a reason to restrict this to `primitive.Value`, instead of just `unknown` like
Reacts comparison? e.g. something like this:

```ts
const arraysEqual<T>(
  arr1: readonly T[],
  arr2: readonly T[],
): boolean {
  if (arr1.length !== arr2.length) return false;
  for (let i = 0; i < arr1.length; i++)
    if (!Object.is(next[i], prev[i])) return false
  return true
}

export const useMemoArray = <T extends unknown>(
  value: readonly T[],
): readonly T[] => {
  const ref = useRef<readonly T[]>(null);
  if (ref.current == null || !arraysEqual(ref.current, value)) ref.current = value
  return ref.current
}

File: pluto/src/aether/store.ts [hunk 3]
```diff
@@ -0,0 +1,473 @@
...
+import { UnexpectedError, ValidationError } from "@synnaxlabs/client";
+import { type errors, TimeSpan, zod } from "@synnaxlabs/x";
+import { type z } from "zod";
+
+import { aether } from "@/aether/aether";
+import { type state } from "@/state";
+
+const DEFAULT_INVOKE_TIMEOUT = TimeSpan.seconds(5);
+
+const reconstructError = (payload: errors.NativePayload): Error => {
```
pjdotson: this should be in x

pluto/src/aether/store.ts
```diff
@@ -0,0 +1,473 @@
...
+  clearCounter(key: string): void {
+    this.counters.delete(key);
+  }
+}
+
+/** Setter argument accepted by {@link Handle.setState}: a new state value or a function
+ * that derives one from the previous value. */
+export type RawSetArg<State extends z.ZodType<state.State>> =
+  | (z.input<State> | z.infer<State>)
+  | ((prev: z.infer<State>) => z.input<State> | z.infer<State>);
```
pjdotson: shouldn't the types for both of these only be `z.input<State`>? Since the result of this
gets passed to `z.parse`, it should only be the input and not the output.

pluto/src/aether/store.ts
```diff
@@ -0,0 +1,473 @@
...
+
+type Listener = () => void;
+
+/** Live worker component tracked by the store. Generic over its schema so reads through
+ * {@link Store.getEntry} recover the per-entry state and callback types without
+ * re-asserting at every field access. */
+interface Entry<State extends z.ZodType<state.State> = z.ZodType<state.State>> {
```
pjdotson: imo type arguments like this and the above one should be named `StateSchema`

pluto/src/aether/store.ts
```diff
@@ -0,0 +1,473 @@
...
+    }
+    this.worker.handle((msg) => this.handleWorkerMessage(msg));
+  }
+
+  private setError(err: Error): void {
+    if (this.currentError != null) {
+      console.error(
+        "[aether] received new error after error was already set, but before previous error was thrown.",
+      );
+      console.error(err);
```
pjdotson: should just be one `console.error` call with two arguments

pluto/src/aether/aether/message.ts
```diff
@@ -0,0 +1,132 @@
...
+  variant: "update";
+  path: readonly string[];
+  type: string;
+  state: state.State;
+}
+
+/** Main → worker: delete the component at `path`. */
+export interface MainDeleteRequest {
+  variant: "delete";
+  type: string;
```
pjdotson: can you look if we need all of these? e.g. it you look at `handle` in `aether.ts`,
delete, only needs the path and not the type.

pluto/src/aether/main.tsx
```diff
@@ -7,633 +7,287 @@
...
 import { context } from "@/context";
+import { useSyncedRef } from "@/hooks";
 import { useUniqueKey } from "@/hooks/useUniqueKey";
-import { useMemoCompare } from "@/memo";
+import { useMemoPrimitiveArray } from "@/memo";
 import { type state } from "@/state";
-import { Worker } from "@/worker";
 
 export type { CallersFromSchema, EmptyMethodsSchema, MethodsSchema };
+export { Store } from "@/aether/store";
```
pjdotson: should be exported from an external.ts

pluto/src/aether/main.spec.tsx
```diff
@@ -90,28 +102,110 @@ class InvokeLeaf
...
   return [
-    (props: PropsWithChildren) => (
-      <Aether.Provider worker={worker} workerKey="vis" {...props} />
-    ),
+    (props: TestProviderProps) => <Aether.Provider worker={mainSide} {...props} />,
     root,
   ];
 };
 
+class FakeWorker {
```
pjdotson, line 179: is there a reason not to use actual web workers in the tests here?

pjdotson (changes requested):
Mostly looks good

pluto/src/aether/main.spec.tsx
emilbon99: Using real web worker here adds a lot of complexity to the tests with very little real value.

pluto/src/aether/main.tsx
```diff
@@ -7,633 +7,290 @@
...
+  /** URL of the worker script to spawn. Ignored when {@link worker} is set. */
+  workerURL?: string | URL;
+  /** Opt out of spawning a worker. Aether mounts, but all worker-bound operations
+   * no-op. */
+  workerEnabled?: boolean;
+  /** Pre-built comms (e.g. from {@link aether.createMockPair} in tests). Takes
+   * precedence over {@link workerURL} and {@link workerEnabled}. */
+  worker?: MainComms;
+  /** Default timeout for async method invocations. Defaults to 5s. */
+  invokeTimeout?: TimeSpan;
```
pjdotson: should be a `CrudeTimeSpan`

pjdotson (changes requested):

PR #2312, opened 2026-05-12
SY-4158: Refactor Flux to Support React Suspense

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4158](https://linear.app/synnax/issue/SY-4158)

## Description

Adds a Suspense-shaped read path to Flux alongside the existing observable hooks, so consumers can read query results directly with `useRetrieveSuspended(query) -> Data` and gate rendering behind data presence with `useEnsureRetrieved(query)`. Concurrent reads of the same query dedupe to a single fetch through a per-client `QueryCache` keyed by a deterministic hash of the query input. Channel-listener updates push new values into the cache without re-suspending mounted consumers, so the suspending API stays consistent with the observable one.

The substrate's hard dependency on `x/ts/src/status` is the new `status.toError` helper, which bridges a status-shaped failure back to a thrown `Error` (with the original `Status` preserved on `cause`) so suspended retrieves can throw across React error boundaries. `pluto/src/errors/SuspenseBoundary.tsx` bundles `<Suspense>` + `<ErrorBoundary>` and is the recommended consumer-side wrapper.

RFC: `docs/tech/rfc/0036-260507-flux-suspense.md`.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds a Suspense-shaped read path to the Flux query system alongside the existing observable hooks, so consumers can call `useRetrieveSuspended(query)` to return data directly or `useEnsureRetrieved(query)` to pre-warm the cache. Concurrent reads of the same query deduplicate to a single fetch through a per-client `QueryCache` keyed by a deterministic hash of the query input, and channel-listener updates push new values into the cache without re-suspending mounted consumers.

- **`QueryCache<Q,D>`** — new per-`createRetrieve` store with stable `hashQuery` (recursive key-sorting, bigint support, `Hashable` delegation), identity-gated promise settlement (`replaceIfStill`), per-query subscriber map, and `invalidate`.
- **`useSuspended` / `useEnsure`** — suspended hooks integrated into `createRetrieve`; `useSyncExternalStore` return value is now properly used (fixing tearing risk noted in prior review); `undefined` listener values now correctly call `cache.invalidate` instead of being silently dropped.
- **Supporting substrate** — `status.toError` bridges a status-shaped failure to a thrown `Error` for error boundaries; `primitive.Hashable` / `isHashable` support class-instance query fields; `TimeStamp`, `TimeSpan`, `Rate`, `TimeRange`, and `DataType` implement `Hashable`; `Shape` is split into `Query` (hashable) and `Data` (arbitrary state).

The new suspended read path is well-designed and safe to merge; a pre-existing defect in the observable path (handleListenerChange silently discarding undefined setter results) remains unaddressed.

The new QueryCache, useSuspended, useEnsure, and hashQuery implementations are correct and thoroughly tested, addressing all issues from prior review rounds. The one remaining gap is in the pre-existing handleListenerChange callback inside useObservableBase: when a channel listener fires undefined, the observable hooks still silently retain stale data — the fix applied in the new suspended path was not retrofitted here.

pluto/src/flux/retrieve.ts — specifically the handleListenerChange path inside useObservableBase, which was not updated with the undefined-invalidation logic that the new suspended path correctly implements.

| Filename | Overview |
|----------|----------|
| pluto/src/flux/base/queryCache.ts | New QueryCache implementation with stable hash-based deduplication, identity-gated promise settlement via replaceIfStill, per-query subscriber map, and invalidate support. All previously flagged issues addressed. |
| pluto/src/flux/retrieve.ts | Adds useSuspended and useEnsure hooks; useSyncExternalStore return value now properly captured; undefined listener values correctly invalidate the cache. |
| pluto/src/flux/result.ts | LoadingResult gains optional promise and name fields; pendingResult factory added; result creator types split for accurate narrowing. |
| x/ts/src/status/status.ts | toError added to bridge exceptionDetailsSchema status back to a thrown Error with cause preserved. |
| x/ts/src/primitive/primitive.ts | Hashable interface and isHashable type guard added for class-instance query fields. |
| pluto/src/errors/SuspenseBoundary.tsx | New SuspenseBoundary bundles ErrorBoundary + Suspense as the recommended consumer wrapper. |

</details>

1. `pluto/src/flux/retrieve.ts`, line 238-248 ([link](https://github.com/synnaxlabs/synnax/blob/a1e09a79f438f3fbc211ca45d884d3f6bc955359/pluto/src/flux/retrieve.ts#L238-L248)) 

   <a href="#">P1</a> **`useObservableBase` `handleListenerChange` silently drops `undefined` setter results**

   The `handleListenerChange` callback executes the setter through `state.executeSetter`, and if the result is `null`/`undefined` it returns the previous result unchanged. A listener that fires `undefined` to signal "resource deleted" leaves the observable consumer displaying permanently stale data with no re-fetch or error — identical to the problem fixed in the suspended path at line 429–433, which now correctly calls `cache.invalidate(hash)`. The `handleListenerChange` path used by `useRetrieve` / `useRetrieveStateful` is unchanged.

2. `pluto/src/flux/retrieve.ts`, line 241-248 ([link](https://github.com/synnaxlabs/synnax/blob/71f0772725b6a78abff0faef910e51149c644161/pluto/src/flux/retrieve.ts#L241-L248)) 

   <a href="#">P1</a> **`handleListenerChange` silently drops `undefined` setter results**

   `state.executeSetter(value, prev.data)` can return `undefined` when a listener fires to signal a deletion. When that happens, `if (next == null) return prev` leaves the observable consumer (`useRetrieve` / `useRetrieveStateful`) displaying permanently stale data with no re-fetch or error. The equivalent case in the new suspended path was fixed at the `onChange` closure inside `useSuspended` — `undefined` now calls `cache.invalidate(memoQuery)` instead — but `handleListenerChange` was not updated.
</details>

<sub>Reviews (10): Last reviewed commit: ["removed useless type assertion from prim..."](https://github.com/synnaxlabs/synnax/commit/bc8220ca9e1b89d987b4bd840596706572277bc1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31795329)</sub>


x/ts/src/status/status.ts
```diff
@@ -97,6 +97,26 @@ export const fromException = (
...
+ * Use this when bridging the status pipeline back into a context that expects
+ * a real Error — typically before `throw`-ing across an error boundary.
+ */
+export const toError = (
+  s: Status<typeof exceptionDetailsSchema, z.ZodLiteral<"error">>,
+): Error => {
+  const inner = s.details.error;
+  const err = new Error(s.message, { cause: s });
+  err.name = inner.name;
+  if (inner.stack != null) err.stack = inner.stack;
```
pjdotson: shouldn't we always set this, otherwise the stack will be set from where `toError` is called?

pluto/src/flux/base/queryCache.ts
```diff
@@ -0,0 +1,140 @@
...
+import { type destructor } from "@synnaxlabs/x";
+
+import { errorResult, type Result, successResult } from "@/flux/result";
+import { type state } from "@/state";
+
+/// Deterministically serializes a query to a stable string. Keys are sorted
+/// recursively so `{a: 1, b: 2}` and `{b: 2, a: 1}` collapse to the same key.
+/// Works for plain objects, arrays, and primitives. Functions, Maps, Sets, and
+/// other complex types are not supported as query inputs.
+export const hashQuery = (query: unknown): string => {
```
pjdotson: if we only support a certain list of things, (e.g. primitives and arrays / ) we should
type the query to only allow those instead of `unknown`.

pluto/src/flux/base/queryCache.ts
```diff
@@ -0,0 +1,140 @@
...
+/// shape. A single-record retrieve with shape `{key: X}` and a list retrieve
+/// with shape `{workspace, filter}` both live here, addressed by their
+/// respective hashes.
+export class QueryCache {
+  private readonly entries: Map<string, Result<state.State>> = new Map();
+  private readonly subscribers: Map<string, Set<() => void>> = new Map();
+
+  /// Returns the result for the given query hash, or undefined if absent.
+  get<T extends state.State>(hash: string): Result<T> | undefined {
+    return this.entries.get(hash) as Result<T> | undefined;
```
pjdotson: is there a reason not to actually strongly type this query cache and make it per-item?
so you could have an API like this:

```ts
export class QueryCache<Query extends state.State, Data extends state.State> {
    get(query: Query): Result<Data> | undefined;
    set(query: Query, result: Result<Data>, equal?: (a: Data, b: Data) => boolean): void;
    invalidate(query: Query): void;
    subscribe(query: Query, callback: () => void): destructor.Destructor;
}
```

which would mean we have actual type safety.

pluto/src/flux/base/queryCache.ts
```diff
@@ -0,0 +1,140 @@
...
+  ///
+  /// `equal` is an optional value-equality check applied on success-to-success
+  /// transitions. When the new and previous data are equal under it, the entry
+  /// is left in place and no notification fires - subscribers don't re-render.
+  /// Selectors use this to skip re-renders when a derived slice didn't change
+  /// despite the parent record changing.
+  set<T extends state.State>(
+    hash: string,
+    result: Result<T>,
+    equal?: (a: T, b: T) => boolean,
```
pjdotson: is equal going to be a different function for each type or something common? if it is
common we shouldn't allow it as an argument.

pluto/src/flux/retrieve.ts
```diff
@@ -145,6 +153,29 @@ export interface UseRetrieveObservable<
...
+}
+
+/// A Suspense-shaped hook that ensures a query has been retrieved into the
+/// cache and then forgets about it: no cache subscription, no listener
+/// mounting. Use this in a parent that needs to gate rendering on data being
+/// present, while children read the live value via `useRetrieveSuspended` or
+/// a selector (whichever variant subscribes to updates is the right home for
+/// listener-driven freshness). Reach for `useRetrieveSuspended` instead if
+/// the caller itself needs to re-render when the data changes.
+export interface UseEnsureRetrieved<Query extends base.Shape> {
```
pjdotson: why is it a problem if the parent renders but not the children? e.g. I am thinking of
stuff you might see on like Twitter where the recommended toolbar is there but the
actual recommended things haven't loaded yet.

pjdotson (changes requested):

pluto/src/flux/base/queryCache.ts
emilbon99: This means we would have to have this massive map at the top level provider that holds an individual cache for each item. Since you still need to get a query cache for a query from the context, you'd just be moving the type assertion up a level. The current pattern follows what was done in tanstack query, and I think just fits better than setting up individual caches for every single query. 

Not to mention some hooks compose queries, so they'd need to be able to pull several caches at the same time which makes things more complex.

pluto/src/flux/retrieve.ts
emilbon99: Let's talk about this one verbally. I can show you concrete examples.

pluto/src/flux/base/queryCache.ts
emilbon99: I removed.

pluto/src/flux/retrieve.ts
```diff
@@ -355,8 +400,124 @@ export const useObservableRetrieve = <
...
+
+  const promise = retrieve({
+    client: client as AllowDisconnected extends true ? Client | null : Client,
+    query: memoQuery,
+    store,
+    cache,
+  });
+  cache.set(hash, pendingResult(name, promise));
+  use(promise);
+};
```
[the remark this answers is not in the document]
emilbon99: Not an issue.

pluto/src/flux/base/queryCache.ts
emilbon99: Ended up reverting my decision and implementing this.

x/ts/src/primitive/primitive.ts
```diff
@@ -61,6 +61,28 @@ export interface Stringer {
...
+  /** @returns a stable, canonical string representation of the value. */
+  hash: () => string;
+}
+
+/** @returns true if the value implements primitive.Hashable, otherwise returns false. */
+export const isHashable = (value: unknown): value is Hashable =>
+  value != null &&
+  typeof value === "object" &&
+  "hash" in value &&
+  typeof (value as Hashable).hash === "function";
```
pjdotson: why do you need `value as Hashable`?

x/ts/src/primitive/primitive.ts
```diff
@@ -61,6 +61,28 @@ export interface Stringer {
...
+ * Hashable is a duck-typed protocol for class instances that can produce a stable,
+ * canonical string representation of themselves. Implementers commit to: (1) the
+ * returned string uniquely identifies the value's logical content, and (2) two
+ * instances representing the same logical value return equal strings.
+ *
+ * Used by cache-key derivation paths (e.g. flux queryCache) so non-primitive query
+ * fields hash to a stable identifier rather than relying on enumerable own-property
+ * iteration, which is fragile for class instances.
+ */
+export interface Hashable {
```
pjdotson, line 74: is fine but we really should just consider refactoring the telem classes here to be primitives, I don't think any of them have any type of state.

pluto/src/flux/base/client.ts
```diff
@@ -53,6 +62,18 @@ export class Client<ScopedStore extends Store = Store> {
     });
   }
 
+  /// Returns the query cache for the given key, creating one on first use.
+  /// Each createRetrieve operation owns a unique key; reads and writes
+  /// against the returned instance are typed concretely.
+  getCache<Q extends Query, D extends state.State>(key: string): QueryCache<Q, D> {
```
pjdotson, line 68: the type assertions and stuff here are fine for right now, but in the future we probably
want to have some Flux `TypedProvider` concept or something that can provide actually
strongly-typed things for everything, and then there is a top level record of keys to
`TypedProviders` that you can use to make strongly type calls like if we properly
integrated Redux with TypeScript and you can have strongly typed actions.

pjdotson (changes requested):

pluto/src/flux/base/client.ts
emilbon99: I think you have very specific intentions for how you want this sort of high level store/context API designed. This includes the forms. I think you should take some time to do these refactors in the style that you want.

x/ts/src/primitive/primitive.ts
emilbon99: I wonder why the linter doesn't error on this. Regardless, removed.

pjdotson (approved):

PR #2314, opened 2026-05-12
SY-4122: Arc Status STL

sy-nico:
## Linear Issue

[SY-4122](https://linear.app/synnax/issue/SY-4122)

## Description

Implements RFC 0037: redefines `status.set` as a dual-mode (`ExecBoth`) function callable from both Arc func blocks and flow nodes, replacing the deprecated bare `set_status` symbol. Adds a matching `status.delete`. Brings the C++ Arc runtime to parity with the Go runtime so both forms work in the cluster and the driver.

**Arc language changes:**
- `status.set` is now callable as `status.set("key_or_name", "message", "variant")` inside func blocks and as `upstream -> status.set{key_or_name="...", message="...", variant="..."}` in flow chains.
- `status.delete` mirrors the same dual shape: `status.delete("key_or_name")` / `upstream -> status.delete{key_or_name="..."}`.
- Both forms accept a status by UUID key or by human name. On by-name multi-match, `status.set` writes to the first match by key order; `status.delete` removes all matches. Either way, a warning surfaces via the task reporter so the operator is informed without the task failing.
- Variant string literals (`"success"`, `"info"`, `"warning"`, `"error"`, `"loading"`, `"disabled"`) are validated at compile time. Non-literal variants are validated at runtime.
- The deprecated bare `set_status` symbol is removed.
- Empty-string WASM string handles (handle `0`) are now recognized as legitimate empty values rather than corruption, so Arc programs that pass `""` no longer trigger a false `"invalid string handle"` warning.

**Service / API:**
- Added `Service.SetByKeyOrName` and `Service.DeleteByKeyOrName` in `core/pkg/service/status`, plus matching API endpoints (`/api/v1/status/set-by-key-or-name`, `/api/v1/status/delete-by-key-or-name`).
- By-name requests enforce access on the `status` resource type. UUID requests still enforce on the specific row. A row-level grant on a specific UUID no longer authorizes a by-name request; callers with row-level grants must use the by-key form.
- Extracted `identifier.IsKey` in `x/go/identifier` as a shared UUID-form detection helper.
- Centralized `AllowedVariants` in `x/go/status` next to the variant constants so the canonical list has a single source of truth.

**C++ driver / client:**
- C++ Arc runtime gains `status::SetStatus` / `status::DeleteStatus` flow nodes and matching WASM host functions, so the driver can write statuses from Arc programs running in the embedded runtime.
- C++ client adds `statuses.set_by_key_or_name` / `statuses.delete_by_key_or_name`.

**UI:**
- Pluto's "Change Status" SPEC migrates from `set_status` to `status.set`, with its config schema aligned to the Arc symbol's `key_or_name` / `message` / `variant` params.

**Internal cleanup:**
- Reporter message templates hoisted to constants in `core/pkg/service/arc/status/{set,delete}.go` and inline helpers in `driver/arc/status/status.h`, with cross-reference comments so the Go and C++ sides stay in sync.
- `NewModule(ctx, ModuleConfig{...})` replaces the prior 7-arg positional constructor in `core/pkg/service/arc/status`.
- C++ `SetStatus` and `DeleteStatus` output writing standardized to a single `*output(0) = Series(v); *output_time(0) = Series(clock.now())` pattern matching the Go side.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR implements RFC 0037, replacing the deprecated bare `set_status` symbol with a dual-mode `status.set` / `status.delete` pair that works in both Arc func blocks and flow chains. It brings the C++ driver runtime to parity with the Go runtime, adds matching server-side `SetByKeyOrName` / `DeleteByKeyOrName` endpoints, migrates saved console graphs from the old schema, and closes a cross-runtime inconsistency where empty WASM strings were misidentified as invalid handles.

- **Service layer**: `SetByKeyOrName` upserts by key → by name → create in a single `WithTx`; `ErrEmptyKeyOrName` and `ErrInvalidVariant` guard both entry points; all previously flagged non-atomic paths are fixed.
- **String-handle parity**: `Create(\"\")` now returns handle 0 on both Go and C++ runtimes; `Get(0)` returns `(\"\", true)`, eliminating false \"invalid string handle\" warnings.
- **Console migration**: saved `set_status` nodes are rewritten to `status.set` with the new `key_or_name` / `message` / `variant` schema, with tests covering all legacy-field combinations.

Safe to merge; all previously flagged atomicity and empty-key gaps are resolved and test coverage is thorough.

All previously flagged issues are resolved in this revision. The remaining observations are documentation and API-design nits that do not affect runtime correctness.

core/pkg/service/status/writer.go — the exported UpsertByName method has subtly different create behavior from Service.SetByKeyOrName that future callers might not expect.

| Filename | Overview |
|----------|----------|
| core/pkg/service/status/dispatch.go | New SetByKeyOrName and DeleteByKeyOrName service methods; atomicity addressed via WithTx; ErrEmptyKeyOrName guard added for empty inputs |
| core/pkg/service/status/writer.go | Adds Update, retrieveByName, UpsertByName, and DeleteByName; UpsertByName creates rows with random UUID keys, differing from SetByKeyOrName's name-as-key create path |
| core/pkg/service/arc/status/set.go | Major refactor: ExecBoth status.set/delete symbols, compile-time variant validation, WASM host function registration, and taskreporter error surfacing |
| driver/arc/status/status.h | C++ status module rewritten with SetStatus/DeleteStatus flow nodes, WASM host function wiring, and StateConsumer integration for string handle parity |
| arc/cpp/stl/strings/state.h | Empty-string sentinel fix: create/create_config/from_memory return handle 0 for empty input; get/exists treat handle 0 as the empty string |
| arc/go/stl/strings/state.go | Create/CreateConfig return handle 0 for empty strings; Get returns ("", true) for handle 0, fixing false invalid-handle warnings |
| core/pkg/api/status/status.go | Adds SetByKeyOrName and DeleteByKeyOrName API endpoints; access check uses ActionCreate for all upsert operations including updates |
| console/src/arc/types/v2.ts | v2 migration rewrites set_status nodes to status.set with key_or_name/message/variant; tests cover all legacy-field combinations |
| x/go/status/status.go | Adds IsVariant helper as the single source of truth for variant membership |

</details>

<sub>Reviews (14): Last reviewed commit: ["SY-4122: fix stl::str:: -&gt; stl::strings:..."](https://github.com/synnaxlabs/synnax/commit/87b10394a001f0d05f8ae85f4a44fb6c85e55c6f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31848581)</sub>


x/go/identifier/identifier.go
```diff
@@ -0,0 +1,21 @@
...
+
+// Package identifier provides utilities for resolving entity references that
+// may appear as either a UUID-form key or a human-readable name.
+package identifier
+
+import "github.com/google/uuid"
+
+// IsKey reports whether s parses as a UUID-form key. Callers that accept a
+// key-or-name input use it to dispatch between by-key and by-name lookups.
+func IsKey(s string) bool {
```
pjdotson: This won't work. Status keys are not guaranteed to be uuids.

x/go/status/status.go
```diff
@@ -16,6 +16,16 @@ import (
 	"github.com/synnaxlabs/x/gorp"
 )
 
+// AllowedVariants is the canonical list of variant strings.
+var AllowedVariants = []string{
```
pjdotson: this would be better as a function `isVariant` that returns a bool

arc/go/lsp/hover.go
```diff
@@ -327,14 +327,13 @@ var keywordDocs = map[string]string{
...
-	"set_status": deprecatedDoc("set_status", "status.set{}", "sensor -> status.set{status_key=\"ox_alarm\", variant=\"error\", message=\"Overpressure\"}"),
 	"status.set": doc.New(
 		doc.TitleWithKind("status.set", "Function"),
 		doc.Paragraph("Sets a status notification on the cluster. Used to report alarms, warnings, or operational state."),
 		doc.Divider(),
-		arcCode("sensor -> status.set{status_key=\"ox_alarm\", variant=\"error\", message=\"Overpressure\"}"),
+		arcCode("sensor -> status.set{key_or_name=\"ox_alarm\", message=\"Overpressure\", variant=\"error\"}"),
 		doc.Divider(),
-		doc.Paragraph("Accepted variants: success, error, warning, info."),
+		doc.Paragraph("Accepted variants: success, info, warning, error, loading, disabled."),
```
pjdotson: Should figure out some type of way of doing dependency injection here, this doc info
should live with the code for statuses

arc/go/symbol/hooks/hooks.go
```diff
@@ -0,0 +1,24 @@
...
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+// Package hooks declares optional analyzer hook signatures that Symbols may
+// carry. Lives in its own package to keep symbol.Symbol free of any analyzer
+// import cycle.
```
pjdotson: what import cycle?

arc/cpp/stl/stl.h
```diff
@@ -52,6 +55,11 @@ class Module : public runtime::node::Factory {
     /// Registers host functions with the WASM Linker under a named module.
     virtual void bind_to(wasmtime::Linker &linker, wasmtime::Store::Context cx) {}
 
+    /// Injects the runtime's shared string handle state so this module can use
+    /// it from bind_to. Called by the runtime before bind_to for cfg.factories
+    /// that need handle access. Modules that don't need strings ignore this.
+    virtual void set_str_state(std::shared_ptr<str::State>) {}
```
pjdotson: fix the dependency cycle here. `stl` should not depend on `stl/str`.

core/pkg/api/status/status.go
```diff
@@ -94,6 +95,81 @@ func (s *Service) Set(
...
+	}
+	return s.access.Enforce(ctx, access.Request{
+		Subject: auth.GetSubject(ctx),
+		Action:  action,
+		Objects: []ontology.ID{obj},
+	})
+}
+
+// SetByKeyOrNameRequest is a request to upsert a status by UUID key or by name.
+type SetByKeyOrNameRequest struct {
```
pjdotson, line 98: Let's talk through all of the Core changes to the status service synchronously. In
general they are problematic, both the fact that we are adding these new API servers and
the stuff you do inside the Status service.

integration/tests/arc/stl_status.py
```diff
@@ -0,0 +1,224 @@
...
+#  License, use of this software will be governed by the Apache License, Version 2.0,
+#  included in the file licenses/APL.txt.
+
+from typing import Callable, NamedTuple, TypeVar
+
+import synnax as sy
+from framework.utils import create_virtual_channel
+from tests.arc.arc_case import ArcConsoleCase
+
+ARC_STL_STATUS_SOURCE = """
```
pjdotson, line 16: is there a reason you can't define this in a `.arc` file and then just read the file to
a string?

arc/go/text/analyze.go
```diff
@@ -480,7 +480,12 @@ func analyzeFunctionNode(
 		Channels: sym.Channels.Copy(),
 		Config:   slices.Clone(freshType.Config),
 		Outputs:  slices.Clone(freshType.Outputs),
-		Inputs:   slices.Clone(freshType.Inputs),
+	}
+	// STL ExecBoth inputs mirror config; in flow form the upstream is a
+	// trigger, so omit Inputs. User-defined funcs (AST != nil) keep them.
+	upstreamIsTrigger := sym.Exec == symbol.ExecBoth && sym.AST == nil
+	if !upstreamIsTrigger {
+		n.Inputs = slices.Clone(freshType.Inputs)
```
pjdotson, line 502: if any of these changes are unrelated to statuses you might want to cut a quick small PR
with them + tests for them

pluto/src/arc/functions/status/Change.tsx
```diff
@@ -17,10 +17,9 @@ import { Status } from "@/status";
 import { Text } from "@/text";
 
 export const config = z.object({
-  statusKey: z.string(),
+  key_or_name: z.string(),
```
pjdotson, line 20: will this cause an error on stored existing graphical Arcs that use the status node?

pjdotson (changes requested):
Let's get together and talk through these things together. I think the Arc changes are mostly fine, but the status service changes are not and I wanted to make sure the new structures added to the Arc like `AnalyzeCall` and `AnalyzeFlowConfig` are the best solutions here.

arc/go/lsp/hover.go
sy-nico: Implemented. Added a `Doc` field to `symbol.Symbol`, moved the hover string into `core/pkg/service/arc/status/set.go` next to the symbol registration, and the LSP now resolves it via `GlobalResolver` (falling back to the built-in catalog for keywords/operators).

x/go/status/status.go
sy-nico: Replaced with `IsVariant(s string) bool` (exported since `core/` consumes it). The two callsites that needed a human-readable list use a local `allowedVariantsList` const instead.

x/go/identifier/identifier.go
sy-nico: Deleted the whole `x/go/identifier` package and reworked `SetByKeyOrName` / `DeleteByKeyOrName` to `try-key → try-name → create-with-Key=Name=input`, so no UUID-shape detection is needed.

arc/go/symbol/hooks/hooks.go
sy-nico: Inlined `CallHook` / `FlowConfigHook` into `symbol.go` and deleted the hooks package.

arc/cpp/stl/stl.h
sy-nico: Moved `set_str_state` off `stl::Module` into a new `stl::str::StateConsumer` interface; external factories like the status module inherit from both, and the runtime injects via `dynamic_pointer_cast<StateConsumer>`. `stl.h` no longer touches `stl/str`.

arc/go/text/analyze.go
sy-nico: Without this block, `status.set` in flow form fails to analyze (the type's `Inputs` mirror `Config`, so the analyzer would expect N edges and only finds the trigger). The `AST != nil` check just narrows it so user-defined `ExecBoth` funcs keep their input type checking.

Needed for: `time.interval{1s} -> status.set{...} -> log_b -> status.delete{...}`

integration/tests/arc/stl_status.py
sy-nico: Created a new issue for this. I want to a refactor of these tests to decouple from playwright:
[SY-4213: Refactor Arc Integration Tests](https://linear.app/synnax/issue/SY-4213/refactor-arc-integration-tessts)

pluto/src/arc/functions/status/Change.tsx
sy-nico: Yes, and I just verified this by creating a graphical Arc on `origin/rc`, then switching to this branch and relaunching the core. I have fixed by adding a migration.

sy-nico:
PR got tangled and I couldn't cleanly fix with a back-merge or rebase. Going to start over with a clean slate.

PR #2317, opened 2026-05-17
SY-4175: Fix root user role assignment and refactor auth service

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4175](sy-4175-root-user-from-updated-configyaml-has-no-owner-role)

## Description

Refactor user, auth, and RBAC service and clears up behavior for assignment of root users:
- make sure there is only one root user
- make sure that the root user has the `"Owner"` role
- make sure root user will rotate username / password based off of the provided password and key.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR refactors the auth, user, and RBAC services to consolidate credential management and introduce a robust startup-time root-user reconciliation flow. The `InsecureCredentials`/`KV` split is replaced with a single `auth.Service` and a primitive `Writer`, and root-user lifecycle is handled by the new `reconcileRootUser` function in the user service.

- **Auth service rewrite**: `KV` authenticator replaced by `auth.Service` + `Writer`. `hashPassword` now propagates bcrypt errors verbatim (fixing a previous `ErrInvalidCredentials` misclassification); `Authenticate` returns non-`ErrNotFound` storage errors as-is; `ChangePassword` maps `ErrNotFound → ErrInvalidCredentials` so `ensureAuthSync` can correctly distinguish rotate-vs-register paths.
- **Root-user reconciliation**: `provisionRoot` replaced by `reconcileRootUser` which enforces exactly-one-root, username/password rotation, non-root user promotion, and stale-root demotion — all in a single transaction per startup. `rbac.OpenService` gains `assignOwnerToRoots` for idempotent Owner-role assignment.
- **API fixes**: `ChangePassword` in `api/auth/auth.go` now correctly authenticates with the old password before calling `Writer.ChangePassword` with the new one (previously it passed the old password as the new value). `Writer.Create` now returns `(User, error)` and rejects `RootUser=true` through the public API.

Safe to merge. The credential management refactor is well-tested and the root-user reconciliation logic is covered by thorough scenario tests. One writer-level behaviour quirk (self-rename returns ErrRepeatedUsername) is guarded by the API layer but could surprise future callers of the internal writer.

The core credential flows — ChangePassword, hashPassword error propagation, ensureAuthSync, and the full reconcileRootUser startup path — have all been reworked with correct error classification and are backed by new unit tests. No data-loss or auth-bypass paths were found in the changed code.

core/pkg/service/user/writer.go — the ChangeUsername self-rename edge case described in the review comment. core/pkg/api/user/user.go — the old username is read outside the transaction and then used inside it; a concurrent rename between those two points would cause UpdateUsername to fail with ErrNotFound instead of completing the rename.

| Filename | Overview |
|----------|----------|
| core/pkg/service/auth/service.go | New file: replaces KV authenticator with a lean Service struct. Password hashing is now inline bcrypt. ErrInvalidCredentials is returned only for missing-row/bad-password; other storage errors propagate verbatim. SecureCredentials is kept exported to preserve the Gorp key prefix. |
| core/pkg/service/auth/writer.go | New file: Writer primitives (Register, UpdateUsername, ChangePassword, Deactivate) are all identity-free; callers own auth. hashPassword propagates bcrypt errors verbatim instead of wrapping in ErrInvalidCredentials. ChangePassword maps ErrNotFound→ErrInvalidCredentials so ensureAuthSync can distinguish register from rotate correctly. |
| core/pkg/service/user/root.go | New file: startup-time root-user reconciliation. ensureAuthSync correctly distinguishes ErrInvalidCredentials (→rotate or register) from other DB errors (→fail fast). All edge cases covered by comprehensive root_test.go tests. |
| core/pkg/api/auth/auth.go | ChangePassword is now correctly fixed: authenticates with req.Credentials (old password), then calls ChangePassword with {Username, NewPassword}. Previously it used the old password as both the auth check and the new value. |
| core/pkg/service/user/service.go | Adds Auth and RootCredentials to ServiceConfig. Validate() guards Auth being non-nil when RootCredentials.Username is set. provisionRoot replaced by reconcileRootUser which handles all edge cases. Signals was previously missing from layer.go wiring and is now added. |
| core/pkg/service/user/writer.go | Writer.Create now returns (User, error) and rejects RootUser=true via public API. ChangeUsername existence check now runs within the writer's transaction. Private create() bypasses the RootUser guard for startup reconciliation. |
| core/pkg/service/access/rbac/service.go | Adds assignOwnerToRoots called after provisioning. User field added to ServiceConfig with validation. Idempotent Owner-role assignment on every startup. |
| core/pkg/service/layer.go | Layer.Auth field type changed from auth.Authenticator interface to *auth.Service concrete. Signals field correctly wired into user.ServiceConfig (was previously missing). |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[user.OpenService startup] --> B[reconcileRootUser]
    B --> C{RootCredentials set?}
    C -- No --> D[collapseStaleRoots keep 1 root if multiple exist]
    C -- Yes --> E[resolveMatchingRoot]
    E --> F{Matching root found?}
    F -- existing root user with config username --> G[return existing root]
    F -- non-root user with config username --> H[promote to RootUser=true]
    H --> G
    F -- no match --> I[return zero User]
    G --> J[ensureAuthSync]
    I --> J
    J --> K{Authenticate with config creds?}
    K -- success --> L[no-op]
    K -- ErrInvalidCredentials --> M[ChangePassword]
    K -- other error --> N[fail fast]
    M --> O{ChangePassword result?}
    O -- success --> P[log rotation]
    O -- ErrInvalidCredentials user not found --> Q[Register new creds]
    O -- other error --> N
    J --> R{matchingRoot zero?}
    R -- Yes --> S[create root user record]
    R -- No --> T[demote all other roots]
    S --> T
    T --> U[rbac.OpenService assignOwnerToRoots]
```

1. `core/pkg/service/auth/service.go`, line 573-580 ([link](https://github.com/synnaxlabs/synnax/blob/65d3dbe22db9ea043d96873ba19b4e0aea5ba231/core/pkg/service/auth/service.go#L573-L580)) 

   <a href="#">P0</a> **Breaking storage key prefix change on upgrade**

   The `secureCredentials` type was renamed from the previously exported `SecureCredentials`. Gorp derives both the KV key prefix and the migration namespace from the concrete type name via `types.Name[E]()` → `reflect.TypeFor[T]().Name()`. The old code stored every credential row under a key prefix built from `"SecureCredentials"`; the new code looks for credentials under a prefix built from `"secureCredentials"`. On an existing deployment these are two distinct key ranges. Upgrading will silently make every previously registered user credential invisible, causing all users (including the configured root user) to fail authentication until their credentials are re-registered. A `gorp.CodecMigration` or an explicit table-rename migration is required to move entries from the old prefix to the new one before the service opens the new table.

2. `core/pkg/service/auth/writer.go`, line 706-718 ([link](https://github.com/synnaxlabs/synnax/blob/a8efe2e794c3d5a72dde6fcfe0a789ec982fa4ef/core/pkg/service/auth/writer.go#L706-L718)) 

   <a href="#">P1</a> **Unexpected DB errors masked as `ErrInvalidCredentials` in `ChangePassword`**

   When `NewUpdate().Exec()` returns an error other than `query.ErrNotFound` (e.g. a transient DB failure), `errors.Combine(ErrInvalidCredentials, err)` makes the returned error satisfy `errors.Is(err, ErrInvalidCredentials)`. The caller in `ensureAuthSync` uses that check to decide whether to fall through to `Register` — so a real DB error causes `ensureAuthSync` to call `Register`. If the row already exists, `Register` fails with `ErrRepeatedUsername`, and startup aborts with the misleading message "register root credentials: username X already exists". The actual DB error is lost. Non-`ErrNotFound` errors should be returned as-is so `ensureAuthSync` can surface them via the `"rotate root credentials"` path.
</details>

<sub>Reviews (22): Last reviewed commit: ["Fix test flakeyness"](https://github.com/synnaxlabs/synnax/commit/ab4cebc5ecfb8d689f683e5ba0a562d1622952df) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32477842)</sub>


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

core/pkg/service/access/rbac/service.go
```diff
@@ -124,31 +154,63 @@ func OpenService(ctx context.Context, configs ...ServiceConfig) (s *Service, err
...
+				user.OntologyID(r.Key),
+				ownerKey,
+			); err != nil {
+				return errors.Wrapf(err,
+					"assign Owner role to root user %q", r.Username)
+			}
+		}
+		return nil
+	})
+}
```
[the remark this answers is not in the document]
pjdotson: This is fine. The Owner role can be assigned to users besides the root user, so it is not the end of the world if multiple users have it.

pjdotson:
@greptile /rerun and update summary, noting what I've said regarding Owner role and root user reconciliation.

pjdotson:
@greptile /rerun new contract is fine to only be enforced at the API layer.

core/pkg/api/auth/auth.go
```diff
@@ -101,14 +101,19 @@ func (s *Service) Login(ctx context.Context, req LoginRequest) (LoginResponse, e
...
 
 // ChangePassword changes the password for the user with the provided credentials.
 func (s *Service) ChangePassword(ctx context.Context, req ChangePasswordRequest) (types.Nil, error) {
 	return types.Nil{}, s.db.WithTx(ctx, func(tx gorp.Tx) error {
-		return s.authenticator.NewWriter(tx).
-			UpdatePassword(ctx, req.InsecureCredentials, req.NewPassword)
+		if err := s.auth.Authenticate(ctx, tx, req.Credentials); err != nil {
+			return err
+		}
+		return s.auth.NewWriter(tx).ChangePassword(ctx, auth.Credentials{
```
emilbon99, line 114: Nit: Could just do `req.Credentials` here.

core/pkg/service/access/rbac/service.go
```diff
@@ -124,31 +154,63 @@ func OpenService(ctx context.Context, configs ...ServiceConfig) (s *Service, err
...
+			return errors.Wrap(err, "load root users")
+		}
+		w := s.Role.NewWriter(tx, false)
+		for _, r := range roots {
+			if err := w.AssignRole(
+				ctx,
+				user.OntologyID(r.Key),
+				ownerKey,
+			); err != nil {
+				return errors.Wrapf(err,
```
emilbon99: Nit: bad formatting.

core/pkg/service/auth/writer.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (w Writer) assertUsernameAvailable(ctx context.Context, username string) error {
+	exists, err := w.service.table.NewRetrieve().
+		Where(gorp.MatchKeys[string, SecureCredentials](username)).
+		Exists(ctx, w.tx)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return errors.Wrap(
```
emilbon99: Use `errors.Wrapf`. Find any other places in this PR where you made a similar mistake.

emilbon99 (changes requested):
Approved after minor comments addressed.

PR #2324, opened 2026-05-18
SY-4130: Add Python client support for server-side import and export

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4130](https://linear.app/synnax/issue/SY-4130/add-python-client-support-for-server-side-import-and-export)

## Description

Adds Python client support for server-side import and export. A new `sy.Synnax.imex`
client streams resource envelopes to and from the Core:

- `imex.import_(source: FilePath) -> ontology.ID` — streams a file from disk to the
  Core and returns the new resource's ontology ID.
- `imex.export(id: ontology.ID, dest: FilePath) -> None` — streams the exported
  resource straight into a destination file.

To support file streaming, the freighter transport layer gains:

- A new **`FileTransport`** protocol (`upload` / `download`) for streaming a request or
  response body to/from disk when a payload could be too large to buffer in memory. The
  wire content type for a streamed body is negotiated from the file path's extension via
  a small registry in `http.py` (`_FILE_CONTENT_TYPES`), which currently maps only
  `.json` → `application/json`; an unregistered extension raises before any request is
  sent.
- **`HTTPClient`** now implements both `UnaryClient` (`send`) and `FileTransport`
  (`upload` / `download`). A single `codec` still encodes and decodes the typed request
  and response payloads (and supplies the `Content-Type` / `Accept` headers for those);
  for a streamed file body the `Content-Type` (upload) or `Accept` (download) header is
  instead derived from the file path's extension, so the file bytes are streamed
  verbatim and the client never (de)serializes the file. Uploads stream the request body
  from disk in fixed-size blocks; downloads stream the response body straight to disk.
- A new **`x.file.stream_to_file`** helper writes a stream of chunks atomically (temp
  file + `fsync` + `os.replace`), so a mid-stream failure leaves any existing
  destination untouched. It also rejects an empty stream when `allow_empty=False`.

### Retry semantics

- `HTTPClient` accepts a `retries` policy (urllib3 `Retry`) and threads it per-request.
  `send` and `download` retry transient failures; `upload` opts out (`retries=False`).
- Uploads disable retries because an upload mutates server state and is **not assumed
  idempotent**: a transient failure after the Core has begun applying the import could
  replay it on retry and double-apply the data. The opt-out is about request semantics,
  not the stream — urllib3 would in fact rewind the seekable file body and resend it. A
  failed upload surfaces to the caller, leaving the retry decision to them.

### Supporting changes

- `MsgPackCodec` renamed to `MessagePackCodec`; the unused `TracingCodec` and `CODECS`
  list removed.
- `download` and the typed-response path both reject an empty successful response body
  rather than treating it as a valid result — `download` leaves any existing destination
  untouched in that case.

### Testing

- The freighter integration server gains a `/unary/flakyUnavailable` endpoint that
  returns `503` the first time it sees a given message and echoes nominally on every
  retry, plus a `/unary/emptyResponse` endpoint that returns `200` with a genuinely
  empty body. Both are bound as raw Fiber routes (rather than a freighter unary server,
  which maps all handler errors to `400` and never produces an empty body) so they can
  return a genuinely retryable `503` and a genuinely empty `200` respectively.
- Python tests cover `send` recovering via retry, `upload` surfacing the first `503`
  rather than retrying a non-idempotent import, extension-based content negotiation for
  `upload`/`download`, large-body streaming, and rejection of an empty download
  (preserving the existing destination). `x.file.stream_to_file` is covered directly for
  ordering, overwrite, empty-stream, and mid-stream-failure behavior.
- Go integration tests cover the new endpoints' contracts: retryable `503` on first
  sight, recovery on retry with the ID incremented, per-message independence of the
  first-seen state, and a `200` with an empty body.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds Python client support for server-side import and export. The main changes are:

- New `Synnax.imex` client methods for file import and export.
- File upload and download support in the Python HTTP transport.
- Atomic streamed file writes with empty-download rejection.
- Retry behavior tests for unary requests and non-retried uploads.
- Integration endpoints for flaky and empty HTTP responses.

This looks safe to merge.

| Filename | Overview |
|----------|----------|
| freighter/py/freighter/http.py | Adds streamed upload and download support, file content negotiation, retry handling, and empty-response checks. |
| client/py/synnax/imex/client.py | Adds the Python IMEX client methods for importing from and exporting to local files. |
| client/py/synnax/transport.py | Exposes the shared HTTP client as the file transport used by the IMEX client. |
| x/py/x/file/file.py | Adds atomic stream-to-file writes with optional empty-stream rejection. |
| freighter/py/freighter/codec.py | Renames the MessagePack codec and simplifies codec encode/decode helpers. |

<sub>Reviews (18): Last reviewed commit: ["comment formatting fix"](https://github.com/synnaxlabs/synnax/commit/931cae5f37af8517556426d66ca60bbf44db9149) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32654115)</sub>

**Context used:**


freighter/py/freighter/codec.py
```diff
@@ -59,40 +73,14 @@ def decode(self, data: bytes, pld_t: type[P]) -> P:
 class JSONCodec(Codec):
```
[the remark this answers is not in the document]
pjdotson: ignore, this one if fine.

freighter/py/freighter/http.py
```diff
@@ -9,119 +9,267 @@
...
+                return out_ctx, self._decode_error(http_res, http_res.data)
+            if http_res.data is None or len(http_res.data) == 0:
+                return out_ctx, ValueError(
+                    f"expected a non-empty response body from {url!r}"
+                )
+            decoder = self._resolve_decoder(http_res)
+            if isinstance(decoder, Exception):
+                return out_ctx, decoder
+            res_container[0] = decoder.decode(http_res.data, res_t)
+            return out_ctx, None
```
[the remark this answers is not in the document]
pjdotson: ignore: will fix all of these later when we remove the tuple exception return

freighter/py/freighter/http.py
```diff
@@ -9,119 +9,267 @@
...
+
+class FileCodec(Codec, Protocol):
+    """A Codec that also has an on-disk file representation.
+
+    HTTPClient uses file_extension to negotiate a content type from a file path during
+    upload and download. Codecs that only exist on a wire (e.g., a WebSocket framer
+    codec) need only satisfy Codec.
+    """
+
+    def file_extension(self) -> str:
```
emilbon99: Should this just be called `extension`?

freighter/py/freighter/download.py
```diff
@@ -0,0 +1,40 @@
...
+#  included in the file licenses/APL.txt.
+
+from __future__ import annotations
+
+from typing import Protocol
+
+from freighter.transport import RQ, FilePath, Transport
+
+
+class DownloadClient(Transport, Protocol):
```
emilbon99: I don't totally understand why these need a different protocol than `Unary`. Why isn't the DownloadClient something like a `DownloadTransport` that is a concrete implementation of `Unary` whose request type accepts a destination as its file type? Seems like we should be trying to avoid adding different transport types to freighter if possible.

emilbon99 (changes requested):
1) Concerns about the protocol class additions. Why can't we just use `UnaryClient`? The download and upload clients seem like relatively minor variations.
2) What happens when we do import/export of compound data structures i.e. entire system configurations? Do we care about supporting that in the imex client?

freighter/py/freighter/http.py
pjdotson: because this is structural typing `file_extension` makes more sense for the Codecs
implementing this.

pjdotson:
@greptile /rerun

freighter/py/freighter/download.py
pjdotson: One of the things I would really like to avoid here is loading the whole file into
memory. fundamentally that means that HTTPClient will have to handle these cases
different than normal unary request / responses, and will need methods that look
like this:

```py
def upload(self, target: str, req: FilePath, res_t: type[RS]) -> RS: ...
def download(self, target: str, req: RQ, dest: FilePath) -> None: ...
```

We could avoid adding the `FileClient` by just having ImEx directly use `HTTPClient`
instead of a different abstraction (like we do by strongly coupling Framer to the
WebSockets client), but that kinda goes against the purpose of Freighter in the first place.

pjdotson:
@greptile /rerun

pjdotson (commented):
@greptile /rerun

pjdotson:
@greptile /rerun

pjdotson:
> 1. Concerns about the protocol class additions. Why can't we just use `UnaryClient`? The download and upload clients seem like relatively minor variations.

See comment

> 2. What happens when we do import/export of compound data structures i.e. entire system configurations? Do we care about supporting that in the imex client?

Maybe? But it also might be in a project client instead. I don't quite know how this will look yet.

pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

x/py/x/fs/fs.py
emilbon99, line 1: Is it worth standardizing the way we name these packages across languages? i.e. should this be `fs`?

freighter/py/freighter/http.py
```diff
@@ -47,73 +75,164 @@ def __init__(
...
+            http_res, out_ctx = self._request(
+                ctx,
+                url=url,
+                content_type=self._codec.content_type(),
+                accept=_file_content_type(dest),
+                body=self._codec.encode(req),
+                preload_content=False,
+            )
+            try:
+                stream_to_file(http_res.stream(), dest, allow_empty=False)
```
emilbon99, line 148: Why does the typescript implementation of this transport return the stream itself, while the python side writes directly to the file?

freighter/py/freighter/http.py
```diff
@@ -47,73 +75,164 @@ def __init__(
...
+        except MaxRetryError as e:
+            raise Unreachable(url, e.url or "Unreachable") from e
+        out_ctx.params = http_res.headers
+        if not 200 <= http_res.status < 300:
+            try:
+                try:
+                    payload = self._codec.decode(http_res.data, ExceptionPayload)
+                except Exception as e:
+                    raise ValueError(
+                        f"undecodable error response: {http_res.data!r}"
```
emilbon99: I feel like this would read better as `failed to decode error response`

x/py/x/codec.py
emilbon99, line 1: Might be worth taking advantage of the moment and moving these to x…

emilbon99 (changes requested):
Not super happy about a new transport type but I understand why you think it's the right choice. Would like the rest of my comments addressed and another round of reviews before merging.

freighter/py/freighter/http.py
pjdotson: TypeScript needs to take in a general stream so that callers can deal with the fact that in a lot of contexts (like the browser), they don't have direct access to the filesystem themselves, but instead are just have various general ways that they can get "streams of data" they can read. In the Console code we will probably have to have a few different helper functions to deal with reading / writing data in Tauri or the browser, depending on what APIs are available (possibly multiple browser APIs depending on context).

My thought with the Python client is that it would usually be used for scripting to look at some file on the filesystem and then upload it, and this feels a little more Pythonic as it looks more like the `open` API, where you just find some file on a FS and upload it or download to a specific point. What we lose here would be the ability to do some workflow like "connect to some source/sink of data that isn't an FS, and use the Python client as a bridge to move data between these Synnax and here efficiently", but I think that this case is probably pretty uncommon and I would rather have the much nicer API for the more common case of "read/write data to/from a file system".

emilbon99: Ok

emilbon99 (approved):

PR #2329, opened 2026-05-18
SY-4046: Rename Channel via Rename in Arc Context Menu

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4046](https://linear.app/synnax/issue/SY-4046)

## Description

Extends the Arc LSP's Rename request so renaming a channel reference in Arc source also renames the underlying Synnax channel resource. Previously, Rename on a channel identifier returned nil (channels resolve through the global resolver and have no AST), so the operation was a no-op even though the menu item appeared enabled.

The LSP now exposes a generic `OnRename` callback in its `Config`. The handler runs after a rename request resolves to a valid symbol and before source-text edits are produced, and is given the resolved `*symbol.Scope` plus the old and new names so it can decide for itself whether (and how) to propagate the change. Rename on channels is gated on this callback being wired, so a host without channel-propagation logic still gets the old "source-defined symbols only" behavior. The core arc service wires a callback that delegates to `channel.Service.Rename` for `KindChannel` symbols (the channel symbol carries the channel key in `Symbol.ID`, so no name lookup is required); the callback lives in `core/pkg/service/arc` to keep `channel` free of LSP-shaped dependencies.

Out of scope for this PR: cascading the rename to other saved Arc programs that reference the channel by its old name. That can be added later in a separate change.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR extends the Arc LSP rename flow to support channel symbols: a new `OnRename` callback hook in `lsp.Config` is called after reference collection and before text edits are emitted, allowing the host to propagate the rename to the backing Synnax channel resource. The `channelResolver` opts channels into the flow by setting `Renameable: !ch.Internal`, and `Service.NewLSP()` wires a `channelRename` callback that delegates to `channel.Service.Rename` using the key stored in `sym.ID`.

- `arc/go/lsp/rename.go` — `isRenameable` now admits global symbols with `Symbol.Renameable = true`; `findAllReferences` is extended to match global symbols by kind + name (no AST); `OnRename` is invoked between reference collection and text-edit production to ensure the old name is still resolvable during collection.
- `core/pkg/service/arc/rename.go` — new `channelRename` factory produces the `lsp.OnRename` callback; it is a no-op for non-channel kinds and uses `allowInternal: false` as a safety net.
- `core/pkg/service/arc/symbol/resolver.go` — `channelToSymbol` now populates `Renameable: !ch.Internal`, correctly gating rename capability on the channel's internal flag.

The production path through `Service.NewLSP()` always wires `OnRename` and validates `Channel` non-nil, so the Synnax rename behavior is correct end-to-end.

The open `OnRename == nil` gating issue from the prior review remains unaddressed: `isRenameable` admits resolver-supplied `Renameable` symbols regardless of whether `OnRename` is wired, contradicting the documented API contract that `nil` restricts rename to source-defined symbols. A library caller who provides a `GlobalResolver` with `Renameable=true` symbols but omits `OnRename` will get text edits applied while the backing resource goes unmodified, leaving the document broken.

arc/go/lsp/rename.go — `isRenameable` and both `PrepareRename`/`Rename` call sites do not gate resolver-supplied renameables on `OnRename != nil`.

| Filename | Overview |
|----------|----------|
| arc/go/lsp/rename.go | Core rename logic extended for global (resolver-supplied) symbols; `isRenameable` now allows `sym.Renameable` symbols but does not gate on `OnRename != nil`, leaving the documented API contract unenforced (flagged in prior review). |
| arc/go/lsp/rename_test.go | Adds comprehensive unit tests for OnRename: invocation with correct args, multi-reference text edits, error abort, source-defined symbols, and Renameable=false rejection. Test names are consistent with assertions. |
| arc/go/lsp/server.go | Adds `OnRename` type and `Config.OnRename` field; wired cleanly into `Override` with `override.Nil`. No behavioral changes to existing code. |
| arc/go/symbol/symbol.go | Adds `Renameable bool` to `Symbol` with clear doc comment distinguishing it from source-defined (AST-backed) symbols. Clean, non-breaking addition. |
| core/pkg/service/arc/rename.go | New `channelRename` factory wires an `lsp.OnRename` callback that delegates to `channel.Service.Rename` using the key stored in `sym.ID`; correctly ignores non-KindChannel symbols and uses `allowInternal=false` as a safety net. |
| core/pkg/service/arc/rename_test.go | Integration tests validate end-to-end channel rename and rejection for internal channels, using real distributed channel service. |
| core/pkg/service/arc/service.go | Wires `channelRename(s.cfg.Channel)` into `NewLSP`; `Channel` is validated non-nil in `ServiceConfig.Validate()` so the callback is always present in production. |
| core/pkg/service/arc/symbol/resolver.go | Sets `Renameable: !ch.Internal` in `channelToSymbol`, correctly gating rename capability on the channel's internal flag. |
| core/pkg/service/arc/symbol/resolver_test.go | Adds tests verifying that regular channels are `Renameable=true` and internal channels are `Renameable=false`. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant C as LSP Client
    participant S as arc LSP Server
    participant GR as GlobalResolver
    participant OR as OnRename callback
    participant CS as channel.Service

    C->>S: PrepareRename(position)
    S->>GR: Resolve symbol at position
    GR-->>S: "symbol.Scope (Renameable=true)"
    S-->>C: Range (rename enabled)

    C->>S: Rename(position, newName)
    S->>GR: Resolve symbol at position
    GR-->>S: targetSym (Scope)
    S->>S: findAllReferences(doc, targetSym)
    S->>OR: OnRename(ctx, sym, oldName, newName)
    OR->>CS: NewWriter(nil).Rename(ctx, Key(sym.ID), newName, false)
    CS-->>OR: nil / error
    OR-->>S: nil / error
    alt OnRename error
        S-->>C: nil, err
    else success
        S->>S: Build TextEdit list
        S-->>C: WorkspaceEdit
    end
```

1. `arc/go/lsp/rename.go`, line 39-53 ([link](https://github.com/synnaxlabs/synnax/blob/e042512e0ffe241de83e3606e3ab354e55ccfc47/arc/go/lsp/rename.go#L39-L53)) 

   <a href="#">P1</a> **`OnRename == nil` doesn't enforce the documented restriction**

   `Config.OnRename`'s doc says "When nil, rename is restricted to source-defined symbols." But `isRenameable` returns `sym.AST != nil || sym.Renameable`, so a global symbol with `Renameable: true` passes the check even when `OnRename` is nil. `PrepareRename` will then return a valid range (enabling the rename UI item), and `Rename` will produce text edits renaming every reference in the source — while silently skipping the out-of-band resource update. The result is that Arc source now references a name the channel resolver cannot find, breaking the document.

   The fix is to gate resolver-supplied renameables on `OnRename` being wired, e.g. add the check in `PrepareRename` and `Rename` right after `isRenameable`:
   ```go
   // resolver-supplied symbols require OnRename to propagate the backing-resource rename
   if sym != nil && sym.AST == nil && s.cfg.OnRename == nil {
       return nil, nil
   }
   ```
</details>

<sub>Reviews (3): Last reviewed commit: ["added rename channel via arc lsp"](https://github.com/synnaxlabs/synnax/commit/4f939e7c6e2e26aee0a096a26d6b10065af3966d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32689903)</sub>


arc/go/lsp/rename_test.go
```diff
@@ -286,4 +289,163 @@ func main() {
...
+			OpenArcDocument(server, ctx, uri, channelContent)
+
+			_, err := server.Rename(ctx, &protocol.RenameParams{
+				TextDocumentPositionParams: protocol.TextDocumentPositionParams{
+					TextDocument: protocol.TextDocumentIdentifier{URI: uri},
+					Position:     protocol.Position{Line: 1, Character: 14},
+				},
+				NewName: "renamed_sensor",
+			})
+			Expect(err).To(MatchError(renameErr))
```
pjdotson: should call via `Expect(server.Rename(...)).Error().To(...)`

pjdotson (changes requested):

PR #2330, opened 2026-05-18
SY-4182: Fix Missing Custom Schematic Symbol Handling

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4182](https://linear.app/synnax/issue/SY-4182)

## Description

Importing or copying a schematic whose custom symbols are not present on the
destination cluster used to break the editor in two distinct ways:

1. Every custom-symbol node fired a `Failed to retrieve schematic symbol`
   toast on every render, producing a wall of status notifications.
2. Dragging or editing properties on a custom symbol caused the property
   panel to enter a setConfig loop that pegged the main thread and made the
   editor unusable.

This PR fixes both:

- **Missing-symbol UX.** Treat a missing custom-symbol reference as a
  first-class state instead of a recoverable fetch error. The canvas renders
  a dashed placeholder for missing nodes, and the property panel swaps in a
  re-link form (`MissingSymbolForm`) that lets the user point the node at a
  symbol that does exist on the cluster. Symbols arriving via peer sync or
  re-import auto-heal through the existing `schematicSymbols` store
  listeners.
- **Property-panel render loop.** `Form.Form` was passing a freshly-spread
  object as the React context value every render
  (`<Context value={rest as ContextValue}>`). Every consumer that read
  `Form.useContext()` got a new reference each render, so any
  `useCallback(..., [form])` was unstable, `useRetrieveEffect` inside
  `StateOverrideForm` re-fired on every render, each fire wrote
  `stateOverrides` back into the form, and the form's `onChange` dispatched
  `setConfig` — which updated the schematic store and re-rendered the
  property panel, restarting the loop. Memoize the context value so
  `useContext` returns a stable reference, and the cascade stops.
- **Custom-symbol render cost.** `Custom.useRender` was stashing a
  `deep.copy` of the spec and state into its change-detection refs, so the
  identity check on the next render always reported "different" and
  re-ran `applyState` (which queries the DOM per region and re-sets stroke
  / fill attributes). The symbol cache is immutable, so store the original
  reference and the check is now correct.
- Drop the canvas-side `MissingSymbolsBanner` / `useAuditCustomSymbols`
  pair: per-node placeholder plus the property-panel re-link form covers
  the same UX with less plumbing.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes two distinct bugs that made the schematic editor unusable when custom symbol references couldn't be resolved: a toast-notification storm on every render, and a `setConfig` dispatch loop that pegged the main thread. It also refactors the `useRender` hook from an inline render-phase hook to a proper `RefCallback` so that auto-heal correctly mounts the SVG.

- **`Form.Form` context memoization** (`pluto/src/form/Form.tsx`): The render-loop root cause is fixed by memoizing the context value object so `useContext` consumers receive a stable reference across renders.
- **`useRender` → `RefCallback`** (`render.ts`, `Primitive.tsx`x2, `Preview.tsx`): The hook now returns a stable ref callback; attach/detach lifecycle is handled by the callback, fixing the blank-canvas-after-resolve issue.
- **Missing-symbol UX** (`MissingSymbolForm`, `Properties.tsx`, `symbol/queries.ts`): A per-node warning placeholder replaces the broken node canvas; the property panel swaps in `MissingSymbolForm` with re-link and create-new flows.

Safe to merge — the three independent fixes are well-scoped and the test suite was updated to match the new API.

All changed paths are well-tested: render.spec.ts covers attach, detach, re-attach, and the auto-heal transition; registry.spec.ts replaces the deleted Symbols.spec.ts with equivalent coverage. The Form.Form memoization change is minimal and correct. No data-loss or security-relevant code paths are touched.

No files require special attention. The feedback on silent canvas failures for non-NotFoundError cases is a quality-of-life gap, not a functional regression.

| Filename | Overview |
|----------|----------|
| pluto/src/form/Form.tsx | Adds useMemo to stabilize the context value object — the root fix for the setConfig dispatch loop. |
| pluto/src/schematic/node/common/custom/render.ts | Refactors useRender from an inline hook to a RefCallback; fixes specDiffers check by storing original references; switches stateOverrides comparison from JSON.stringify to deep.equal. |
| pluto/src/schematic/node/general/customStatic/Primitive.tsx | Adopts the new useRender ref-callback API and Symbol.isMissing guard; renders a warning placeholder when the symbol is missing. |
| pluto/src/schematic/node/general/customActuator/Primitive.tsx | Same changes as customStatic/Primitive.tsx — adopts ref-callback API and isMissing guard. |
| console/src/schematic/toolbar/MissingSymbolForm.tsx | New component: shows a re-link selector and create-new-symbol flow in the property panel when a custom node's symbol cannot be resolved. |
| console/src/schematic/toolbar/Properties.tsx | Adds CustomVariantForm wrapper that checks isMissing and renders MissingSymbolForm or the real variant form; memoizes initialValues with deep.copy. |
| pluto/src/schematic/symbol/queries.ts | Adds isMissing helper that checks whether a Flux.Result is an ErrorResult wrapping a NotFoundError. |
| pluto/src/schematic/node/registry.ts | Adds CUSTOM_VARIANTS, isCustomVariant, isCustomConfig, and STATIC_SPECS exports, moving custom-variant bookkeeping into the shared registry. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Custom node renders] --> B{Symbol.useRetrieve result}
    B -->|loading| C[Render empty container]
    B -->|success| D[spec available]
    B -->|error NotFoundError| E[Render warning Note placeholder]
    B -->|error other| C
    D --> F[Custom.useRender RefCallback]
    C --> F
    F -->|ref attaches| G[runRender mounts SVG]
    F -->|ref detaches| H[SVG removed, state reset]
    H -->|ref re-attaches| G
    E --> I[Property panel CustomVariantForm]
    I -->|isMissing true| J[MissingSymbolForm shown]
    I -->|isMissing false| K[VariantForm shown]
    J -->|Re-link picker| L[form.set specKey and stateOverrides]
    J -->|Create new symbol| M[createEditLayout with createKey]
    M --> N[Server creates symbol under dangling key]
    N --> O[schematicSymbols store listener fires]
    O --> P[useRetrieve transitions to success]
    L --> P
    P --> G
```

<sub>Reviews (4): Last reviewed commit: ["tuning"](https://github.com/synnaxlabs/synnax/commit/b94afd7febea75c9d7d0272e54c8eeefed685fb1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32695868)</sub>


pjdotson (changes requested):
Approved when CI passes

PR #2332, opened 2026-05-18
SY-4190: Arc - Require Module Imports

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4190](https://linear.app/synnax/issue/SY-4190/make-arc-module-import-required)

## Description

Implements the `import` syntax for Arc per RFC §5. Qualified moduleaccess (`time.now`, `authority.set`, `status.set`, …) now requires anexplicit `import` block at the top of the file. Bare deprecated
aliases (`now`, `set_authority`, `interval`, `wait`, …) stay auto-imported.

Supported forms: single, block, alias:

    import time 
    import ( 
        time
        authority 
        status
    )
    import time as t 

### Implementation

- **Grammar**: new `IMPORT` / `AS` tokens and `importStatement` rule.
- **Symbol layer**: `ImportSet` on the root `Scope`. `Resolve` gates dotted lookups, returning `ModuleNotImportedError` when the alias is unbound. Aliases rewrite the lookup (`t.now` → `time.now`).
- **Analyzer**: new `collectImports` pass diagnoses duplicates, unknown modules, and unused imports. The authority-must-be-first check now skips import statements so `import (…) / authority …`  is accepted. Graph mode and LSP block snippets leave the import set nil and stay ungated.
- **Compiler + IR**: canonical-name rewriting before WASM emit and before populating `Node.Type`, so aliased calls dispatch to the real module at runtime.
- **Formatter, LSP, TextMate grammar, syntax.mdx**: `import` and `as` treated as keywords; completion snippet + hover doc; new "Imports" section in syntax.mdx.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR implements the `import` syntax for Arc per RFC §5, requiring explicit `import` blocks before using qualified module members (`time.now`, `authority.set`, etc.). Bare deprecated aliases remain auto-imported for backwards compatibility.

- **Grammar/Lexer**: New `IMPORT` and `AS` tokens with an `importStatement` rule supporting bare, block, and aliased forms; `AUTHORITY` is accepted as an importable name.
- **Symbol layer**: `ImportSet` on the root `Scope` gates dotted-name lookups and rewrites aliases to canonical paths before resolver dispatch; nil `Imports` means ungated (graph mode, LSP snippets).
- **Analyzer**: New `collectImports` pass populates the `ImportSet` and diagnoses duplicates/unknowns; `reportUnusedImports` flags unused imports; the authority-must-be-first check skips import statements.
- **Formatter/LSP**: Collapse logic for single-item parens, multi-line expansion for two or more, plus keyword docs, hover, and TextMate grammar updates.

Safe to merge; all semantic and resolution logic is correct and well-tested end-to-end.

The core import gating, alias rewriting, canonical-name propagation to IR, and unused/duplicate/unknown diagnostics are correctly implemented and thoroughly covered. The one gap is a formatter edge case where import ( authority time ) would print both modules on the same line — formatting only, semantic meaning preserved.

arc/go/formatter/printer.go — the multi-line import newline logic does not account for AUTHORITY as a line-terminating token when it precedes another module in a block.

| Filename | Overview |
|----------|----------|
| arc/go/symbol/imports.go | New ImportSet type with nil-safe methods; MarkUsed is called at the gate site so failed GlobalResolver lookups don't produce spurious unused import errors. |
| arc/go/symbol/scope.go | Resolve gates dotted names on ImportSet and rewrites alias prefixes; MarkUsed is correctly placed before the GlobalResolver call. |
| arc/go/analyzer/imports.go | collectImports always initialises Imports to a non-nil gated set; unknown modules are immediately marked used to suppress spurious unused diagnostics. |
| arc/go/analyzer/imports_test.go | Good coverage overall; two test descriptions are misleading — one contradicts its own assertion, another claims to test a nil path it never reaches. |
| arc/go/formatter/printer.go | Import paren formatting is correct for most cases but the handleDefault newline check misses AUTHORITY as a line-terminating token in multi-item blocks. |
| arc/go/parser/ast.go | ImportEntry, ImportPathText, ImportAlias, and Imports helpers are clean and well-tested. |
| arc/go/import_test.go | Comprehensive end-to-end coverage including alias rewriting in IR node types and authority-block ordering. |
| integration/tests/arc/edge_cases.py | Adds import diagnostic integration cases; replaces interval{} triggers with channel triggers throughout. |

</details>

<sub>Reviews (2): Last reviewed commit: ["SY-4190: Allow bare imports, update form..."](https://github.com/synnaxlabs/synnax/commit/12816d38182137d7c837a1f11783e316d77b3177) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32716673)</sub>


arc/go/analyzer/imports.go
```diff
@@ -0,0 +1,75 @@
...
+// duplicates and unknown modules. Must run before any resolution pass.
+func collectImports(ctx acontext.Context[parser.IProgramContext]) {
+	imports := symbol.NewImportSet()
+	root := ctx.Scope.Root()
+	root.Imports = imports
+
+	known := knownModuleSet(root.GlobalResolver)
+
+	for _, entry := range parser.Imports(ctx.AST) {
+		rec := &symbol.ImportRecord{
```
emilbon99: Can you explain to me why an `import` is not just another type of symbol i.e. it becomes like a `module` symbol? I feel like that would be useful for things like validating a module is imported.

arc/go/symbol/imports.go
```diff
@@ -0,0 +1,145 @@
...
+	Path  string
+	Alias string
+	AST   antlr.ParserRuleContext
+	Used  bool
+}
+
+// ImportSet gates dotted-name lookups on the root Scope. AutoAll skips
+// the gate (graph mode, REPLs, anywhere without source-level imports).
+type ImportSet struct {
+	aliases map[string]*ImportRecord
```
emilbon99: Why do we need to store pointers here? It's generally recommended not to store maps of pointers or slices of pointers, since they are already held by reference.

arc/go/symbol/imports.go
```diff
@@ -0,0 +1,145 @@
...
+	autoAll bool
```
emilbon99: Can you add a doc comment to explain what this is for?

arc/go/text/analyze.go
```diff
@@ -471,10 +471,16 @@ func analyzeFunctionNode(
 		))
 		return nodeResult{}, false
 	}
+	// Node.Type must be the canonical module path so factories find it;
+	// rewrite aliased prefixes (`t.now` → `time.now`).
+	nodeType := name
+	if root := ctx.Scope.Root(); root.Imports != nil {
```
emilbon99: Should `root.Imports` ever be nil? You could also explicitly handle the nil case within root.Imports. It would be much nicer if this API did not need this `nil` check.

arc/go/compiler/expression/compiler.go
```diff
@@ -252,6 +252,9 @@ func compileFunctionCallExpr(
 	if scope.Deprecated != "" {
 		emitName = scope.Deprecated
 	}
+	if root := ctx.Scope.Root(); root.Imports != nil {
```
emilbon99: See my other comments. Would be nice to make it so root.Imports is never nil and/or it handles the nil case internally. Would remove several of these duplicated nil checks.

arc/go/formatter/formatter_test.go
emilbon99, line 1: Imports should be formatted in the same style as follows. 

```
// single import, no parenthesis
import time

// single import, alias
import time as t

// multiple imports
import (
     time
     math
)

// multiple imports, aliased
import (
     time as t
     math
)
```

This follows the conventions of (a) the authority blocks and (b) other languages that use this import style.

emilbon99 (changes requested):
Syntax looks fine
Formatting needs to be fixed

arc/go/analyzer/imports.go
sy-nico: I looked into this but it got messy:

1. `Symbol` would need a `Path` field meaningful only for `KindModule`. The  alias goes in `Name`, but the canonical path ("time") has no home. Adding `Path string` adds noise to a struct shared by ~14 other kinds.

2. "Used" tracking has the same problem. A `Used bool` on `Symbol` only applies to one `Kind`.

3. `CanonicalName`, `AutoAll`, and `Unused` don't naturally hang off `Symbol`. They want a container that knows about all imports together,  which is what `ImportSet` is.

arc/go/text/analyze.go
sy-nico: Fixed. `ImportSet` methods are nil-safe internally now (every method checks `s == nil` and returns a sensible zero), so the call sites just do `ctx.Scope.Root().Imports.CanonicalName(name)` without a guard.

arc/go/compiler/expression/compiler.go
sy-nico: Implemented

arc/go/symbol/imports.go
sy-nico: Fixed by splitting the metadata from the mutable state.`ImportRecord` is now an immutable value type, `ImportSet` holds them by value, and "used" lives in a separate `set.Set[string]`. Matches the `MapResolver` pattern already in this package.

arc/go/formatter/formatter_test.go
sy-nico: Updated parser to allow for imports without parens, updated formatting, fixed docs.

arc/go/analyzer/imports.go
emilbon99: Copy, thanks for the explanation.

emilbon99 (changes requested):
1) There should be a positional constraint on the `import` block. It has to be the first thing in the file (except for comments). They should be required before the `authority` block IMO.
2) Should consider whether or not we allow multiple `import` statements in one file. Should the formatter auto-collapse?
3) LSP auto-complete should probably work better. If I type `time.now` and hit enter it should auto import time right? 
4) Formatting doesn't seem to be working properly. This:

```
import (
    time
)

func cat() {
}
```

Formatted to this

```
import
time
func cat() {}
```

Import block bad and there should be a space between import block and function declaration.
5) Should formatting auto-remove unused imports? Probably?

PR #2339, opened 2026-05-20
SY-4223: Expand TimestampFormat Enum and Fix shortDate Format

pjdotson:
## Description

Expands the Oracle `TimestampFormat` enum from a three-value subset to the
eight formats actually supported by `TimeStamp.toString`, so the schema and
the renderer are aligned. Removes `ISOTime` (which produced output identical
to `preciseTime`) and adds a missing `case "shortDate"` to the `toString`
switch — previously `toString("shortDate", ...)` silently fell through to
the default ISO 8601 output, contradicting the documented `"Nov 5"`
behavior and breaking `formatBySpan` for spans ≥ 30 days.

### Linear issue

- [SY-4223](https://linear.app/synnax/issue/SY-4223/add-timestampformat-type-to-x)

### Changes

- `schemas/telem.oracle` — `TimestampFormat` enum gains `ISODate`, `time`,
  `date`, `preciseDate`, `dateTime`, `shortDate`; loses `ISOTime`.
- `x/go/telem/types.gen.go`, `x/ts/src/telem/types.gen.ts` — regenerated.
- `x/ts/src/telem/telem.ts`:
  - `TimeStampStringFormat` becomes a type alias for the generated
    `TimestampFormat`, so the renderer and the persisted schema can no
    longer drift.
  - `toString` switch handles `"shortDate"` (renders as `dateString()`,
    e.g., `Dec 15`).
  - `formatBySpan` returns `preciseTime` for sub-second spans instead of
    `ISOTime`; the two produced identical output, so behavior is
    unchanged.
- `x/ts/src/telem/external.ts` — re-exports `TIMESTAMP_FORMATS`,
  `timestampFormatZ`, `TimestampFormat`.
- `x/ts/src/telem/telem.spec.ts` — drops the `ISOTime` test case, adds a
  `shortDate` test case, updates `formatBySpan` expectations.

### Why not the bigger refactor

A cleaner long-term shape would split format from precision (so users can
ask for nanosecond resolution on any time-bearing format) and drop the
date-only formats that don't include the year. Out of scope here — this
PR is the minimum cleanup that lets logs (and any other consumer) pick
from the full set of formats the renderer can already produce.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR expands the `TimestampFormat` enum from the 3-value subset that lived locally in `pluto/log` to all 7 formats actually supported by `TimeStamp.toString`, and fixes a long-standing silent fall-through bug where `formatBySpan` returned `"shortDate"` for ≥30-day spans but `toString` had no matching case.

- **Schema & generated types** — `telem.oracle` gains `ISODate`, `time`, `date`, `preciseDate`, `dateTime`; drops `ISOTime` (identical to `preciseTime`). Both `x/ts/src/telem/types.gen.ts` and `x/go/telem/types.gen.go` are regenerated accordingly.
- **Core fix in `telem.ts`** — `TimeStampStringFormat` is removed in favour of the schema-derived `TimestampFormat`, the broken `"shortDate"` branch is replaced by `"date"` (which has a real `toString` case), and `formatBySpan` is simplified from 5 branches to 4.
- **New UI component** — `SelectTimestampFormat` replaces the hard-coded 3-option selector in `console/log/toolbar/Channels.tsx` with a reusable picker covering all 7 formats.

Safe to merge; the core bug fix (shortDate silent fall-through) is correct and all enum values are consistent across schema, generated code, and renderer. One outstanding concern is that dateString() — now the output function for the primary ≥30-day format — always uses the host local timezone regardless of the timeZone argument, which was already flagged on this PR.

The schema expansion and type unification are clean, and the shortDate→date rename correctly closes the fall-through gap. The dateString() timezone-blindness issue affects the new default format for long-span display; a timestamp at midnight UTC can silently render as the previous day in a UTC-negative environment.

x/ts/src/telem/telem.ts — the date/preciseDate/dateTime cases all delegate to dateString(), which ignores the timeZone parameter passed to toString().

| Filename | Overview |
|----------|----------|
| schemas/telem.oracle | Adds the TimestampFormat enum with 7 values; shortDate is intentionally absent as it is collapsed into date. |
| x/ts/src/telem/telem.ts | Removes TimeStampStringFormat type alias; fixes silent shortDate fall-through by replacing with date; simplifies formatBySpan from 5 to 4 branches. The date/preciseDate/dateTime cases call dateString() which ignores timeZone (pre-existing, already flagged). |
| x/ts/src/telem/types.gen.ts | Generated file adds TIMESTAMP_FORMATS tuple, timestampFormatZ, and TimestampFormat type mirroring the 7-value oracle schema. |
| pluto/src/telem/SelectTimestampFormat.tsx | New component exposing all 7 formats, merging className from dialogProps correctly. |
| pluto/src/log/aether/Log.ts | Drops local 3-value timestampFormatZ in favour of the shared one from @synnaxlabs/x; backward-compatible. |
| console/src/log/toolbar/Channels.tsx | Replaces inline 3-option TimestampFormatSelect with Telem.SelectTimestampFormat (7 options). |
| pluto/src/select/Dialog.tsx | Forwards className prop through CSS() so callers can append custom classes without losing the base class. |
| x/ts/src/telem/telem.spec.ts | Drops ISOTime test case and updates formatBySpan expectations to match the simplified 4-branch logic. |
| pluto/src/telem/SelectTimestampFormat.spec.tsx | Tests render, all 7 format options, and onChange callback. |

</details>

1. `x/ts/src/telem/telem.ts`, line 206-211 ([link](https://github.com/synnaxlabs/synnax/blob/3f3f11adbcfde8a09fb9472a9abdaa55b4097b6d/x/ts/src/telem/telem.ts#L206-L211)) 

   <a href="#">P1</a> **`dateString()` ignores the `timeZone` argument**

   `toString` accepts a `timeZone` parameter and routes it to `timeString`, but the `"date"`, `"shortDate"`, `"preciseDate"`, and `"dateTime"` cases all call `dateString()`, which always uses `this.date().toLocaleString(...)` against the host's local timezone. A timestamp at `2022-12-15 01:00 UTC` in a UTC-5 environment would render `"Dec 14"` for the date portion while the time portion correctly renders `"01:00:00 UTC"`. This inconsistency predates this PR, but the PR adds `"shortDate"` to the same code path, making it the primary format for long-span date display, which makes the discrepancy more impactful.
</details>

<sub>Reviews (11): Last reviewed commit: ["Merge branch &#39;rc&#39; into sy-4223-add-times..."](https://github.com/synnaxlabs/synnax/commit/dc74ea47b15da5d5f9dcd7a30f289abbe93c482a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32924221)</sub>


emilbon99 (changes requested):
Do you plan on making the `TimeStamp` type itself be `Timestamp`? What's the naming convention?

pjdotson:
> Do you plan on making the `TimeStamp` type itself be `Timestamp`? What's the naming convention?

Yes, "timestamp" is one word so we should make the types be one word: https://en.wikipedia.org/wiki/Timestamp. In contrast, "time span" and "time range" should stay as two words. https://linear.app/synnax/issue/SY-4224/rename-timestamp-to-timestamp-across-codebase.

emilbon99 (approved):

PR #2340, opened 2026-05-20
SY-4199: Arc Select Update Part 2 -  Inline Stage/Sequence Bodies

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4199](https://linear.app/synnax/issue/SY-4199/arc-select-updates)

## Description

Adds support for inline (anonymous) `stage { ... }` and `sequence { ... }` bodies as routing-table case bodies in Arc, enabling parallel or sequential actions.

```arc
sensor -> decide{low=30.0, high=80.0} -> {
    vent: stage { 1 -> vent_cmd, "vent" -> log },
    press: sequence {
        1 -> press_cmd
        0 -> press_cmd
        wait{2s} => next
    },
    abort: abort_handler
}
```

**Behavior**

- **Anonymous only.** `yes: stage foo { ... }` is a compile error. Inline bodies cannot be referenced by name from sibling flows.
- **No auto-advance on completion.** Firing and completing an inline routing case body does not advance the enclosing sequence to its next stage. This is consistent with the named-scope form (`high: alarm`) and is verified by a regression test.
- **Cross-stage navigation works.** An inline body can transition to a peer top-level stage as its terminal write, e.g. `true: stage { time.wait{1s} -> following_stage }`. The synth scope's transition resolves through the same activation pathway as a hand-written `=> following_stage`.
- **`next` is rejected inside inline bodies at any depth.** Because synth scopes are root-level siblings rather than children of the enclosing sequence, there is no `next` target to resolve to. The analyzer rejects `=> next` whether it appears in a top-level inline, a doubly-nested inline, or deeper.
- **Arbitrary N-level nesting.** Inline routing tables may appear inside other inline routing tables. Each level activates in the same scheduler cycle, one stratum past its activator (verified through 4-level chain tests).
- **Mixed inline + named entries.** A routing table may freely mix inline bodies and named scope references in the same dispatch.

**Implementation**

Inline bodies are desugared into synthesized root-level sibling scopes during declaration collection, then activated through the existing transition pathway used by named scope references. This replaces the previous approach, which carried inline bodies through every pipeline stage as a special case.

- `ArcParser.g4`: `routingEntry` now also accepts `stageDeclaration` and `sequenceDeclaration` alternatives.
- `analyzer/sequence`: `CollectDeclarations` runs a single AST walk that registers each inline body as a synth scope under its lexically enclosing scope, with an auto-generated `__inline_N` name. Named inlines emit a diagnostic. A second pass, `AnalyzeSynthInlines`, runs after the top-level analyze loop so bodies can resolve forward references to peer top-level scopes.
- `text/analyze`: a single branch in `analyzeOutputRoutingTable` dispatches inline entries by emitting a `transitionIntent{activateKey: synth.Name}` against the source output. Synth scopes emit as additional `Root.Strata[0]` members through the unchanged `Analyze` / `analyzeStage` paths, and the existing activation-binding loop handles the rest.
- `stratifier`: an activation-dep rule keeps inline synth scopes at stratum 0 when activated by a non-inline scope (so their bodies pre-walk for cross-scope marks in the same cycle), but enforces the dep when one inline activates another (so nested inlines walk after their outer activator). Supports arbitrary N-level nesting.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR replaces the previous per-pipeline-stage special-casing of inline `stage`/`sequence` bodies with a clean desugar-then-reuse approach: a single AST walk in `desugarInlineDecls` registers each anonymous body as a synthesized `__inline_N` root-level sibling scope, and the existing activation/stratifier machinery handles the rest.

- `ArcParser.g4` adds `stageDeclaration` and `sequenceDeclaration` as `flowNode` alternatives; `sequence.go` adds `desugarInlineDecls`, `registerInlineBody` (with named-inline rejection), and the post-pass `AnalyzeSynthInlines`.
- `text/analyze.go` adds `collectSynthByAST`, `processInlineBody`, `analyzeInlineBody`, and fixes the previously-flagged spurious auto-wire by setting `p.transitionEmitted = true` unconditionally in `processRoutingTable`; `bindActivations` is made recursive to reach inline scopes nested inside flow-step and stage scopes.
- All five previously-flagged review issues (auto-wire spurious transition, missing nil guard, name-rejection guard, top-level-stage registration, nested-sequence registration in `collectTopLevelStage`) are addressed in this iteration.

The inline-body desugar pass, IR-lowering path, and activation-binding loop all look correct for the cases exercised by the extensive new test suite; the five previously-flagged issues are addressed.

The implementation is large and touches the compiler's collect, semantic-analysis, IR-lowering, and stratifier layers simultaneously. The recursive `bindActivations` walk is new and critical — it must visit every nesting depth of inline scopes; the test suite exercises 4-level chains, which gives reasonable confidence. A small diagnostic style nit (plural kind name) is the only remaining finding.

arc/go/text/analyze.go (recursive `bindActivations` and `processInlineBody` scope wiring) and arc/go/analyzer/sequence/sequence.go (`desugarInlineDecls` walk correctness for all nesting combinations) warrant the most attention on final review.

| Filename | Overview |
|----------|----------|
| arc/go/text/analyze.go | Main IR-lowering changes: `collectSynthByAST`, `processInlineBody`, `analyzeInlineBody`, `addInlineMembers`, `inlineBoundaryBlocksNext`; fixes previously-flagged spurious auto-wire by setting `p.transitionEmitted = true` in `processRoutingTable`; adds recursive `bindActivations` to stamp activations onto nested inline scopes. |
| arc/go/analyzer/sequence/sequence.go | Adds `desugarInlineDecls` (AST walk registering all inline bodies as synth scopes), `registerInlineBody` (registers + rejects named inlines), `AnalyzeSynthInlines` (post-pass semantic analysis), and adds inline-routing support to `collectTopLevelStage`; fixes previously-flagged missing registrations. |
| arc/go/ir/ir.go | Adds the `InlinePrefix = "__inline_"` constant used to mark synthesized inline-body scopes throughout the compiler pipeline. |
| arc/go/parser/ArcParser.g4 | Adds `stageDeclaration` and `sequenceDeclaration` alternatives to the `flowNode` rule so inline anonymous bodies are valid routing-entry targets in the grammar. |
| arc/go/sequence_test.go | Extensive new runtime harness tests covering: anonymous stage/sequence inline dispatch, named-gated activation, no-auto-advance regression, `next`-rejection, nested inline routing tables (4-level chains), and cross-scope terminal transitions. |
| arc/go/text/text_test.go | Adds IR-lowering golden-path tests for inline stage/sequence bodies, including mixed inline+named entries and named-inline rejection diagnostics. |
| integration/tests/arc/inline_bodies.py | Consolidates the deleted `inline_sequence_in_stage.py` and `inline_stage_in_sequence.py` into a single integration test file covering both directions. |
| arc/go/analyzer/flow/flow.go | Minor: updates the comment in `analyzeNode` to acknowledge inline stage/sequence declarations as valid flow nodes deferred to sequence analysis. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["CollectDeclarations\n(sequence.go)"] --> B["collectTopLevelStage /\ncollectSequenceDecl\n— registers named scopes"]
    A --> C["desugarInlineDecls\n— AST walk, registers\n__inline_N synth scopes"]
    C --> D["registerInlineBody\n— rejects named inlines\n— adds synth to parent scope"]

    E["analyzeDeclarations\n(analyzer.go)"] --> F["AnalyzeTopLevelStage /\nAnalyze\n— semantic pass on named scopes"]
    E --> G["AnalyzeSynthInlines\n— semantic pass on synth scopes\n(runs after top-level loop)"]

    H["Analyze / text/analyze.go"] --> I["collectSynthByAST\n— builds AST→synth map"]
    H --> J["analyzeFlow → analyzeFlowNode\n— detects inline flowNode"]
    J --> K["analyzeInlineBody\n— looks up synth via synthByAST map\n— calls processInlineBody"]
    K --> L["processInlineBody\n— lowers inline body to IR\n— returns ir.Scope (Gated)\n— accumulates inlineNodes/Edges"]
    L --> M["transitionIntent\n{activateKey, suppressExit:true}"]
    M --> N["applyTransitionIntent\n— registerActivation(synth.Name, handle)\n— NO exit transition (suppressExit)"]
    J --> O["processRoutingTable\n— sets transitionEmitted=true\n— prevents spurious autoWire"]
    H --> P["bindActivations (recursive)\n— stamps Activation handle\nonto inline ir.Scope"]
```

1. `arc/go/analyzer/sequence/sequence.go`, line 184-205 ([link](https://github.com/synnaxlabs/synnax/blob/6d7baff856cec06a3d6c21cae4ac5988627f031e/arc/go/analyzer/sequence/sequence.go#L184-L205)) 

   <a href="#">P1</a> **Top-level stage inline routing bodies are never registered**

   `collectTopLevelStage` registers the stage itself but, unlike `collectStageDecl`, never calls `collectInlineRoutingDecls` for the flow statements in its body. If a user writes a top-level stage with an inline case body — valid syntax after the grammar change — the inline's scope is never added to the symbol table. In the semantic pass, `analyzeInlineRoutingBodies` runs but `GetChildByParserRule` silently falls back to the wrong scope. In the IR-lowering pass, `findInlineScopeByAST` returns `nil` and `emitInlineRoutingDispatch` emits `"internal: inline declaration scope not found"`, aborting compilation with a misleading internal error for otherwise valid code.

   The fix mirrors what `collectStageDecl` already does: iterate over the top-level stage's body items and call `collectInlineRoutingDecls` for each `FlowStatement`.

2. `arc/go/parser/ArcParser.g4`, line 497-501 ([link](https://github.com/synnaxlabs/synnax/blob/6d7baff856cec06a3d6c21cae4ac5988627f031e/arc/go/parser/ArcParser.g4#L497-L501)) 

   <a href="#">P2</a> **Promised name-rejection guard is not implemented**

   The grammar comment states "the analyzer rejects any provided name to keep them strictly anonymous and one-shot," but neither `collectInlineRoutingDecls` nor `analyzeInlineRoutingBodies` in `sequence.go` contain any such check. A user can write `yes: stage my_step { ... }` and `my_step` is silently registered in the enclosing stage's scope, making it referenceable (via `=> my_step`) from other flows within the same stage — contrary to the stated design intent of strict anonymity. A check should be added in `analyzeInlineRoutingBodies` to emit a diagnostic if the inline stage or sequence carries an `IDENTIFIER`.

3. `arc/go/text/analyze.go`, line 1323-1346 ([link](https://github.com/synnaxlabs/synnax/blob/94e6508d5cc1c32600eaa202e3839e3168cc0823/arc/go/text/analyze.go#L1323-L1346)) 

   <a href="#">P1</a> **Spurious auto-wire transition when inline-only routing table is in a sequence body**

   When a flow statement in a sequence body ends in a routing table whose entries are *all* inline (`stage`/`sequence`), `analyzeOutputRoutingTable` returns zero nodes (it calls `continue` for each inline entry). That means `nodes` only contains the pre-table nodes (e.g., `flag_read`, `gate`). Because `transitionEmitted` is never set by `processRoutingTable`, the guard `!transitionEmitted && len(nodes) > 0` is `true`, so `autoWireTransition` fires on `firstOutputParam(gate.Outputs)` — the same handle already claimed by `emitInlineRoutingDispatch`. The sequence frame ends up with two transitions on the same output handle: one dispatching to the inline peer step and one advancing (or exiting) the sequence, causing both to fire simultaneously at runtime.

   The fix is to mark `transitionEmitted = true` in `processRoutingTable` when at least one inline entry is present, so `analyzeSequence` skips the auto-wire. This path is only reachable when inline routing appears **directly** in a sequence body (not inside a stage body, where `analyzeStage` already discards `transitionEmitted`).
</details>

<sub>Reviews (12): Last reviewed commit: ["SY-4199: Consolidate and optimize tests"](https://github.com/synnaxlabs/synnax/commit/b6379f700361925a214bb8089e5d85f20e746ea6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32958973)</sub>


emilbon99 (changes requested):
In the example you have in your description where does `next` go to? Take this example:

```
 sequence main {
      stage hold {
          pressure -> threshold{} -> {
              critical: sequence { 
                   1 -> alarm; 
                   wait{500ms}
                   1 -> shutdown 
               }
          }
      }
  }
```

In this example if we enter `critical` the sequence will end and then park here, not go back to hold. If we add support for anonymous sequences/select stages in routing tables, then why isn't this allowed? Technically a direct flow node with no routing table is just a routing table with only one case (wired to the default output).

```
valve_cmd -> sequence {
      1 -> vlv_1_cmd
      wait{500ms}
      0 -> vlv_1-cmd
}
```

This is currently a syntax error right? 

I'm fine with what you're going after here, I just think you haven't thought about this from a consistent language design perspective.

sy-nico:
@emilbon99 , I've definitely thought about this. See [this slack thread](https://synnaxlabs.slack.com/archives/C09AT1QJEF3/p1779250343514429) that didn't get any responses.


The existing behavior also doesn't take into account for stage transitions. Given that the stages are inline, unless there is an explicit `next` or stage transition, this PR keeps the same behavior.

```
sequence main_1 {
    stage first_stage {
        "first_stage" -> log_ch
    }
    stage second_stage {
        "second_stage" -> log_ch
    }
}

sequence main_2 {
    stage first_stage {
        1 -> select{} -> {
            true: "critical" -> log_ch,
            false: "stable" -> log_ch,
        }   
    }
    stage second_stage {
        "second_stage" -> log_ch
    }
}
```

For that reason, I didn't want to add implicit stage navigation after a select statement where it didn't exist before. That being said, I have been thinking about resolving this by enabling the following behavior:

image

PR #2346, opened 2026-05-22
SY-4198: Arc LSP Hover, Rename, and Diagnostic UX Improvements

emilbon99:
## Description

A collection of Arc LSP polish:

**Hover**
- Move stdlib documentation to the symbol definition site (new `symbol.Symbol.Doc` field, `Document`/`WithDoc` helpers, `symbol.Arc("...")` code-block helper) and drop the parallel `keywordDocs` map for migrated entries.
- Cursor-aware hover via `GetQualifiedPrefixWordAtPosition`: hovering `time` in `time.now` resolves the module; hovering `now` resolves the canonical function.
- Dotted-name resolution walks head → tail through the symbol tree (with `IncludeInternal` so bare module names are reachable) and renders modules with a member list + module description.
- Deprecated aliases inline the canonical's Doc beneath the deprecation banner.

**Rename**
- Advertise `RenameOptions{PrepareProvider: true}` on the server so the LSP client gates rename on `prepareRename`.
- Console's custom context menu now hides (not just disables) the Rename item when the symbol can't be renamed, via an async `ILanguageFeaturesService.resolveRenameLocation` check tracked on cursor change (debounced 80ms, service resolved once per editor lifetime).
- A positional module alias (`import (math)`) is treated as non-renameable since its name is the canonical module identifier; `import (math as t)` remains renameable.

**STL surface**
- Mark `constant` and the `error` module + `panic` as Internal (lowering targets, not user-callable).
- Add Doc to `op.{ge,gt,le,lt,eq,ne,and,or,not}` to clarify they're the lowering targets of `>=`, `&&`, etc.

**Editor styling**
- Stacked diagnostic markers butt up cleanly (square top on a marker following a marker).
- Zero per-row `border-radius` so a diagnostic + hover-docs in the same popup read as one continuous box (Monaco's default rounds each row, creating notches).
- Hide the diagnostic source chip (e.g., `arc-analyzer`) — the source is already implied by the editor.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR is a substantial LSP polish pass for the Arc language server: cursor-aware hover with dotted-name resolution, `prepareRename` gating for the rename action (client- and server-side), diagnostic/hover UX tweaks in the Monaco editor, and a refactor of all STL symbol packages from package-level singletons to per-call `NewSymbols()` factories (fixing a latent concurrent-mutation race on shared parent pointers).

- **Hover**: `GetQualifiedPrefixWordAtPosition` replaces the old full-span extractor so hovering on `time` in `time.now` resolves the module while hovering on `now` resolves the canonical function. Documentation is moved to the definition site (`symbol.Doc`) and removed from the parallel `keywordDocs` map.
- **Rename**: The server advertises `RenameOptions{PrepareProvider: true}` and the React editor gates the context-menu rename item on an async `prepareRename` round-trip (debounced 80 ms), with the `debounce` utility gaining a new `.cancel()` / `.flush()` API.
- **STL refactor**: All packages switch from `var Symbols []*symbol.Symbol` to `func NewSymbols() []*symbol.Symbol`, and `symbol.Counter` / `symbol.GlobalResolver` are lowercased to unexported fields; `stateful` module is renamed from `"state"` to `"stateful"` with the C++ header updated in lockstep.

Safe to merge — the hover, rename, and STL refactor logic is correct and well-tested, with no data-loss or crash paths identified.

The STL singleton-to-factory migration, symbol field encapsulation, and cursor-aware hover resolution are all consistent throughout the codebase. The debounce.cancel() fix directly addresses the previous review finding. The one substantive logic gap — module completions appearing at the alias slot of import (M as alias) — is a minor UX issue with a straightforward one-line fix and no correctness or security impact.

arc/go/lsp/context.go — the isImportPathContext block-form path warrants a small guard for the AS token.

| Filename | Overview |
|----------|----------|
| arc/go/lsp/context.go | Adds ContextImportPath, isImportPathContext, isDeclarationNameContext, and isExpressionLevelComma. The block-form path in isImportPathContext incorrectly triggers for the alias position in import (M as alias), surfacing module completions where a user-chosen local name is expected. |
| arc/go/lsp/hover.go | Cursor-aware hover via GetQualifiedPrefixWordAtPosition; dotted-name resolution walks head to tail; module hover shows member list and module doc; deprecated aliases inline canonical doc. Logic is correct. |
| arc/go/lsp/completion.go | Large expansion: ContextImportPath completions, module member completions with auto-import edits, applyInvocationSuffix, isModuleImportedInSource token-level fallback, and buildAutoImportEditFromSnapshot using per-statement deletes. |
| arc/go/lsp/rename.go | PrepareRename handler already existed; isRenameable now correctly blocks positional module aliases. renameTextEdits collapses the Location to TextEdit hop into a single pass. |
| arc/go/symbol/symbol.go | Counter/GlobalResolver lowercased; NewRoot signature changed from variadic to slice; AddChild made variadic; NewModule/Deprecate helpers removed; Doc field added. All changes consistent. |
| console/src/code/Editor.tsx | Async prepareRename check (debounced 80ms) gates context-menu rename item visibility; debounce.cancel() now called on cleanup. Logic is correct. |
| x/ts/src/debounce/debounce.ts | DebouncedFn interface adds cancel/flush; zero-delay passthrough also gets no-op cancel/flush; NodeJS.Timeout replaced with ReturnType<typeof setTimeout> for portability. |
| x/go/lsp/document.go | Adds GetQualifiedPrefixWordAtPosition: left walk crosses dots, right walk stops at word boundary — correct cursor-word extraction for hover/goto-def. |
| console/src/code/Editor.css | Diagnostic source chip hidden; hover rows set to border-radius:0; stacked-marker and markdown padding rules added; h4+h5 baseline alignment. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Editor as Console Editor
    participant LSP as Arc LSP Server
    participant SymTree as Symbol Tree

    Note over Editor: Cursor moves
    Editor->>Editor: onDidChangeCursorPosition
    Editor->>Editor: debounce 80ms
    Editor->>LSP: prepareRename(position)
    LSP->>SymTree: resolveSymbolAtPosition
    SymTree-->>LSP: sym / err
    LSP->>LSP: isRenameable(sym)?
    alt renameable
        LSP-->>Editor: Range
        Editor->>Editor: setCursorRenameable(true)
    else not renameable
        LSP-->>Editor: null
        Editor->>Editor: setCursorRenameable(false)
    end

    Note over Editor: User triggers rename
    Editor->>LSP: rename(position, newName)
    LSP->>SymTree: renameTextEdits(doc, targetSym, newName)
    SymTree-->>LSP: []TextEdit
    LSP-->>Editor: WorkspaceEdit

    Note over Editor: Hover
    Editor->>LSP: hover(position)
    LSP->>LSP: GetQualifiedPrefixWordAtPosition
    LSP->>SymTree: resolveDotted(scope, name)
    SymTree-->>LSP: Symbol
    LSP->>LSP: getUserSymbolHover
    LSP-->>Editor: MarkupContent
```

1. `console/src/code/Editor.tsx`, line 272-283 ([link](https://github.com/synnaxlabs/synnax/blob/26e5dbd94bc75b48a67f5c3a8774b573e066de98/console/src/code/Editor.tsx#L272-L283)) 

   <a href="#">P2</a> **Debounce timer not cancelled on unmount**

   The `debouncedRenameCheck` is a closure with an internal `setTimeout`. When the editor's effect cleanup runs, `cursorDispose.dispose()` stops new cursor events and `renameCheckAbort?.abort()` cancels any in-flight async requests — but a pending debounce timer can still fire up to 80ms later. When it does, it calls `runRenameCheck`, which calls `setCursorRenameable` on an already-unmounted component. While React 18 silently drops such updates, it still wastes work and can produce stale results if the component re-mounts quickly in a different context.

   The `debounce` utility from `@synnaxlabs/x` returns a plain function with no `.cancel()` method, so the fix would require either a custom debounce variant that exposes a cancel handle or a mounted-flag guard inside `runRenameCheck`.
</details>

<sub>Reviews (9): Last reviewed commit: ["Merge branch &#39;rc&#39; of https://github.com/..."](https://github.com/synnaxlabs/synnax/commit/990ae80fe0121bbfd7e681492dab1834021e2523) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33511306)</sub>


arc/go/stl/channels/channels.go
```diff
@@ -23,55 +23,53 @@ import (
...
-)
-
-// Symbols are the symbols this package contributes to a program's ambient
-// prelude: the channels module plus `on` and `write` as bare globals so
+// NewSymbols returns a fresh slice of ambient prelude symbols this package
+// contributes: the channels module plus `on` and `write` as bare globals so
 // flow-mode programs can reference them without an import.
-var Symbols = []*symbol.Symbol{module, &userOnSymbol, &userWriteSymbol}
+func NewSymbols() []*symbol.Symbol {
+	numConstraint := new(types.NumericConstraint())
```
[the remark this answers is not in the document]
emilbon99: Not true.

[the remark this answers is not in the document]
emilbon99: I don't know what to tell you queen. This compiles totally fine.

arc/go/lsp/completion.go
```diff
@@ -583,12 +619,299 @@ func symbolCompletionItem(name string, t types.Type) protocol.CompletionItem {
...
+	// the user just inserted. A textual scan answers "is moduleName
+	// already imported?" without needing a valid AST.
+	if isModuleImportedInSource(doc.displayContent(), moduleName) {
+		return nil
+	}
+	var (
+		existingNames []string
+		stmts         []antlr.ParserRuleContext
+		seenStmts     = make(map[antlr.ParserRuleContext]bool)
+	)
```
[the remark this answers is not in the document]
sy-nico: `buildAutoImportEdit` still does a single replacement from `minStart` to `maxStop`, so anything sitting between two `import` statements (comments, in-progress edits) gets nuked when the consolidated block
is written.
Probably want one delete per statement plus one insert at `minStart` instead of one big replacement.

Before quick fix:
Image

After quick fix:

Image

console/src/code/Editor.css
sy-nico, line 304: Can you pin the "View Problem"/"Quick Fix" footer so that it's always visible, even when the hover content is too large and needs a scrollbar?

Image

Image

console/src/code/Editor.tsx
```diff
@@ -184,12 +225,52 @@ const use = ({
...
+    runRenameCheck();
+
     const extensionDisposables = extensions?.map((ext) => ext(editor)) ?? [];
 
     return () => {
       contentDispose.dispose();
       triggerDispose.dispose();
       contextMenuDispose.dispose();
+      cursorDispose.dispose();
+      renameCheckAbort?.abort();
```
sy-nico, line 274: P2 from Greptile still unresolved. Cleanup disposes the cursor listener and aborts the in-flight async, but the pending `setTimeout` inside `debounce` keeps running. `debounce` from `@synnaxlabs/x` returns a plain function with no `.cancel()` handle, so there's nothing to clear here.

Fix would be either a custom debounce variant that exposes a cancel, or a mounted-flag guard inside `runRenameCheck`.

### Repro
1. Bump `RENAME_CHECK_DEBOUNCE` to 5000ms.
2. Add `console.log("rename check fired", Date.now())` at the top of `runRenameCheck` and `console.log("editor unmounted", Date.now())` at  the top of this cleanup function.
3. Open an Arc editor, move the cursor, then close the editor within 5s.
4. Console shows `editor unmounted` followed by `rename check fired` several seconds later. Timer fired post-unmount.

arc/go/lsp/rename.go
sy-nico: `findAllReferences` re-tokenizes the entire document on every rename (tokenizeContent on line 118). For larger files this is O(N) per Rename call. Consider reusing the tokens already parsed for IR.

sy-nico: The locations list gets walked three times: once in `findAllReferences` to build, once in `toDocLocations` to convert, once here to map to `TextEdits`. Could be a single loop that produces TextEdits directly, or toDocLocations could yield `TextEdits`.

console/src/code/Editor.tsx
sy-nico, line 1: This file has no spec test file. The rename/debounce/cleanup logic added is non-trivial React-with-Monaco glue and has no regression/unit test coverage. At minimum add a test for the unmount-cancels-pending-rename-check path.

sy-nico (changes requested):
In `x/ts/src/debounce/debounce.ts`, the debounce helper has no `.cancel()` method, which is what surfaced the Editor.tsx leak. Worth adding a `.cancel()` (and maybe `.flush()`) to the returned function so callers can clean up pending timers on unmount. Same shape as `lodash.debounce`.

arc/go/lsp/code_action.go
sy-nico, line 1: `unusedImportQuickFix`, `deprecatedSymbolQuickFix`, and `missingImportQuickFix` (plus their helpers `findAliasAtDiagnostic`, `findDeprecatedBareRef`, `firstIdentifierInRange`) all read `doc.IR.Symbols` and `doc.Content` without holding `s.mu`, while `runAnalysis` writes both under `s.mu.Lock` (`server.go:352-357`). That's a data race and `go test -race` will flag it. 

Fix: at the top of `CodeAction`, take `s.mu.RLock`, copy `doc.IR.Symbols` and `doc.Content` into locals, release, then pass those locals to the handlers instead of letting each one read off the live `doc`.

arc/go/lsp/code_action.go
```diff
@@ -0,0 +1,493 @@
...
+
+// eatsTrailingNewline reports whether the character at (line, char) in
+// content is a newline. Both \n and \r\n collapse to a single LSP newline
+// because the End position sits before the \n in either case.
+func eatsTrailingNewline(content string, line, char int) bool {
+	offset := offsetAt(content, line, char)
+	return offset >= 0 && offset < len(content) && content[offset] == '\n'
+}
+
+func offsetAt(content string, line, char int) int {
```
sy-nico: Could you not use  the existing helper`lsp.PositionToOffset` from `x/go/lsp/document.go` instead?

arc/go/lsp/code_action.go
```diff
@@ -0,0 +1,493 @@
...
+			return t
+		}
+	}
+	return nil
+}
+
+// isAmbientModuleInScope reports whether name matches a non-Internal
+// KindModule child of an ancestor KindAmbient scope reachable from start.
+// This duplicates symbol.isAmbientModule, which is unexported.
+func isAmbientModuleInScope(start *symbol.Symbol, name string) bool {
```
sy-nico: Rather than duplicate, why not export as `Symbol.IsAmbientModule` instead?

arc/go/lsp/code_action.go
```diff
@@ -0,0 +1,493 @@
...
+func (s *Server) CodeAction(
+	ctx context.Context,
+	params *protocol.CodeActionParams,
+) ([]protocol.CodeAction, error) {
+	doc, ok := s.getDocument(params.TextDocument.URI)
+	if !ok {
+		return nil, nil
+	}
+	var actions []protocol.CodeAction
+	for _, diag := range params.Context.Diagnostics {
```
sy-nico, line 61: O(N*L) per CodeAction request. Every diagnostic causes `missingImportQuickFix` (and on a hit, `deprecatedSymbolQuickFix`) to re-tokenize the entire document via `firstIdentifierInRange` / `findDeprecatedBareRef`. 
Condiser tokenizing once at the top of `CodeAction` and pass the slice down.

arc/go/lsp/code_action.go
```diff
@@ -0,0 +1,493 @@
...
+			if action := s.deprecatedSymbolQuickFix(ctx, doc, diag); action != nil {
+				actions = append(actions, *action)
+			}
+		}
+		// Missing-import is offered for any diagnostic whose range contains
+		// a leading identifier that names an unimported ambient module.
+		// The handler's own filters (local resolution succeeds, no ambient
+		// match, already imported) reject non-candidates, so we don't need
+		// a separate code to gate it.
+		if action := s.missingImportQuickFix(ctx, doc, diag); action != nil {
```
sy-nico: `missingImportQuickFix` fires for every diagnostic regardless of code. Gate it on a specific code (e.g. `UndefinedSymbol`) the way `unusedImportQuickFix` and deprecated`SymbolQuickFix` are gated, otherwise every diagnostic eats a tokenize + scope lookup with no chance of producing a fix.

arc/go/lsp/code_action.go
```diff
@@ -0,0 +1,493 @@
...
+			},
+			End: protocol.Position{
+				Line:      uint32(endLine),
+				Character: uint32(endChar),
+			},
+		},
+	}
+}
+
+func indexOfItem(items []parser.IImportItemContext, target parser.IImportItemContext) int {
```
sy-nico: `indexOfItem` is a linear scan over the same slice `items` came from (`stmt.AllImportItem()` in the caller). You should track the index where `item` was found and pass it down to skip this.

arc/go/lsp/hover.go
```diff
@@ -507,55 +425,108 @@ func cleanDocComment(comments []string) string {
...
 	docComment := s.extractDocComment(content, sym)
 
+	displayName := name
+	if displayName == "" {
+		displayName = sym.Name
+	}
+
 	var d doc.Doc
 	switch sym.Kind {
 	case symbol.KindFunction:
```
sy-nico, line 469: Missing test coverage for `KindOutput`, `KindConfig`, `KindModule`, or the default branch of this switch. Add a `DescribeTable` over `(kind, expected title)` so a missing case doesn't silently regress to `default`

arc/go/lsp/code_action_test.go
```diff
@@ -0,0 +1,262 @@
...
+		return &protocol.CodeActionParams{
+			TextDocument: protocol.TextDocumentIdentifier{URI: uri},
+			Range:        protocol.Range{},
+			Context:      protocol.CodeActionContext{Diagnostics: diags},
+		}
+	}
+
+	Describe("Unused import quick fix", func() {
+		It("should delete the whole statement when it imports a single module", func(ctx SpecContext) {
+			content := "import time\n\nfunc test() i64 { return 0 }\n"
```
sy-nico, line 44: The `import (time control)` case only exercises 2-item blocks where every item is removed. Mid-item deletion in a 3+ item block (the `idx>0 && idx<len-1` branch of `deleteItemEdit`) isn't covered. Add an entry that removes the middle of `(a b c)` and asserts the surrounding whitespace.

sy-nico (changes requested):

arc/go/lsp/code_action.go
emilbon99: Fixed for `CodeAction` by snapshotting `doc.Content` + `doc.IR.Symbols` under `s.mu.RLock` at handler entry and threading a snapshot struct through the quick-fix helpers. The same race shape lives in Hover, Rename, and Completion. They all `getDocument` and then read live `doc` fields. Plan is to lift the snapshot pattern into a shared helper and migrate those handlers in a follow-up rather than expand this PR.

arc/go/lsp/code_action.go
emilbon99: This only happens when you open the quick fix menu. Doesn't seem super concerning to me to micro-optimize this until it actually causes performance problems.

arc/go/lsp/rename.go
emilbon99: I also don't think this is worth the optimization in this PR. The reference finding happens not very often so I'd rather leave as is for now?

arc/go/lsp/code_action.go
```diff
@@ -32,16 +45,27 @@ func (s *Server) CodeAction(
...
+	content := doc.Content
+	symbols := doc.IR.Symbols
+	s.mu.RUnlock()
+	snap := codeActionSnapshot{
+		URI:     doc.URI,
+		Content: content,
+		Symbols: symbols,
+		Tokens:  tokenizeContent(content),
+		IsBlock: doc.isBlock(),
+	}
```
sy-nico, line 59: Suggestion to move `doc.isBlock()` into the `s.mu.RLock` block above so the snapshot pattern stays consistent and doesn't become a real race the moment `doc.metadata` gains a write path.

arc/go/lsp/code_action.go
```diff
@@ -50,7 +74,7 @@ func (s *Server) CodeAction(
 		// The handler's own filters (local resolution succeeds, no ambient
 		// match, already imported) reject non-candidates, so we don't need
 		// a separate code to gate it.
-		if action := s.missingImportQuickFix(ctx, doc, diag); action != nil {
+		if action := missingImportQuickFix(ctx, snap, diag); action != nil {
```
sy-nico, line 78: Does it make sense to fire `missingImportQuickFix` on every diagnostic, including ones already handled by the `switch` above? An `UnusedImport` diagnostic whose range starts at an ambient module name would produce both "Remove unused import" and a contradictory "Add import 'X'" on the same range.

arc/go/lsp/code_action.go
```diff
@@ -77,30 +101,43 @@ func (s *Server) unusedImportQuickFix(
...
+	}
+	if len(items) > 1 && idx < 0 {
+		return nil
+	}
+	edit := buildUnusedImportEdit(snap.Content, stmt, items, idx)
 	return &protocol.CodeAction{
 		Title:       "Remove unused import",
 		Kind:        protocol.QuickFix,
 		Diagnostics: []protocol.Diagnostic{diag},
 		IsPreferred: true,
```
sy-nico, line 121: Why mark every quick-fix kind `IsPreferred: true` (here, line 300, line 473)? LSP says one preferred per diagnostic; when multiple actions mark themselves preferred the client picks one arbitrarily. If `missingImportQuickFix` and `unusedImportQuickFix` both fire on the same diag (see line 77). Which gets surfaced to the user?

arc/go/lsp/code_action.go
```diff
@@ -166,18 +208,18 @@ func deleteStatementEdit(
...
+// parenthesized import block. When idx has a following sibling the
+// deletion eats the whitespace between them; for the last item the
+// deletion eats the whitespace after the previous sibling instead.
 func deleteItemEdit(
 	items []parser.IImportItemContext,
-	item parser.IImportItemContext,
+	idx int,
 ) protocol.TextEdit {
-	idx := indexOfItem(items, item)
+	item := items[idx]
```
sy-nico, line 220: `items[idx]` will panic if `idx >= len(items)`. The caller's guard at line 112 only catches `idx < 0`. Will this break the first time a future caller passes a stale or wrongly-computed index? Worth either a defensive bound check or a precondition note in the godoc.

console/src/code/Editor.spec.tsx
```diff
@@ -0,0 +1,162 @@
...
+  ILanguageFeaturesService: Symbol("ILanguageFeaturesService"),
+}));
+
+import { Editor } from "@/code/Editor";
+
+const flushMicrotasks = async (): Promise<void> => {
+  for (let i = 0; i < 8; i++) await Promise.resolve();
+};
+
+const DEBOUNCE_PADDING = TimeSpan.milliseconds(500);
```
sy-nico, line 110: `DEBOUNCE_PADDING = TimeSpan.milliseconds(500)` doesn't reference `RENAME_CHECK_DEBOUNCE`. If the production debounce is ever raised past 500ms, this test will pass without actually advancing past the debounce, so the cancel/re-run paths stop being exercised. Should this derive from the actual constant?

sy-nico (changes requested):
Not sure how many of these are valid or just nitpicks.

arc/go/lsp/code_action.go
emilbon99: For the UnusedImport case specifically, this is already prevented by the resolve filter at lines 460-462 — the alias is in scope at its own declaration, so `scope.Resolve` succeeds and `missingImportQuickFix` returns nil before reaching the ambient-module check. The existing tests at `code_action_test.go:42-109` assert exactly one action per UnusedImport diagnostic, so a regression in that filter would surface as a test failure. The design (opportunistic fix gated by self-filters rather than by diagnostic code) is intentional and documented at lines 72-76. Happy to add an explicit regression test asserting no "Add import" action accompanies UnusedImport diagnostics if you'd like firmer proof.

arc/go/lsp/code_action.go
emilbon99: Tied to the same filter logic as the previous comment: in every code path today, each diagnostic gets exactly one quick-fix action (validated at `code_action_test.go:52, 79, 109, 180, 209, 224` — `HaveLen(1)` for single-diagnostic cases). The LSP/VSCode rule applies per-diagnostic ("multiple preferred for the same diagnostic"), not globally across the response, so three quick-fix functions each marking their action preferred is fine when only one ever applies to a given diagnostic. Each is also the canonical fix for its diagnostic (remove the unused import, replace the deprecated symbol, add the missing ambient module), which matches the LSP guidance for `isPreferred`.

arc/go/lsp/code_action.go
emilbon99: `deleteItemEdit` is only reachable from `buildUnusedImportEdit` when `len(items) > 1` (the `<=1` branch routes to `deleteStatementEdit` and ignores `idx`), and `idx` is set exclusively inside the loop at lines 107-111 that iterates `items` — so when control reaches the helper, `0 <= idx < len(items)` is guaranteed by construction. The guard at line 113 covers the only remaining failure mode (multi-item statement where the alias's AST node isn't in the parsed item list). The precondition is also already stated in the godoc at lines 160-166. A defensive bound check would be adding a runtime guard for a scenario that can't happen with the current single caller, which we generally avoid in this codebase.

PR #2351, opened 2026-05-24
SY-4210: Fix Flakey Aspen KV Observable Dedup Test

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4210](https://linear.app/synnax/issue/SY-4210)

## Description

The `IgnoreHostLeaseholder` observable test in `aspen/internal/kv/kv_test.go` was flaking in CI because aspen's gossip-ingress pipeline could persist (and observe) the same `(key, version)` tuple more than once. The `versionFilter` segment read a key's digest from the engine and decided accept/reject; the `persist` segment then wrote the op and the digest. The two stages ran on separate goroutines connected by a buffered channel, so two duplicate gossip deliveries could both pass the filter before either committed — both then persisted and the observer fired twice.

Rather than relax the test, this PR fixes the underlying race by fusing the two stages on the gossip-ingress path into a single `filterPersist` segment that opens one pebble transaction per inbound `TxRequest` and performs the digest check + the op/digest write through that transaction. Because reads inside the transaction see prior writes from the same transaction, and the segment processes one `TxRequest` at a time on a single goroutine, the dedup invariant is now structural rather than coordinated through timing.

The local write path is unchanged. `versionAssigner -> persist` still handles locally-originated ops without a digest check, because the assigner already stamps fresh monotonic versions by construction. Single-node clusters do not receive remote ops and therefore pay no overhead for the new logic.

`DB.NewObservable` is now documented as at-most-once per `(key, version)`. A new deterministic test drives duplicate gossip RPCs through the mock `OpNet` and asserts the observer fires exactly once — this catches future regressions of the dedup invariant without depending on CI load to surface the race.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes a race condition where two duplicate gossip deliveries could both pass the version filter before either committed a digest, causing the observable to fire more than once for the same `(key, version)` tuple. The fix fuses the version-filter and persist stages for the gossip-ingress path into a single `filterPersist` segment that opens one pebble transaction per inbound `TxRequest`, performing the digest-check and the op/digest write atomically under that transaction.

- **`filter_persist.go`**: New segment that replaces the old two-stage (`versionFilter` → `persist`) gossip-ingress path; the dedup invariant is now structural rather than timing-dependent.
- **`kv.go`**: Pipeline topology updated — `versionAssigner → persistAddr` now routes exclusively to `persistDeltaAddr`, and `filterPersistAddr` is wired directly to `persistDeltaAddr`, bypassing the old `persistAddr` step for gossip ops.
- **`kv_test.go`**: New deterministic dedup test drives 5 identical gossip RPCs through the mock `OpNet` and asserts the observer fires exactly once.

Safe to merge — the fix is structurally sound and the new deterministic test directly verifies the corrected invariant.

The race was caused by two pipeline stages running on separate goroutines sharing mutable engine state; fusing them into a single transactional segment eliminates the window entirely. The local write path is untouched. The routing topology changes are minimal and correct: gossip ops bypass persistAddr (they are already persisted by filterPersist), and local ops continue through versionAssigner → persist → persistDelta unchanged. xkv.WithTx correctly defers txn.Close(), so the concern from the previous review thread is already addressed by the helper itself.

| Filename | Overview |
|----------|----------|
| aspen/internal/kv/filter_persist.go | New segment fusing gossip-ingress version-filter and persistence into a single atomic transaction; well-documented and clearly structured with correct use of xkv.WithTx (which defers Close). |
| aspen/internal/kv/kv.go | Pipeline topology correctly updated: gossip-ingress now routes operationReceiver/operationSender → filterPersist → persistDelta, while the local path versionAssigner → persist → persistDelta is unchanged. |
| aspen/internal/kv/kv_test.go | New deterministic dedup test directly drives 5 identical gossip RPCs and asserts exactly-once observer behavior; avoids the timing sensitivity of the original flaky test. |
| aspen/internal/kv/version.go | versionFilter removed; getDigestFromKV signature generalized from xkv.DB to xkv.Reader so it can be called inside a transaction in filterPersist. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph Gossip Ingress ["Gossip-Ingress Path (NEW)"]
        OR[operationReceiverAddr] --> FP[filterPersist]
        OS[operationSenderAddr] --> FP
        FP -- "atomic txn: digest check + op write" --> FP_ACCEPTED{accepted?}
        FP_ACCEPTED -- "yes to persistDelta" --> PD[persistDeltaAddr]
        FP_ACCEPTED -- "no duplicate/stale" --> FS[feedbackSenderAddr]
    end
    subgraph Local Write Path ["Local Write Path (unchanged)"]
        EX[executorAddr] --> LP[leaseProxyAddr]
        LP --> VA[versionAssignerAddr]
        VA --> P[persistAddr]
        P --> PD
    end
    subgraph Fan-out
        PD --> OBS[observableAddr]
        PD --> SS[storeSinkAddr]
    end
```

<sub>Reviews (3): Last reviewed commit: ["Log non-NotFound digest read errors in f..."](https://github.com/synnaxlabs/synnax/commit/ee665ca8a2b8f0ff2ced3d910da6ad1df135a1fa) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33667096)</sub>

### DIFF
--- a/aspen/internal/kv/filter_persist.go
+++ b/aspen/internal/kv/filter_persist.go
@@ -1,0 +1,126 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package kv
+
+import (
+	"context"
+
+	"github.com/synnaxlabs/x/address"
+	"github.com/synnaxlabs/x/confluence"
+	"github.com/synnaxlabs/x/errors"
+	xkv "github.com/synnaxlabs/x/kv"
+	"github.com/synnaxlabs/x/query"
+)
+
+// filterPersist fuses version filtering and persistence into a single segment
+// for the gossip-ingress path. It opens one transaction per inbound TxRequest,
+// reads each digest inside that transaction, and writes accepted ops and their
+// digests through the same transaction before committing. Because reads inside
+// the transaction see prior writes from the same transaction, two duplicate
+// ops in one TxRequest cannot both pass. Because the segment processes one
+// TxRequest at a time on a single goroutine and the engine state cannot change
+// between the digest read and the digest write of any single op, duplicate
+// deliveries across TxRequests are also rejected. This is the structural
+// guarantee that backs the exactly-once observable contract documented on
+// DB.NewObservable.
+//
+// The local write path (versionAssigner -> persist) does not flow through
+// filterPersist: locally-assigned versions are monotonic by construction, so
+// the digest check would always pass and would burden single-node deployments
+// with a Get they cannot benefit from.
+type filterPersist struct {
+	confluence.BatchSwitch[TxRequest, TxRequest]
+	db         xkv.Atomic
+	acceptedTo address.Address
+	rejectedTo address.Address
+	Config
+}
+
+func newFilterPersist(
+	cfg Config,
+	acceptedTo address.Address,
+	rejectedTo address.Address,
+) segment {
+	s := &filterPersist{
+		Config:     cfg,
+		db:         cfg.Engine,
+		acceptedTo: acceptedTo,
+		rejectedTo: rejectedTo,
+	}
+	s.Switch = s._switch
+	return s
+}
+
+func (fp *filterPersist) _switch(
+	_ context.Context,
+	b TxRequest,
+	o map[address.Address]TxRequest,
+) (err error) {
+	ctx, span := fp.T.Debug(b.Context, "tx-filter-persist")
+	defer span.End()
+
+	var (
+		accepted = TxRequest{Sender: b.Sender, doneF: b.doneF, Context: b.Context, span: b.span}
+		rejected = TxRequest{Sender: b.Sender, Context: b.Context, span: b.span}
+	)
+
+	txn := fp.db.OpenTx()
+	defer func() {
+		if err != nil {
+			err = errors.Combine(err, txn.Close())
+		} else if commitErr := txn.Commit(ctx); commitErr != nil {
+			err = errors.Combine(commitErr, txn.Close())
+		}
+		accepted.done(err)
+	}()
+
+	for _, op := range b.Operations {
+		keep, filterErr := fp.filter(ctx, txn, op)
+		if filterErr != nil {
+			err = filterErr
+			return
+		}
+		if !keep {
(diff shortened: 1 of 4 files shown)


aspen/internal/kv/filter_persist.go
```diff
@@ -0,0 +1,113 @@
...
+
+// supersedes reports whether op should replace the digest stored for op.Key in
+// r. An op supersedes when its version is strictly newer, or when versions tie
+// and its leaseholder wins the higher-key tiebreak. A missing digest is treated
+// as superseded. Any other read error is treated as not-superseding so a
+// transient read failure cannot cause an op to overwrite stored state.
+func supersedes(ctx context.Context, r xkv.Reader, op Operation) bool {
+	dig, err := getDigestFromKV(ctx, r, op.Key)
+	if err != nil {
+		return errors.Is(err, query.ErrNotFound)
```
pjdotson: should be logging / doing something with non-`query.ErrNotFound` errors.

aspen/internal/kv/filter_persist.go
```diff
@@ -0,0 +1,113 @@
...
+) error {
+	ctx, span := fp.T.Debug(b.Context, "tx-filter-persist")
+	defer span.End()
+
+	accepted := TxRequest{Sender: b.Sender, doneF: b.doneF, Context: b.Context, span: b.span}
+	rejected := TxRequest{Sender: b.Sender, Context: b.Context, span: b.span}
+
+	err := xkv.WithTx(ctx, fp.db, func(txn xkv.Tx) error {
+		for _, op := range b.Operations {
+			if !supersedes(ctx, txn, op) {
```
pjdotson: in the case where the versions are equal but the leaseholders are higher, why are we
"applying" the operations? Isn't that just extra "stuff" that has to happen even though
the version is the same? am I misunderstanding this?

pjdotson (changes requested):

aspen/internal/kv/filter_persist.go
emilbon99: Just because the version are the same does not mean the value is the same. 

Let's say you have Nodes 1 and 2:

- Both write to the key "foo"
      - Node 1 has the value "a" with version 1 
      - Node 2 has the value "b" also with version 1
- There is a temporary partition (natural due to gossip) and they heal by gossiping ops to each other. The rule that the higher leaseholder wins means the cluster should settle on "b"
- If we didn't apply on tie then Node 1 would keep "a" while Node 2 would keep "b"
- The critical thing here is that the counter is not shared across nodes as that would require extra synchronization mechanisms, so you need to use the node's unique key guarantee to ensure convergence.

pjdotson (approved):

PR #2354, opened 2026-05-24
SY-4066 (1/3): Strongly Type Tables on the Core

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4066](https://linear.app/synnax/issue/SY-4066)

## Description

First PR in the SY-4066 table strong-typing series. Moves the table resource from an opaque `data: record` blob into a fully typed shape on the server (`Cell`, `Row`, `Column` with `rows`/`columns`/`cells` as direct typed fields on `Table`), and brings server-side migration support to parity with what the console already has.

Highlights:

- **`schemas/table.oracle`** rewritten as the strongly-typed shape. Client TS, Go, and protobuf types regenerate from it. `Cell.props` remains an opaque `record` carrying `@ts preserve_case` so per-variant props (text vs value) round-trip arbitrary key casing intact. The whole `cells` map is preserve-cased so the marker propagates through the nested map values.
- **`v55_lift_typed_table` migration** at `core/pkg/service/table/migrations/v55/` snapshots the pre-typed shape; `MigrateTable` decodes the console v0 blob, flattens row layouts from `CellLayout[]` to `[]string`, drops the per-cell `selected` UI flag, and projects into the typed Table. UI-only top-level fields (`lastSelected`, `editable`, `remoteCreated`) are dropped — they stay in the console slice. Wired into the gorp `Migrations` slice after `msgpack_to_orc`.
- **`SetData(ctx, key, Table)`** on Writer and API both take the typed body. Key and Name are preserved from the existing entry on update.
- **Client SDK** gains a typed `SetDataBody` and a `ZERO_NEW` constant for tests / defaults.
- **Console state shape stays at v0 internally.** `toWire(state, name)` / `fromWire(table)` helpers project Redux ↔ typed wire at the four boundary points (`Table.tsx` `useSyncComponent`, `export.ts`, `services/link.ts`, `services/ontology.tsx`, `workspace/services/ontology.tsx`). The slice gut and the `selectedCells` UI-only array move with the Pluto-owned state PR.

Test coverage:

- **14 server specs** (Writer Create/Rename/SetData + Retrieve) exercising the typed shape.
[...]

PR #2355, opened 2026-05-24
SY-4066 (2/3): Add Table Action Codec and Dispatch Endpoint

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4066](https://linear.app/synnax/issue/SY-4066)

## Description

Second PR in the SY-4066 table series. Lays the server-side foundation for granular table mutations: an oracle-generated discriminated-union action codec, a hand-written reducer, and a new \`Writer.Dispatch\` endpoint that applies an action sequence atomically. The existing \`SetData\` endpoint is **not** removed; both coexist so console and pluto continue to work unchanged through the next split.

**Stacked on [#2354](https://github.com/synnaxlabs/synnax/pull/2354).** Targets \`sy-4066-strongly-type-tables\`. Flatten in order.

### Action set

Eight actions declared in \`schemas/table.oracle\`: \`Rename\`, \`AddRow\`, \`RemoveRow\`, \`AddCol\`, \`RemoveCol\`, \`ResizeRow\`, \`ResizeCol\`, \`SetCell\`.

- \`AddRow\` / \`AddCol\` carry the full set of \`Cell\` values for the inserted row/column; the reducer derives \`Row.cells\` from the incoming Cell keys and registers each Cell in the table's cells map.
- \`RemoveRow\` / \`RemoveCol\` drop both the structural slot and every cell it referenced.
- \`SetCell\` is a no-op for an unknown key — \`Add*\` are the only path that introduces cells.
- Indices are \`uint32\` with end-clamping on out-of-range Adds and no-op on out-of-range Removes/Resizes.

### Server (Go)

- \`actions.go\`: \`ScopedAction{Key, DispatchKey, Seq, Actions}\` envelope and per-payload \`Handle\` methods.
- \`Writer.Dispatch\` applies the sequence inside a single gorp \`Update\` transaction and notifies \`actionObserver\` after \`Exec\` returns.
- \`service.go\`: gains an \`actionObserver\` + atomic \`seq\` counter; \`OpenService\` wires \`signals.PublishFromObservable\` for \`sy_table_set\` (JSON) and a separate gorp delete publisher for \`sy_table_delete\`.
- \`core/pkg/api/table\`: registers \`/api/v1/table/dispatch\` alongside the existing endpoints; grpc noop parity included.

### Client TS

- \`actions.ts\`: \`Handlers\` impl mirroring the Go reducer with inverse + targets per action — used by the Flux undoable store in PR 3. \`SetCell\`'s inverse restores the prior cell verbatim; \`AddRow\`/\`AddCol\` pair with \`RemoveRow\`/\`RemoveCol\`; resize stores the old size on the inverse. \`isUndoable\` returns true for every variant (no layout-derived equivalent of schematic's \`set_node_measured\`).
- \`client.tables.dispatch(key, dispatchKey, actions)\` sends the action sequence over \`/table/dispatch\` with the client-generated \`dispatch_key\` carried verbatim so the originator can recognize its own echo race-safely.
- \`scopedActionZ\` + \`SET_CHANNEL_NAME\` exported for the Pluto echo listener in PR 3.

### Test coverage

- **12 new Go dispatch specs** (one per action plus multi-action atomicity plus an observer-notification spec asserting monotonic \`Seq\`).
- **9 new TS dispatch integration specs** exercising every action through the real \`/table/dispatch\` endpoint.
- **5 reduceAll inverse specs** (rename inverse name, addRow inverse is removeRow, setCell inverse restores prior cell, setCell on unknown key returns no inverse, isUndoable true for every variant).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds a `Dispatch` endpoint to the table service that applies a sequence of granular mutations (Rename, AddRow/Col, RemoveRow/Col, ResizeRow/Col, SetCell) atomically inside a gorp transaction, then broadcasts the action envelope to subscribers via a new `actionObserver`. It also ships the matching TypeScript client method, TS action handlers with undo-stack inverses, and thorough integration + unit tests for both layers.

- **Go side**: Oracle-generated `actions.gen.go` defines the discriminated-union `Action` type and `Reduce` function; `actions.go` supplies the per-payload `Handle` methods; `service.go` wires `actions.State` + `signals.PublishSignals` for `sy_table_set` and a gorp delete publisher for `sy_table_delete`; `writer.go` adds `Writer.Dispatch` that calls `Reduce` inside a `ChangeErr` callback then notifies observers.
- **TypeScript side**: `actions.gen.ts` mirrors the wire schema via Zod discriminated unions; `actions.ts` implements `Handlers` with correct Immer-draft–safe inverses for every variant (using `snapshot` to freeze draft values before storing on the undo stack); `client.ts` adds `Client.dispatch` and exports `SET_CHANNEL_NAME`.

This PR is safe to merge; all changes are additive, the existing SetData endpoint is untouched, and the new Dispatch path is guarded by the same access enforcement.

The dispatch logic is correct end-to-end: Go insert-into-slice patterns avoid aliasing, Immer draft handling in TypeScript correctly freezes values before storing inverses, the NO_OP constant is never mutated by createReduceAll, and signals wiring degrades gracefully when Signals is nil. Every action variant is exercised by both Go (Ginkgo) and TypeScript (Vitest) tests, including out-of-range clamping, multi-action atomicity, and observer seq monotonicity.

| Filename | Overview |
|----------|----------|
| schemas/table.oracle | Adds 8 action definitions to the oracle schema; doc strings match the generated code; all actions have appropriate field types (uint32 indices, float64 sizes, Cell arrays). |
| core/pkg/service/table/actions.gen.go | Oracle-generated discriminated-union Action type and Reduce function; nil-payload guard returns union.MissingPayload for known types with missing pointers, unknown types are skipped safely. |
| core/pkg/service/table/actions.go | Hand-written Handle methods for all 8 payload types; insert-into-slice patterns are correct, nil-map deletes are safe, bounds clamping/no-op guards match the spec. |
| core/pkg/service/table/writer.go | Adds Writer.Dispatch; uses ChangeErr to apply Reduce inside the gorp transaction, then calls Dispatcher.Notify after Exec succeeds (pre-commit, mirrors the established schematic pattern). |
| core/pkg/service/table/service.go | Gains optional Signals wiring (action publisher + gorp delete publisher with DisableSet), actions.State, and io.MultiCloser; service.NewOpener pattern correctly defers cleanup on partial failure. |
| core/pkg/api/table/table.go | Dispatch endpoint added with access.ActionUpdate enforcement matching SetData; delegates to Writer.Dispatch inside db.WithTx. |
| client/ts/src/table/actions.gen.ts | Oracle-generated Zod schemas and factory functions; discriminated union uses snake_case type literals that the HTTP transport's camelToSnake codec maps correctly to Go JSON tags. |
| client/ts/src/table/actions.ts | Hand-written Handlers with correct Immer-draft-safe inverse computation; snapshot helper correctly uses isDraft check before current(); NO_OP constant is safe since createReduceAll only reads (never mutates) the returned HandlerResult arrays. |
| client/ts/src/table/client.ts | Adds Client.dispatch sending to /table/dispatch and exports SET_CHANNEL_NAME for the Pluto echo listener in PR 3. |
| core/pkg/service/table/writer_test.go | 12 new Dispatch specs covering every action, out-of-range clamping/no-op, multi-action atomicity, and monotonic seq observer notification. |
| client/ts/src/table/table.spec.ts | 9 integration dispatch tests plus 4 inverse unit tests and 8 round-trip tests; comprehensive coverage of every action variant through the real endpoint. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant C as TS Client
    participant API as API Layer (table.Service.Dispatch)
    participant W as table.Writer.Dispatch
    participant G as gorp (WithTx)
    participant R as Reduce()
    participant D as actions.Dispatcher.Notify
    participant O as actions.State observer
    participant S as signals.PublishSignals

    C->>API: "POST /api/v1/table/dispatch {key, dispatchKey, actions}"
    API->>API: access.Enforce (ActionUpdate)
    API->>G: db.WithTx(callback)
    G->>W: Writer.Dispatch(ctx, key, dispatchKey, actions)
    W->>G: tbl.NewUpdate().ChangeErr(Reduce).Exec(tx)
    G->>R: Reduce(currentTable, actions...)
    R-->>G: (updatedTable, nil)
    G-->>W: Exec returns nil
    W->>D: dispatcher.Notify(ctx, key, dispatchKey, actions)
    D->>O: observer.Notify(Scoped)
    O->>S: PublishSignals to sy_table_set channel
    G-->>API: WithTx commits
    API-->>C: 200 OK
```

<sub>Reviews (9): Last reviewed commit: ["SY-4066: Migrate table dispatch to the s..."](https://github.com/synnaxlabs/synnax/commit/92c1c45e448392c1e058651cbc01d5c7f1a11947) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33676984)</sub>


client/ts/src/table/actions.gen.ts
```diff
@@ -0,0 +1,210 @@
...
+  z.object({ type: z.literal("resize_row"), resizeRow: resizeRowPayloadZ }),
+  z.object({ type: z.literal("resize_col"), resizeCol: resizeColPayloadZ }),
+  z.object({ type: z.literal("set_cell"), setCell: setCellPayloadZ }),
+]);
+
+export type Action = z.infer<typeof actionZ>;
+
+export const rename = (payload: RenamePayload): Action => ({
+  type: "rename",
+  rename: payload,
```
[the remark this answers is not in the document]
emilbon99: False positive. The TS HTTP transport runs `caseconv.camelToSnake` on every outgoing payload (see `x/ts/src/binary/codec.ts:71`) and `caseconv.snakeToCamel` on every incoming one. The discriminated-union key `addRow` is serialized as `add_row` on the wire, matching Go's `json:"add_row,omitempty"` tag.

Proof: the 9 dispatch integration tests in `client/ts/src/table/table.spec.ts` exercise every multi-word action (`add_row`, `add_col`, `remove_row`, `remove_col`, `resize_row`, `resize_col`, `set_cell`) against a real server and assert the resulting state. They all pass — which couldn't happen if the payload arrived as `{ "add_row": null, "addRow": {...} }`.

core/pkg/service/table/writer.go
```diff
@@ -95,6 +106,37 @@ func (w Writer) SetData(
...
+	}
+	if w.actionObserver == nil {
+		return nil
+	}
+	w.actionObserver.Notify(ctx, ScopedAction{
+		Key:         key,
+		DispatchKey: dispatchKey,
+		Seq:         w.seq.Add(1),
+		Actions:     actions,
+	})
```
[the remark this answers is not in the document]
emilbon99: Intentional, mirrors the schematic pattern (`core/pkg/service/schematic/writer.go` does the same thing in `Dispatch`). Greptile flagged this on #2290 too and the team accepted it. The originator client deduplicates via `dispatch_key`, and any subscriber that processed a not-yet-committed action will see the next read (or the next dispatch echo) reconcile the state. Moving the notify outside `WithTx` is a cross-resource change that should land in a separate PR touching schematic and tables together; out of scope for this stack.

client/ts/src/table/table.spec.ts
```diff
@@ -125,4 +125,294 @@ describe("Table", () => {
...
+        [
+          table.addRow({
+            index: 0,
+            size: 30,
+            cells: [{ key: "x", variant: "text", props: {} }],
+          }),
+        ],
+      );
+      expect(inverse).toHaveLength(1);
+      expect(inverse[0].type).toEqual("remove_row");
```
pjdotson, line 289: might be worth it to see if we can add a fuzzing test to typescript and then testing
that state + action + inverse = state.

client/ts/src/table/actions.ts
```diff
@@ -0,0 +1,154 @@
...
+      targets: [payload.cell.key],
+    };
+  },
+};
+
+export const reduceAll = createReduceAll(handlers);
+
+// Every table action carries user intent; none are layout-derived. All are
+// undoable.
+export const isUndoable = (_action: Action): boolean => true;
```
pjdotson: ?? what's the point of this? why does it need to take in an `Action` if it isn't being used?

pjdotson (changes requested):

client/ts/src/table/table.spec.ts
emilbon99: Should this be addressed in this PR?

emilbon99: Seems like this is test infrastructure that should be done in a separate PR

pjdotson: separate PR

pjdotson (approved):

PR #2357, opened 2026-05-25
SY-4066 (3/3): Migrate Table to Pluto-Owned State

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4066](https://linear.app/synnax/issue/SY-4066)

## Description

Third PR in the SY-4066 table series. Moves the structural table state (rows, columns, cells) into the Pluto-owned flux store via the action codec landed in PR 2, and collapses the console-side iteration and mutation plumbing accordingly.

**Stacked on [#2355](https://github.com/synnaxlabs/synnax/pull/2355).** Targets \`sy-4066-1-table-actions\`. Flatten in order.

### Pluto: self-contained \`<Table resourceKey ...>\`

\`pluto/src/table/Table.tsx\` becomes the connected component:

\`\`\`tsx
<Base.Table
  resourceKey={key}
  selected={selectedCells}
  onSelectionChange={(next) => dispatch(setSelectedCells({key, cells: next}))}
  editable={canEdit}
  visible={visible}
  onContextMenu={(e, target) => ...}
/>
\`\`\`

Inside, it reads rows/columns/cells from flux, iterates the grid, renders each cell through the variant registry (\`@/table/cells\`), and dispatches \`ResizeRow\`/\`ResizeCol\` on resize gestures and \`SetCell\` on cell content edits. Selection state stays in the consumer (Console Redux); the component computes shift-click region geometry against an internal \`lastSelectedRef\` and emits \`onSelectionChange\`. The internal \`Row\` / \`ColumnIndicators\` / \`Indicator\` primitives drop their exports; the \`Cell\` primitive (a styled \`<td>\` wrapper) stays exported because the variant cell components in \`@/table/cells\` compose it.

### Pluto queries.ts: action-builder hooks + geometry

- \`useAddRow({key})\` / \`useAddCol({key})\` / \`useRemoveRow({key})\` / \`useRemoveCol({key})\` — handle default-cell-creation + the empty-grid bootstrap so consumers don't repeat the invariant logic (\"AddRow needs N default cells where N = column count\").
- \`findCellPosition(rows, cellKey)\` / \`cellsInRegion(rows, start, end)\` — pure geometry helpers.
- \`useCellPosition({key, cellKey})\` — selector hook for the (x, y) grid coordinates.

### Console: slice gut + collapsed render path

- **v0 → v1 state migration** drops \`layout\`, \`cells\`, and \`remoteCreated\`. Selection moves from a per-cell \`selected: boolean\` to a per-table \`selectedCells: string[]\` array. The slice keeps only UI reducers: \`create\`, \`remove\`, \`selectCells\`, \`setSelectedCells\`, \`setEditable\`.
- The legacy \`toWire\` / \`fromWire\` / \`copyBuffer\` / paste flow is removed, along with the inline \`addRow\`/\`addCol\`/\`deleteRow\`/\`deleteCol\`/\`setCellProps\`/\`setCellType\`/\`resizeRow\`/\`resizeCol\`/\`clearSelected\`/\`copySelected\`/\`pasteSelected\` reducers.
- \`console/src/table/Table.tsx\` drops from 633 lines to ~220: reads \`selectedCells\` and \`editable\` from Redux, renders \`<Base.Table>\`, wires the context menu through Pluto's action-builder hooks and \`useCellPosition\`, wraps with the existing toolbar and edit-toggle controls.
- \`Toolbar.tsx\` reads the selected cell via \`Base.useSelectCell\` and dispatches variant/props edits through \`Base.useDispatch\`.
- Layout.Creator moves into a dedicated \`layout.ts\`. New \`useCreate.ts\` wraps \`PTable.useCreate\` + \`placeLayout\` to mirror the schematic create flow; \`Selectable.tsx\` adopts the same pattern. Palette / ontology / link / import / export services switch to the typed wire format and no longer hydrate Redux from server-shaped data.

### Deferred to follow-up

- Copy / paste / delete keyboard handling. The v0 implementation lived in the console slice (\`copyBuffer\` + \`copySelected\`/\`pasteSelected\`/\`clearSelected\` reducers) and used Redux-owned cell positions for translation. The Pluto-owned equivalent is a \`useClipboard\` hook mirroring \`pluto/src/schematic/clipboard.ts\` — a self-contained PR that builds the \`SetCell\`/\`AddRow\`/\`AddCol\` sequence at paste time. Tracked separately.

### Test coverage

- 8 Pluto specs (existing table queries pass against the upgraded undoable store).
- 8 Console specs (slice + selector specs rewritten for v1 shape).
- 25 client TS integration specs (unchanged from PR 2; verifies no regression in the action codec).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR completes the SY-4066 table migration series by moving structural table state (rows, columns, cells) from the console's Redux slice into the Pluto-owned undoable flux store. The console slice is trimmed to UI-only state (`selectedCells`, `editable`, `hideIndicators`, `pendingUpload`), and a v0→v1 migration path handles existing tables with unpublished structural data.

- **Pluto `Table.tsx`** becomes the connected component: reads rows/columns/cells from the flux store, dispatches `ResizeRow`/`ResizeCol`/`SetCell` actions, and computes shift-click region geometry internally using a `lastSelectedRef` anchor.
- **Console slice** is reduced from ~633 to ~220 lines, dropping all structural reducers; selection changes to a `selectedCells: string[]` array; `useAutoUpload` handles v0 tables that were never synced to the server by re-creating them on first render.
- **Clipboard** (`clipboard.ts`) is implemented as a self-contained `useClipboard` hook that builds `AddRow`/`AddCol`/`SetCell` action sequences at paste time, mirroring the schematic clipboard pattern.

Safe to merge once the known flux-store-before-placeLayout gaps in link.ts and ontology.tsx (tracked in existing threads) are addressed; the core migration path and new structural logic are sound.

The migration, new Pluto-owned rendering loop, v0→v1 state migration, and clipboard implementation all look correct. The open link/ontology threads mean tables opened via link or drag-drop will crash on first render until those stores are populated — those are the only blocking concerns in the changeset.

console/src/table/services/link.ts and console/src/table/services/ontology.tsx still lack store.tables.set before placeLayout — tracked in existing review threads.

| Filename | Overview |
|----------|----------|
| pluto/src/table/Table.tsx | Connected component rewrite — reads rows/columns/cells from flux store, handles region selection via internal lastSelectedRef, gates all mutations on editable; paste handler correctly gated |
| console/src/table/Table.tsx | Reduced to ~220 lines; delegates structural rendering to Base.Table, handles editable/hideIndicators controls; TableControls contains an unreachable null guard |
| console/src/table/Toolbar.tsx | Significantly extended with MultiCellForm; readCellColor/cellColorPatch switch statements lack default branches; handleLevelChange applies level to all cell variants indiscriminately |
| console/src/table/slice.ts | Correctly stripped to UI-only reducers; create guard prevents overwriting existing entries; migrateSlice now wired into store.ts |
| pluto/src/table/queries.ts | New selectors and action-dispatch plumbing added; findCellPosition, cellsInRegion geometry helpers correct; useCreate wrapper correctly seeds 2x2 default layout |
| pluto/src/table/clipboard.ts | New self-contained clipboard hook using custom MIME type; correctly pre-builds keyAt map before dispatching; correctly syncs selectedRef via useSyncedRef |
| console/src/table/types/v1.ts | v1 state shape correct; migration preserves remoteCreated logic for pendingUpload; anyStateZ handles both v0 and v1 inputs |
| console/src/table/services/link.ts | Still does not populate store.tables before placeLayout (existing thread); handleLink retrieves table but never writes it to flux store |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant ConsoleTable as Console Table.tsx
    participant ReduxSlice as Console Redux Slice
    participant PlutoTable as Pluto Table.tsx
    participant FluxStore as Pluto Flux Store
    participant Server

    Note over User,Server: Cell edit path
    User->>PlutoTable: edit cell content
    PlutoTable->>FluxStore: dispatch actions SetCell
    FluxStore->>Server: tables.dispatch

    Note over User,Server: Selection change
    User->>PlutoTable: shift-click cell
    PlutoTable->>PlutoTable: cellsInRegion(lastSelectedRef, end)
    PlutoTable->>ConsoleTable: onSelectionChange(cellKeys)
    ConsoleTable->>ReduxSlice: setSelectedCells

    Note over User,Server: v0 pending upload path
    ConsoleTable->>ConsoleTable: useAutoUpload detects pendingUpload
    ConsoleTable->>FluxStore: PTable.useCreate with structural data
    FluxStore->>Server: tables.create
    Server-->>FluxStore: created table stored
    FluxStore->>ReduxSlice: clearPendingUpload
```

<sub>Reviews (33): Last reviewed commit: ["lots and lots of cleanup"](https://github.com/synnaxlabs/synnax/commit/aeab77273140d27c0675f102c2968815189472cd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33691989)</sub>


console/src/table/useUpload.ts
```diff
@@ -0,0 +1,37 @@
...
+        dispatch(clearPendingUpload({ key }));
+      },
+      [dispatch],
+    ),
+  });
+  useEffect(() => {
+    if (pendingUpload == null) return;
+    create({ ...pendingUpload, workspace: workspaceKey ?? undefined, name });
+  }, [pendingUpload, workspaceKey, key, create, name]);
+  return pendingUpload == null;
```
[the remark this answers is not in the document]
emilbon99: Acknowledged. After Comment 2 is fixed the only tables that hit `pendingUpload` are genuinely-never-synced v0 tables, which is a narrow population. `flux.update` already calls `addStatus` on failure, so the user gets a status toast with the error, and the slice persists `pendingUpload` so quitting and reopening the console rehydrates and retries.

You're right that an in-session retry is missing (a tab-reopen won't requeue because `create` is a stable `useCallback` and `pendingUpload` doesn't change). Tracking as a follow-up rather than fixing in this PR.

console/src/table/slice.ts
```diff
@@ -49,582 +42,114 @@ export interface RemovePayload {
...
   reducers: {
     create: (state, { payload }: PayloadAction<CreatePayload>) => {
-      const table: State = {
+      if (state.tables[payload.key] != null) return;
+      let migrated: State | undefined;
+      if (payload.data != null) {
+        const adjusted =
+          typeof payload.data === "object" && payload.data !== null
+            ? { ...(payload.data as Record<string, unknown>), remoteCreated: false }
+            : payload.data;
```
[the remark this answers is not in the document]
emilbon99: Good catch on the override. After tracing call sites, the entire `if (payload.data != null)` branch is unreachable: `Table.create` in `layout.ts` is the only `internalCreate` caller and never passes `data`. Purged the branch and the `data` field on `CreatePayload` in e0ef59f752.

Paired with that, wired `Table.migrateSlice` into `store.ts:migrateState` (it was the only slice missing). That is the correct home for v0 to v1 conversion at rehydration time, and it preserves `state.remoteCreated` from the disk shape instead of overriding it. Was a real latent bug for v0 not-remoteCreated tables, whose structural data would have been silently stranded.

Schematic carries the same dead branch from SY-3833-3. Cleaning that up in a separate PR off rc.

core/pkg/service/table/actions.go
```diff
@@ -112,25 +117,26 @@ func (p RemoveColPayload) Handle(state Table) (Table, error) {
...
-// range.
+// Handle resizes the row at the given index. Sizes below the minimum cell
+// dimension are clamped up to the floor. No-op if the index is out of range.
 func (p ResizeRowPayload) Handle(state Table) (Table, error) {
 	idx := int(p.Index)
 	if idx >= len(state.Rows) {
 		return state, nil
 	}
-	state.Rows[idx].Size = p.Size
+	state.Rows[idx].Size = max(p.Size, minCellDim)
```
pjdotson, line 164: why should we clamp this instead of rejecting with an error?

console/src/table/services/palette.tsx
```diff
@@ -12,10 +12,11 @@ import { Access, Table as PTable } from "@synnaxlabs/pluto";
...
 
 import { Palette } from "@/palette";
-import { Table } from "@/table";
+import { useCreate } from "@/table/useCreate";
 
-const CreateCommand: Palette.Command = ({ placeLayout, ...listProps }) => {
-  const handleSelect = useCallback(() => placeLayout(Table.create()), [placeLayout]);
+const CreateCommand: Palette.Command = (listProps) => {
+  const create = useCreate({});
+  const handleSelect = useCallback(() => create(), [create]);
```
pjdotson: pointless callback

pjdotson (changes requested):
Bugs. Not sure which is new and which is preexisting but most/all should probably be fixed:
1) i delete all cells, then hit the add column button and not all render:
2) when editable is off, I still get cursor changes when i hover over column borders that makes it look like i can drag them.
Pressing f highlights the table but doesn't even highlight properly, not sure what it should be doing then:

Image
4) If i select a cell and then hit tab, I just highlight random things:

Image

5) there is a way i can just move the bottom of the row border but not actually? see image:

Image

6) text rendering in the cell doesn't look good when you squash things:

Image

7) these context menu options:
- Add row below / Add row above
- Add column right / Add column left
appear in the opposite order you would expect (e.g. opposite of google sheets) and should be swapped

8) "Disable editing" context menu option should still appear even if you select "below" the table

Image

9) no way to scroll / pan if the table overflows the window.

10) search variants is not wide enough:

Image

11) when i have multiple cells selected, the toolbar at the bottom doesn't make sense, e.g. i am only editing one cell if there are multiple selected.

12) when i am not in edit mode, the table toolbar should say something about editing edit mode since i can't just select a cell

Image

13) for telem, sometimes they don't render in the right place. in image below the number is rendering above the cell instead of inside it:

Image

core/pkg/service/table/actions.go
emilbon99: Resize fires per-mouse-move during a drag (~60fps via useCursorDrag in Indicator.tsx). Rejecting would mean every below-min frame produces a rejected action, rolling back the optimistic update and flickering the UI mid-gesture. The min-dim is a soft UX floor, not a contract violation. AddRow/AddCol's `size` field clamps the same way, and the TS reducer clamps identically so optimistic and server states agree.

console/src/table/services/palette.tsx
emilbon99: Dropped in ede8d3d. `create` from useCreate is already useCallback-memoized internally, and TS variance handles `() => void` accepting `(init?: Partial<table.Table>) => void`.

PR #2358, opened 2026-05-25
SY-4065 (2/3): Add Line Plot Action Codec and Dispatch Endpoint

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4065](https://linear.app/synnax/issue/SY-4065/strongly-type-line-plots)

## Description

Adds the granular action codec and \`Dispatch\` endpoint for line plots, analog of schematic #2290. Twelve typed actions replace whole-state \`SetData\` writes for the cases where the console wants partial mutations: Rename, SetTitle, SetLegend, AddChannel, RemoveChannel, SetXChannel, AddRange, RemoveRange, SetAxis, SetLine, SetRule, RemoveRule. Each action is generated end-to-end (oracle schema -> Go envelope + Reduce dispatcher -> TS discriminated union + reduceAll factory) and the hand-written \`Handle\` methods enforce axis-key constraints with \`validate.ErrValidation\` at apply time.

This is the second PR in the SY-4065 series. PR 3 will move console state into Pluto's flux store and migrate the toolbar to dispatch.

**Stacked on [#2356](https://github.com/synnaxlabs/synnax/pull/2356)** (typed line plots). Targets \`sy-4065-strongly-type-line-plots\`; will retarget \`rc\` once that and its own base (PR #2353, oracle generator fixes) land.

### What changes

**Oracle schema + codegen**
- 12 \`action\` blocks declared inside \`LinePlot\` in \`schemas/lineplot.oracle\`. The plugin emits \`actions.gen.go\` with the Action envelope + Reduce dispatcher and \`actions.gen.ts\` with payload zod schemas + actionZ discriminated union + createReduceAll factory.
- The oracle generator extensions in PR #2353 (\`casing.TypePascal\`/\`TypeCamel\`) make \`SetXChannel\` render as \`SetXChannel\` / \`setXChannel\` / \`set_x_channel\` everywhere instead of collapsing into \`SetXchannel\` via lo's leading-acronym blind spot.

**Server**
- \`actions.go\` (hand-written): \`ScopedAction{Key, DispatchKey, Seq, Actions}\` envelope and per-payload \`Handle\` methods. Add/Remove channel and range handlers reject misrouted axis keys (y-only or x-only) with \`validate.ErrValidation\`. SetAxis, SetLine, and SetRule are full-replace / upsert.
[...]

PR #2356, opened 2026-05-24
SY-4065 (1/3): Strongly Type Line Plots on the Core

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4065](https://linear.app/synnax/issue/SY-4065/strongly-type-line-plots)

## Description

Replaces the opaque `data: record` blob on `lineplot.LinePlot` with typed `Title`, `Legend`, `Channels`, `Ranges`, `Axes`, `Line`, and `Rule` structs generated from `schemas/lineplot.oracle`. Existing clusters lift their stored blobs forward on first read after upgrade via a per-version v0..v4 legacy ladder mirroring the console TS migration chain, followed by a single `v55_lift_typed_lineplot` gorp migration.

This is the first PR in the SY-4065 series and the analog of #2283 for schematics. Action codec + Pluto-owned state will follow in separate PRs.

**Stacked on [#2353](https://github.com/synnaxlabs/synnax/pull/2353).** Targets `sy-4065-oracle-snake-preserve-digits`; will re-target `rc` once that lands.

### What changes

**Oracle schema and codegen**
- `schemas/lineplot.oracle` rewritten as the strongly-typed shape. Client TS (`client/ts/src/lineplot/types.gen.ts`), Go (`core/pkg/service/lineplot/types.gen.go`), and protobuf regenerate from it.
- Console `types/v5.ts` aliases the oracle-generated `titleZ`/`legendZ`/`channelsZ`/`rangesZ`/`axesZ`/`lineZ`/`ruleZ` so the wire format and console state can never drift. v4 to v5 state migration extracts `selectedRules: string[]` from the dropped per-rule `selected` field.

**Server-side migration ladder**
- `migrations/legacy/v0..v4` freeze each historical wire shape and own a `Migrate` step. The v0 to v1 step attaches the default legend; v1 to v2 sets x-axes to `time` tick type and flips y-axes to `labelDirection: "y"`; v3 and v4 are typed passthroughs that bump the version string (the fields they added are UI-only).
- `legacy.MigrateData` decodes the opaque blob, dispatches on the version string, and walks forward to `v4.Data`. Nil blobs and blobs without a version field fall through to v0 and walk the full chain.
- `MigrateLinePlot` projects `v4.Data` into the typed `LinePlot` body. UI-only fields (viewport, selection, mode, control, toolbar, measure, annotations, the wire-format key, remoteCreated) are dropped here.
[...]

PR #2360, opened 2026-05-25
SY-4065 (3/3): Migrate Line Plot Component to Pluto-Owned State

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4065](https://linear.app/synnax/issue/SY-4065/strongly-type-line-plots)

## Description

Final leg of SY-4065. Console line plots stop owning body state in Redux and instead read/write through Pluto's flux store, with dispatched actions roundtripping to the server. Multi-client echo, undo/redo, and per-field selectors now behave the same way they do for schematics.

**Stacked on [#2358](https://github.com/synnaxlabs/synnax/pull/2358)** (action codec), which stacks on [#2356](https://github.com/synnaxlabs/synnax/pull/2356) (typed line plots). Targets `sy-4065-line-plot-action-codec`; will retarget `rc` once PRs 1 and 2 land.

### What changes

**Pluto**
- `pluto/src/lineplot/queries.ts`: read-only `UnaryStore` upgraded to full `UndoableUnaryStore`. Per-field selectors for channels, ranges, axes, lines, rules, title, legend, and name. `useDispatch` / `useUndo` / `useRedo` wired to `client.lineplots.dispatch`. `useCreate` accepts an optional workspace and defaults to `uuid.ZERO`, matching schematic, so auto-upload works before the user has joined a workspace.

**Schema**
- `schemas/lineplot.oracle`: `Ranges.x1`/`x2` and `AddRange`/`RemoveRange` payload `range` fields flip from `uuid` to `string`. Console layers synthetic rolling-window keys (`recent`, `rolling1m`, etc.) on top of persisted ranges and they need to round-trip through the dispatch path without server-side UUID rejection.

**Console**
- v6 slice migration strips `title` / `legend` / `channels` / `ranges` / `axes` / `lines` / `rules` into a one-shot `pendingUpload` field. Pluto owns the body once upload completes.
- `useAutoUpload` sends `pendingUpload` via `PLinePlot.useCreate` on first mount and dispatches `clearPendingUpload` on success.
- Two-layer mount gate (applied to both `LinePlot.tsx` and `toolbar/Toolbar.tsx`): outer wrapper checks `pendingUpload` synchronously and returns `null` while it is set; inner component calls `useEnsureRetrieved` to suspend until the canonical record is in cache before reading via selectors.
- Toolbar tabs Data, Axes, Lines, Annotations, Properties dispatch through Pluto `useDispatch` and read state via Pluto selectors.
- `LinePlot.tsx` drops ~150 lines (`useSyncComponent` body sync, `useLoadRemote`, `buildLines` helper, `linePlotBody` projection). Body field flow runs through `PLinePlot` dispatch + render-time theme palette fallback for line colors.
- `useAssignedDispatch` wrapper prepends `AddRange(activeRange)` on the first channel/x-channel dispatch when `ranges.x1` is empty (the legacy `assignActiveRangeEffect` behavior). Console-only synthetic ranges ride through unchanged since the schema is now string-typed.
- `useDerivedLines` computes the cartesian product of channels x ranges merged with stored Line overrides, replacing the slice's auto-derive logic.
- Cmd+Z / Cmd+Shift+Z wired to `useUndo` / `useRedo` via `Triggers.use`.
- `addChannelsToActivePlot` helper for `channel/services/ontology.tsx` (which runs outside React) goes through `client.lineplots.dispatch` directly.
- `export.ts` emits a v5-shaped state for backwards compatibility with saved JSON files; the v5 -> v6 migration on import lifts the body into `pendingUpload`.
- `services/link.ts`, `services/ontology.tsx`, and `workspace/services/ontology.tsx` use the new `remote: true` option on `LinePlot.create` to skip pendingUpload when opening an existing plot.

**Server (Go)**
- `writer_test.go` and `actions.go` updated for string range type.
- `migrate.go` v0 -> v55 lift converts legacy `uuid.UUID` range slices to strings.

### Verification

- 22-case integration suite at `integration/tests/console/plot/line_plot.py` runs green end-to-end (create, add channels, set line thickness/label/title, drag-drop, range creation, csv export, json export/import, context-menu operations, multi-delete, open from resources/mosaic).
- 45 server-side specs in `core/pkg/service/lineplot` pass.
- Console + Pluto + Client type-check and lint clean.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR completes the SY-4065 migration by removing Redux ownership of the line-plot body: `title`, `legend`, `channels`, `ranges`, `axes`, `lines`, and `rules` are now owned exclusively by Pluto's undoable flux store and round-trip through the server. A v5 → v6 Redux slice migration lifts the body into a `pendingUpload` field, `useAutoUpload` uploads it on first mount, and a two-layer gate (`pendingUpload != null → null`, then `useEnsureRetrieved` suspends) ensures neither the canvas nor the toolbar renders body selectors before the canonical record is in cache.

- **State migration**: v6 slice strips all body fields from Redux; non-`remoteCreated` v5 plots are projected into `pendingUpload` which `useAutoUpload` sends via `PLinePlot.useCreate` on first mount.
- **Toolbar/canvas refactor**: all five toolbar tabs and the canvas dispatch through `LinePlot.useDispatch` and read through new Pluto per-field selectors; `assignColorsEffect`/`assignActiveRangeEffect` middleware and `useSyncComponent`/`useLoadRemote`/`buildLines` are deleted.
- **Go backend**: adds `SetChannels` and `SetRanges` action handlers and changes range fields from `uuid` to `string` to allow synthetic rolling-window keys to pass through.

Several flows introduced in this PR can silently fail or produce diverged state: a freshly placed plot is invisible to the server until useAutoUpload completes, yet both the range-add path and the resource-panel drag-drop path dispatch against the server key during that window.

The architectural migration is well-executed and extensively tested, but three interaction paths that were working before the PR now contain races against the pending-upload window: useAddToActivePlot dispatches to client.lineplots.dispatch without checking pendingUpload, addChannelsToActivePlot calls client.lineplots.retrieve which fails with not found, and useAutoUpload has no retry path on server failure, permanently blocking the plot for the session.

console/src/lineplot/useUpload.ts (no failure recovery path), console/src/range/useAddToActivePlot.ts and console/src/lineplot/addChannelsToActivePlot.ts (race against pending upload), console/src/channel/services/ontology.tsx (same race for resource-panel drag).

| Filename | Overview |
|----------|----------|
| console/src/lineplot/useUpload.ts | New hook that uploads pendingUpload on first mount; has no retry/failure path and can double-fire when workspaceKey or name changes while the first create is in-flight (previously flagged). |
| console/src/lineplot/LinePlot.tsx | Canvas renderer refactored to use PLinePlot selectors; drops ~150 lines of body-sync logic. The enableTriggers callback correctly gates Cmd+Z to the focused plot. |
| console/src/lineplot/addChannelsToActivePlot.ts | New out-of-React helper that dispatches AddChannel actions; races with pending upload (flagged in previous review) and lacks active-range injection (also previously flagged). |
| console/src/range/useAddToActivePlot.ts | Migrated to dispatch through client.lineplots.dispatch; guard still does not check pendingUpload so dispatches against non-existent plots (previously flagged). |
| pluto/src/lineplot/queries.ts | FluxStore upgraded to UndoableUnaryStore; comprehensive per-field selectors, useDispatch/useUndo/useRedo, kindOfTransaction coalescing, and delete made optimistic. Well-designed. |
| console/src/lineplot/slice.ts | Body actions removed; new clearPendingUpload and setLineVisible actions added; purgeSliceState exported for Redux Persist to clear hiddenLines on save. |
| console/src/lineplot/types/v5.ts | v5 to v6 migration strips all body fields; non-remoteCreated plots project their body into pendingUpload via buildPendingUpload; remoteCreated forced to false in parseImport prevents import failures on old exports. |
| console/src/lineplot/services/import.ts | Import now creates the plot on the server immediately rather than staging via pendingUpload; parseImport tries legacy schema first (with remoteCreated:false override) to avoid silently dropping body data. |
| console/src/lineplot/export.ts | Export now always fetches from the server and emits the typed format; requires a live connection where the old path could serve from Redux cache while disconnected. |
| core/pkg/service/lineplot/actions.go | Adds SetChannels and SetRanges handlers with axis validation and line reconciliation; mirrors the existing AddChannel/AddRange pattern. |
| core/pkg/service/lineplot/writer.go | Create now skips DefineRelationship when ws==uuid.Nil (zero-workspace plots), fixing a previously unreachable return err after a no-op. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant LinePlot as LinePlot React
    participant Redux as Redux Store
    participant PFlux as Pluto Flux Store
    participant Server as Synnax Server

    User->>LinePlot: Place new line plot
    LinePlot->>Redux: internalCreate ZERO_STATE key
    LinePlot->>LinePlot: useAutoUpload fires pendingUpload not null
    LinePlot->>Server: client.lineplots.create workspace pendingUpload
    Server-->>PFlux: store.lineplots.set key data
    Server-->>LinePlot: afterSuccess clearPendingUpload
    LinePlot->>Redux: clearPendingUpload key

    User->>LinePlot: Edit axis label toolbar
    LinePlot->>PFlux: LinePlot.useDispatch dispatch setAxisLabel
    PFlux->>Server: client.lineplots.dispatch key dispatchKey action
    Server-->>PFlux: broadcast store.lineplots.set updated
    PFlux-->>LinePlot: selector re-renders with new label

    User->>LinePlot: Cmd+Z undo
    LinePlot->>PFlux: useUndo key undo
    PFlux->>Server: client.lineplots.dispatch key dispatchKey inverseAction
    Server-->>PFlux: store.lineplots.set reverted
```

<sub>Reviews (17): Last reviewed commit: ["efficiency tuning"](https://github.com/synnaxlabs/synnax/commit/03dc6dbab53f011f0e9cfdd4665854433ba8c63d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33708783)</sub>


emilbon99:
@greptile re-run

pluto/src/channel/AliasInput.tsx
```diff
@@ -22,16 +22,23 @@ export interface AliasInputProps extends Input.TextProps {
...
   range?: string;
   shadow?: boolean;
+  // isDefault reports whether value is the derived default rather than a stored
+  // override. The reset button is shown only when an override is present.
+  isDefault?: boolean;
+  // onReset clears the override so the value reverts to its derived default.
+  onReset?: () => void;
 }
 
 export const AliasInput = ({
```
pjdotson, line 32: add tests

console/src/lineplot/services/palette.tsx
```diff
@@ -15,7 +15,8 @@ import { LinePlot } from "@/lineplot";
 import { Palette } from "@/palette";
 
 const CreateCommand: Palette.Command = ({ placeLayout, ...listProps }) => {
-  const handleSelect = useCallback(() => placeLayout(LinePlot.create()), [placeLayout]);
+  const create = LinePlot.useCreate({});
+  const handleSelect = useCallback(() => create(), [create]);
```
pjdotson: useless

pluto/src/flux/select.ts
```diff
@@ -50,10 +60,15 @@ export const createSelector =
     );
 
     const getSnapshot = useCallback(() => versionRef.current, []);
-    const selector = useCallback(
-      () => params.select(store, memoArgs),
-      [store, memoArgs],
-    );
+    const selector = useCallback((): Selected => {
```
pjdotson, line 63: this file needs tests

pluto/src/lineplot/Frame.tsx
```diff
@@ -0,0 +1,297 @@
...
+      >
+    >,
+    Omit<HTMLDivProps, "ref">,
+    Aether.ComponentProps {
+  resizeDebounce?: CrudeTimeSpan;
+  onHold?: (hold: boolean) => void;
+  ref?: Ref<FrameRef>;
+}
+
+export const Frame = ({
```
pjdotson, line 125: Why do LinePlots need a `Frame` when other things don't?

pluto/src/lineplot/Frame.tsx
```diff
@@ -0,0 +1,297 @@
...
+import { useEffectCompare } from "@/hooks";
+import { lineplot } from "@/lineplot/aether";
+import { useMemoDeepEqual } from "@/memo";
+import { type Viewport } from "@/viewport";
+import { Canvas } from "@/vis/canvas";
+import { grid } from "@/vis/grid";
+
+type HTMLDivProps = DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>;
+
+export interface ContextValue {
```
pjdotson, line 50: this file needs tests

pjdotson (changes requested):
Issues, some might be from this PR some might not:
1) It looks like I can drag the "Properties" toolbar tab name and it floats? very odd
2) when I look at the "Axes - X1" tab in a browser, the default "X1" for "Label" does not appear properly grayed out.
3) If "Title" is just the name of the line plot now, does it make sense to even allow to edit it from the bottom toolbar? we don't do that for other visualizations.
4) Some _very_ janky things happen when you create a new annotation, have multiple Consoles open, and try to drag it up and down.

pluto/src/lineplot/Frame.tsx
emilbon99: The key thing here is the component design. Line plot was designed as a composition based API (which I think is the better approach), to the old line plot did.

`
<LinePlot.LinePlot>
      <LinePlot.XAxis>
        <LinePlot.YAxis>
           <LinePlot.Line />    
      </LinePlot.YAxis>
     </LinePlot.XAxis>
</LinePlot.LinePlot>
`

This lets you compose whatever pieces of a plot you may need and leave out the ones that don’t, very `radix` or `shadcn` style. 

We previously had `Channel.LinePlot` which was a wrapper around this composition API to let it take a structured document i.e. the stored line plot. I refactored to get rid of `Channel.LinePlot` since that was just a bad design to put it in that package.

This means that we now have conflicting line plot definitions, one designed around a composition API (general purpose, no ties to flux queries or pluto) and a wrapper that expects a specific structured document from the core.

The wrapper became `LinePlot.LinePlot` an the generic, composition styled one became `LinePlot.Frame`. My take: *keep them separate*. In the future, even if we might plan on releasing pluto components as a standalone library keeping this separation of concerns is good.

Even if you mildly disagree, I think they should stay separate just because it keeps this PR focused on the objective.

pluto/src/lineplot/Frame.tsx
emilbon99: This is the original line plot implementation, which also didn't have unit tests. Do you want me to solve that problem in this file? 

Writing high quality tests that are not just low qual smoke tests is going to take a considerable amount of effort, especially considering that integration tests already hit all of this directly.

emilbon99:
> Issues, some might be from this PR some might not:
> 
> 1. It looks like I can drag the "Properties" toolbar tab name and it floats? very odd
> 2. when I look at the "Axes - X1" tab in a browser, the default "X1" for "Label" does not appear properly grayed out.
> 3. If "Title" is just the name of the line plot now, does it make sense to even allow to edit it from the bottom toolbar? we don't do that for other visualizations.
> 4. Some _very_ janky things happen when you create a new annotation, have multiple Consoles open, and try to drag it up and down.

1) solved
2) solved
3) left as is
4) working on replicating now. re-requesting review for other things.

pluto/src/lineplot/Frame.tsx
pjdotson: I get worried when doing refactors that don't have good test coverage. Fine to merge
this PR but we should make sure all of Pluto code for visualizations has good unit tests
before release.

pjdotson (approved):

PR #2364, opened 2026-05-26
SY-4122: Arc Statuses

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4122](https://linear.app/synnax/issue/SY-4122)

## Description

Implements RFC 0037: redefines `status.set` as a dual-mode (`ExecBoth`) function callable from both Arc func blocks and flow nodes, replacing the deprecated bare `set_status` symbol. Adds a matching `status.delete`. Brings the C++ Arc runtime to parity with the Go runtime so both forms work in the cluster and in the driver.

**Arc language:**

- `status.set` is callable as `status.set("key_or_name", "message", "variant")` inside func blocks and as `upstream -> status.set{key_or_name="...", message="...", variant="..."}` in flow chains.
- `status.delete` mirrors the same dual shape: `status.delete("key_or_name")` / `upstream -> status.delete{key_or_name="..."}`.
- Both forms accept a status by UUID key or by human name. On a by-name multi-match, `status.set` writes to the first match by key order and `status.delete` removes all matches; either way a warning surfaces through the task reporter so the operator is informed without failing the task.
- Variant string literals (`"success"`, `"info"`, `"warning"`, `"error"`, `"loading"`, `"disabled"`) are validated at compile time; non-literal variants are validated at runtime.
- The deprecated bare `set_status` symbol is removed.
- Empty-string WASM string handles (handle `0`) are now treated as a legitimate empty value rather than corruption, so Arc programs that pass `""` no longer trigger a false `"invalid string handle"` warning. The Go and C++ strings runtimes share this contract.

**Analyzer / runtime infrastructure:**

- STL modules can now own their own analyzer logic via `Symbol.AnalyzeCall` and `FunctionType.AnalyzeFlowConfig` hooks, so variant validation lives next to the symbol instead of in core analyzer packages.
- Adds a `taskreporter.Reporter` callback that lets Arc stdlib modules surface runtime failures (multi-match, invalid variant, DB error) as task statuses prefixed with the task name, without failing the task.

**Service / API:**

- Adds `Service.SetByKeyOrName` and `Service.DeleteByKeyOrName` in `core/pkg/service/status`, plus matching endpoints (`/api/v1/status/set-by-key-or-name`, `/api/v1/status/delete-by-key-or-name`). Upserts run inside a single `WithTx`.
- By-name requests enforce access on the `status` resource type; UUID requests stay row-level. A row-level grant on a specific UUID does not authorize a by-name request.
- Adds `IsVariant` in `x/go/status` as the single source of truth for variant membership, used by both the server and the Arc runtime.
- Fixes a latent task-retrieve failure: manually deleting a task's status row no longer hard-fails task retrieves; the bulk lookup skips `ErrNotFound` and auto-heals.

**C++ driver / client:**

- C++ Arc runtime gains `status::SetStatus` / `status::DeleteStatus` flow nodes and matching WASM host functions.
- C++ client adds `statuses.set_by_key_or_name` / `statuses.delete_by_key_or_name`.
- The status module's dependency on the strings runtime is made opt-in via a `StateConsumer` interface; output writing is standardized to `*output(0) = Series(v); *output_time(0) = Series(clock.now())` to match the Go side.

**Console / Pluto:**

- Pluto's "Change Status" block migrates from `set_status` to `status.set`, with its config aligned to the `key_or_name` / `message` / `variant` params.
- Adds a console v1 -> v2 state migration that rewrites saved `set_status` graph nodes into `status.set` so persisted graphs render and roundtrip.
- Routes the editor create-sites through `ZERO_STATE.version` instead of hardcoded version literals, so the schema version has a single definition and the next bump is a one-constant change.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR implements RFC 0037, redefining `status.set` as a dual-mode (`ExecBoth`) Arc function callable from both func blocks and flow chains, and introduces the `SetByKeyOrName` upsert operation throughout the Go service, gRPC/HTTP transport, and C++ client layers. The deprecated bare `set_status` symbol is removed and a v1→v2 console migration rewrites saved graph nodes.

- **Core upsert logic**: `Service.SetByKeyOrName` and `ResolveKeyOrName` are added to `core/pkg/service/status`; the API handler enforces access on the resolved key inside the transaction, so multi-name matches are authorized against the actual row being written rather than the raw input string.
- **Arc analyzer hooks**: `Symbol.AnalyzeCall` and `FunctionType.AnalyzeFlowConfig` callbacks let STL modules own compile-time validation; `status.set` uses these to validate variant string literals at compile time.
- **String-handle 0 contract**: Both Go and C++ string states now treat handle `0` as a canonical empty-string sentinel (counters start at 1), eliminating false "invalid string handle" warnings for Arc programs that pass `""`.

The change is broadly correct and well-tested; the two C++ driver findings are design-fragility concerns that do not cause current failures.

The Go service logic, API access-enforcement inside the transaction, string-handle-0 contract, and console migration are all correct. The C++ driver Reporter lambda silently drops the variant argument and bind_to uses str_state without a null guard — neither causes a current failure but both represent fragility worth fixing before the next extension.

driver/arc/task.h and driver/arc/status/status.h — the Reporter variant drop and the unguarded str_state dereference in bind_to.

| Filename | Overview |
|----------|----------|
| core/pkg/service/status/service.go | Adds ResolveKeyOrName, SetTarget, and SetByKeyOrName; variant validation before the TX is correct; multi-match semantics (first by key order) are well-defined and tested. |
| core/pkg/api/status/status.go | Adds SetByKeyOrName endpoint; access check correctly placed inside the transaction on the resolved status key. |
| core/pkg/service/arc/status/status.go | New file replacing set.go; ExecBoth symbol with compile-time variant validation hooks, WASM host registration, and flow-node implementation. |
| driver/arc/status/status.h | C++ SetStatus refactored; str_state injected via StateConsumer; dispatch_set correct. bind_to uses str_state without a null guard before first use. |
| driver/arc/task.h | Reporter lambda discards variant and always calls send_warning; Go-side routes by variant. Currently harmless but asymmetric with the Go contract. |
| arc/go/stl/strings/state.go | Handle-0-equals-empty-string contract correctly implemented; counter starts at 1. |
| arc/cpp/stl/strings/state.h | Mirrors Go contract: counter starts at 1, handle 0 short-circuits to empty string. StateConsumer cleanly decouples injection from construction. |
| console/src/arc/types/v2.ts | v1→v2 migration rewrites set_status nodes with safe defaults for missing fields. Non-set_status nodes pass through unchanged. |
| arc/go/analyzer/expression/expression.go | Calls AnalyzeCall hook after generic func-form validation. |
| arc/go/analyzer/flow/flow.go | Calls AnalyzeFlowConfig hook after generic flow-form config validation. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant ArcTask as Arc Task (Go/C++)
    participant StatusMod as Arc status Module
    participant Svc as status.Service
    participant DB as DB (gorp)
    participant API as API SetByKeyOrName

    Note over ArcTask,DB: Flow/Func form: status.set(key_or_name, msg, variant)

    ArcTask->>StatusMod: Next() / WASM host call
    StatusMod->>Svc: SetByKeyOrName(keyOrName, msg, variant)
    Svc->>Svc: Validate variant (IsValid)
    Svc->>DB: WithTx
    Svc->>DB: ResolveKeyOrName - key lookup first
    alt key found
        DB-->>Svc: "matches = [existing row]"
    else name scan
        DB-->>Svc: "matches = [0..N rows by name]"
    end
    Svc->>DB: NewWriter(tx).Set(SetTarget(matches,...))
    Svc-->>StatusMod: key, multipleMatches, err
    alt multipleMatches
        StatusMod->>ArcTask: Reporter(VariantWarning, multi-match msg)
    end
    StatusMod-->>ArcTask: resolved key on output(0)

    Note over API,DB: HTTP/gRPC API path

    API->>API: Validate variant
    API->>DB: WithTx
    API->>Svc: ResolveKeyOrName(tx, keyOrName)
    Svc-->>API: matches
    API->>API: "SetTarget -> st"
    API->>API: Enforce(ActionCreate, OntologyID(st.Key))
    API->>DB: NewWriter(tx).Set(st)
    DB-->>API: ok
    API-->>API: "SetByKeyOrNameResponse{key, multipleMatches}"
```

<sub>Reviews (7): Last reviewed commit: ["SY-4122: fix test (missing import in arc..."](https://github.com/synnaxlabs/synnax/commit/83ff7ce67e379865f3b65da2934aeab32be32057) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33851488)</sub>


arc/go/stl/strings/string.go
```diff
@@ -125,6 +125,9 @@ func registerFormat[T any](
 }
 
 func formatWithSpec(memory api.Memory, ptr, length uint32, value any) string {
+	if memory == nil {
```
pjdotson: why is this getting called when `memory` is `nil`?

pluto/src/arc/functions/status/external.ts
```diff
@@ -12,13 +12,13 @@ import { Form } from "@/arc/functions/status/Form";
 import { type Spec } from "@/arc/functions/types/spec";
 
 export const SPEC: Spec<Config> = {
-  key: "set_status",
+  key: "status.set",
   name: "Change Status ",
```
pjdotson: why does this end with a space?

pluto/src/arc/functions/status/external.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+  it("Should expose the canonical key", () => {
+    expect(SPEC.key).toBe("status.set");
+  });
+
+  it("Should sit at zIndex 100", () => {
+    expect(SPEC.zIndex).toBe(100);
+  });
+
+  it("Should default to key_or_name='', variant='success', message='Notification'", () => {
+    expect(SPEC.defaultProps({} as Theming.Theme)).toEqual({
```
pjdotson: remove type assertion

pluto/src/arc/functions/status/Change.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+import { describe, expect, it } from "vitest";
+
+import { config } from "@/arc/functions/status/Change";
+
+describe("Change config schema", () => {
+  it("Should accept a valid configuration", () => {
+    expect(() =>
+      config.parse({ key_or_name: "alarm", variant: "success", message: "ok" }),
+    ).not.toThrow();
+  });
```
pjdotson, line 20: for the success tests, you should make assert on the actual parsed object, not `not.toThrow()`

pluto/src/arc/functions/status/Change.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  it("Should accept each allowed variant", () => {
+    for (const variant of [
+      "success",
```
pjdotson: should use `VARIANTS` from x/status

pluto/src/arc/functions/status/Change.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    ])
+      expect(() =>
+        config.parse({ key_or_name: "k", variant, message: "m" }),
+      ).not.toThrow();
+  });
+
+  it("Should reject an unknown variant", () => {
+    expect(() =>
+      config.parse({ key_or_name: "k", variant: "bogus", message: "m" }),
+    ).toThrow();
```
pjdotson: For failure tests, you should assert what specific failure(s) it should throw

pluto/src/arc/functions/status/Form.tsx
```diff
@@ -16,7 +16,7 @@ import { Status } from "@/status";
 
 export const Form = (): ReactElement => (
   <Flex.Box y grow empty>
-    <Base.Field<string> path="statusKey" label="Status">
+    <Base.Field<string> path="key_or_name" label="Status">
       {(p) => <Status.Select {...p} />}
```
pjdotson: Should define these two function as constant render props. Also should make the below
style something in a CSS file or a constant

console/src/arc/types/v2.ts
```diff
@@ -0,0 +1,118 @@
...
+  ...v1.ZERO_SLICE_STATE,
+  version: VERSION,
+  arcs: {},
+};
+
+// Rewrites the deprecated set_status node to status.set. The bare set_status
+// STL symbol was removed in SY-4122; saved graphs must remap the props.
+const rewriteSetStatusProp = (p: NodeProps): NodeProps => {
+  if (p.key !== "set_status") return p;
+  const legacy = p as NodeProps & {
```
pjdotson: shouldn't use type assertions

console/src/arc/types/v2.ts
```diff
@@ -0,0 +1,118 @@
...
+    statusKey?: string;
+    variant?: string;
+    message?: string;
+  };
+  return {
+    key: "status.set",
+    key_or_name: legacy.statusKey ?? "",
+    variant: legacy.variant ?? "success",
+    message: legacy.message ?? "",
+  } as NodeProps;
```
pjdotson: shouldn't use type assertions

arc/go/symbol/symbol.go
```diff
@@ -34,13 +34,20 @@ import (
...
 	"github.com/synnaxlabs/x/errors"
 	"github.com/synnaxlabs/x/query"
 	"github.com/synnaxlabs/x/set"
 )
 
+// CallHook runs after the generic func-form validation passes.
+type CallHook func(ctx any, funcCall parser.IFunctionCallSuffixContext)
+
+// FlowConfigHook runs after the generic flow-form config validation passes.
+type FlowConfigHook func(ctx any, config parser.IConfigValuesContext)
```
pjdotson: why should these two types take in `ctx any` instead of a strongly-typed context?

arc/go/stl/strings/string_test.go
```diff
@@ -284,6 +280,24 @@ var _ = Describe("Strings", func() {
 			Entry("format_f32 with OOB spec", "format_f32", testutil.F32(3.14)),
 			Entry("format_f64 with OOB spec", "format_f64", testutil.F64(3.14)),
 		)
+
+		DescribeTable("Should return handle 0 when memory is nil",
+			func(ctx SpecContext, fn string, value uint64) {
+				rt2 := testutil.NewRuntime(ctx)
+				defer func() { Expect(rt2.Close(ctx)).To(Succeed()) }()
```
pjdotson: use `DeferClose`

arc/go/stl/strings/state_test.go
```diff
@@ -130,9 +128,10 @@ var _ = Describe("ProgramState", func() {
 			Expect(ok).To(BeFalse())
 		})
 
-		It("Should return false for handle zero", func() {
-			_, ok := s.Get(0)
-			Expect(ok).To(BeFalse())
+		It("Should return empty string with ok=true for handle zero", func() {
+			str, ok := s.Get(0)
+			Expect(ok).To(BeTrue())
```
pjdotson: use `MustBeOK`

arc/go/stl/strings/state.go
```diff
@@ -43,6 +43,9 @@ func NewProgramState() *ProgramState {
 
 // Create stores a string and returns a transient handle for later retrieval.
 func (s *ProgramState) Create(str string) uint32 {
+	if str == "" {
```
pjdotson, line 46: Could this change cause issue with existing customer Arcs?

arc/go/symbol/symbol.go
```diff
@@ -34,13 +34,20 @@ import (
...
+	"github.com/synnaxlabs/arc/parser"
 	"github.com/synnaxlabs/arc/types"
 	"github.com/synnaxlabs/x/compare"
 	"github.com/synnaxlabs/x/errors"
 	"github.com/synnaxlabs/x/query"
 	"github.com/synnaxlabs/x/set"
 )
 
+// CallHook runs after the generic func-form validation passes.
+type CallHook func(ctx any, funcCall parser.IFunctionCallSuffixContext)
```
pjdotson: are you sure the right way to handle this problem is to have symbols be allowed to
implement their own analysis code? It doesn't make sense to me that symbols should be
able to do generic analysis, they should be using built-in things from the language.

arc/cpp/stl/strings/state.h
```diff
@@ -27,41 +28,49 @@ class State {
     uint32_t config_counter = CONFIG_HANDLE_BASE;
 
 public:
-    /// Creates a transient string handle from a C++ string.
+    /// Creates a transient string handle. Empty input returns handle 0
+    /// (mirrors arc/go/stl/strings.Create).
```
pjdotson: silly comment, of course it mirrors the go function.

arc/cpp/stl/strings/state.h
```diff
@@ -27,41 +28,49 @@ class State {
...
     uint32_t create(const std::string &s) {
+        if (s.empty()) return 0;
```
pjdotson, line 33: do we have Python tests that assert for this behavior of empty strings returning 0? The
Python tests are the closest thing we have to ensuring that both language
implementations function the same.

arc/cpp/runtime/runtime.h
```diff
@@ -281,6 +281,16 @@ load(const Config &cfg, errors::Handler error_handler = errors::noop_handler) {
...
     };
 
+    // External factories that are also stl::Modules get bind_to here;
+    // StateConsumer opt-ins receive str_state first.
+    for (const auto &f: cfg.factories) {
+        if (auto m = std::dynamic_pointer_cast<stl::Module>(f)) {
+            if (auto sc = std::dynamic_pointer_cast<stl::strings::StateConsumer>(m))
+                sc->set_str_state(str_st);
+            stl_modules.push_back(m);
+        }
```
pjdotson, line 291: instead of adding a specific `strings::StateConsumer` interface, there might be a better
solution like turning the externally-provided modules into a builder function that takes
in the WASM runtime instead? happy to talk-through what might be a better pattern.

integration/tests/arc/stl_status.py
```diff
@@ -0,0 +1,226 @@
...
+        return [
+            s for s in self.client.statuses.retrieve(search_term=name) if s.name == name
+        ]
+
+    def _poll(
+        self,
+        check: Callable[[], T | None],
+        what: str,
+        timeout_s: float = 5.0,
+    ) -> T:
```
pjdotson: why isn't this a generic function?

integration/tests/arc/stl_status.py
```diff
@@ -0,0 +1,226 @@
...
+func del_via_func() {
+    status.delete("integ_status_func")
+}
+trigger_func_del -> del_via_func{}
+
+// Chain pattern: chain_key is both a sink for status.set's output and the
+// source that drives status.delete. Each trigger creates the row, writes the
+// key, then immediately deletes it, so the next trigger creates a fresh row
+// with a new UUID and the channel value changes across cycles.
+trigger_chain -> status.set{
```
pjdotson: we don't have tests for testing via key?

core/pkg/api/task/task.go
```diff
@@ -152,11 +154,35 @@ func (s *Service) Retrieve(
 		if err = status.NewRetrieve[task.StatusDetails](s.status).
 			Where(status.MatchKeys[task.StatusDetails](ontology.IDsToKeys(task.OntologyIDsFromTasks(res.Tasks))...)).
 			Entries(&statuses).
-			Exec(ctx, nil); err != nil {
+			Exec(ctx, nil); errors.Skip(err, query.ErrNotFound) != nil {
 			return res, err
 		}
-		for i, stat := range statuses {
-			res.Tasks[i].Status = &stat
+		byKey := make(map[string]*task.Status, len(statuses))
```
pjdotson: all of this logic should be added into the service layer.

core/pkg/service/task/writer.go
```diff
@@ -48,6 +48,12 @@ func resolveStatus(t *Task, provided *status.Status[StatusDetails]) *status.Stat
 	return provided
 }
 
+// SetDefaultStatus writes the default "status unknown" row for the task.
+func (w Writer) SetDefaultStatus(ctx context.Context, t *Task) (*status.Status[StatusDetails], error) {
```
pjdotson: should delete this once you refactor API layer task calls

core/pkg/service/task/writer.go
```diff
@@ -48,6 +48,12 @@ func resolveStatus(t *Task, provided *status.Status[StatusDetails]) *status.Stat
...
+	fresh := resolveStatus(t, nil)
+	return fresh, w.status.Set(ctx, fresh)
```
pjdotson: if the error from `w.status.Set` is non-`nil`, then you should return `nil` instead of
`fresh`. This is a common pattern and also what the Ginkgo/Gomega assertions check: when
the last value returned is an error, every other return should be a zero value.

core/pkg/service/status/service_test.go
```diff
@@ -0,0 +1,293 @@
...
+	"github.com/synnaxlabs/synnax/pkg/service/status"
+	"github.com/synnaxlabs/x/gorp"
+	"github.com/synnaxlabs/x/kv/memkv"
+	"github.com/synnaxlabs/x/query"
+	xstatus "github.com/synnaxlabs/x/status"
+	"github.com/synnaxlabs/x/telem"
+	. "github.com/synnaxlabs/x/testutil"
+)
+
+var _ = Describe("Dispatch", Ordered, func() {
```
pjdotson, line 30: why is this file called `dispatch_test`?

core/pkg/service/status/retrieve.go
```diff
@@ -153,6 +155,18 @@ func MatchLabels[D any](matchLabels ...xlabel.Key) Filter[D] {
...
+// retrieveByName returns all statuses sharing the given name.
+func retrieveByName[D any](ctx context.Context, tx gorp.Tx, name string) ([]Status[D], error) {
+	var matches []Status[D]
+	err := gorp.NewRetrieve[string, status.Status[D]]().
+		Where(gorp.Match(func(_ gorp.Context, s *status.Status[D]) (bool, error) {
+			return s.Name == name, nil
+		})).
+		Entries(&matches).
+		Exec(ctx, tx)
+	return matches, errors.Skip(err, query.ErrNotFound)
```
pjdotson: 1) if returning an `error`, first return should be nil
2) do we even return `query.ErrNotFound` if we are matching by name?

core/pkg/service/status/service.go
```diff
@@ -20,17 +20,27 @@ import (
...
+	"github.com/synnaxlabs/x/telem"
 	"github.com/synnaxlabs/x/validate"
 )
 
+var (
+	// ErrInvalidVariant signals a variant string rejected by xstatus.IsVariant.
+	ErrInvalidVariant = errors.New("invalid status variant")
+	// ErrEmptyKeyOrName signals a missing key_or_name parameter.
+	ErrEmptyKeyOrName = errors.New("key_or_name is required")
+)
```
pjdotson: We shouldn't be creating new error types here, instead you should inline wrap a
validation error for both of these

core/pkg/api/status/status.go
```diff
@@ -94,6 +94,68 @@ func (s *Service) Set(
...
+}
+
+// DeleteByKeyOrNameResponse is a response to a DeleteByKeyOrNameRequest.
+type DeleteByKeyOrNameResponse struct {
+	// Count is the number of statuses deleted.
+	Count int `json:"count" msgpack:"count"`
+}
+
+// DeleteByKeyOrName deletes statuses matched by key or by name.
+func (s *Service) DeleteByKeyOrName(
```
pjdotson: I don't think we should be adding the `DeleteByKeyOrName` endpoint. Setting by name is a
real current issues user have, so it's worth the non-ideal implementation to get it out.
We can still get access to keys in Arc, and we should have a better strategy before
introducing a different `Delete` pattern from the rest of our metadata APIs.

core/pkg/api/status/status.go
```diff
@@ -94,6 +94,68 @@ func (s *Service) Set(
...
+	req DeleteByKeyOrNameRequest,
+) (res DeleteByKeyOrNameResponse, err error) {
+	if err = s.access.Enforce(ctx, access.Request{
+		Subject: auth.GetSubject(ctx),
+		Action:  access.ActionDelete,
+		Objects: []ontology.ID{status.OntologyID(req.KeyOrName)},
+	}); err != nil {
+		return res, err
+	}
+	res.Count, err = s.internal.DeleteByKeyOrName(ctx, req.KeyOrName)
```
pjdotson: can you look all through your Go code for this pattern here - when you have an error and
that error is non nil, then you need to make sure the non-error return values are zero values?

core/pkg/service/arc/status/status.go
```diff
@@ -0,0 +1,349 @@
...
+)
+
+// Reporter message templates. Keep in sync with driver/arc/status/status.h.
+const (
+	setMultiMatchMsg    = "status.set: multiple statuses named %q; updated first match (%s)"
+	setFailureMsg       = "status.set: %v"
+	deleteFailureMsg    = "status.delete: %v"
+	deleteNotFoundMsg   = "status.delete: no status found %q"
+	deleteMultiMatchMsg = "status.delete: multiple statuses named %q; deleted all (%d)"
+)
```
pjdotson, line 38: what in the world is the point of these constants?

core/pkg/api/status/status.go
```diff
@@ -94,6 +94,68 @@ func (s *Service) Set(
...
 }
 
+// SetByKeyOrNameRequest is a request to upsert a status by key or by name.
+type SetByKeyOrNameRequest struct {
+	// KeyOrName is either an existing status key or a status name to match.
+	KeyOrName string `json:"key_or_name" msgpack:"key_or_name"`
+	// Message is the new status message.
+	Message string `json:"message" msgpack:"message"`
+	// Variant is the new status variant.
+	Variant string `json:"variant" msgpack:"variant"`
```
pjdotson: shouldn't this be a `status.Variant`?

core/pkg/api/status/status.go
```diff
@@ -94,6 +94,68 @@ func (s *Service) Set(
...
+
+// SetByKeyOrName upserts a status by key or by name.
+func (s *Service) SetByKeyOrName(
+	ctx context.Context,
+	req SetByKeyOrNameRequest,
+) (res SetByKeyOrNameResponse, err error) {
+	if err = s.access.Enforce(ctx, access.Request{
+		Subject: auth.GetSubject(ctx),
+		Action:  access.ActionCreate,
+		Objects: []ontology.ID{status.OntologyID(req.KeyOrName)},
```
pjdotson: this doesn't work as an access check - if they passed by name for example

client/cpp/status/status.h
```diff
@@ -174,10 +188,59 @@ class Client {
...
+    /// @param resolved_key Output: the resulting status key.
+    /// @param multiple_matches Output: true when the by-name path matched multiple
+    /// rows.
+    [[nodiscard]] x::errors::Error set_by_key_or_name(
+        const std::string &key_or_name,
+        const std::string &message,
+        const std::string &variant,
+        std::string &resolved_key,
+        bool &multiple_matches
+    ) const {
```
pjdotson, line 205: why does this take in `resolved_key` and `multiple_matches` as arguments instead of returns?

pjdotson (changes requested):

core/pkg/service/status/retrieve.go
```diff
@@ -153,6 +155,18 @@ func MatchLabels[D any](matchLabels ...xlabel.Key) Filter[D] {
...
 
+// retrieveByName returns all statuses sharing the given name.
+func retrieveByName[D any](ctx context.Context, tx gorp.Tx, name string) ([]Status[D], error) {
+	var matches []Status[D]
+	err := gorp.NewRetrieve[string, status.Status[D]]().
+		Where(gorp.Match(func(_ gorp.Context, s *status.Status[D]) (bool, error) {
+			return s.Name == name, nil
+		})).
+		Entries(&matches).
+		Exec(ctx, tx)
```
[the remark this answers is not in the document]
sy-nico: By-name lookups now resolve through a `status_by_name` `LookupIndex` (O(1)) instead of the full-table `gorp.Match` scan.

pluto/src/arc/functions/status/external.ts
sy-nico: Pre-existing, but fixed.

pluto/src/arc/functions/status/Change.spec.ts
sy-nico: updated

pluto/src/arc/functions/status/Change.spec.ts
sy-nico: Updated

arc/go/stl/strings/string.go
sy-nico: `memory` is nil only during the `wasmRT.Instantiate` -> `SetMemory` window (task.go:182-187): instantiation runs the guest's init section, so a guest that calls a `format_*` import during init hits the host before memory is wired. The guard degrades that to handle `0` instead of a nil-deref panic, same as the `from_literal` guard above.

pluto/src/arc/functions/status/Form.tsx
sy-nico: Implemented. Went with a constant rather than dedicating a whole CSS file to a single rule.

arc/go/stl/strings/string_test.go
sy-nico: fixed

arc/go/stl/strings/state_test.go
sy-nico: fixed

arc/go/symbol/symbol.go
sy-nico: The `ctx any` was there for two reasons. First, `symbol` can't reference the analyzer's `Context[T]` type directly: `analyzer/context` already imports `symbol`, so naming it here would create an import cycle. Second, the hook is called from two different analyzer phases that each pass a different `Context[T]` (one over `IPostfixExpressionContext`, one over `IFunctionContext`), so no single concrete `Context[T]` type would fit both call sites.

Since both hooks only read `ctx.Diagnostics`, I changed the parameter to `*diagnostics.Diagnostics` directly. That removes the `any`, removes the type assertions in the hook implementations, and sidesteps the import cycle because `x/diagnostics` is a leaf package.

core/pkg/api/task/task.go
sy-nico: Moved it into the service as `(*task.Service).AttachStatuses`. The API handler now just calls that when `IncludeStatus` is set.

core/pkg/service/task/writer.go
sy-nico: Deleted. Its body is now inlined in the auto-heal path of `AttachStatuses` in the service layer, and the API no longer calls it.

core/pkg/api/status/status.go
sy-nico: The reason delete stays key-only on most resources is that callers just retrieve by name to get the key, then delete by key. Arc can't do that: no globals to hold a key between nodes, and no Arc-side retrieve to look one up. So this is the one environment where the usual "resolve then delete by key" fallback isn't available, which is why delete-by-name earns its place here.

It's also better UX to keep a consistent interface across functions in the same STL module: `status.set` takes key-or-name, so `status.delete` should accept the same. Forcing one to take a name and the other a key would be a surprising asymmetry within `status`.

Name-based addressing isn't foreign to our clients either: retrieve-by-name is supported across channel, range, task, user, rack, and device, and channel's `delete` already accepts names.

core/pkg/api/status/status.go
sy-nico: The failure mode is the commit path: the closure populates the named returns and returns nil, but if the commit then fails, `WithTx` returns a non-nil error while those values are already set, so the handler returned populated values alongside the error.

Audited and fixed it everywhere it occurred. New to this PR: service `SetByKeyOrName`/`DeleteByKeyOrName`, api `SetByKeyOrName`/`DeleteByKeyOrName`, and the `IncludeStatus` path in `task.Retrieve`. Pre-existing but caught by the same audit: api `status.Set`. Each now returns explicit zero values on the error path.

core/pkg/api/status/status.go
sy-nico: Done, it's `xstatus.Variant` now. The internal `SetTarget`/`IsVariant` calls stay on `string` since the service `SetByKeyOrName` also takes raw strings from the Arc runtime, and the grpc translator converts at the boundary.

core/pkg/service/arc/status/status.go
sy-nico: removed

core/pkg/api/status/status.go
sy-nico: Fixed in `b9961f7e58`. The enforce now runs after `ResolveKeyOrName` against the resolved `st.Key` inside the tx, not against the raw key-or-name. `DeleteByKeyOrName` does the same on the resolved match keys.

core/pkg/service/status/retrieve.go
sy-nico: Both fixed. The `errors.Skip` was dead. A multi-entry `Entries` query never returns `query.ErrNotFound` (only single-entry bound queries do, per gorp's `Exec` contract), so by-name resolution returns an empty slice on no matches. 
Dropped it and now return `nil` on a real error.

core/pkg/service/status/service_test.go
sy-nico: Leftover name from when these methods lived in a `dispatch.go` source file.
They got consolidated into `service.go` during the `v1 -> v2` port but the test file kept the old name. Renamed to `service_test.go`.

core/pkg/service/task/writer.go
sy-nico: The `SetDefaultStatus` helper this was on is gone. The default-status write is now inline in `Service.maybeSetDefaultStatuses`, which returns the error and only assigns the status after `Set` succeeds. Also fixed a pre-existing instance of the same pattern in `Copy` (returned a populated `res` on the error path).

arc/go/symbol/symbol.go
sy-nico: These don't do generic analysis. After the `any` -> `*diagnostics.Diagnostics` change, a hook can only emit diagnostics. Generic validation (arity, types) still runs
first and owns all of it. The hook does one thing: check that `status.set`'s `variant`
arg is a known variant (`xstatus.IsVariant`). Arc has no enum or constrained-string
type to express that, and the valid set lives in `x/status`, not the language. Without
the hook the only check is a runtime warning. If there's a built-in mechanism for
"this string must be one of N values," happy to switch.

arc/cpp/runtime/runtime.h
sy-nico: Messed around with a builder version, but it still hardcodes `strings::State` as the one dependency it passes in, so it doesn't actually solve the coupling until there's a second external module. Decided not to go with it for this PR, but I can create a  new linear issue to revisit the deps-struct version when that comes up.

arc/go/stl/strings/state.go
sy-nico: No, string handles are ephemeral runtime values within a single execution cycle, not something baked into saved Arc source or graphs, so this can't invalidate an existing program. The round-trip is unchanged: `Create("")` -> `0`, `Get(0)` -> `("", true)`, so `""` in still gives `""` out, we just skip allocating a junk handle.

Separately, `status.set` is the only user-callable function that returns a string (the `strings` module fns are all `Internal`), and string formatting support isn't on `main` yet, so there's no released surface this could affect regardless.

core/pkg/service/status/retrieve.go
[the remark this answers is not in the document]
sy-nico: Tried the suggested secondary index (`status_by_name` `LookupIndex`) and reverted it because it broke correctness. The index's committed state is updated by an async DB-change observer, and the status `Writer` writes through the package-level `gorp.NewCreate` (no index registration), so writes never stage index deltas inside the write tx. Against the distribution DB that makes the index eventually consistent: a by-name lookup immediately after a `status.set` reads stale (empty) state. That breaks read-after-write, which both `SetByKeyOrName`'s own name resolution and `status.delete` depend on -> deterministic failure in the `status.delete` runtime test.

The name->key cache option has the same defect: anything not updated synchronously inside the write transaction returns stale reads here. A correct version would have to be maintained in the write tx itself, which is more than this path warrants.

The perf point is fair, but status is low-cardinality metadata and the scan only runs on a key-miss, so the linear cost is acceptable. Keeping the `gorp.Match` scan; a write-tx-maintained index is out of scope for this PR.

client/cpp/status/status.h
sy-nico: Switched both to return values. `set_by_key_or_name` now returns `pair<SetByKeyOrNameResult, Error>` (the struct holds `key` + `multiple_matches`), and `delete_by_key_or_name` returns `pair<int, Error>` for the count. The out-params were a holdover; pair-returns match the rest of the client (e.g. `retrieve`). Call sites and
tests updated to `ASSERT_NIL_P`.

sy-nico:
Integration test (stl_status.py) was just ported over from the v1 branch, but still needs fixing. I update at the end when everything else is in a better state. Same goes for documentation.

x/go/status/status.go
```diff
@@ -16,6 +16,15 @@ import (
 	"github.com/synnaxlabs/x/gorp"
 )
 
+// IsVariant reports whether s names a known status variant.
+func IsVariant(s string) bool {
```
pjdotson: if you backmerge with `rc` there is an `IsValid` method now on `Variant`s that is
generated via Oracle and you could use instead of this.

pluto/src/arc/functions/status/external.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+// included in the file licenses/APL.txt.
+
+import { describe, expect, it } from "vitest";
+
+import { SPEC, SYMBOLS } from "@/arc/functions/status/external";
+import { Theming } from "@/theming";
+
+const THEME = Theming.themeZ.parse(Theming.SYNNAX_LIGHT);
+
+describe("status.set SPEC", () => {
```
pjdotson: 1) we shouldn't have an `external.spec.ts` file, this is the only example of this file
   in the entire codebase. So you should maybe make a different file `spec.ts` and have
   this be `spec.test.ts`
2) You should really only think about testing "function" code, not "object" / "constant"
   code. e.g. you have a test that the key is `"status.set"`. That's nice - but say we
   ended up changing the key here? then we have to change the test? it is basically a
   pointless test. So I would remove all tests in this file except for the
   `defaultProps` test as that is actually testing a function.

integration/tests/arc/stl_status.py
pjdotson: as in this function is not specific to Arc or statuses, so it shouldn't in this file.

arc/cpp/runtime/runtime.h
pjdotson: sounds good

arc/go/symbol/symbol.go
pjdotson: My point is that I am not sure it "makes sense" to provide a hook / diagnostics
mechanism versus e.g. a string enum / union string literal type for something like this.
The reason is is that we are expressing things that CANNOT be done just through the Arc
language itself with this.

But the other question is, say in the future you would want to validate keys at
compile-time for ranges, statuses, etc. should that be something those nodes own? I am
not entirely sure. This is worth a little bit of thought though.

arc/go/stl/strings/string.go
pjdotson: let's talk through this one in person, am still a little confused.

driver/arc/task.h
```diff
@@ -166,7 +166,18 @@ class Task final : public task::Task {
...
+                    std::make_shared<::driver::arc::status::Module>(
+                        ctx->client,
+                        [task_ptr = task.get()](
+                            const std::string &variant,
+                            const std::string &message
+                        ) {
+                            if (variant == x::status::VARIANT_ERROR)
+                                task_ptr->state.send_error({"arc.status", message});
+                            else
+                                task_ptr->state.send_warning(message);
```
pjdotson: what if the variant is not a warning or an error?

driver/arc/status/status.h
```diff
@@ -17,46 +17,225 @@
...
+}
+inline std::string delete_not_found_msg(const std::string &key_or_name) {
+    return "status.delete: no status found \"" + key_or_name + "\"";
+}
+inline std::string delete_multi_match_msg(const std::string &key_or_name, int count) {
+    return "status.delete: multiple statuses named \"" + key_or_name +
+           "\"; deleted all (" + std::to_string(count) + ")";
+}
+
+/// @brief Upserts a status via the cluster API and surfaces failures via report.
```
pjdotson: Core not cluster

driver/arc/status/status.h
```diff
@@ -17,46 +17,225 @@
...
+                   << " error=" << err.data;
+        report(x::status::VARIANT_WARNING, set_failure_msg(err.data));
+        return "";
+    }
+    if (res.multiple_matches)
+        report(x::status::VARIANT_WARNING, set_multi_match_msg(key_or_name, res.key));
+    return res.key;
+}
+
+/// @brief Deletes a status via the cluster API and surfaces failures via report.
```
pjdotson: Core not cluster

driver/arc/status/status.h
```diff
@@ -17,46 +17,225 @@
...
+/// Returns true if at least one row was deleted.
```
pjdotson: why should we have a boolean return instead of a void return?
every other delete call in out codebase (looking at the API layer) will return
`(types.Nil{}, error)`. you are introducing a new contract / form of deleting. Is there
a real use case for this added complexity? Does this added complexity makes things
easier to reason about, or more confusing? Are we getting real "wins" by adding this?

Happy to talk through this point a little more sync too.

core/pkg/api/status/status.go
```diff
@@ -91,7 +93,109 @@ func (s *Service) Set(
...
+// DeleteByKeyOrNameRequest is a request to delete statuses by key or by name.
+type DeleteByKeyOrNameRequest struct {
+	// KeyOrName is either an existing status key or a status name to match.
+	KeyOrName string `json:"key_or_name" msgpack:"key_or_name"`
+}
+
+// DeleteByKeyOrNameResponse is a response to a DeleteByKeyOrNameRequest.
+type DeleteByKeyOrNameResponse struct {
+	// Count is the number of statuses deleted.
+	Count int `json:"count" msgpack:"count"`
```
pjdotson: Adding a delete response is also new added complexity that is not used by any other type
of deletion. I think the delete response should be `types.Nil{}`.

core/pkg/api/status/status.go
pjdotson: >>> Arc can't do this

Agreed, but we don't urgently need Arc to do this right now. We could add an Arc-side
retrieve or global variables in the next release, which would keep our metadata APIs
more similar.

>> It's also better UX

Partially? But it also good UX for every deletion across Synnax
(Arc,C++,Python,TypeScript) to "look" the same way. You are introducing a new way for
deletes to "look". This is added complexity + confusion for the user. 

>>> Name-based addressing isn't foreign

Yes, if we add a status.retrieve API we should absolutely allow retrieval by name. But
the right comparison is just to the other delete APIs. Right now, every delete API but
channel's just accepts keys, and channels has the guarantee of unique names. 

My bigger point is that the code here for set-by-name / delete-by-name is pretty
inconsistent with the rest of our APIs. I am fine with that for set-by-name because it
is a high priority customer issue to solve right now. I don't want to add extra
inconsistency for a `status.delete` that our customers won't be using right now when we
could, say, add global variables in the next release and obviate the hard need for deleting
by name.

core/pkg/service/task/service.go
```diff
@@ -216,6 +218,44 @@ func (s *Service) NewRetrieve() Retrieve {
 	}
 }
 
+// AttachStatuses populates the Status field of each task, writing and attaching
+// a default "status unknown" row for any task that has no persisted status.
+func (s *Service) AttachStatuses(ctx context.Context, tasks []Task) error {
```
pjdotson: can you explain to me why we even need this? doesn't every task have a status at
creation time?

Now, the task.retrieve API is NOT read-only as a user would assume, but actually
potentially creates statuses as well.

core/pkg/api/task/task_test.go
```diff
@@ -0,0 +1,103 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package task
```
pjdotson: should be in package task_test

core/pkg/api/task/task_suite_test.go
```diff
@@ -0,0 +1,123 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package task
```
pjdotson: should be in package task_test

core/pkg/service/status/retrieve.go
```diff
@@ -153,6 +153,22 @@ func MatchLabels[D any](matchLabels ...xlabel.Key) Filter[D] {
 	})
 }
 
+// retrieveByName returns all statuses sharing the given name.
+func (s *Service) retrieveByName(ctx context.Context, tx gorp.Tx, name string) ([]Status[any], error) {
```
pjdotson: isn't the right way to do this to have a `MatchNames[D any](names ...string) Filter[D]`
matcher function?

pjdotson (changes requested):
See comments

integration/tests/arc/stl_status.py
sy-nico: This was removed

integration/tests/arc/stl_status.py
sy-nico: added

arc/cpp/runtime/runtime.h
sy-nico: [SY-4244](https://linear.app/synnax/issue/SY-4244/refactor-stringsstateconsumer-in-runtimeh)

x/go/status/status.go
sy-nico: Back-merged and applied

pluto/src/arc/functions/status/external.spec.ts
sy-nico: Removed `external.spec.ts`.

I tried to keep the `defaultProps` test but every placement was worse than dropping it:

1. Move `defaultProps` into `Change.tsx` and test it in `Change.spec.ts`. Rejected because all other arc functions define `defaultProps` inline in their `external.ts` SPEC, so this would make `status` the only exception.
2. Test it from `Change.spec.ts` by importing `SPEC`. Rejected because `Change.spec.ts` is scoped to `Change.tsx`, so reaching into the barrel is a scope smell.

`defaultProps` lives in the barrel by convention, so its only co-located home is `external.spec.ts` (the file being removed). It also just returns a constant, and no sibling function tests its equivalent, so I dropped it.

driver/arc/task.h
sy-nico: This is a bug. I originally added the error branch here for parity with the Go reporter, which ports the set/delete result variant straight into the task's setStatus (and that can represent errors, warnings, etc.). But that parity was misguided for the C++ side: this reporter only surfaces non-fatal set/delete failures, and those should always be a task warning so the task keeps running, never an error. So the VARIANT_ERROR arm was dead. 

Removed the dead branch. The reporter now just calls send_warning.

This also sets the right pattern going forward: if a future stdlib function(say ranges.start{...}) fails, it should surface the same way, as a task warning via this reporter, not a task error.

Image

sy-nico: Also added coverage for these warnings in the integration test case

driver/arc/status/status.h
sy-nico: Agreed and updated. `status.delete` is now output-less (void), matching every other delete. I kept `Count` in the response only so the driver runtime can still surface the multi-match and not-found warnings; it's no longer exposed as a function output.

core/pkg/api/status/status.go
sy-nico: Kept the response instead of `types.Nil{}` because the driver runtime needs `Count` to surface the multi-match and not-found warnings on delete. I documented that as the sole reason on the field so it's clear it isn't a general delete-result contract.

core/pkg/api/task/task_test.go
sy-nico: Moved to package task_test.

core/pkg/api/task/task_suite_test.go
sy-nico: Moved to package task_test; construct the Service via apitask.NewService since the fields are unexported.

core/pkg/service/status/retrieve.go
sy-nico: Added MatchNames and dropped retrieveByName for it.

core/pkg/service/task/service.go
sy-nico: I added this because I discovered a bug  whenever a task status is deleted. (more details in the following replies). I reverted this file back to the rc state.

sy-nico: [SY-4247](https://linear.app/synnax/issue/SY-4247/bug-tasks-break-when-you-delete-their-statuses): task statuses live in a separate table, and `task.retrieve` with `IncludeStatus` joins them back on positionally (`for i, stat := range statuses`). If a status row is deleted out-of-band, the fetched `statuses` slice no longer lines up with `res.Tasks`, so statuses get attached to the wrong task or the retrieve misbehaves. Every task does get a status at creation, so this only happens when a row is removed after the fact. Tracking the proper fix (read-only, key-matched join plus repair on `configure`) in a separate ticket.

After configuring, deleting the corresponding status, and then configuring again:
image

```
ERROR: Failed to create Arc

Description: Statuses with keys [task:281483566645249] not found: not found: query

Stack Trace:
decode$1@http://localhost:5173/@fs/Users/nicoalba/Desktop/synnaxlabs/synnax/client/ts/dist/client.js:12857:76
decode@http://localhost:5173/@fs/Users/nicoalba/Desktop/synnaxlabs/synnax/x/ts/src/errors/errors.ts:111:33
@http://localhost:5173/@fs/Users/nicoalba/Desktop/synnaxlabs/synnax/client/ts/dist/client.js:12568:36
```

arc/go/symbol/symbol.go
sy-nico: I think you're right, and I'm maybe leaning the same way. The hard part is that I don't think I'll be able to properly weigh the hook approach against a language-level construct until we actually hit the case you're describing, like validating range/status keys at compile-time, since that's where the tradeoffs become concrete. For now the hook is constrained to only emitting diagnostics, so I'll keep it as-is for this PR and open a ticket to revisit this once we start on that.

[SY-4248](https://linear.app/synnax/issue/SY-4248/reconsider-compile-time-constrained-string-validation-in-the-language)

core/pkg/api/status/status.go
sy-nico: I'm going to leave this one open for now and revisit after everything else is addressed and cleaned up. Whenever I play around and think how this would work in a "real" Arc program, I always go back to needing to delete by name. Until we release some other important features, it's not viable to do anything by key. 

We already received similar criticism from a customer about setting the pager_duty task by key.  Consistency between module functions should be weighed more than consistency with clients, and even go as far as making a plan to update the other clients to be consistent with what the users needs.

I'll keep trying to make clean examples that delete statuses by key and keep this thread open while I resolve the rest.

arc/go/stl/strings/string.go
sy-nico: I went back and dug into this more because I started to confuse myself too.

My original reasoning was wrong I'd said a guest could call `format_*` during init before memory is wired, but that can't actually happen right now. Arc doesn't emit a WASM start function, so no guest code runs at instantiation. Anything that calls `format_*` or `from_literal` lives inside a flow-node function, and those only run at runtime, well after `SetMemory` has wired up guest memory. So `memory` is never nil at that point and the guard never trips.

The guard is defensive parity with the existing `from_literal` guard above it. This is going to be important once we add `global` variables with initializers like `global i64 x := some_chan`. Those evaluate at load time, which would compile to a start function and could reach the host before memory is set.

I'm going to keep this here, rather than place this in another PR, because it's a single nil check, so the runtime cost is negligible, and it brings us to parity now while covering globals later. Plus, I've already deferred a few fixes to other issues, so I'd like to make sure this simple fix gets in.

sy-nico:
Will resolve conflicts last.

arc/go/stl/strings/string.go
pjdotson: discussed inperson, fix as you will

core/pkg/api/status/status.go
pjdotson: we can talk through this again when we talk about whether deleting by name should exist
at all, but I don't think we return the count response. For the rest of our APIs it is
completely fine to delete things that are not found, and if I am in general thinking
about Arc deletion clients, it's hard to justify informing the user about this scenario
for Arc but not C++,TypeScript and Python.

arc/go/symbol/symbol.go
pjdotson: yeah, probably fine for right now but we should think about it more.

arc/go/text/analyze.go
```diff
@@ -494,7 +494,12 @@ func analyzeFunctionNode(
 		Channels: sym.Channels.Copy(),
 		Config:   slices.Clone(freshType.Config),
 		Outputs:  slices.Clone(freshType.Outputs),
-		Inputs:   slices.Clone(freshType.Inputs),
+	}
+	// STL ExecBoth inputs mirror config; in flow form the upstream is a
+	// trigger, so omit Inputs. User-defined funcs (AST != nil) keep them.
+	upstreamIsTrigger := sym.Exec == symbol.ExecBoth && sym.AST == nil
+	if !upstreamIsTrigger {
+		n.Inputs = slices.Clone(freshType.Inputs)
```
pjdotson, line 502: add test

driver/arc/status/status_test.cpp
```diff
@@ -33,205 +38,346 @@ ::arc::runtime::node::Context make_context() {
...
-    EXPECT_NE(timestamps[0], 0);
-    EXPECT_NE(timestamps[1], 0);
-    EXPECT_GE(timestamps[1], timestamps[0]);
-}
+    ASSERT_NIL(created->next(ctx));
+    ASSERT_EQ(calls.size(), 1u);
+    EXPECT_NE(calls[0].second.find("no status found"), std::string::npos);
 }
+
+} // namespace driver::arc::status
```
pjdotson: bad comment

core/pkg/api/grpc/status/handler.go
```diff
@@ -199,6 +219,70 @@ func (t deleteRequestTranslator) Backward(
 	return apistatus.DeleteRequest{Keys: msg.Keys}, nil
 }
 
+func (t setByKeyOrNameRequestTranslator) Forward(
```
pjdotson: nit: no reason to declare the receiver `t` for these

core/pkg/api/status/status_suite_test.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package status
```
pjdotson: should be in package `status_test`

core/pkg/api/status/status_test.go
```diff
@@ -0,0 +1,266 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package status
```
pjdotson: should be in package `status_test`

core/pkg/service/arc/status/status.go
```diff
@@ -0,0 +1,331 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package status
```
pjdotson, line 10: i think for files like this, we should split them up by functions, so we have a
`set.go`, `delete.go`, and general `status.go` or `module.go` files. don't feel strongly
so you tell me if you agree / disagree

core/pkg/service/arc/status/status.go
```diff
@@ -0,0 +1,331 @@
...
+
+const (
+	setMemberName    = "set"
+	deleteMemberName = "delete"
+	moduleName       = "status"
+)
+
+// allowedVariantsList is the human-readable list used in compile-time and
+// runtime diagnostics. Variant.IsValid is the source of truth for membership.
+const allowedVariantsList = "success, info, warning, error, loading, disabled"
```
pjdotson, line 42: just flagging as another reason it would be good to have the string enum support - this
is not needed, compiler message doesn't need to list available variants because you see
based off of the type

core/pkg/service/arc/runtime/task.go
```diff
@@ -330,6 +342,12 @@ func (t *taskImpl) Stop() error {
 	return nil
 }
 
+func (t *taskImpl) reporter() taskreporter.Reporter {
+	return func(ctx context.Context, variant status.Variant, message string) {
+		t.setStatus(ctx, variant, t.isRunning(), fmt.Sprintf("[%s] %s", t.task.Name, message))
+	}
+}
```
pjdotson, line 348: what happens with errors in the other ARC STL? how do we get these to change status of
the Arc? just wondering if this `taskreporter.Reporter` type makes the most sense

core/pkg/service/arc/status/status.go
```diff
@@ -0,0 +1,331 @@
...
+	stat *status.Service,
+	ins alamos.Instrumentation,
+	report taskreporter.Reporter,
+	keyOrName, message, variantStr string,
+) string {
+	key, multi, err := stat.SetByKeyOrName(ctx, keyOrName, message, variantStr)
+	if err != nil {
+		ins.L.Error("status.set failed",
+			zap.String("key_or_name", keyOrName), zap.Error(err))
+		report(ctx, xstatus.VariantWarning, fmt.Sprintf("status.set: %v", err))
```
pjdotson, line 191: why do we need the alamos instrumentation and the taskreporter?

pjdotson (changes requested):

arc/go/stl/strings/string.go
sy-nico: Deleted

core/pkg/service/arc/status/status.go
sy-nico: I'm leaning towards keeping in one file. I generally agree with you, though. 

The convention across the other A modules is one file per module holding all of its symbols, and none split per-function.
- channels: 5, 
- math: 3, 
- time: 3 
  
The multi-file modules (strings, series) split by concern like state.go, not by symbol. status.go follows that pattern, so splitting into set.go would make it the odd one out. 

Removing `status.delete` also makes this less of an issue. I absolutely want to reorganize... but after we refactor the ExecBoth shim so that we only have to define a function once. Then it will be a lot cleaner.

pjdotson: sounds good

core/pkg/service/arc/status/status.go
sy-nico: Removed the alamos side. I was being a little overly cautious, but we don't need to write this to the core logs.

core/pkg/service/arc/runtime/task.go
sy-nico: The flow form gets `ctx.ReportError` from the node context, but the func form runs as a WASM host function that only receives a plain `context.Context`, so `ReportError` isn't reachable there and a host fn can't trap without killing the task. The reporter was the surgical way to give both forms a single path to surface a non-fatal warning to the task status. `status.set` is also the first stdlib op that hits the DB and can fail at runtime while wanting to show the user what went wrong, so there wasn't an existing pattern to lean on.

That said, I'm exploring whether we can unify this onto `ctx.ReportError` so we don't introduce a parallel mechanism.

sy-nico: Tried it locally: unifying onto `ctx.ReportError` is small. The WASM node injects `ctx.ReportError` into the host-fn context, so func and flow forms share one path and `taskreporter` drops out. Other STLs would get the same surfacing path for free (future `ranges.set` etc.). Caveat: that path is warning-only, so a non-fatal *error* variant would need a wider contract.

This is another piece that will likely change after the ExecBoth refactor.

sy-nico:
Removed `status.delete` support per the offline discussion.

PR #2376, opened 2026-05-27
SY-4240: Fix Node Metrics Writer Failing on Backward Clock Steps

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4240](https://linear.app/synnax/issue/SY-4240)

## Description

A backward step in the host's wall clock (e.g. an NTP correction, common on Windows) would permanently fault a node's metrics writer. The collector stamped each sample with the raw wall clock, but Cesium requires commits within a domain to be monotonically non-decreasing. When the clock jumped backward, the next sample's timestamp landed before the previous commit, Cesium rejected it, and the stream writer latched the error for the rest of the process lifetime — the node's metrics channels stopped updating and the logs flooded with the same validation error (observed repeating ~1100 times within a single session on a customer's machine). The node never recovered without a restart, and a clock still behind the persisted data could re-trigger the fault on the next startup.

This stamps collected samples through `telem.MonoClock`, which clamps each timestamp to `previous + 1ns` whenever the underlying clock would otherwise stall or move backward, guaranteeing strictly increasing timestamps. The writer's `Start` is derived from the same clock so the start-of-domain validation can't be tripped either. The collection cadence is unchanged (still a monotonic `time.Ticker`) — only the sampled timestamp is sourced through the monotonic clock. As a side benefit this also fixes coarse clock-resolution duplicate timestamps on Windows.

The time source is injectable via `ServiceConfig.Now` (defaulting to `telem.Now`) purely so the regression test can simulate a backward clock step; production is unaffected.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes a permanent fault in the node metrics writer caused by backward wall-clock steps (e.g. NTP corrections on Windows). The collector previously stamped each sample with the raw wall-clock time; when the clock jumped backward, Cesium's monotonicity requirement rejected the sample and the stream writer latched that error for the rest of the process lifetime.

- **Timestamp delegation to `AutoIndex`**: The collector's manual timestamp emission is removed entirely. The writer is now opened with `AutoIndex: new(true)`, which generates timestamps on the leaseholder side using a per-leaseholder high-water mark that guarantees strict monotonicity — effectively preventing the backward-clock fault.
- **Frame structure simplification**: `idx` is demoted from a `collector` struct field to a local variable in `OpenService`, and the writer's `Keys` now only contains data channels (the index is implicitly opened by `AutoIndex`). The `Start: telem.Now()` that could itself be tripped by a stale clock is removed.
- **Test update**: The \"Should write metrics at configured interval\" test is updated to look up series by channel key rather than by positional index, since AutoIndex no longer places the index channel at a deterministic frame offset.

Safe to merge. The approach is idiomatic and the core change is a straightforward delegation to the existing AutoIndex monotonicity guarantee.

The change removes the wall-clock timestamp from the collector and delegates timestamp generation to the writer's AutoIndex feature, which maintains a per-leaseholder high-water mark to guarantee strict monotonicity. The logic is sound, the frame structure simplification is clean, and the test update correctly adapts to the new frame layout.

No files require special attention. All three changed files are consistent with each other and correctly implement the AutoIndex delegation pattern.

| Filename | Overview |
|----------|----------|
| core/pkg/service/metrics/service.go | Switches the stream writer from explicit timestamp keys + `Start: telem.Now()` to `AutoIndex: new(true)` with no `Start`, cleanly removing the wall-clock dependency from domain open. |
| core/pkg/service/metrics/collector.go | Removes the ticker-time capture and manual timestamp series from the emitted frame; metrics-only frames are built and sent to the AutoIndex-enabled writer. |
| core/pkg/service/metrics/metrics_test.go | Test updated to resolve series by channel key lookup instead of positional index, correctly adapting to AutoIndex frame layout. No regression test for backward clock step is included. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Ticker as time.Ticker
    participant Collector as collector
    participant Writer as StreamWriter (AutoIndex=true)
    participant Cesium as Cesium (storage)

    Note over Collector,Writer: Before: collector stamped raw wall-clock time
    Ticker->>Collector: tick (wall-clock time t)
    Collector->>Writer: "Frame{idx:t, cpu:v, mem:v, ...}"
    Writer->>Cesium: "write (Start=t, domain must be monotonic)"
    Note over Cesium: Clock steps backward → t < prev → FAULT (permanent)

    Note over Collector,Writer: After: AutoIndex generates monotonic timestamps
    Ticker->>Collector: tick (unused)
    Collector->>Writer: "Frame{cpu:v, mem:v, ...} (no timestamp)"
    Writer->>Writer: "AutoIndex: now = max(wallClock, lastTS+1)"
    Writer->>Cesium: "write (Start=autoNow, guaranteed monotonic)"
    Cesium-->>Writer: OK
```

<sub>Reviews (3): Last reviewed commit: ["preallocate frame"](https://github.com/synnaxlabs/synnax/commit/3f300c49440a5a8ffbc7e9f9b8cec4afd55be3bb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34108485)</sub>


emilbon99 (changes requested):
Is there any reason that we don't just use the new timestamp auto-generation feature?

emilbon99 (approved):

PR #2388, opened 2026-05-28
SY-4254: Add Stack Traces to C++ Driver Crashes

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4254](https://linear.app/synnax/issue/SY-4254)

## Description

The C++ driver previously died silently on a crash — a segfault, an unhandled exception, or a failed assertion produced no diagnostic, just a missing process. This adds a process-wide crash handler (`x::crash`, installed at startup in `driver/cmd/exec.cpp`) that dumps a stack trace to stderr before the process exits.

**What it captures**

- **POSIX** (`x/cpp/crash/unix.cpp`): `SIGSEGV`/`SIGABRT`/`SIGILL`/`SIGFPE`/`SIGBUS` via `sigaction`, plus unhandled C++ exceptions via `std::set_terminate` (including the exception's `what()`). The signal path is async-signal-safe (`backtrace` + `backtrace_symbols_fd`, no allocation) and runs on an alternate signal stack so **stack-overflow** crashes are still traced; the terminate path demangles symbols. Fatal signals are re-raised against the default handler to preserve core dumps and the 128+signal exit code.
- **Windows** (`x/cpp/crash/windows.cpp`): `SetUnhandledExceptionFilter` + `std::set_terminate`, symbolized via DbgHelp, with a reserved thread stack guarantee for stack overflow.
- `SIGINT`/`SIGTERM` are deliberately left to `x::shutdown` for graceful shutdown.

**Embedded-driver log integration**

The driver runs as a subprocess of the core server, which captures its stderr and re-logs it through `PipeToLogger`. Crash output is raw (not glog-formatted), so `PipeToLogger` now detects the crash banner and logs the dump **verbatim at Error level** — fixing severity (was INFO) and preventing the glog parser from mangling bracketed Linux frame addresses.

**Trace quality:** optimized release builds are stripped, so field traces show addresses that must be resolved offline (`addr2line` / `.pdb`); `-rdynamic` is added on POSIX to name exported frames. See the new section in `driver/DEBUGGING.md`, which also documents the `crash_demo` harness.

**Cross-platform note for reviewers:** verified locally on macOS (arm64). The feature-branch CI only runs the C++ suite on Linux (Windows/macOS are excluded until the `rc → main` PR); `windows.cpp` has not yet been compiled in CI. A `workflow_dispatch` on `test.x.cpp.yaml` will exercise all three OSes if you want that signal before the release PR.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds a process-wide crash handler (`x::crash`) to the C++ driver that dumps a symbolized stack trace to stderr before the process dies, along with a refactored Go-side supervisor that replaces the old recursive restart approach with an iterative `restart.Policy` backed by configurable backoff and a healthy-uptime reset heuristic. The previously-flagged `close of closed channel` panic on driver restart is fixed by lifting `startedOnce` out of the per-run scope so `d.started` is closed exactly once regardless of how many times the driver is restarted.

- **C++ crash handler** (`x/cpp/crash`): POSIX path uses `sigaction`+`SA_ONSTACK`+async-signal-safe `backtrace_symbols_fd`, plus a `std::set_terminate` handler with demangling; Windows path uses `SetUnhandledExceptionFilter`+DbgHelp, returning `EXCEPTION_CONTINUE_SEARCH` to preserve WER/minidump semantics. Both serialize concurrent crashes via `atomic_flag`.
- **Go log piping** (`log.go`): detects the crash banner and accumulates raw dump lines until pipe EOF, emitting them as a single `ERROR` entry; also suppresses the Go-side stacktrace on forwarded driver errors via `zap.AddStacktrace(FatalLevel)`.
- **Supervisor refactor** (`enabled.go`, `restart/`): iterative restart loop with exponential backoff, `MaxRetries` cap, healthy-uptime counter reset, and `HardShutdown` to avoid deadlocking against a backoff sleep during `Close`.

Safe to merge; the crash handler, log piping, and supervisor refactor are all well-tested and the previously-reported double-close regression is correctly fixed.

All three changed subsystems have targeted unit and integration tests. The startedOnce fix is clean and exercised by a dedicated regression test. The two findings are low-probability diagnostic quality issues that do not affect normal operation or correctness.

No files require special attention, though core/pkg/driver/internal/log/log.go carries a minor heuristic risk in isCrashBanner and x/cpp/crash/windows.cpp silently degrades to address-only traces when SymInitialize fails.

| Filename | Overview |
|----------|----------|
| core/pkg/driver/enabled.go | Refactors the supervisor loop: replaces recursive restart-on-signal-exit with a clean iterative loop over `runOnce`, introduces `restart.Policy` for backoff/retry logic, and fixes the double-close panic by sharing `startedOnce` across all restarts. The stdin STOP write is moved into `cmd.Cancel` so `exec.CommandContext` owns graceful-stop semantics. |
| core/pkg/driver/internal/log/log.go | Adds crash-dump detection and accumulation: once a banner line is seen, all subsequent lines are buffered and emitted as a single ERROR entry on EOF. Also suppresses the Go-side stacktrace on forwarded driver errors via `zap.AddStacktrace(FatalLevel)`. The `isCrashBanner` heuristic uses `strings.Contains`, which can false-positive on normal driver log messages. |
| core/pkg/driver/internal/restart/restart.go | New package wrapping `breaker.Breaker` with a `HealthyUptime` reset: an unexpected exit after long uptime resets the consecutive-failure counter so a single failure after a long healthy run does not count toward `MaxRetries`. Logic is straightforward and well-tested. |
| x/cpp/crash/unix.cpp | POSIX crash handler: installs a shared `signal_handler` via `SA_ONSTACK` for five fatal signals, plus a `terminate_handler` via `std::set_terminate`. Alternate signal stack pre-allocated statically; unwinder pre-warmed at install time. `claim_crash` serializes concurrent crashes via `atomic_flag`. `demangle_line` handles both Apple and Linux `backtrace_symbols` output formats. |
| x/cpp/crash/windows.cpp | Windows crash handler using `SetUnhandledExceptionFilter` + `std::set_terminate`. Returns `EXCEPTION_CONTINUE_SEARCH` correctly to preserve WER/minidump semantics. `capture_trace` initializes and cleans up the DbgHelp session on every call, but does not check the `SymInitialize` return value — a failure leaves addresses without symbol names with no indication in the output. |
| x/cpp/crash/crash_test.cpp | Comprehensive test suite using death tests for signal and unhandled-exception paths, plus deterministic in-process tests for handler registration, alternate-signal-stack, and `capture_trace`. The `noexcept` boundary trick to reach the terminate handler without gtest intercepting the throw is well-designed. |
| core/pkg/driver/enabled_test.go | New restart-behavior tests exercise transient crashes, exhaust-and-give-up, graceful stop, and the specific double-close regression. Uses a thread-safe `syncBuffer` to avoid data races between logging goroutines and test assertions. |
| core/pkg/driver/driver.go | Adds `RestartBaseInterval`, `RestartMaxRetries`, and `RestartHealthyUptime` config fields (all tagged `json:"-"` so they are core-side only) with sensible defaults: 2s base, 100 retries, 1-minute healthy-uptime window. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Open as driver.Open()
    participant Sup as supervisor goroutine
    participant RunOnce as runOnce()
    participant Proc as C++ driver process
    participant Pipe as PipeToLogger (stderr)
    participant Policy as restart.Policy

    Open->>Sup: launch (sCtx, startedOnce shared)
    loop restart loop
        Sup->>RunOnce: call(ctx, policy, startedOnce)
        RunOnce->>Proc: cmd.Start()
        Proc-->>Pipe: log lines / crash banner
        Pipe-->>Open: startedOnce.Do(close started) [first run only]
        Open-->>Open: unblock caller
        alt normal exit / crash
            Proc-->>RunOnce: cmd.Wait() returns
            Pipe-->>Pipe: flush crashLines as single ERROR
            RunOnce->>Policy: "Decide(expected=false, uptime)"
            Policy-->>RunOnce: Restart / GiveUp
        else Close() called
            Open->>Sup: cancel sCtx (HardShutdown)
            Note over Proc: cmd.Cancel writes STOP, WaitDelay escalates to kill
            Proc-->>RunOnce: cmd.Wait() returns
            RunOnce-->>Sup: "ctx.Err()!=nil, return nil"
        end
        alt "action == Restart"
            Note over Sup: backoff elapsed inside Policy.Decide
        else "action == GiveUp"
            Sup-->>Open: return error
        end
    end
```

<sub>Reviews (7): Last reviewed commit: ["improved enabled mode"](https://github.com/synnaxlabs/synnax/commit/34f5629ad5472067b17363630720a7896de57f95) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34432135)</sub>


x/cpp/crash/unix.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/// @brief captures and writes the current stack trace to fd. Async-signal-safe:
+/// backtrace and backtrace_symbols_fd do not allocate. Symbol names are left mangled
+/// because demangling requires the allocator.
+void print_trace_safe(const int fd) {
+    void *frames[MAX_FRAMES];
+    const int n = backtrace(frames, MAX_FRAMES);
+    backtrace_symbols_fd(frames, n, fd);
+}
```
sy-nico, line 89: The `crash.h` doc says the signal path "allocates no memory and calls only async-signal-safe functions," but `print_trace_safe()` calls `backtrace()`, and on glibc the first `backtrace()` call lazily `dlopen`s libgcc and calls `malloc`. If the crash is a SIGSEGV taken while the malloc lock is held, the handler can deadlock inside that first `backtrace()`. Suggest pre-warming with a throwaway `backtrace()` call in `install()` to force the lazy load, or softening the doc claim.

Repro: build the driver/demo for Linux (glibc), then crash while the allocator lock is held, e.g. corrupt the heap and call `free()`/abort inside `malloc`, or in `crash_demo` add a mode that does:

    void *p = malloc(32); ((size_t *)p)[-1] = 1; free(p); // fault with allocator lock held

Without the pre-warm, the first `backtrace()` in the handler tries to `malloc` and hangs.

core/pkg/driver/enabled.go
```diff
@@ -196,19 +205,27 @@ func (d *Driver) start(ctx context.Context) error {
...
-			}
-		}
-		if isSignal {
+		expected := d.closing.Load()
+		if !expected {
 			d.cfg.L.Warn("embedded driver process exited unexpectedly", zap.Error(err))
+		}
+		switch policy.Decide(expected, time.Since(startedAt)) {
+		case restart.Restart:
+			return runProcess(ctx)
```
sy-nico: I think you should still convert the supervisor's restart from recursion to a for-loop so the goroutine's call stack and its accumulated `defer` cleanups no longer grow one frame per restart over the process lifetime. This also runs each iteration's temp-file cleanup between restarts instead of piling them up to fire at unwind, eliminating the spurious "failed to remove" errors.

core/pkg/driver/driver.go
```diff
@@ -123,15 +134,18 @@ var (
...
-		TaskOpTimeout:       time.Second * 60,
-		TaskPollInterval:    time.Second * 1,
-		TaskShutdownTimeout: time.Second * 30,
-		TaskWorkerCount:     4,
+		Integrations:         []string{},
+		Enabled:              new(true),
+		Debug:                new(false),
+		StartTimeout:         time.Second * 10,
+		StopTimeout:          10 * time.Second,
+		RestartBaseInterval:  time.Second,
```
sy-nico: On a crash loop the driver repeatedly drops and re-opens its writer/streamer connections to the Core. How does the Core handle a rapid reconnect storm, i.e. the same channels' writer torn down and re-opened many times per second? 

Is the crashed instance's writer/lease reliably reaped before the restarted driver opens a new one, or can we transiently end up with conflicting or orphaned writers (and churned device-create requests from the scan task)? `RestartBaseInterval` defaults to `1s`, which throttles this, but the loop can still cycle fast under a low base interval. Mostly asking to confirm the floor is high enough to protect the cluster, or if it should be raised to something like `5s`.

core/pkg/driver/enabled.go
```diff
@@ -237,6 +254,9 @@ func (d *Driver) Close() error {
 }
 
 func (d *Driver) close() error {
```
sy-nico, line 255: `close()` reads `d.cmd` and `d.stdInPipe` without holding `d.mu`, but the supervisor goroutine reassigns both on every restart: `d.cmd` in `setupCmd` under `d.mu` (L336), and `d.stdInPipe` at L172 without the lock at all. Now that restart is the common path, a `Close()` that lands while a restart is re-running setup races on these fields, risking a write to a stale/closed pipe or a nil deref. 

Suggest guarding these reads with `d.mu` (and taking the lock around the `d.stdInPipe` assignment too), and running the suite with `-race`.

core/pkg/driver/enabled.go
```diff
@@ -196,19 +205,27 @@ func (d *Driver) start(ctx context.Context) error {
...
-			isSignal = strings.Contains(err.Error(), "signal") || strings.Contains(err.Error(), "exit status")
-			if !isSignal && bre.Wait() {
-				d.cfg.L.Warn("embedded driver process crashed", zap.Error(err))
-				return runProcess(ctx)
-			}
-		}
-		if isSignal {
+		expected := d.closing.Load()
+		if !expected {
 			d.cfg.L.Warn("embedded driver process exited unexpectedly", zap.Error(err))
```
sy-nico: On a clean exit `err` is nil, so this logs `"exited unexpectedly"` with `error: null`; consider gating the log on `err != nil`.

x/cpp/crash/windows.cpp
```diff
@@ -0,0 +1,121 @@
...
+    SymInitialize(process, nullptr, TRUE);
+
+    void *frames[MAX_FRAMES];
+    const USHORT n = CaptureStackBackTrace(0, MAX_FRAMES, frames, nullptr);
+
+    // SYMBOL_INFO is variable-length (the name trails the struct); allocate an aligned
+    // buffer large enough for the struct plus the name, per the DbgHelp contract.
+    ULONG64 buffer
+        [(sizeof(SYMBOL_INFO) + MAX_SYMBOL_NAME + sizeof(ULONG64) - 1) /
+         sizeof(ULONG64)];
```
sy-nico, line 112: None of the Windows path is built in CI here: this file, the `/DEFAULTLIB:dbghelp.lib` linkopt, and this `SYMBOL_INFO` buffer math have never been compiled. The `windows-build-bot` matrix entry in test.x.cpp.yaml isn't actually runnable, so CI can't cover this before the rc -> main PR. Can someone compile `bazel test //x/cpp/crash:crash_test` on a Windows machine locally and confirm it builds and the handler test passes before merge, rather than shipping windows.cpp unverified?

x/cpp/crash/windows.cpp
```diff
@@ -0,0 +1,121 @@
...
+    ULONG guarantee = 64 * 1024;
+    SetThreadStackGuarantee(&guarantee);
+    SetUnhandledExceptionFilter(exception_filter);
+    std::set_terminate(terminate_handler);
+}
+
+std::string capture_trace() {
+    const HANDLE process = GetCurrentProcess();
+    SymSetOptions(SYMOPT_UNDNAME | SYMOPT_DEFERRED_LOADS | SYMOPT_LOAD_LINES);
+    SymInitialize(process, nullptr, TRUE);
```
sy-nico, line 103: `SymInitialize`/`SymCleanup` operate on a global per-process DbgHelp session and DbgHelp is single-threaded, so two concurrent `capture_trace()` calls will race (one's `SymCleanup` tears down the session the other is using). That's fine for the single-shot crash path, but `crash.h:48-49` advertises `capture_trace()` for voluntary diagnostics on non-fatal paths, which invites concurrent use. Either guard the body with a mutex, or document in `crash.h` that `capture_trace()` is not safe for concurrent use on Windows.

sy-nico (changes requested):

x/cpp/crash/windows.cpp
emilbon99: Already done.

core/pkg/driver/enabled.go
emilbon99: This never gets logged on a nominal shutdown path, so regardless of whether it exited with a clean code or not, it's still unexpected.

core/pkg/driver/driver.go
emilbon99: I bumped this to `2s`, but after extra research i'm not very concerned.

core/pkg/driver/enabled.go
```diff
@@ -254,8 +284,9 @@ func (d *Driver) close() error {
 		return err
 	case <-time.After(d.cfg.StopTimeout):
 		d.cfg.L.Warn("embedded driver did not stop within grace period, escalating to kill")
-		if d.cmd != nil && d.cmd.Process != nil {
-			if killErr := d.cmd.Process.Kill(); killErr != nil {
+		// Re-read the process: a restart may have replaced cmd while we waited.
+		if cmd, _ := d.currentProcess(); cmd != nil && cmd.Process != nil {
+			if killErr := cmd.Process.Kill(); killErr != nil {
 				d.cfg.L.Error("failed to kill embedded driver process", zap.Error(killErr))
```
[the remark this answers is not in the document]
sy-nico: +1, this needs to be fixed before merge:

`d.started` is created once in Open; `startedOnce` is fresh per `runOnce`, so each restart runs `close(d.started)` on an already-closed channel and panics. Recover keeps us alive but logs a panic + stacktrace on every restart, which is bad in a PR about clean crash diagnostics.

Only the first start matters (it's all Open blocks on), so passing `started`/`startedOnce` on the first iteration only and `nil` after is the simplest fix. PipeToLogger already nil-guards `startedOnce`.

sy-nico (changes requested):

PR #2397, opened 2026-05-29
SY-4260: Fix Bad Schematic & Table Import

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4260](https://linear.app/synnax/issue/SY-4260)

## Description

Fixes several issues with importing and opening schematics (and the analogous table paths) after the move to server-typed schematic/table state.

- **Blank import of legacy files.** `parseImport` tried the typed `schematicZ` / `tableZ` first. Those schemas are lenient — they strip unknown keys and default `configs` / `rows`/`columns`/`cells` to empty — so a legacy console export (which stores symbol state under `props`, or table structure under `layout`) parsed "successfully" with everything dropped, producing a blank schematic/table. The strict legacy state schemas are now tried first, falling back to the typed direct parse only when the data isn't a legacy export.

- **Required `workspaceKey` on the file ingester.** `Import.FileIngesterContext.workspaceKey` is now a required `workspace.Key` (no longer `| undefined`); the value is defaulted to `uuid.ZERO` at each context-construction site, so ingesters can rely on it.

Adds regression tests for both the schematic and table import paths.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes two bugs in the schematic and table import paths introduced when the system moved to server-typed state: blank imports of legacy files (caused by the lenient typed schema silently accepting and stripping legacy data), and a required `workspaceKey` field on `FileIngesterContext`.

- **Blank legacy import fix**: `parseImport` now tries the strict legacy `anyStateZ` schema first (requiring specific version literals and console-only fields). When that succeeds it extracts data from `pendingUpload`; when it fails it falls through to the typed parse path.
- **`workspaceKey` fix**: The field is now typed as required `workspace.Key` on `FileIngesterContext`; callers default to `uuid.ZERO` instead of forwarding `undefined`.
- **Regression tests**: Both schematic and table import paths get spec files covering legacy, typed, and silent-drop scenarios.

Safe to merge — the parse-order inversion is logically sound and the key invariant (strict legacy schemas require console-only fields that typed server exports never carry) is verified by the new regression tests.

The legacy-first approach is correct: every legacy state schema requires a specific version literal plus console-only fields that typed server exports never carry, so the two branches are mutually exclusive in practice. The workspaceKey change is mechanical and all three construction sites are updated consistently.

| Filename | Overview |
|----------|----------|
| console/src/schematic/services/import.ts | Core fix: legacy-first parse order correctly prevents lenient typed schema from silently consuming legacy files; workspaceKey nullish guard removed after upstream default. |
| console/src/table/services/import.ts | Mirrors schematic fix for table import; same legacy-first parse order. |
| console/src/schematic/services/import.spec.ts | New regression tests covering legacy v2, typed v6, and silent-drop prevention for schematic import. |
| console/src/table/services/import.spec.ts | New regression tests covering legacy v0, typed v1, and silent-drop prevention for table import. |
| console/src/import/ingester.ts | Makes workspaceKey required on FileIngesterContext, removing the optionality that forced downstream guards. |
| console/src/import/import.ts | Defaults workspaceKey to uuid.ZERO before constructing context. |
| console/src/import/dataTransferItem.ts | Second context-construction site updated with the same uuid.ZERO default. |
| console/src/schematic/selectors.ts | Replaces Diagram.Viewport import with locally-defined Viewport type that includes the mode field; no logic change. |
| console/src/schematic/types/v6.ts | Renames ViewportState to Viewport and minor formatting cleanup; no logic changes. |
| console/src/schematic/types/index.ts | Re-exports the new Viewport interface from v6 through the types barrel. |
| console/src/schematic/slice.ts | Re-exports Viewport from latest types so selectors can import it from the slice module. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Import file data] --> B{data is object?}
    B -- No --> E
    B -- Yes --> C[anyStateZ.safeParse with remoteCreated: false]
    C -- success --> D{pendingUpload == null?}
    D -- Yes --> ERR[throw: no graph data]
    D -- No --> OK1[Extract nodes/edges/configs - Return schematic.New]
    C -- fail --> E[schematicZ.parse / tableZ.parse]
    E -- success --> OK2[Return typed schematic.New / table.New]
    E -- fail --> ZERR[throw ZodError]
```

1. `console/src/schematic/selectors.ts`, line 82-95 ([link](https://github.com/synnaxlabs/synnax/blob/d45cc01a5992fd26f1feccde9b02e3b4c0620c74/console/src/schematic/selectors.ts#L82-L95)) 

   <a href="#">P1</a> **Selectors used in `Internal` still lack null guards**

   The PR description says the crash is fixed by "reading through individual selectors with graceful fallbacks." Four of the seven selectors now used in `Internal` do have `?.` guards (`selectControlStatus`, `selectAuthority`, `selectLegend`, `selectSelected`), but `selectEditable`, `selectViewport`, and `selectFitViewOnResize` access `select(state, key)` directly without `?.`.

   `useAutoUpload` returns `true` when the Redux state entry doesn't exist, because `selectPendingUpload` uses `?.pendingUpload` — so `undefined == null` evaluates to `true` and `Internal` renders. On that first render `select(state, key)` returns `undefined`, and `undefined.editable` / `undefined.viewport` / `undefined.fitViewOnResize` all throw — producing the same crash the PR is trying to eliminate, just with a different error message.
</details>

<sub>Reviews (2): Last reviewed commit: ["go back to nondesctructure"](https://github.com/synnaxlabs/synnax/commit/6d41ad50365c90a31f0aa7f1bf4e55afedc2f649) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34615586)</sub>

### DIFF
--- a/console/src/schematic/Schematic.tsx
+++ b/console/src/schematic/Schematic.tsx
@@ -25,16 +25,7 @@ import { Layout } from "@/layout";
 import { Controller } from "@/schematic/Controller";
 import { Controls } from "@/schematic/Controls";
 import { useHandleNodeClickAction } from "@/schematic/navigate";
-import {
-  selectEditable,
-  useSelectAuthority,
-  useSelectControlStatus,
-  useSelectEditable,
-  useSelectFitViewOnResize,
-  useSelectLegend,
-  useSelectSelected,
-  useSelectViewport,
-} from "@/schematic/selectors";
+import { selectEditable, useSelect } from "@/schematic/selectors";
 import {
   setEditable,
   setFitViewOnResize,
@@ -50,13 +41,15 @@ const Internal: Layout.Renderer = ({ layoutKey: key, visible }) => {
   Base.useEnsureRetrieved({ key });
   const isSnapshot = Base.useSelectSnapshot({ key });
   const dispatch = useDispatch();
-  const editable = useSelectEditable(key);
-  const viewport = useSelectViewport(key);
-  const controlStatus = useSelectControlStatus(key);
-  const selected = useSelectSelected(key);
-  const legend = useSelectLegend(key);
-  const authority = useSelectAuthority(key);
-  const fitViewOnResize = useSelectFitViewOnResize(key);
+  const {
+    editable,
+    viewport,
+    controlStatus,
+    selected,
+    legend,
+    authority,
+    fitViewOnResize,
+  } = useSelect(key);
 
   const hasUpdatePermission =
     Access.useUpdateGranted(schematic.ontologyID(key)) && !isSnapshot;
--- a/console/src/schematic/Schematic.tsx
+++ b/console/src/schematic/Schematic.tsx
@@ -25,7 +25,16 @@ import { Layout } from "@/layout";
 import { Controller } from "@/schematic/Controller";
 import { Controls } from "@/schematic/Controls";
 import { useHandleNodeClickAction } from "@/schematic/navigate";
-import { selectEditable, useSelect } from "@/schematic/selectors";
+import {
+  selectEditable,
+  useSelectAuthority,
+  useSelectControlStatus,
+  useSelectEditable,
+  useSelectFitViewOnResize,
+  useSelectLegend,
+  useSelectSelected,
+  useSelectViewport,
+} from "@/schematic/selectors";
 import {
   setEditable,
   setFitViewOnResize,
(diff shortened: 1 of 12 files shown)


console/src/schematic/Schematic.tsx
```diff
@@ -41,15 +50,13 @@ const Internal: Layout.Renderer = ({ layoutKey: key, visible }) => {
...
-  const {
-    editable,
-    viewport,
-    controlStatus,
-    selected,
-    legend,
-    authority,
-    fitViewOnResize,
-  } = useSelect(key);
+  const editable = useSelectEditable(key);
```
emilbon99: Can you explain why this is better? Seems like we're just mounting a lot more hooks for react to track and re-evalute for no real reason.

console/src/schematic/services/import.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { type record } from "@synnaxlabs/x";
+import { describe, expect, it } from "vitest";
+
+import { parseImport } from "@/schematic/services/import";
+
+const configsOf = (s: { configs?: unknown }): record.Unknown =>
```
emilbon99, line 15: Kind of a weird helper. We're sure we want it?

emilbon99 (changes requested):

console/src/schematic/services/import.spec.ts
pjdotson: yeah makes the code slightly easier to write via avoiding type assertions

console/src/schematic/Schematic.tsx
pjdotson: Originally switched for debugging something, removed.

emilbon99 (approved):

PR #2400, opened 2026-05-29
SY-4265: Gate Arc LSP Hover & Highlighting on Module Imports

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4265](https://linear.app/synnax/issue/SY-4265)

## Description

The Arc LSP treated standard-library and host-injected modules (`time`, `math`, `control`, …) as ambient: they live above the program root and are always present, so hover and semantic highlighting surfaced them even when the document had not imported them. The analyzer correctly reports an unimported `time.now` as an undefined reference, so users saw an error squiggle while simultaneously getting full hover docs and function highlighting — the "weird with modules" behavior.

This change makes hover and semantic highlighting respect the same import gate the analyzer enforces:

- **Hover** (`resolveDotted`) now resolves a qualified head with normal user-code resolution (no `IncludeInternal`). A module reachable only through the ambient prelude — i.e. not imported — no longer resolves, so hovering `time.now` without `import time` returns nothing, matching the diagnostic. Imported modules still hover via their alias.
- **Semantic tokens** no longer color a post-`.` member as a function unconditionally. The member is colored by its symbol kind only when the qualifier resolves to an imported module that actually has that member; an unimported or unknown qualifier yields no token.

Completion is intentionally unchanged: offering `module.member` with an auto-import edit (which inserts the missing `import`) is an existing feature, not the bug.

A few pre-existing hover/semantic tests relied on the old behavior (hovering `time.now`, `math.avg`, etc. with no import). They are preserved, updated to import the module and use parseable program forms.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR gates the Arc LSP's hover and semantic-token highlighting behind the same module-import check the analyzer uses. Previously, modules in the ambient prelude (e.g. `time`, `math`, `control`) were always resolvable for hover and coloring, producing a confusing UX where `time.now` showed docs and was colored as a function even though the analyzer simultaneously raised an "undefined" diagnostic.

- **Hover** (`resolveDotted`): drops `IncludeInternal` so un-imported module heads no longer resolve, making hover return `nil` to match the diagnostic.
- **Semantic tokens** (`classifyTokenAt`): instead of unconditionally coloring a post-`.` identifier as a function, the qualifier is now looked up in the symbol table; coloring is applied only when the qualifier resolves to an imported module that actually has the member. All previously passing tests are updated to include the required `import` statement, and new regression tests cover both the negative (unimported → no hover/color) and positive (imported → hover/color) cases.

Safe to merge — the changes are narrowly scoped to hover and semantic token paths, with no impact on the analyzer, completion, or any data-flow logic.

Both hover and semantic token paths now correctly mirror the analyzer's import gate. The qualifier extraction via `allTokens[i-2]` is reliable for normal `module.member` patterns and fails safely for unusual token arrangements. The existing test suite was fully updated to require import statements, and new regression tests verify both the negative and positive behaviors for each path.

No files require special attention; all four files are straightforward and the changes are self-contained within the LSP layer.

| Filename | Overview |
|----------|----------|
| arc/go/lsp/hover.go | Removes `IncludeInternal` from `resolveDotted`'s head resolution, correctly blocking hover for un-imported ambient modules. Logic is clean and minimal. |
| arc/go/lsp/semantic.go | Adds qualifier-extraction and scope-resolution before coloring a post-DOT identifier. The `allTokens[i-2]` index assumption is correct for normal `module.member` patterns and fails safely for unusual token arrangements; no false positives possible. |
| arc/go/lsp/hover_test.go | Existing tests updated to include required `import` statements with adjusted positions; two new tests (imported/unimported) clearly cover the gated behavior. |
| arc/go/lsp/semantic_test.go | Two new tests pin the negative (unimported qualifier → no function token at col 29) and positive (imported qualifier → function token at col 29) cases; column arithmetic verified as correct. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[LSP Request\nhover or semantic tokens] --> B{Request type?}

    B -->|Hover| C[resolveDotted\nhead = 'time', tail = 'now']
    C --> D[scope.Resolve ctx, head\nno IncludeInternal]
    D -->|Not imported| E[err != nil → return nil\nNo hover rendered]
    D -->|Imported| F[recurse: resolveDotted ctx, sym, tail]
    F --> G[Return symbol → render hover docs]

    B -->|Semantic tokens| H[extractSemanticTokens\nfor each token i]
    H --> I{prevType == DOT\n&& i >= 2?}
    I -->|No| J[classifyToken normal path]
    I -->|Yes| K[qualifier = allTokens i-2 .GetText]
    K --> L[findScopeAtInternalPosition\nposition of member token]
    L --> M[scope.Resolve ctx, qualifier\nno IncludeInternal]
    M -->|Not imported / err| N[return nil\nNo semantic token emitted]
    M -->|Imported| O[mod.Resolve ctx, memberText]
    O -->|member not found| N
    O -->|member found| P[mapSymbolKind member.Kind\nEmit colored token]
```

<sub>Reviews (1): Last reviewed commit: ["SY-4265: Gate Arc LSP hover and highligh..."](https://github.com/synnaxlabs/synnax/commit/991ea5c579877a37c0e430f82ba9372173fef3ac) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34628506)</sub>

### DIFF
--- a/arc/go/lsp/hover.go
+++ b/arc/go/lsp/hover.go
@@ -428,15 +428,18 @@ func cleanDocComment(comments []string) string {
 // resolveDotted resolves a possibly-qualified name like "time.now" by
 // splitting on dots and walking: the head is resolved in scope, each
 // subsequent segment is resolved against the previous result's children.
-// IncludeInternal is passed so bare module names (filtered out of normal
-// user-code resolution) are reachable for hover and similar tooling.
+// Resolution mirrors user-code resolution (no IncludeInternal), so a
+// module that is reachable only through the ambient prelude — i.e. one the
+// document has not imported — does not resolve. Hovering a member of an
+// unimported module therefore returns nothing, matching the analyzer's
+// "undefined" diagnostic instead of rendering docs as if it were valid.
 func resolveDotted(
 	ctx context.Context,
 	scope *symbol.Symbol,
 	name string,
 ) (*symbol.Symbol, error) {
 	head, tail, hasDot := strings.Cut(name, ".")
-	sym, err := scope.Resolve(ctx, head, symbol.IncludeInternal)
+	sym, err := scope.Resolve(ctx, head)
 	if err != nil {
 		return nil, err
 	}
--- a/arc/go/lsp/hover_test.go
+++ b/arc/go/lsp/hover_test.go
@@ -116,13 +116,13 @@ var _ = Describe("Hover", func() {
 		})
 
 		It("should provide hover for 'control.set_authority' function", func(ctx SpecContext) {
-			content := "control.set_authority{value=255}"
+			content := "import control\n\ntrig -> control.set_authority{value=255}"
 			OpenArcDocument(server, ctx, uri, content)
 
 			hover := MustSucceed(server.Hover(ctx, &protocol.HoverParams{
 				TextDocumentPositionParams: protocol.TextDocumentPositionParams{
 					TextDocument: protocol.TextDocumentIdentifier{URI: uri},
-					Position:     protocol.Position{Line: 0, Character: 12},
+					Position:     protocol.Position{Line: 2, Character: 20}, // control.set_a|uthority
 				},
 			}))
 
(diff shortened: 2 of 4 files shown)


pjdotson:
still a little odd to get the suggestion for `time.now` when I haven't yet imported `time`

image

Compare to Go

image

pjdotson (approved):

PR #2406, opened 2026-05-29
SY-4269: Reject Arc Flow Nodes With Unconnected Required Inputs

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4269](https://linear.app/synnax/issue/SY-4269)

## Description

A text-mode Arc program such as `trigger -> (1 > 0) select{} -> { true: ... }` compiled successfully but crashed the entire server node at run time. `select`'s required `u8` selector input was never wired (the `(1 > 0) select{}` juxtaposition produced no edge into it), so at runtime `node.New(...).Node()` tried to materialize a series from the input's `nil` default and panicked with `cannot cast <nil> to uint8`, taking down the process.

This is a semantic error, not a syntax error: a node with a required input (one with no default value) and no incoming edge can never run. The graphical frontend (`graph.Analyze`) already rejected this, but the text frontend (`text.Analyze`) did not. Rather than duplicate the check, the validation is hoisted into the shared `analyzer.ResolveNodeTypes` that **both** frontends call, and the now-redundant copy is removed from `graph.Analyze`. Such programs now fail compilation with `node '<key>' (<type>) missing required input '<param>'` instead of producing IR that panics.

This also formalizes that a flow node receives a single value from its upstream: a multi-input function used as a bare flow node (e.g. `add{}(a, b)`) can never have all of its inputs sourced and is now reported as an error.

Test fixtures that relied on the old leniency (flow functions whose required data input was never sourced) were updated to be complete programs — feeding a source where the construct is valid (single-input pipes) and asserting rejection where it is not (multi-input flow nodes).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR hoists the missing-required-input check from `graph.Analyze` into the shared `analyzer.ResolveNodeTypes` so both the graph and text frontends reject programs where a required (default-less) node input has no incoming edge, preventing the runtime nil-cast panic described in SY-4269.

- **`analyzer.ResolveNodeTypes`** now builds a `connected` set from edge targets after type-checking and emits a diagnostic for every unconnected required input, collecting all errors before returning; the now-redundant Step 5B in `graph/analyze.go` is removed and Step 5A is simplified to a flat `set.New[ir.Handle]()`.
- **Test suite** is brought in line with the new stricter validation: single-input flow-node tests gain a source channel, the old "binary operator parameter names" integration test is replaced by a multi-input-bare-flow-node rejection test, and a direct regression test for the `trigger -> (1 > 0) select{}` panic is added.
- **C++ `call_func` helper** is simplified to compile and invoke function declarations directly via their WASM export, removing the now-invalid practice of scaffolding a virtual channel and an unconnected flow node.

Safe to merge — the change moves an existing validation check into shared infrastructure and adds validation that was previously absent from the text frontend, with good test coverage for all new and updated paths.

The core logic in ResolveNodeTypes is straightforward: build a set of connected edge targets, then flag any node input that has no default value and no incoming edge. All new error paths are exercised by unit tests and integration tests. The graph frontend removes its now-redundant duplicate check cleanly. No existing runtime behavior is silently altered — programs that previously compiled and panicked now fail at compile time with an actionable message.

| Filename | Overview |
|----------|----------|
| arc/go/analyzer/resolve.go | Adds missing-required-input validation to the shared ResolveNodeTypes pass; iterates edges a second time to build the connected set after type-checking, then reports every unconnected required (default-less) input as a diagnostic instead of returning early. |
| arc/go/graph/analyze.go | Removes the now-redundant Step 5B (missing-required-input check) and simplifies Step 5A to use a flat set.New[ir.Handle]() for duplicate-target detection, replacing the old nested map[string]set.Set[string]. |
| arc/go/analyzer/resolve_test.go | Adds three new unit tests covering the new validation (required-input error, optional-input pass-through, edge-satisfied pass-through); fixes the skip-missing-target-param test by giving the unconnected value param a default so the new check doesn't flag it. |
| arc/go/text/text_test.go | Updates integration tests to supply source channels for single-input flow nodes, converts the old binary-operator-parameter-names test into a rejection test for multi-input bare flow nodes, and adds the regression test for the original runtime-panic bug. |
| arc/cpp/runtime/wasm/node_test.cpp | Simplifies call_func to compile and call function declarations directly via WASM export rather than scaffolding a virtual channel and flow node — aligns with the new rule that multi-input functions cannot be expressed as bare flow nodes. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Parse source text / graph] --> B[Analyze: build IR nodes & edges]
    B --> C[Step 5: ResolveNodeTypes]
    C --> D{Type-check each edge}
    D -- mismatch --> E[Return false / emit error]
    D -- ok --> F[Unify constraints]
    F -- error --> E
    F -- ok --> G[Apply substitutions to all params]
    G --> H[Build connected set from edge targets]
    H --> I{For each node input: Value == nil AND not in connected?}
    I -- yes --> J[Emit missing required input diagnostic]
    I -- no --> K[Continue]
    J --> K
    K --> L{All nodes checked?}
    L -- no --> I
    L -- yes --> M{missingRequiredInput?}
    M -- yes --> E
    M -- no --> N[Step 5A: Duplicate edge target check]
    N --> O[Stratify & return IR]
```

1. `arc/go/graph/analyze.go`, line 196-213 ([link](https://github.com/synnaxlabs/synnax/blob/ad731ab6bfe7769057c3684f3222075e97a4f456/arc/go/graph/analyze.go#L196-L213)) 

   <a href="#">P2</a> **Stale comment and dead allocation**

   After Step 5B was removed, `connectedInputs` is only read inside the duplicate-detection loop (via `.Contains`/`.Add`) and is never consumed after Step 5A completes. The label "and Build Connected Inputs Map" in the comment is now misleading, because the map's purpose was to feed the missing-required-input check that no longer lives here. The data is allocated and populated but goes unused by any subsequent step.

</details>

<sub>Reviews (4): Last reviewed commit: ["fixed C++ tests"](https://github.com/synnaxlabs/synnax/commit/04c6a72c4c624a0a28efdbd411b955724e498dcd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34659692)</sub>

### DIFF
--- a/arc/cpp/runtime/wasm/node_test.cpp
+++ b/arc/cpp/runtime/wasm/node_test.cpp
@@ -91,28 +91,16 @@ std::vector<std::shared_ptr<stl::Module>> build_stl_modules(
     };
 }
 
-/// @brief compiles a single Arc function definition, calls it, and returns the first
-/// output value. Parses the function name and return type from the definition to create
-/// the necessary flow and channel automatically.
+/// @brief compiles a single Arc function definition, calls its exported WASM
+/// function directly with the given params, and returns the first output value.
+/// A function declaration compiles and exports to WASM on its own, so the
+/// definition is compiled as-is with no flow or channel scaffolding.
 template<typename T>
 T call_func(
     const synnax::Synnax &client,
     const std::string &func_def,
     const std::vector<x::telem::SampleValue> &params = {}
 ) {
-    static const std::unordered_map<std::string, x::telem::DataType> arc_types = {
-        {"i8", x::telem::INT8_T},
-        {"i16", x::telem::INT16_T},
-        {"i32", x::telem::INT32_T},
-        {"i64", x::telem::INT64_T},
-        {"u8", x::telem::UINT8_T},
-        {"u16", x::telem::UINT16_T},
-        {"u32", x::telem::UINT32_T},
-        {"u64", x::telem::UINT64_T},
-        {"f32", x::telem::FLOAT32_T},
-        {"f64", x::telem::FLOAT64_T},
-    };
-
     // Parse function name: first word after "func "
     auto func_pos = func_def.find("func ");
     if (func_pos == std::string::npos)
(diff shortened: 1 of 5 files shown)


sy-nico (approved):
LGTM. I independently arrived at the same fix on `resolve.go`. Only thing to note, this fixes a specific panic, but we still don't have a way to prevent a crash on generic issues.

emilbon99:
See https://github.com/synnaxlabs/synnax/pull/2405

PR #2409, opened 2026-06-01
SY-4171: Re-fire Constants on Upstream Trigger (Matching Expressions)

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4171](https://linear.app/synnax/issue/SY-4171/fix-arc-node-update-bug)

## Description

Constants and bare expressions in Arc flow chains behaved inconsistently depending on their position and on what else the program contained. Chained constants behaved as one-shots regardless of upstream activity: `time.interval{1s} -> "true" -> log` only wrote `"true"` to `log` on the first interval tick and silently stopped, while the equivalent expression form `time.interval{1s} -> "true"+"!" -> log` fired on every tick. Separately, standalone entry nodes never fired at startup, so a constant-only or input-driven program such as `"c" -> log` produced nothing until the first channel input arrived, which could be seconds away or never. All of these should be indistinguishable from the program author's perspective.

The first bug came from the constant runtime (`arc/go/stl/constant/constant.go`) applying an unconditional `initialized`-flag guard in `Next()`: once a constant emitted, it short-circuited every subsequent scheduler cycle and never propagated `MarkChanged` downstream. Entry-node detection was also duplicated and inconsistent. The WASM runtime used a string-prefix hack that treated every `expression_` node as non-entry, so bare expressions never became one-shots, while the constant runtime had no edge awareness at all. The startup bug came from the runtime granting the t=0 startup tick only to programs that contained an interval.

This change centralizes the rule in `ir.Node.IsEntryNode(edges)`: a node is an entry node when it has no incoming edges and reads no channels. Both `constant.go` and `arc/go/stl/wasm/wasm.go` now consult it, replacing the prefix hack, and the one-shot guard applies only when it returns true. Separately, every program now runs on the ticker runtime (`core/pkg/service/arc/runtime/task.go`), whose t=0 startup tick fires entry nodes once regardless of program shape; with no active timers the ticker idles on input rather than busy-looping. The now-unused `dataRuntime.Flow` was removed. No analyzer or grammar changes are required; the IR gains a single helper method.

The entry-node case is deliberately preserved: edge-fed constants re-fire on every upstream trigger, while standalone constants at stratum-0 and constants inside sequence stages (e.g., `stage active { 1 -> vlv_cmd }`) continue to fire exactly once per activation, with `Reset()` clearing the flag on stage re-entry.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes a bug where chained constant nodes in Arc flow programs behaved as one-shots regardless of upstream activity. The constant runtime now computes `isEntryNode` at construction by checking whether the node has any incoming edges, and only applies the one-shot guard when acting as a true entry node.

- `constant.go`: `isEntryNode` is derived from `cfg.Program.Edges.GetInputs(cfg.Node.Key)` at construction time, mirroring the existing WASM node pattern; `Next()` skips emit-and-return only when `isEntryNode == true`, so non-entry constants propagate `MarkChanged` on every scheduler cycle.
- `constant_test.go` / `routing_table_test.go`: existing entry-node tests are renamed for clarity; two new unit tests cover the non-entry firing behavior, and an integration test verifies the full `trig -> 1 -> log` chain with strictly increasing timestamps across four upstream ticks.

Safe to merge — the change is small and well-contained, correctly mirrors the existing WASM node pattern, and backward compatibility for entry-node (one-shot) constants is preserved.

The diff is minimal: a single boolean flag set at construction, a guarded early-return restructured inside an if, and Reset() left untouched. Entry-node behavior for stratum-0 constants and sequence-stage constants is preserved exactly as before. The new code path for non-entry nodes is straightforward — no guard, emit every cycle — and is directly validated by both unit and integration tests. No edge-case gaps are apparent.

| Filename | Overview |
|----------|----------|
| arc/go/stl/constant/constant.go | Adds `isEntryNode` flag computed at construction from incoming-edge count; the one-shot guard now only applies when true, allowing non-entry constants to re-fire every cycle. |
| arc/go/stl/constant/constant_test.go | Existing one-shot tests renamed to clarify "entry node" context; two new tests verify that non-entry constants emit on every `Next()` call and don't require `Reset()`. |
| arc/go/routing_table_test.go | Integration test validates the end-to-end `trig -> 1 -> log` chain fires on every upstream tick and produces strictly monotonically increasing timestamps. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["constant.Create()"] --> B{{"len(GetInputs(key)) == 0?"}}
    B -- "yes (no incoming edges)" --> C["isEntryNode = true"]
    B -- "no (has upstream)" --> D["isEntryNode = false"]

    C --> E["Next() called"]
    D --> F["Next() called"]

    E --> G{{"initialized?"}}
    G -- "true" --> H["return (no-op)"]
    G -- "false" --> I["initialized = true\nwrite output\nMarkChanged"]

    F --> J["write output\nMarkChanged (every cycle)"]

    K["Reset()"] --> L["initialized = false\n(no-op for non-entry nodes)"]
```

<sub>Reviews (1): Last reviewed commit: ["SY-4171: Add test"](https://github.com/synnaxlabs/synnax/commit/6d377303024796867c9fb728413e5b92a5a743b6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34934365)</sub>

### DIFF
--- a/arc/go/stl/constant/constant.go
+++ b/arc/go/stl/constant/constant.go
@@ -57,7 +57,7 @@ func (h *Host) Create(_ context.Context, cfg node.Config) (node.Node, error) {
 	return &constant{
 		State:       cfg.State,
 		value:       cfg.Node.Config[0].Value,
-		isEntryNode: len(cfg.Program.Edges.GetInputs(cfg.Node.Key)) == 0,
+		isEntryNode: cfg.Node.IsEntryNode(cfg.Program.Edges),
 	}, nil
 }
 
--- a/arc/go/stl/constant/constant.go
+++ b/arc/go/stl/constant/constant.go
@@ -54,23 +54,30 @@ func (h *Host) Create(_ context.Context, cfg node.Config) (node.Node, error) {
 	if cfg.Node.Type != symbolName {
 		return nil, query.ErrNotFound
 	}
-	return &constant{State: cfg.State, value: cfg.Node.Config[0].Value}, nil
+	return &constant{
+		State:       cfg.State,
+		value:       cfg.Node.Config[0].Value,
+		isEntryNode: len(cfg.Program.Edges.GetInputs(cfg.Node.Key)) == 0,
+	}, nil
 }
 
 type constant struct {
 	*node.State
 	clock       telem.MonoClock
 	value       any
+	isEntryNode bool
 	initialized bool
 }
 
 var _ node.Node = (*constant)(nil)
 
 func (c *constant) Next(ctx node.Context) {
-	if c.initialized {
-		return
+	if c.isEntryNode {
+		if c.initialized {
+			return
+		}
+		c.initialized = true
 	}
-	c.initialized = true
 	d := c.Output(0)
 	*d = telem.NewSeriesFromAny(c.value, d.DataType)
 	t := c.OutputTime(0)
--- a/arc/go/stl/constant/constant_test.go
+++ b/arc/go/stl/constant/constant_test.go
@@ -287,7 +287,7 @@ var _ = Describe("Constant", func() {
 			Expect(vals[0]).To(Equal(int64(-42)))
 		})
 
-		It("Should only emit once across multiple Next calls when entry node", func(ctx SpecContext) {
+		It("Should only emit once across multiple Next calls when node has no incoming edges", func(ctx SpecContext) {
 			cfg := node.Config{
 				Node:  ir.Node{Type: "constant", Config: types.Params{{Name: "value", Type: types.I64(), Value: int64(42)}}},
 				State: s.Node("const"),
@@ -305,7 +305,7 @@ var _ = Describe("Constant", func() {
 			Expect(marked).To(HaveLen(1))
 		})
 
-		It("Should emit again after Reset is called when entry node", func(ctx SpecContext) {
+		It("Should emit again after Reset is called when node has no incoming edges", func(ctx SpecContext) {
 			cfg := node.Config{
 				Node:  ir.Node{Type: "constant", Config: types.Params{{Name: "value", Type: types.I64(), Value: int64(42)}}},
 				State: s.Node("const"),
--- a/arc/go/stl/constant/constant_test.go
+++ b/arc/go/stl/constant/constant_test.go
@@ -137,7 +137,7 @@ var _ = Describe("Constant", func() {
 				Node:  ir.Node{Type: "constant", Config: types.Params{{Name: "value", Type: types.I64(), Value: 42}}},
 				State: s.Node("const"),
 			}
-			n, _ := factory.Create(ctx, cfg)
+			n := MustSucceed(factory.Create(ctx, cfg))
 			n.Next(node.Context{Context: ctx, MarkChanged: func(i int) {
 				marked = append(marked, i)
 			}})
@@ -149,7 +149,7 @@ var _ = Describe("Constant", func() {
 				Node:  ir.Node{Type: "constant", Config: types.Params{{Name: "value", Type: types.I64(), Value: int64(100)}}},
 				State: s.Node("const"),
 			}
-			n, _ := factory.Create(ctx, cfg)
+			n := MustSucceed(factory.Create(ctx, cfg))
 			n.Next(node.Context{Context: ctx, MarkChanged: func(int) {}})
 			out := s.Node("const").Output(0)
 			Expect(out.Len()).To(Equal(int64(1)))
@@ -163,7 +163,7 @@ var _ = Describe("Constant", func() {
 				},
 				State: s.Node("const"),
 			}
-			n, _ := factory.Create(ctx, cfg)
+			n := MustSucceed(factory.Create(ctx, cfg))
 			n.Next(node.Context{Context: ctx, MarkChanged: func(int) {}})
 			outTime := s.Node("const").OutputTime(0)
 			Expect(outTime.Len()).To(Equal(int64(1)))
@@ -178,7 +178,7 @@ var _ = Describe("Constant", func() {
 			}
 			constNode := s.Node("const")
 			*constNode.Output(0) = telem.NewSeriesV[float64](0)
-			n, _ := factory.Create(ctx, cfg)
+			n := MustSucceed(factory.Create(ctx, cfg))
 			n.Next(node.Context{Context: ctx, MarkChanged: func(int) {}})
 			out := constNode.Output(0)
 			vals := telem.UnmarshalSeries[float64](*out)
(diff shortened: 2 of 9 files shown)


arc/go/stl/constant/constant_test.go
```diff
@@ -286,7 +287,7 @@ var _ = Describe("Constant", func() {
 			Expect(vals[0]).To(Equal(int64(-42)))
 		})
 
-		It("Should only emit once across multiple Next calls", func(ctx SpecContext) {
+		It("Should only emit once across multiple Next calls when entry node", func(ctx SpecContext) {
```
pjdotson: when entry node??

arc/go/stl/constant/constant_test.go
```diff
@@ -304,7 +305,7 @@ var _ = Describe("Constant", func() {
 			Expect(marked).To(HaveLen(1))
 		})
 
-		It("Should emit again after Reset is called", func(ctx SpecContext) {
+		It("Should emit again after Reset is called when entry node", func(ctx SpecContext) {
```
pjdotson: when entry node??

arc/go/stl/constant/constant_test.go
```diff
@@ -322,6 +323,55 @@ var _ = Describe("Constant", func() {
...
+				},
+				State: s.Node("const"),
+				Program: program.Program{IR: ir.IR{Edges: ir.Edges{{
+					Source: ir.Handle{Node: "upstream", Param: ir.DefaultOutputParam},
+					Target: ir.Handle{Node: "const", Param: ir.DefaultInputParam},
+				}}}},
+			}
+			constNode := s.Node("const")
+			*constNode.Output(0) = telem.NewSeriesV[int64](0)
+			n, _ := factory.Create(ctx, cfg)
```
pjdotson: don't ignore errors

arc/go/stl/constant/constant_test.go
```diff
@@ -322,6 +323,55 @@ var _ = Describe("Constant", func() {
...
```
pjdotson: don't ignore errors

pjdotson (changes requested):
1) is there any other node types that has similar behavior we should be looking to change
2) Does documentation need to be updated

arc/go/stl/constant/constant.go
```diff
@@ -54,23 +54,30 @@ func (h *Host) Create(_ context.Context, cfg node.Config) (node.Node, error) {
 	if cfg.Node.Type != symbolName {
 		return nil, query.ErrNotFound
 	}
-	return &constant{State: cfg.State, value: cfg.Node.Config[0].Value}, nil
+	return &constant{
+		State:       cfg.State,
+		value:       cfg.Node.Config[0].Value,
+		isEntryNode: len(cfg.Program.Edges.GetInputs(cfg.Node.Key)) == 0,
```
pjdotson: maybe this should be a method on `ir.Node` instead of a bool you pass around

arc/go/stl/constant/constant_test.go
sy-nico: Follows existing pattern in tests and you said to ignore this comment.

arc/go/stl/constant/constant_test.go
sy-nico: Follows existing pattern in tests and you said to ignore this comment.

arc/go/stl/constant/constant_test.go
sy-nico: "when entry node" ~= "when _it_ is an entry node".
I've updated to be more clear.

arc/go/stl/constant/constant_test.go
sy-nico: Updated

arc/go/stl/constant/constant.go
sy-nico: Moved to `ir.Node` and changed logic to return `0 inputs && no channels.read`. It is now decoupled from `"expression_"`.

`str(sy_node_1_cpu_percent)` returns False because it has a channel as an input.

sy-nico:
Replying to [this comment](https://github.com/synnaxlabs/synnax/pull/2409#pullrequestreview-4411465542)

1. Not that I have come across. One behavior that I am questioning is what happens for a simple program like
```
"some_constant" -> log
```
We discussed this in person and resolved by firing once at startup.

2. No documentation updates required. We do not have existing documentation about behavioral differences for expressions vs constants. Additionally, it would seem weird/redundant to add a section just to mention that constants and expressions will fire on an upstream trigger. If you think I need it, I can expand the existing [`Expressions in Flows` section.](https://docs.synnaxlabs.com/reference/control/arc/reference/flow#expressions-in-flows). Though it would likely change to be more generic.

arc/go/stl/constant/constant_test.go
pjdotson: Thought we were talking about the other comments on the test files. You should definitely be using `MustSucceed` here. If the existing pattern is to ignore these errors then the other tests are broken as well.

core/pkg/service/arc/runtime/task.go
```diff
@@ -227,10 +226,10 @@ func (t *taskImpl) start(ctx context.Context) (err error) {
 
 	pipeline := plumber.New()
 
-	var runtime confluence.Segment[framer.StreamerResponse, framer.WriterRequest] = &drt
-	if hasIntervals := timeMod.BaseInterval != telem.TimeSpan(math.MaxInt64); hasIntervals {
-		runtime = &tickerRuntime{dataRuntime: drt}
-	}
+	// The ticker's t=0 startup tick fires entry nodes; an input-driven program
+	// would otherwise not fire them until the first input is recieved.
```
pjdotson: typo. Would also probably turn on CSpell VSCode extension

core/pkg/service/arc/runtime/task.go
```diff
@@ -227,10 +226,10 @@ func (t *taskImpl) start(ctx context.Context) (err error) {
...
+	ticker := &tickerRuntime{dataRuntime: drt}
+	var runtime confluence.Segment[framer.StreamerResponse, framer.WriterRequest] = ticker
```
pjdotson: not any reason to declare this variable `runtime`, could just pass `ticker` below

pjdotson (changes requested):

arc/go/stl/constant/constant_test.go
sy-nico: Understood. Fixed the added tests and the pre-existing issues.

PR #2414, opened 2026-06-02
SY-4291: Implement Extends Keyword for Oracle Enums

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4291](https://linear.app/synnax/issue/SY-4291)

## Description

Adds an `extends` clause to Oracle enums so an enum can be composed as the union of one or more other enums. Each member's value is declared once in its home enum; the wide enum is built from the narrow ones:

```
XAxisKey enum { x1 = "x1"  x2 = "x2" }
YAxisKey enum { y1 = "y1"  y2 = "y2"  y3 = "y3"  y4 = "y4" }
AxisKey  enum extends XAxisKey, YAxisKey {}
```

This lets a payload field that may only carry a y-axis key be typed `YAxisKey` for compile-time safety, while a single wide `AxisKey` remains available for broad uses. It is the foundation for narrowing the line plot action payloads (removing the `requireYAxis`/`isYAxis` runtime guards) and for deriving the spatial `Location` lattice instead of hand-duplicating it.

- **Grammar:** `enumDef` gains an optional `extends typeRefList` clause (reusing the list used by `struct extends`); the ANTLR parser is regenerated. Member values are still required.
- **Analyzer:** an extending enum's members expand to `union(parents) + own`, with diagnostics for unknown or non-enum parents, mixing integer and string kinds, parents contributing conflicting member values, and cyclic `extends` chains.
- **IR / formatter / ordering:** `EnumForm.Extends` + `IsExtension()`; parents are ordered before children for codegen; the formatter renders the `extends` clause.
- **No language-plugin changes.** Because the analyzer fully populates an extending enum's member list, all five targets (Go, TypeScript, Python, C++, protobuf) emit it as an ordinary standalone enum. Protobuf uses 0-based positional numbering, so the generated enum satisfies proto3's zero-value requirement.

Applying this to `schemas/lineplot.oracle` (narrowing the channel/range action payloads and deleting the runtime axis guards) lands in a follow-up once this merges and `oracle sync` is run.

## Basic Readiness


oracle/analyzer/analyzer.go
```diff
@@ -186,12 +186,123 @@ func analyze(c *analysisCtx) {
...
+		parent, ok := c.table.Get(form.Extends[i].Name)
+		if !ok {
+			addEnumDiag(c, typ, "enum %s extends unknown enum %q", typ.QualifiedName, form.Extends[i].Name)
+			return nil, false, false
+		}
+		if _, ok := parent.Form.(resolution.EnumForm); !ok {
+			addEnumDiag(c, typ, "enum %s extends %s, which is not an enum", typ.QualifiedName, parent.QualifiedName)
+			return nil, false, false
+		}
+		pVals, pInt, ok := effectiveEnumValues(c, parent, visited)
```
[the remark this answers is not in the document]
emilbon99: Good catch — real bug. Fixed in c574982: each parent branch now recurses with `visited.Copy()` so the set tracks the ancestor path rather than every node seen, which keeps diamonds valid. Added a diamond regression test that fails without the fix.

oracle/analyzer/analyzer_test.go
```diff
@@ -152,6 +152,113 @@ var _ = Describe("Analyzer", func() {
...
+				C enum extends A, B {}
+			`
+			_, diag := analyzer.AnalyzeSource(ctx, source, "x", loader)
+			Expect(diag.Ok()).To(BeFalse())
+			Expect(diag.Error()).To(ContainSubstring("conflicting values"))
+		})
+
 		It("Should collect field domains", func(ctx SpecContext) {
 			source := `
 				User struct {
```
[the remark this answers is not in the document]
emilbon99: Added both: direct cycle tests (self-extension and A↔B) and a diamond test (`Diamond extends Left, Right` where both extend `Base`) asserting it's accepted with `Base` de-duplicated. The diamond test fails on the pre-fix shared-set code.

oracle/plugin/ts/types/types_test.go
```diff
@@ -252,6 +252,32 @@ var _ = Describe("TS Types Plugin", func() {
...
+					y2 = "y2"
+				}
+
+				AxisKey enum extends XAxisKey, YAxisKey {}
+			`
+			table, diag := analyzer.AnalyzeSource(ctx, source, "lineplot", loader)
+			Expect(diag.Ok()).To(BeTrue())
+
+			resp, err := typesPlugin.Generate(&plugin.Request{Resolutions: table})
+			Expect(err).To(BeNil())
```
pjdotson: use MustSucceed

oracle/plugin/py/types/types_test.go
```diff
@@ -250,6 +250,31 @@ var _ = Describe("Python Types Plugin", func() {
...
+					y2 = "y2"
+				}
+
+				AxisKey enum extends XAxisKey, YAxisKey {}
+			`
+			table, diag := analyzer.AnalyzeSource(ctx, source, "lineplot", loader)
+			Expect(diag.Ok()).To(BeTrue())
+
+			resp, err := typesPlugin.Generate(&plugin.Request{Resolutions: table})
+			Expect(err).To(BeNil())
```
pjdotson: use `MustSucceed`

oracle/plugin/cpp/types/types_test.go
```diff
@@ -963,6 +963,36 @@ var _ = Describe("C++ Types Plugin", func() {
...
+
+				Plot struct {
+					axis_key AxisKey
+				}
+			`
+			table, diag := analyzer.AnalyzeSource(ctx, source, "lineplot", loader)
+			Expect(diag.Ok()).To(BeTrue())
+
+			resp, err := cppPlugin.Generate(&plugin.Request{Resolutions: table})
+			Expect(err).To(BeNil())
```
pjdotson: use `MustSucceed`

pjdotson (changes requested):
approved after comments addressed

PR #2415, opened 2026-06-02
SY-4292: Identify Aether Store Components by Path, Not Key

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4292](https://linear.app/synnax/issue/SY-4292)

## Description

Editing the channels on one line plot could mangle or evict state on a *different* line plot whenever both plots charted the same channel. The two plots rendered fine with different channels; the corruption only appeared when a channel was shared.

**Root cause.** Every aether component has two names: its `path` (its position in the worker tree, `[...parentPath, key]`, unique by construction) and its `key` (the last path segment, unique only among siblings). The worker tree is addressed by `path`, but the main-thread `Store` indexed every component by the bare `key` alone (`entries`/`listeners`/`snapshots` all keyed by leaf key). Line-plot lines deliberately use a *deterministic, channel-derived* key (`${yAxis}---${xAxis}---${range}---${x}---${y}`) so that line properties survive channel reconfiguration and workspace reloads — but that key is only unique *within* a plot. Two plots charting the same channel produced lines with identical keys, which collapsed onto a single store entry: registering the second plot's line deleted the first's worker component and re-pointed the shared entry, so the two plots thrashed each other's state.

**Fix.** Identify components in the main-thread `Store` by their full `path` instead of the leaf `key`, mirroring the worker tree that was already path-addressed. The store now flattens the path to a collision-proof string identity; keys only need to be unique among siblings (React's own `key` contract), which the existing channel-derived line keys already satisfy. This removes an implicit, undocumented "keys must be globally unique" invariant rather than adding one, and requires **no** change to line keying or the persisted slice format.

Since `key` is always `path[path.length - 1]`, it is no longer passed alongside `path` through the constructor / `RegisterParams` / `Handle`; it is derived where needed. The worker→main `update` message now carries `path` instead of `key`.

### Changed
- `aether/store.ts`: entries/listeners/snapshots keyed by `pathID(path)`; `subscribe`/`getSnapshot`/`unregister` take a path; `RegisterParams`/`Handle` drop the redundant `key`.
- `aether/aether/aether.ts`: components take and store `path`, derive `key` from it, and push state back by `path`.
- `aether/aether/message.ts`: `WorkerUpdateRequest` carries `path`.
- `aether/main.tsx`: lifecycle calls pass `path`.

### Tests
- New regression pin in `main.spec.tsx`: two parents each holding a leaf with the same key stay independent (fails under the old key-based store). Full aether suite green (101 passing).

> Note: hotfix targeting `main` (branched off main for a production-critical bug, matching recent console hotfixes). A back-merge into `rc` will be needed.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes a production bug where editing channels on one line plot would corrupt state on another plot that charted the same channel. The root cause was that the main-thread `Store` indexed components by bare leaf `key` instead of full `path`, so two plots' lines with the same channel-derived key collapsed to one store entry, causing each registration to evict the other.

- **`store.ts`**: Entries, listeners, and snapshots are now keyed by `pathID(path)` (a dot-joined path string). `register`, `unregister`, `subscribe`, and `getSnapshot` all take the full `path`; the `key` field is removed from `RegisterParams` and `Handle`.
- **`aether/aether.ts`** and **`message.ts`**: Worker-side components now store and transmit `path` instead of `key`; `key` is derived lazily as `path[path.length - 1]`.
- **`main.spec.tsx`**: A new regression test verifies that two parents each holding a leaf with the same key remain independent after the fix.

Safe to merge. The change is a self-contained identity-layer replacement with a focused scope, a new regression test that directly pins the reported bug, and no changes to persisted state formats or cross-thread wire protocols beyond swapping one field name.

The fix correctly mirrors the worker tree's existing path-addressed routing on the main-thread store. The PATH_SEP invariant is explicitly enforced at registration time (both empty-leaf and dot-in-segment guards are present). The new regression test fails under the old scheme and passes under the new one. All 101 existing aether tests continue to pass. No production data formats or API surfaces are affected.

No files require special attention. The main logic change in store.ts is straightforward and well-commented.

| Filename | Overview |
|----------|----------|
| pluto/src/aether/store.ts | Core fix: entries/listeners/snapshots now keyed by pathID(path) instead of bare key. Validation correctly catches empty leaf keys and dot-containing segments. PATH_SEP invariant is enforced at register time. |
| pluto/src/aether/aether/aether.ts | Worker-side components now store path and derive key as last element. setState pushes back with full path. Constructor takes path instead of key. Clean implementation with no regressions. |
| pluto/src/aether/main.tsx | useLifecycle's unregister/subscribe/getSnapshot calls all updated to pass path arrays. useLayoutEffect dependency updated from key to path. Behavior is correct for stable aetherKey usage. |
| pluto/src/aether/main.spec.tsx | New regression test correctly pins the cross-parent key collision bug. Existing tests updated to path-based API. StrictMode snapshot caching test updated and preserved. |
| pluto/src/aether/aether/message.ts | WorkerUpdateRequest.key replaced by path (readonly string[]). All other message types unchanged. |

</details>

<sub>Reviews (2): Last reviewed commit: ["SY-4292: Make component path private, dr..."](https://github.com/synnaxlabs/synnax/commit/c93e0e94e657420955231cfb1626eef410977979) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35215321)</sub>


emilbon99:

console/src-tauri/tauri.conf.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "../node_modules/@tauri-apps/cli/config.schema.json",
   "productName": "Synnax",
-  "version": "0.56.2",
```
pjdotson, line 4: Need Core version update as well

pluto/src/aether/aether/aether.ts
```diff
@@ -65,6 +65,8 @@ export interface InvokeMethodParams extends Omit<MainInvokeRequest, "variant"> {
 export interface Component {
   type: string;
   key: string;
+  /** Path from the root; its cross-thread identity. Last element is `key`. */
+  path: readonly string[];
```
pjdotson: shouldn't really have both `key` and `path` then.

pjdotson (changes requested):

pjdotson (approved):

PR #2416, opened 2026-06-02
SY-3999: Add Struct Union Support to Oracle

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-3999](https://linear.app/synnax/issue/SY-3999)

## Description

Adds first-class **discriminated struct unions** to the Oracle schema language, the missing piece needed to strongly type hardware task configurations (NI, Modbus, LabJack, OPC UA, EtherCAT, HTTP), whose channel arrays are deeply nested, internally-tagged unions dispatched on a `type` field. Today these live as opaque `record` JSON with hand-maintained per-language schemas (~2,000 lines of Zod for NI alone), no shared source of truth, no migrations, and no protobuf. This PR makes `union` a declarable Oracle type and generates idiomatic, type-safe representations in every target.

Design and rationale: [RFC 0032](docs/tech/rfc/0032-260331-oracle-struct-unions.md).

### New language construct

```oracle
Scale union on type {
    linear LinearScale
    map    MapScale
    none   NoneScale
}

AIChannel union on type extends BaseAIChannel {
    ai_voltage AIVoltageFields
}
```

A variant value may be a reserved word (e.g. `map`) without quoting. Unions support `extends` for shared base fields and nest (a variant field may itself be a union).

### What is implemented

- **Language + resolution + analyzer + formatter.** `UnionForm`/`UnionVariant`, `UnifiedVariantFields` flattening, dependency and topological-sort wiring, and validation (at least one variant, unique values, struct-only extends/variants, discriminator ownership).
- **All five type-layer codegens**, each idiomatic for its language and all emitting the same internally-tagged wire:
  - **TypeScript:** `z.discriminatedUnion` + per-variant schemas + discriminator enum + `<UNION>_SCHEMAS` map.
  - **Python:** one Pydantic model per variant + `Annotated[Union[...], Field(discriminator=...)]`.
  - **C++:** `std::variant` alias + variant structs + `parse`/`to_json`, including array-of-union element dispatch.
  - **Go:** sealed variant interface + per-variant structs (embedding the shared base and the payload) + a concrete wrapper that owns two-pass internally-tagged `MarshalJSON`/`UnmarshalJSON`, so a union field round-trips like any struct field.
  - **Protobuf:** `.proto` wrapper message with shared base fields outside a `oneof` of variant messages.
- Union-typed fields resolve correctly in every target.
- Fixes a pre-existing Go generator bug that mangled acronym-named `extends` bases (`BaseAIChan` to `BaseAiChan`, uncompilable), repairing regular struct `extends` as well.

### Deferred (forward-looking; nothing exercises these until a schema adopts `union`)

- Zero-value constants, blocked on a separate Oracle field-defaults feature; emitting type-zeros would silently diverge from the semantic defaults consumers rely on.
- Go msgpack dispatch and protobuf `to_proto`/`from_proto` conversion. See RFC section 7.5.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds first-class discriminated struct unions to the Oracle schema language, generating idiomatic, internally-tagged representations across all five code-gen targets (Go, TypeScript, Python, C++, Protobuf). It also fixes a pre-existing Go generator bug that mangled acronym-named `extends` bases (`BaseAIChan` → `BaseAiChan`).

- **New language construct**: `union on <discriminator> [extends <bases>]` with full analyzer, formatter, resolver, and validation support including cycle detection, struct-only extends/variants, and discriminator ownership.
- **All five code-gen targets**: Go (sealed interface + two-pass JSON marshal), TypeScript (`z.discriminatedUnion` + `SCHEMAS` map), Python (Pydantic `Annotated[Union[...]]`), C++ (`std::variant` alias + free `parse`/`to_json`), Protobuf (`oneof` wrapper message).
- **Cross-cutting plumbing**: migration auto-copy skips union-ness mismatches, pb translator handles union fields/arrays/map-values, `casing.VariantTypeName` factors out leading acronyms for readable variant struct names.

Safe to merge; the new union construct is purely additive and no existing schema uses it yet, so generated output is unchanged until a schema adopts the keyword.

All five code-gen targets are exercised by new tests, the analyzer validates every structural constraint (cycles, struct-only extends/variants, discriminator ownership), and the only findings are cosmetic: a wrong example in a doc comment, a confusing secondary error on an already-diagnosed failure path, and a MarshalJSON gap that only affects user code that stores pointer variants (not generated code).

oracle/plugin/go/types/types.go (MarshalJSON pointer-variant gap) and oracle/analyzer/analyzer.go (duplicate error on union-extends-union failure) are worth a second look before union types see production schema use.

| Filename | Overview |
|----------|----------|
| oracle/analyzer/analyzer.go | Adds union collection, extension flattening (finalizeUnionExtensions / effectiveUnionVariants), and validateUnion; renames addEnumDiag → addTypeDiag. Validation logic is thorough, but a union-extends-union failure path leaves the form unmodified and triggers a secondary misleading non-struct error from validateUnion. |
| oracle/resolution/type.go | Introduces UnionForm, UnionVariant, and UnifiedVariantFields. Design and semantics are clean; discriminator ownership, first-wins deduplication, and struct-only validation are all correct. |
| oracle/plugin/go/types/types.go | Adds union codegen to the Go types template: sealed interface, per-variant structs with embeds, internally-tagged MarshalJSON/UnmarshalJSON, and discriminator constants. The MarshalJSON type switch only covers value variants; pointer variants fall to a runtime error rather than a compile-time one. Also fixes the BaseAIChan acronym-mangling bug in resolveExtendsType. |
| oracle/plugin/go/pb/pb.go | Adds union pb translators (ToPB/FromPB/slice helpers), map-value and array-element union conversions, and NeedsRecordArrayHelpers. Correctly defers unions-with-extends and emits an actionable error. |
| oracle/plugin/ts/types/union.go | TypeScript union codegen: z.discriminatedUnion with per-variant z.object schemas, TYPES array, enum schema, and SCHEMAS map. The template assumes ParentSchemas is non-empty (guaranteed by validateUnion's struct enforcement). |
| oracle/plugin/cpp/naming/naming.go | New shared naming package that delegates to casing.VariantTypeName. Logically correct; the doc comment example AIChannelAIVoltage is wrong — actual output is AIVoltageChannel. |
| oracle/plugin/internal/casing/casing.go | Adds PascalAcronym and VariantTypeName with acronym-factoring logic. The leadingAcronym implementation correctly handles all-caps, mixed, and single-char unions. Well-tested. |
| oracle/plugin/pb/types/types.go | Protobuf union codegen: wrapper message with base fields plus a oneof. Hardcoded oneof name variant is now validated against base struct fields. |
| oracle/plugin/framework/generator.go | Adds CollectUnions to the Generator, inserting union handling in both the struct-shared-path and the union-only path. Correctly removes collected unions from the unionCollector when they share an output path with structs. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Oracle schema
'union on disc'] --> B[collectUnion
analyzer.go]
    B --> C[resolveTypeRefs]
    C --> D{Extends type?}
    D -- union base --> E[finalizeUnionExtensions
flatten variants]
    D -- struct base --> F[validateUnion
check discriminator ownership]
    E --> F
    F --> G[resolution.UnionForm
in type table]
    G --> H[Go types plugin
sealed interface +
MarshalJSON/UnmarshalJSON]
    G --> I[TypeScript plugin
z.discriminatedUnion +
SCHEMAS map]
    G --> J[Python plugin
Pydantic Annotated Union]
    G --> K[C++ types plugin
std::variant alias]
    G --> L[Proto plugin
oneof wrapper message]
    H --> M[Go pb translator
ToPB / FromPB per variant]
    L --> M
```

<sub>Reviews (2): Last reviewed commit: ["SY-3999: Renumber Struct Union RFC to 00..."](https://github.com/synnaxlabs/synnax/commit/9aad0cc8bab4343bb3ce2468415cafeeafd50ac1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35241023)</sub>


oracle/plugin/ts/types/union.go
```diff
@@ -0,0 +1,108 @@
...
+		variantTSName := tsName + pascalCase(v.Name)
+		vd := unionVariantData{
+			Value:      v.Name,
+			TypeName:   variantTSName,
+			SchemaName: camelCase(variantTSName) + "Z",
+			Doc:        doc.Get(v.Domains),
+		}
+		for _, f := range resolution.UnifiedVariantFields(entry, v, table) {
+			vd.Fields = append(vd.Fields, p.processField(f, entry, table, data, false, false))
+		}
```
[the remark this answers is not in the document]
emilbon99: Addressed in c5a873c58f: `validateUnion` now rejects a variant value equal to the discriminator field name, so this produces a clear schema error instead of colliding type names in the generated TypeScript.

oracle/plugin/pb/types/types.go
```diff
@@ -539,11 +583,84 @@ func (p *Plugin) typeToProto(typeRef resolution.TypeRef, data *templateData) (st
...
+		})
+		fieldNumber++
+	}
+	md.Oneof = oneof
+	return md, nil
+}
+
 func (p *Plugin) resolveStructType(typeRef resolution.TypeRef, resolved resolution.Type, data *templateData) (string, error) {
 	form, ok := resolved.Form.(resolution.StructForm)
 	if !ok {
```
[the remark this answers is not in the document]
emilbon99: Addressed in c5a873c58f: `validateUnion` now rejects a union base struct that declares a field named `variant` (reserved for the generated proto oneof).

emilbon99:
@greptile re-run

docs/tech/rfc/0042-260331-oracle-struct-unions.md
```diff
@@ -0,0 +1,479 @@
...
+**Status:** Partially Implemented (language + all five type-layer codegens; see section
+7.5)
+
+**Related:** [RFC 0028 - Oracle Schema System](./0028-251229-oracle-schema-system.md)
+
+---
+
+# 0 - Summary
+
+This RFC proposes adding first-class support for **discriminated struct unions** to the
```
pjdotson, line 14: what is the point of adding an RFC with the implementation? If it's something you want
to get feedback on before implementing, add it in its own PR, otherwise just put this
stuff in the PR description

docs/tech/rfc/0042-260331-oracle-struct-unions.md
```diff
@@ -0,0 +1,479 @@
...
+A new top-level definition type for discriminated unions:
+
+```oracle
+Name union on discriminatorField {
+    variant_value_1 VariantType1
+    variant_value_2 VariantType2
+}
+```
+
+The `on` clause names the JSON field that serves as the discriminator. Each entry maps a
```
pjdotson, line 202: so struct unions _have_ to have a type discriminator? that is not hurting us in anything
we may want to represent with struct unions (schematic symbols, workspaces, etc.), that
doesn't yet have it, right?

docs/tech/rfc/0042-260331-oracle-struct-unions.md
```diff
@@ -0,0 +1,479 @@
...
+```go
+type ScaleType string
+
+// Sealed variant interface.
+type ScaleVariant interface{ isScaleVariant() }
+
+// One concrete struct per variant. Each embeds the variant payload struct (and,
+// for a union with extends, the shared base struct ahead of it). The
+// discriminator is NOT stored on the variant; the wrapper owns it.
+type ScaleLinear struct{ LinearScale }
```
pjdotson: where is `LinearScale` defined? what is `LinearScale` vs `ScaleLinear`?

oracle/plugin/cpp/naming/naming.go
```diff
@@ -0,0 +1,25 @@
...
+import "github.com/synnaxlabs/oracle/plugin/internal/casing"
+
+// VariantTypeName returns the C++ struct name for one variant of a discriminated
+// union: the union's C++ name followed by the PascalCased discriminator value
+// (e.g. union "Scale" + value "linear" -> "ScaleLinear", union "AIChannel" +
+// value "ai_voltage" -> "AIChannelAIVoltage"). cpp/types declares the variant
+// struct under this name and cpp/json emits its parse/to_json bodies, so both
+// MUST derive it identically; routing through this one function guarantees they
+// cannot drift.
+func VariantTypeName(unionCppName, variantValue string) string {
```
pjdotson, line 22: needs test

oracle/plugin/cpp/json/json_test.go
```diff
@@ -652,3 +652,123 @@ var _ = Describe("C++ JSON Plugin", func() {
...
+				none NoneScale
+			}
+		`
+		resp := MustGenerate(ctx, source, "ni", loader, jsonPlugin)
+		ExpectContent(resp, "json.gen.h").
+			ToContain(
+				`inline Scale parse_scale(x::json::Parser parser) {`,
+				`const auto discriminator = parser.field<std::string>("type");`,
+				`if (discriminator == "linear") return ScaleLinear::parse(parser);`,
+				`if (discriminator == "none") return ScaleNone::parse(parser);`,
```
pjdotson, line 685: what should it do if the discriminator doesn't match?

oracle/plugin/framework/generator.go
```diff
@@ -103,13 +113,18 @@ func (g *Generator) Generate(req *plugin.Request) (*plugin.Response, error) {
 		if typeDefCollector != nil && typeDefCollector.Has(outputPath) {
 			typeDefs = typeDefCollector.Remove(outputPath)
 		}
+		var unions []resolution.Type
```
pjdotson, line 116: needs test

oracle/resolution/type.go
```diff
@@ -74,6 +74,40 @@ func (EnumForm) typeForm() {}
...
+	// Variants lists every concrete shape the union admits, in declaration order.
+	Variants []UnionVariant
+	// Extends names struct types whose fields are shared across all variants.
+	// Resolved later by the analyzer; struct-form is validated there.
+	Extends []TypeRef
+}
+
+func (UnionForm) typeForm() {}
+
+func (f UnionForm) Variant(name string) (UnionVariant, bool) {
```
pjdotson, line 101: needs test

oracle/resolution/table.go
```diff
@@ -133,6 +133,13 @@ func (t *Table) DistinctTypes() []Type {
 	})
 }
 
+func (t *Table) UnionTypes() []Type {
```
pjdotson, line 150: needs test

oracle/plugin/resolver/resolver.go
```diff
@@ -104,11 +104,31 @@ func (r *Resolver) ResolveTypeRef(typeRef resolution.TypeRef, ctx *Context) stri
...
+	case resolution.UnionForm:
+		return r.resolveUnionType(resolved, ctx)
 	default:
 		return r.Formatter.FallbackType()
 	}
 }
 
+// resolveUnionType resolves a discriminated union type to a language-specific
+// string. A union is referenced by name, like an enum or distinct type.
+func (r *Resolver) resolveUnionType(resolved resolution.Type, ctx *Context) string {
```
pjdotson, line 116: add test

oracle/plugin/go/marshal/marshal_test.go
```diff
@@ -380,6 +380,30 @@ var _ = Describe("Go Marshal Plugin", func() {
 		})
 
 		Context("map field nil preservation", func() {
+			It("Should encode union map values as length-prefixed JSON", func(ctx SpecContext) {
```
pjdotson: is there a reason to encode as this versus some specific encode / decode for the
specific variant that is more efficient?

pjdotson (changes requested):

docs/tech/rfc/0042-260331-oracle-struct-unions.md
emilbon99: Because it's useful documentation for future readers (and agents), and helps create a design trail for potential customers who want to understand the care we put into our engineering process.

docs/tech/rfc/0042-260331-oracle-struct-unions.md
emilbon99: I mean the field doesn't necessarily have to be `type` but it does have to be an explicit field. This is for a few reasons:

1) It dramatically simplifiers the oracle analyzer
2) Implementing this in any language other than typescript would be a PITA, especially go. How do you decide which one in the union is correct? You'd need to implement zod like parsing logic in go. 
3) In go, you can get better performance by having an explicit discriminator because you don't have to do a bunch of logic to figure out which union is valid.

oracle/plugin/cpp/naming/naming.go
emilbon99: Added in 6aab375724 (`naming_test.go`): covers the plain-prefix, acronym-factored, non-repeating-acronym, and reserved-word cases. Also fixed the doc comment here, which claimed `AIChannel` + `ai_voltage` -> `AIChannelAIVoltage` when the function factors the acronym and returns `AIVoltageChannel`.

oracle/plugin/framework/generator.go
emilbon99: Added in 6aab375724 (`generator_test.go`): unions sharing the structs' output path land in `GenerateContext.Unions`, `CollectUnions: false` collects nothing, and a union at a struct-less output path generates its own union-only file.

oracle/resolution/type.go
emilbon99: Added in 6aab375724 (`union_test.go`): `Variant` returns the matching variant and reports a missing discriminator value.

oracle/resolution/table.go
emilbon99: Added in 6aab375724 (`union_test.go`): `UnionTypes` filters union forms out of a table holding struct, enum, and union types.

oracle/plugin/resolver/resolver.go
emilbon99: Added in 6aab375724 (`resolver_test.go`, Union Types describe): same-output resolution to the bare name, cross-output qualification with an import, and fallback when the union has no output path.

docs/tech/rfc/0042-260331-oracle-struct-unions.md
emilbon99: **TLDR: It's not the best solution, but it produces the type safety and ergonomics we need. Downside: API footprint is increased for generated code (for now)**

This is a bit of a nasty thing that the syntax has right now. So you have:

```
struct LInearScaleFields {
       cat string
}

struct MapScaleFields {
       dog string
}

type Scale union on variant {
       linear LinearScaleFields
       map   MapScaleFields
}
```

Will generate the following (partial) go code:

```
type Scale struct {}

type ScaleLinear struct{}

type ScaleMap struct {}

type MapScaleFields struct{}

type LinearScaleFields struct {}
```

It's not totally ideal, and I don't really like it TBH, but it's probably the best thing I can think of right now that doesn't (a) require weird oracle syntax or new domain rules and (b) doesn't take a ton of time refactor. I.e. it's not perfect and extends the public API surface area, but I don't think it's worth investing energy to resolve right now. 

Regardless, refactored the RFC to be clearer.

oracle/plugin/cpp/json/json_test.go
emilbon99: It records a parse error on the accumulating parser (`parser.field_err("type", "unknown Scale type: ...")`) and returns a default-constructed variant, matching how all generated parses report bad fields. Callers check the parser's error state before using the result. The fallback branch was generated but untested, so a09d2bb197 adds assertions for it here.

oracle/plugin/go/marshal/marshal_test.go
emilbon99: You're right, there wasn't a good one. The JSON blob was hedging against variant evolution, but frozen codecs already solve that for structs and unions are no different. Reworked in dcc6c4f129: unions now encode fully binary as a length-prefixed discriminator string followed by the active variant's base and payload structs encoded positionally through their own codecs. The string tag (rather than a variant index) keeps stored bytes stable under variant addition and reordering; field changes inside a variant version through frozen codecs like any struct change. Nil variants fail encode and unknown tags fail decode, matching the JSON codec. Covered by substring tests plus a runtime round-trip fixture mirroring the generated wrapper.

pjdotson (approved):

PR #2418, opened 2026-06-03
SY-4276: Gate Visualization Toolbars Against Missing Slice State

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4276](https://linear.app/synnax/issue/SY-4276)

## Description

Switching away from a workspace and back could crash the schematic toolbar with `TypeError: undefined is not an object (evaluating 'select(state, key).toolbar')`, dropping the visualization into the error boundary until refresh. The schematic slice entry is briefly absent (purged on workspace change, recreated on return) while the toolbar — which renders in the nav drawer, independently of the main renderer's gate — mounts and dereferences it.

The earlier commits patched this by making the schematic selectors defensive (`selectOptional`, `?? ZERO_STATE`). This reworks that into the explicit pattern already used by the main renderers and the lineplot/log toolbars: **selectors declare a non-null contract, and each toolbar gates entry existence at its wrapper before rendering its inner component.** A component gate is a single, removable line, whereas scattered selector-level null-safety becomes dead defensive code once the entry is guaranteed at the reducer level.

Changes:

- Add `selectExists` / `useSelectExists` to the schematic, table, log, and lineplot slices as the canonical existence gate, and point the existing renderer `useExists` configs at it.
- Revert the schematic and table field selectors to non-null contracts, and gate the schematic and table toolbars on `useSelectExists`.
- Split the log and lineplot toolbars into a thin `useSelectExists` wrapper (`Toolbar`) plus a non-null inner component (`Internal`), uniform with schematic and table.
- Replace whole-state drills (`useSelect(key).toolbar.activeTab`, `useSelect(key).axes.axes[axisKey]`) with granular selectors (`useSelectActiveToolbarTab`, `useSelectAxis`) so toolbars subscribe only to the fields they read.
- Audited every consumer of a non-null per-key selector across the console (schematic, table, log, lineplot, arc); all are now behind an existence gate (arc's graph toolbar is gated by its `arc/editor/Toolbar.tsx` mode dispatcher).
- Add selector unit tests across all four visualization slices.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR gates visualization toolbars (lineplot, log, schematic, table) against missing Redux slice entries to eliminate a crash that occurred when switching workspaces caused a brief window where the toolbar mounted and dereferenced an absent entry. It standardizes on an explicit component-level gate (`useSelectExists`) rather than scattered selector-level null-safety.

- **Existence gate pattern**: Each toolbar now has a thin `Toolbar` wrapper that calls `useSelectExists` and returns `null` when absent, while the renamed `Internal` component holds the real implementation with non-null selector contracts.
- **Granular selectors added**: `selectExists`/`useSelectExists`, `selectActiveToolbarTab`/`useSelectActiveToolbarTab`, and `selectAxis`/`useSelectAxis` are introduced across all four slices, replacing whole-state reads with field-scoped subscriptions.
- **Selector unit tests**: New spec files cover all four slices, verifying both the happy path and the absent-entry fallback for defensive selectors.

Safe to merge — the existence-gate pattern is applied consistently and all non-null selector call sites are confirmed to be behind a gate.

Every toolbar now has a thin wrapper that returns null when the Redux slice entry is absent, and every non-null selector is only reachable from inside an Internal component that is conditionally rendered by that wrapper. The previous concern about selectLegendVisible, selectFitViewOnResize, and selectViewport dereferencing without a guard is resolved: useSelectLegendVisible is called only inside Control.tsx which is rendered from the gated Internal; useSelectFitViewOnResize and useSelectViewport have zero consumers in the codebase today. Defensive selectors that are still called before the gate retain optional-chaining fall-backs. New unit tests across all four slices confirm both the happy path and the absent-entry behaviour.

| Filename | Overview |
|----------|----------|
| console/src/schematic/selectors.ts | Adds selectExists/useSelectExists and selectActiveToolbarTab/useSelectActiveToolbarTab; reverts selectSelectedSymbolGroup to a non-null contract. All callers of non-null selectors are verified to be behind the useSelectExists toolbar gate. |
| console/src/schematic/toolbar/Toolbar.tsx | Splits into Toolbar (gate) + Internal (impl); gates on !exists || pendingUpload != null; replaces useSelectToolbar().activeTab with useSelectActiveToolbarTab, and removes the === true comparison on useSelectEditable. |
| console/src/lineplot/toolbar/Toolbar.tsx | Splits into Toolbar + Internal; gates on useSelectExists; replaces full toolbar state subscription with useSelectActiveToolbarTab; removes the null-check early return from the inner component. |
| console/src/log/toolbar/Toolbar.tsx | Same wrapper/internal split as the other toolbars; also fixes log.ontologyID(state.key) to log.ontologyID(layoutKey) now that the state object is no longer in scope. |
| console/src/table/selectors.ts | Adds selectExists/useSelectExists; converts selectEditable, selectHideIndicators, selectSelectedCellKeys, and selectLastSelected to non-null contracts — all callers are inside Loaded which is gated by useEnsureState. |
| console/src/schematic/toolbar/Properties.tsx | Removes ?.zoom ?? 1 fall-back in two getLayouts* helpers in favour of direct .zoom access — safe because Properties only renders from within the gated Internal. |
| console/src/lineplot/selectors.ts | Adds selectExists, selectActiveToolbarTab, and selectAxis families; all new non-null selectors are consumed only inside the gated Internal components. |
| console/src/schematic/selectors.spec.ts | New spec file covering selectExists, all toolbar selectors, and defensive selectors under the absent-entry case; confirms the boundary between null-contract and defensive selectors. |
| console/src/lineplot/selectors.spec.ts | New spec file with broad lineplot selector coverage including the new selectExists, selectActiveToolbarTab, selectAxis, and selectAxisBounds. |
| console/src/log/selectors.spec.tsx | New spec with Redux-provider-based hook tests for useSelect and useSelectExists, plus plain-function tests for all log selectors. |
| console/src/table/selectors.spec.tsx | Extends existing spec with selectExists, selectHideIndicators, selectVersion, and selectPendingUpload coverage; removes obsolete defensive-fallback tests that no longer apply to the non-null contracts. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph "Before (crash path)"
        OT[Toolbar mounts] --> OS[useSelectOptional / whole-state read]
        OS --> ONR{entry present?}
        ONR -- No --> ONC[null-check scattered in body]
        ONC -- missed --> OCRASH["TypeError: undefined.toolbar"]
        ONR -- Yes --> ORENDER[render]
    end

    subgraph "After (this PR)"
        NT[Toolbar wrapper mounts] --> NE["useSelectExists(layoutKey)"]
        NE -- false --> NNULL[return null]
        NE -- true --> NI[render Internal]
        NI --> NA["useSelectActiveToolbarTab(layoutKey)"]
        NA --> NRENDER[render tabs]
    end

    subgraph "Selector contracts"
        DEF["Defensive selectors\n(selectSelected, selectLegend, selectPendingUpload)\nuse ?. / ?? fallback"]
        NN["Non-null selectors\n(selectEditable, selectActiveToolbarTab, selectViewport)\nno fallback — callers must gate"]
    end
```

<sub>Reviews (2): Last reviewed commit: ["SY-4276: Gate visualization toolbars wit..."](https://github.com/synnaxlabs/synnax/commit/b2f3f199d0a517d90677560e13f7bcb4c1327429) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35281824)</sub>

### DIFF
--- a/console/src/hooks/selectors.spec.tsx
+++ b/console/src/hooks/selectors.spec.tsx
@@ -1,0 +1,137 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { configureStore, createSlice, type PayloadAction } from "@reduxjs/toolkit";
+import { act, renderHook } from "@testing-library/react";
+import { type PropsWithChildren, type ReactElement } from "react";
+import { Provider } from "react-redux";
+import { describe, expect, it } from "vitest";
+
+import { useMemoSelect } from "@/hooks/selectors";
+
+interface Item {
+  key: string;
+  value: number;
+}
+
+interface TestState {
+  tracked: { value: number };
+  untracked: { value: number };
+  items: Record<string, Item>;
+}
+
+const ZERO_STATE: TestState = {
+  tracked: { value: 1 },
+  untracked: { value: 1 },
+  items: { a: { key: "a", value: 1 }, b: { key: "b", value: 2 } },
+};
+
+const testSlice = createSlice({
+  name: "test",
+  initialState: ZERO_STATE,
+  reducers: {
+    setTracked: (state, { payload }: PayloadAction<number>) => {
+      state.tracked = { value: payload };
+    },
+    setUntracked: (state, { payload }: PayloadAction<number>) => {
+      state.untracked = { value: payload };
+    },
+    setItem: (state, { payload }: PayloadAction<Item>) => {
+      state.items[payload.key] = payload;
+    },
+  },
+});
+
+const { setTracked, setUntracked, setItem } = testSlice.actions;
+
+const makeStore = () => configureStore({ reducer: testSlice.reducer });
+
+const wrapperFor = (store: ReturnType<typeof makeStore>) => {
+  const Wrapper = ({ children }: PropsWithChildren): ReactElement => (
+    <Provider store={store}>{children}</Provider>
+  );
+  return Wrapper;
+};
+
+describe("useMemoSelect", () => {
+  it("returns a stable reference across a dispatch that does not touch the accessed state", () => {
+    const store = makeStore();
+    const { result } = renderHook(
+      () => useMemoSelect((s: TestState) => s.tracked, []),
+      { wrapper: wrapperFor(store) },
+    );
+    const first = result.current;
+    act(() => {
+      store.dispatch(setUntracked(99));
+    });
+    expect(result.current).toBe(first);
+  });
+
+  it("returns a new value when the accessed state changes", () => {
+    const store = makeStore();
+    const { result } = renderHook(
+      () => useMemoSelect((s: TestState) => s.tracked, []),
+      { wrapper: wrapperFor(store) },
+    );
+    const first = result.current;
+    act(() => {
+      store.dispatch(setTracked(2));
+    });
+    expect(result.current).not.toBe(first);
+    expect(result.current).toEqual({ value: 2 });
+  });
+
+  it("keeps a constructed object stable until its inputs change", () => {
+    const store = makeStore();
+    const { result } = renderHook(
+      () => useMemoSelect((s: TestState) => ({ doubled: s.tracked.value * 2 }), []),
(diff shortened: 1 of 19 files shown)


emilbon99 (changes requested):
Have you checked whether the other visualizations need these guards?

sy-nico:
@emilbon99 , Yes, I audited before I made the fix:
image

emilbon99:

PR #2426, opened 2026-06-03
SY-3036: Schematic Symbols - Resize by Dragging

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-3036](https://linear.app/synnax/issue/SY-3036)

## Description

Schematic symbols can now be resized by dragging when selected. Resizing was previously non-functional: a few primitives declared `onResize` props that were wired to nothing.

### Plumbing

- The shared `Grid` renders React Flow `NodeResizeControl` handles when a symbol is selected and given an `onResize`: four edge "line" handles plus four aspect-locked corner handles, with emitted dimensions rounded to integers.
- A `keepAspectRatio` option locks every handle for symbols sized by a single value, so their top and bottom edges resize too instead of doing nothing.
- `Schematic.css` reveals the handles only on selection, and the `createLabeled` factory forwards a resize to a config patch.

### Adoption

Each resizable symbol maps a resize to its own config field:

| Symbol | Config field |
| --- | --- |
| tank / box / cylinder | `dimensions` |
| value | `inlineSize` |
| light | `scale` |
| circle | `radius` |
| polygon | `sideLength` |

Single-dimension symbols pass `keepAspectRatio`, and the now-dead `onResize` props were removed from the tank/box/cylinder primitives.

### Cleanup

Light's hardcoded `51.2` scale-1 width was a pre-multiplied copy of two real constants. It is now the derived, exported `WIDTH_PER_SCALE = DIMENSIONS.width * BASE_SCALE`, so it cannot silently drift.

### Tests

- Grid plumbing: control gating, positions, and edge/corner variants on real React Flow DOM, with a minimal spy only for the `keepAspectRatio` and rounding the DOM does not expose.
- The `createLabeled` wiring, the real circle/polygon transforms, and light's width-to-scale conversion.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR wires up schematic symbol resizing end-to-end: `Grid` now renders React Flow `NodeResizeControl` handles when a symbol is selected and has an `onResize` prop, each symbol maps its drag to its own config field (dimensions, radius, sideLength, scale, or inlineSize), and a `resizing` guard in `nodeChangesToActions` prevents React Flow's node-measurement events from racing with the drag updates.

- **Grid plumbing**: eight `NodeResizeControl` handles (four edge "line" + four aspect-locked corners) are memoized and rendered conditionally on `editable && onResize != null`; a full-viewport cursor-overlay portal keeps the cursor from flickering as the mouse leaves the node boundary during drag.
- **Symbol adoption**: `createLabeled` defaults to a ratio-based `useScaleResize` hook; symbols with a single scalar dimension (circle, polygon, light) pass `keepAspectRatio` and a direct inversion formula; tank/box/cylinder pass `onResize` straight to the Grid, replacing the now-dead primitive-level `onResize` props.
- **`scale` consolidation**: the `scale` field is moved from each factory's own schema into the shared `labeledConfigZ`, and `Light`'s hardcoded `51.2` is replaced with the derived, exported `WIDTH_PER_SCALE`.

Safe to merge — the resize path is well-isolated and does not touch data persistence, auth, or any cross-service boundary.

All changed code is confined to UI-layer schematic components. The `resizing` guard correctly prevents write races during drag, `useScaleResize` captures start state via a ref before any config change fires, the cursor overlay is self-cleaning, and the CSS selector change matches the new DOM structure. Test coverage addresses render-gating, handle variants, and the resizing-flag drop path.

| Filename | Overview |
|----------|----------|
| pluto/src/schematic/node/common/grid/Grid.tsx | Core resize plumbing added: NodeResizeControl handles, useScaleResize hook, cursor overlay portal, and roundDimensions helper. Logic is sound and hook rules are respected. |
| pluto/src/schematic/node/common/label/Label.tsx | createLabeled now defaults to useScaleResize and accepts a typed onResize override; LabeledOverrides narrowed to only allowRotate/keepAspectRatio in grid to prevent silent override discards. |
| pluto/src/vis/diagram/aether/types.ts | Adds optional resizing flag to NodeChange dimensions branch and propagates it through translateNodeChangeForward with a ?? false default. |
| pluto/src/schematic/Diagram.tsx | Adds an early-return guard on ch.resizing in the dimensions case to prevent React Flow measurement events from racing with onResize config updates during active drags. |
| pluto/src/schematic/node/general/polygon/Symbol.tsx | New dedicated Symbol component for polygon with keepAspectRatio and a direct sideLength inversion formula; replaces the createLabeled shortcut that lacked resize support. |
| pluto/src/schematic/node/common/toggle/toggle.tsx | createToggle and createDummyToggle now spread useScaleResize props onto Grid, giving toggle-based symbols scale-resize behavior with no further changes needed per symbol. |
| pluto/src/schematic/node/general/light/Symbol.tsx | Light resize now correctly inverts the scale using the exported WIDTH_PER_SCALE constant rather than the previous hardcoded 51.2. |
| pluto/src/schematic/Schematic.css | Resize-control visibility rule lifted from inside .pluto-symbol-primitive to the node-level selector, which is the correct ancestor now that controls render directly in Grid. |
| pluto/src/schematic/node/common/grid/Grid.spec.tsx | New Grid resize controls test suite covers render-gating, correct count, and edge/corner variant+position assertions using real React Flow DOM. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant NodeResizeControl
    participant Grid
    participant Symbol
    participant Diagram
    participant Store

    User->>NodeResizeControl: drag resize handle
    NodeResizeControl->>Grid: onResizeStart(width, height)
    Grid->>Grid: setResizeCursor(cursor)
    Grid->>Grid: createPortal(overlay to document.body)
    Grid->>Symbol: onResizeStart?(roundDimensions)

    loop during drag
        NodeResizeControl->>Grid: onResize(width, height)
        Grid->>Symbol: onResize(roundDimensions)
        Symbol->>Store: onConfigChange dimensions/radius/scale/sideLength/inlineSize
        Note over Diagram: NodeDimensionChange(resizing=true) DROPPED
    end

    User->>NodeResizeControl: release handle
    NodeResizeControl->>Grid: onResizeEnd()
    Grid->>Grid: setResizeCursor(null) overlay removed
    Note over Diagram: NodeDimensionChange(resizing=false) setNodeMeasured
```

<sub>Reviews (3): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/rc&#39;..."](https://github.com/synnaxlabs/synnax/commit/ecf4b7ea7ed689f69c9e0091c9af50d44437ac56) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35324508)</sub>


pluto/src/schematic/node/general/circle/external.ts
```diff
@@ -34,7 +34,10 @@ export const spec: Spec<typeof VARIANT, Config> = {
   key: VARIANT,
   name: NAME,
   Form: CircleForm,
-  Node: Label.createLabeled<Config>(Circle, { grid: { allowRotate: false } }),
+  Node: Label.createLabeled<Config>(Circle, {
+    grid: { allowRotate: false, keepAspectRatio: true },
+    onResize: ({ width }) => ({ radius: width / 2 }),
```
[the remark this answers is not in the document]
sy-nico: `radius = width / 2` keeps radius in the symbol's native unit, and the `BASE_SCALE` shrink applies uniformly to every `Primitive.SVG` symbol (box, cylinder, polygon), so this is consistent rather than off. The suggested fix hardcodes `0.8` and a `- 3` padding that only holds at the default `strokeWidth`, so we're leaving it as-is to avoid the magic numbers.

pluto/src/schematic/node/general/polygon/external.ts
```diff
@@ -38,7 +38,10 @@ export const spec: Spec<typeof VARIANT, Config> = {
   key: VARIANT,
   name: "Polygon",
   Form: CommonPolygonForm,
-  Node: Label.createLabeled<Config>(Polygon),
+  Node: Label.createLabeled<Config>(Polygon, {
+    grid: { keepAspectRatio: true },
+    onResize: ({ width }) => ({ sideLength: width / 2 }),
```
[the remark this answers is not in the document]
sy-nico: Same as circle: `BASE_SCALE` applies to every `Primitive.SVG` symbol, so `sideLength = width / 2` is consistent with box and cylinder rather than a one-off bug. The suggested `width / (2 * 0.8)` drops the `sin(π / numSides)` term and hardcodes `0.8`, so it would not actually align the edge and introduces magic numbers we don't want.

pluto/src/schematic/node/general/light/Symbol.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    NodeResizeControl: (props: SpyResizeControlProps): ReactElement => {
+      resizeControls.set(props.position, props);
+      return <Real {...props} />;
+    },
+  };
+});
+
+// Light's telemetry source is irrelevant to resizing; stub the hook so the Symbol
+// renders without an Aether provider.
+vi.mock("@/vis/light", () => ({ Light: { use: () => ({ enabled: false }) } }));
```
pjdotson: Can you please please please make sure not to do mocks for things like this? We should only ever be mocking for something we can't use in tests - and really we probably should be doing dependency injection instead of mocking for that (look at `pagerduty/factory.go` with the `EventSender` interface as an example of this). This has popped up on a bunch of different PRs, you should probably edit the Claude instructions and this should also be something you personally catch before sending for review.

sy-nico: dropped it and swapped the `@/vis/light` stub for a real worker-disabled `Aether.Provider`, so the telemetry hook now runs for real. The only mock left is the `@xyflow/react` spy on line 32, which still renders the real `NodeResizeControl` and only captures its `onResize` prop since jsdom can't fire an actual drag. Is that one acceptable, or would you prefer I extract the rounding and transform logic into pure functions and test those directly? I deliberated on that for a while yesterday.

sy-nico:
@greptile, please update your review.

pluto/src/schematic/node/general/light/Symbol.spec.tsx
pjdotson: Ideally you have some type of prop for the Diagram that in prod you pass in react flow, and for tests we pass in some type of test prop / function. Probably is worth the work to make that change.

pluto/src/vis/diagram/Context.ts
```diff
@@ -20,18 +23,22 @@ export interface ContextValue {
...
-    editable: true,
-    fitViewOnResize: false,
-    fitViewOptions: diagram.FIT_VIEW_OPTIONS,
-    onEditableChange: () => {},
-    onViewportModeChange: () => {},
-    setFitViewOnResize: () => {},
-    viewportMode: "select",
-    visible: true,
-  },
+  defaultValue: ZERO_CONTEXT_VALUE,
```
pjdotson: i don't think you want to do this because different contexts are now the same object.

pluto/src/schematic/node/common/grid/Grid.spec.tsx
```diff
@@ -10,15 +10,46 @@
...
   type ReactElement,
   type ReactNode,
   useState,
 } from "react";
-import { describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import { Haul } from "@/haul";
 import { Grid } from "@/schematic/node/common/grid";
+import { Context as DiagramContext, ZERO_CONTEXT_VALUE } from "@/vis/diagram/Context";
```
pjdotson: you shouldn't really be calling the diagram context and the zero context value directly from outside of the diagram module. That is more of an implementation detail of the `Diagram.Diagram`.

pluto/src/schematic/node/common/grid/Grid.tsx
```diff
@@ -238,6 +271,19 @@ export const Grid: FC<GridProps> = ({
...
+          <ResizeControl
+            key={position}
+            position={position}
+            variant={variant}
+            keepAspectRatio={keepAspectRatio || corner}
+            onResize={(_, { width, height }) =>
+              onResize(roundDimensions(width, height))
+            }
+          />
+        ))}
```
pjdotson: the map should be memoized on `ResizeControl`.

pluto/src/schematic/node/general/light/Symbol.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+  resizeControls.set(props.position ?? "", {
+    triggerResize: (width, height) =>
+      props.onResize?.(RESIZE_EVENT, { x: 0, y: 0, width, height, direction: [] }),
+  });
+  return <NodeResizeControl {...props} />;
+};
+
+const diagramCtx = { ...ZERO_CONTEXT_VALUE, resizeControl: SpyResizeControl };
+
+const Wrap = ({ children }: { children: ReactNode }): ReactElement => (
```
pjdotson: we have generic test helpers you could use for things like the aether provider.

pluto/src/vis/diagram/Diagram.tsx
```diff
@@ -452,6 +453,7 @@ export const create = ({
         fitViewOptions,
         viewportMode,
         onViewportModeChange,
+        resizeControl: NodeResizeControl,
```
pjdotson: probably just want to remove all of this passing `resizeControl` as a prop and do the dependency injection properly in another PR.

pjdotson (changes requested):
1) resizing for some things like values does not work properly
2) a lot of elements can't be resized at all
3) when i resize by dragging to the right, I often find that my cursor is further to the right than the right edge of the node. I would expect them to be at the same x-position.
4) This feels "jumpy" when I resize and not smooth. You should download Figma and place down a rectangle and hit the "Lock aspect ratio" button and then resize it in the editor - that is the ideal experience we want to hit.

pluto/src/vis/diagram/Context.ts
sy-nico: Backed this out entirely. `resizeControl` and the shared `ZERO_CONTEXT_VALUE` are gone and the default is inline again as before.

pluto/src/schematic/node/common/grid/Grid.spec.tsx
sy-nico: Removed the `DiagramContext` and `ZERO_CONTEXT_VALUE` imports so the spec no longer reaches into the diagram module, and the proper injectable seam is moving to the follow-up issue ([SY-4302](https://linear.app/synnax/issue/SY-4327/refactor-diagram-tests-to-use-dependency-injection-for-xyflowreact)) you filed.

pluto/src/schematic/node/common/grid/Grid.tsx
sy-nico: Wrapped the control array in a u`seMemo` keyed on `editable`, `onResize`, and `keepAspectRatio`.

pluto/src/schematic/node/general/light/Symbol.spec.tsx
sy-nico: Deleted this spec entirely as part of backing out the injection, so the aether-provider setup is gone too.

pluto/src/vis/diagram/Diagram.tsx
sy-nico: Removed the `resizeControl` injection here and in the context; the proper DI is moving to the follow-up issue.

sy-nico:
Issues addressed:
- **Jumpy (4):** dual source of truth... React Flow's measured size fought the config-driven `onResize` each frame. Suppressing the drag-driven measure plus a drag-time cursor overlay fixed it.
- **Cursor past the edge (3):** uncompensated `BASE_SCALE`, now handled for the whole SVG family by a ratiometric `useScaleResize` hook so the edge tracks the cursor at any size.
- **Value (1):** resizes cleanly from the right and widens from the left; top/bottom/corner handles removed. Left-shrink is limited by the async text-width worker and is documented, not hacked.
- **Can't be resized (2):** now wired across value, select, state indicator, box, tank, cylinder, circle, polygon, and the entire scale family (meters, pumps, valves, fittings, process, safety) through the shared factories.

sy-nico:
@greptile, please update the feedback. Discussed synchronously with reviewer and scoped the DI as its own fix/improvement.

sy-nico:
@greptile, update the PR description with your new score.

pluto/src/vis/diagram/aether/types.ts
```diff
@@ -142,6 +143,7 @@ export const translateNodeChangeForward = (
         type: "dimensions",
         key: change.id,
         dimensions: change.dimensions,
+        resizing: change.resizing ?? false,
```
pjdotson, line 146: could you explain why we need this? and should be a test for this

pluto/src/schematic/Diagram.tsx
```diff
@@ -77,6 +77,8 @@ export const nodeChangesToActions = (
         actions.push(schematic.setNodePosition({ key: ch.key, position: ch.position }));
         return;
       case "dimensions":
+        // onResize drives the symbol's size during a drag; skip the competing write.
+        if (ch.resizing) return;
```
pjdotson, line 81: have you tested how well this works with cross-Console editing?

pjdotson (changes requested):
mostly looks good

pluto/src/vis/diagram/aether/types.ts
sy-nico: React Flow tags every dimension change during an active drag with `resizing: true`. We forward that flag here so `nodeChangesToActions` in `Diagram.tsx` can skip the `"dimensions"` change while a resize is in flight.

The intent is to make the symbol size flow through one path only: a resize drag writes the new size into the symbol's `config`, that `config` comes back down as props, and the symbol re-renders and re-measures from those props. Without this flag, RF also writes the size directly via `setNodeMeasured`, so two writers race every frame and the drag jumps (this is why it was originally super choppy). Forwarding `resizing` lets the config/props path be the sole writer mid-drag, and once the drag ends the flag goes back to `false` so RF's normal measurement resumes.

sy-nico: Added tests

pluto/src/schematic/Diagram.tsx
sy-nico: Yes, and just re-tested again to verify. Looks good across multiple symbols and sessions.

pjdotson (approved):

PR #2433, opened 2026-06-03
K. SY-4299: Strongly-Typed NI Task Types (Part 1)

emilbon99:
## Issue

Closes [SY-4299](https://linear.app/synnax/issue/SY-4299/).

## Description

Rebuilt from scratch on top of the task refactor chain (autosave stack + Arc convergence), replacing the June diff. This is Phase 2 of RFC 0051 for NI: the schema plus inert generated clients, nothing wired.

- Adds `schemas/synnax/ni.oracle`: 32 enums, 4 unions, 57 structs covering the five NI task configs and their channel, scale, and excitation shapes. Channels carry their own device (multiport shape).
- Adds shared config bases to `schemas/synnax/task.oracle`: `BaseStartConfig` (`auto_start`, `data_saving_disabled`), `BaseReadConfig` (adds `sample_rate`, `stream_rate`), `BaseWriteConfig` (adds `device`), generated into `core/pkg/service/task/common` and `client/cpp/task/common`.
- Boolean polarity now satisfies the RFC 0043 invariant (defaults always zero): channel `enabled` (default true) flips to `disabled`, and `data_saving` (default true) flips to `data_saving_disabled`. Console and Driver rewires in the follow-up PRs map the old wire shape through their migration transforms.
- The thermocouple channel models cold-junction compensation as a nested `CJC` union on `source` (`built_in`, `const_val.val`, `chan.port`), replacing the legacy flat `cjc_source`, `cjc_val`, and `cjc_port` fields. The Driver still reads the flat shape, so the Driver rewire migrates the union out to flat fields on the wire.
- Drops the `ai_rosette_strain_gage` channel type from the schema and the hand-written Python client. It was broken before this chain and is unused.
- Corrects the charge units to `Coulombs` and `PicoCoulombs`, and drops the terminal config from `ai_force_bridge_polynomial`. Neither matched the DAQmx signature.
- Generated outputs land in `client/ts/src/ni`, `client/py/synnax/ni`, `client/cpp/ni`, and `core/pkg/service/ni`. NI config types stay unversioned: they are not gorp-persisted until the config-table phase.

## Reviewer Checklist

- [ ] Sufficient test coverage
- [ ] Docs updated
- [ ] Changelog entry


client/cpp/ni/json.gen.h
```diff
@@ -0,0 +1,2741 @@
...
+
+inline NoneScale NoneScale::parse(x::json::Parser parser) {
+    return NoneScale{};
+}
+
+inline x::json::json NoneScale::to_json() const {
+    x::json::json j;
+    return j;
+}
+
```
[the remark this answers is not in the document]
emilbon99: Addressed in c5a873c58f: the C++ JSON codec now leaves the parse parameter unnamed for field-less structs (`NoneScale::parse(x::json::Parser)`), so there's no unused-parameter warning.

client/cpp/ni/json.gen.h
```diff
@@ -0,0 +1,2741 @@
...
+    j["thermocouple_type"] = this->thermocouple_type;
+    j["cjc_source"] = this->cjc_source;
+    j["cjc_val"] = this->cjc_val;
+    j["cjc_port"] = this->cjc_port;
+    return j;
+}
+
+inline AITorqueBridgeTableFields
+AITorqueBridgeTableFields::parse(x::json::Parser parser) {
+    AITorqueBridgeTableFields result;
```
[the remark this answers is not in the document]
emilbon99: Addressed: changed `cjc_val`/`cjc_port` to hard-optional (`??`) in the schema, so C++ now generates `std::optional<...>` and emits `null` (which Python reads as `None`) instead of `0`. Regenerated in 1d3eee627f. Note: the key is still present as `null` rather than omitted; fully omitting primitive optionals would be a separate, broader codec change.

client/ts/src/ni/types.gen.ts
```diff
@@ -0,0 +1,2194 @@
...
+  "InchOunces",
+  "FootPounds",
+] as const;
+export const unitsZ = z.enum(UNITSS);
+export type Units = z.infer<typeof unitsZ>;
+
+export const TERMINAL_CONFIGS = [
+  "Cfg_Default",
+  "RSE",
+  "NRSE",
```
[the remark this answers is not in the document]
emilbon99: Addressed in c5a873c58f: the enum values const no longer doubles a trailing S, so e.g. `UNITS` stays `UNITS` instead of `UNITSS`.

schemas/ni.oracle
```diff
@@ -71,68 +285,1051 @@ Scale union on type {
...
+        @doc value "is the scale applied to the raw measurement."
+    }
+}
+
 BaseAIChannel struct {
-    key     string {
+    key     string = "" {
         @doc value "uniquely identifies the channel within the task."
         @key
     }
```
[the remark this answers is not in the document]
emilbon99: Deferring this one: oracle has no support yet for a default on a union-typed field, so `custom_scale Scale = none` would require a new generator feature (union-field defaults across the analyzer + all four plugins). Nothing is broken without it — the console already derives its zero value via `parse({ type: 'none' })`. Tracking as a separate follow-up rather than expanding this PR.

schemas/synnax/ni.oracle
```diff
@@ -0,0 +1,1458 @@
...
+        units          ForceUnits = newtons {
+            @doc value "are the units of the force measurement."
+        }
+        physical_units ForceUnits = newtons {
+            @doc value "are the physical units of the polynomial calibration."
+        }
+
+        @doc value "reads force from a bridge with polynomial calibration."
+    }
+    ai_freq_voltage extends MinMaxVal, CustomScale {
```
sy-nico, line 737: The variant key here is `ai_freq_voltage`, but the driver registers this channel as `ai_frequency_voltage` (`driver/ni/channel/channels.h:2455`). It is the only driver factory key with no matching schema variant, and `parse_input` rejects anything missing from the table, so the pairing fails with "unknown channel type" until one side moves.

Python uses `ai_freq_voltage` (`client/py/synnax/ni/types.py:730`) and the Console never emits this channel type, so the schema agrees with the only client that names it, and that client cannot drive the channel today either way. Either name works.
Schema side:

```
    ai_frequency_voltage extends MinMaxVal, CustomScale {
```

Driver side:

```cpp
    INPUT_CHAN_FACTORY("ai_freq_voltage", AIFrequencyVoltage),
```

schemas/synnax/ni.oracle
```diff
@@ -0,0 +1,1458 @@
...
+        shunt_resistor_loc     ShuntResistorLocation = default {
+            @doc value "selects where the shunt resistor is located."
+        }
+        ext_shunt_resistor_val float64 = 1 {
+            @doc value "is the external shunt resistor value, in ohms."
+        }
+
+        @doc value "reads RMS current."
+    }
+    ai_force_bridge_polynomial extends MinMaxVal, Terminal, Bridge, VoltageExcitation, BridgePolynomial, CustomScale {
```
sy-nico: Drop `Terminal` from both of these, and the `units` field from `ai_rosette_strain_gage`.

`DAQmxCreateAIForceBridgePolynomialChan` and `DAQmxCreateAIRosetteStrainGageChan` take no `terminalConfig` parameter, and the rosette function takes no units parameter either. The driver parses `terminal_config` for neither channel, and it parses `units` for the rosette only to drop it before the create call. `parse_units` defaults to `"Volts"`, so removing the field does not break the driver.

Python carries `terminal_config` on these two plus `AIStrainGageChan` and `AIBridgeChan`, and none of the four DAQmx functions accept it. The schema already dropped it from `ai_strain_gauge` and `ai_bridge`, so these two look missed rather than deliberate.

Both extends lines should be:

```
    ai_force_bridge_polynomial extends MinMaxVal, Bridge, VoltageExcitation, BridgePolynomial, CustomScale {
```

```
    ai_rosette_strain_gage extends MinMaxVal, VoltageExcitation {
```

schemas/synnax/ni.oracle
```diff
@@ -0,0 +1,1458 @@
...
+        @doc value "reads pressure from a bridge with polynomial calibration."
+    }
+    ai_rosette_strain_gage extends MinMaxVal, Terminal, VoltageExcitation {
+        rosette_type            RosetteType = rectangular {
+            @doc value "selects the rosette geometry."
+        }
+        gage_orientation        float64 = 0 {
+            @doc value "is the orientation of gauge element 0, in degrees."
+        }
+        rosette_meas_types      RosetteMeasType[] = [] {
```
sy-nico: `rosette_meas_types` has nowhere to land. `BaseAIChannel` gives this variant one `channel` key and one `port`, but NI emits one output channel per requested measurement type, and the call spans several gages ("Creates channels to measure two-dimensional strain using a rosette strain gage", with `physicalChannel` naming the gages). N measurement types means N streams and one Synnax channel to write them to.

The cheap resolution is a scalar. It matches the single channel key, and it matches the driver, which already hardcodes the count to 1:

```
        rosette_meas_type       RosetteMeasType = principle_strain_1 {
```

Keeping the array means pairing each measurement type with its own `channel.Key`, so the element type becomes a struct instead of an enum, and the variant needs a way to address more than one gage port.

Separately, `AIRosetteStrainGauge` parses the singular key `rosette_meas_type` with no default, so a config carrying `rosette_meas_types` fails with "This field is required" either way.

schemas/synnax/ni.oracle
```diff
@@ -0,0 +1,1458 @@
...
+        @doc value "reads temperature from the device's built-in sensor."
+    }
+    ai_thermocouple extends MinMaxVal {
+        units             TemperatureUnits = deg_c {
+            @doc value "are the units of the temperature measurement."
+        }
+        thermocouple_type ThermocoupleType = j {
+            @doc value "selects the thermocouple alloy type."
+        }
+        cjc               CJC              {
```
sy-nico: `cjc` has the same missing default as `custom_scale`, since Oracle has no union-field defaults yet. Whatever ticket covers that should cover this field too.

schemas/synnax/ni.oracle
```diff
@@ -0,0 +1,1458 @@
...
+    r = "R"
+    s = "S"
+    t = "T"
+    b = "B"
+    e = "E"
+
+    @doc value "selects the thermocouple alloy type."
+}
+
+CJC union on source {
```
sy-nico, line 196: The `cjc` paragraph in the description is backwards. It says the schema flattens the nested `cjcSource` union into `cjc_source`, `cjc_val`, and `cjc_port`, but the schema nests `cjc` as a union on `source`, and the generated TS emits `z.discriminatedUnion("source", ...)`. The driver and the Console both read the flat shape today, so the follow-up migration gets written against the wrong target if the description stands.

sy-nico (changes requested):

schemas/synnax/ni.oracle
```diff
@@ -0,0 +1,1458 @@
...
+    delta       = "DeltaRosette"
+    tee         = "TeeRosette"
+
+    @doc value "selects the geometric arrangement of a strain-gauge rosette."
+}
+
+RosetteMeasType enum {
+    principle_strain_1        = "PrincipleStrain1"
+    principle_strain_2        = "PrincipleStrain2"
+    principle_strain_angle    = "PrincipleStrainAngle"
```
sy-nico: These should be `Principal`. Cross-check against NI headers.

sy-nico (changes requested):

schemas/synnax/ni.oracle
emilbon99: These gets fixed in PR part 3. This was broken beforehand and we're going to standardize on `ai_freq_voltage`.

sy-nico (approved):
Followup: [SY-4574](https://linear.app/synnax/issue/SY-4574/fix-undefined-behavior-in-ni-parse-units-on-unknown-units)

PR #2434, opened 2026-06-03
L. SY-4299: Strongly-Typed NI Task Types (Part 2 — Console)

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Part 2 rewires the Console onto the Oracle-generated NI types from Part 1 and deletes
the hand-written channel definitions.

- `console/src/feature/ni/task/types/index.ts` now sources channel types, unions,
  per-type schemas, and enums from the `ni` namespace of `@synnaxlabs/client`. Deletes
  `v0.ts` (~1800 lines) and `v1.ts`.
- The generated schemas follow the RFC 0043 zero-default boolean rule, so the stored
  shape flips polarity: channels carry `disabled` and configs carry
  `dataSavingDisabled`. The Console parses only the generated shape. Configs stored
  with the old keys migrate once at Core boot (the storage cutover PR later in this
  chain); there are no client-side compat arms.
- Platform task components (`EnableDisableButton`, `ChannelList` context menu,
  `ListAndDetailsChannelItem`, `Fields.DataSaving`) take a `polarity` prop
  (default `"enabled"`). NI passes `"disabled"`; other integrations stay on `enabled`
  because their driver parsers still read it. The prop dies once every integration is
  on generated types.
- `ZERO_*` defaults derive from the generated schemas' baked-in defaults
  (`schema.parse({ type, customScale: NO_SCALE })`) rather than hand-maintained tables.
- Cross-field validation Oracle does not generate is re-attached on the deploy schemas:
  map/table scale refinements, per-device port uniqueness, digital port/line
  uniqueness, and stream-rate bounds. Draft (stored) schemas stay lax so drafts
  round-trip.
- The Console keeps its existing 18 AI / 11 CI / 3 AO / digital channel types; the 13
  driver-only AI variants and the polynomial scale still parse on the wire but are not
  offered in the UI. Icon maps are total over the generated union, so retrieved
  configs with unoffered types render without crashing.

`pnpm check-types:console` and `pnpm lint:console` are clean; the NI suite and the
platform task suite pass against a live core.


console/src/hardware/ni/task/AnalogRead.tsx
```diff
@@ -66,7 +66,7 @@ const ChannelListItem = ({ onTare, ...rest }: ChannelListItemProps) => {
   const isRunning = Common.Task.useIsRunning();
   const hasTareButton = channel !== 0 && !isSnapshot;
   const canTare = enabled && isRunning;
-  const Icon = AI_CHANNEL_TYPE_ICONS[type];
+  const Icon = AI_CHANNEL_TYPE_ICONS[type]!;
```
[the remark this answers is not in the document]
emilbon99: Addressed in b28c0d29a4: `AI_CHANNEL_TYPE_ICONS` is now a total `Record<AIChannelType, Icon.FC>` (the 13 driver/Python-only types reuse their measurement-category icon), so the non-null assertion is gone and a Python-created channel of an unsupported type renders instead of crashing.

console/src/feature/ni/task/AIChannelForm.tsx
```diff
@@ -978,6 +986,7 @@ export interface AIChannelFormProps {
 
 export const AIChannelForm = ({ type, prefix }: AIChannelFormProps) => {
   const Form = CHANNEL_FORMS[type];
+  if (Form == null) throw new Error(`No form for AI channel type ${type}`);
```
sy-nico: `AIChannelForm` throws for the 12 driver-only AI types. You made `AI_CHANNEL_TYPE_ICONS` total so those configs render in the list, but selecting one crashes the details pane. Should these be rendered in the base fields instead of throwing?

console/src/feature/ni/task/types/index.ts
```diff
@@ -7,119 +7,496 @@
...
+
+// A complete default channel for every generated type, derived from the schema
+// defaults. The console only offers the types in AI_CHANNEL_TYPE_NAMES; the rest exist
+// so configs created by the Python client or the Driver still round-trip.
+export const ZERO_AI_CHANNELS = Object.fromEntries(
+  ni.AI_CHANNEL_TYPES.map((t) => [
+    t,
+    ni.AI_CHANNEL_SCHEMAS[t].parse({
+      type: t,
+      customScale: NO_SCALE,
```
sy-nico: `customScale` and `cjc` prefault in the generated schemas now (`{type:"none"}` and `{source:"built_in"}`), so both arguments are no-ops. Five AI types have no `customScale` field at all and just strip it. Drop `NO_SCALE`, `BUILT_IN_CJC`, and the same arguments at lines 186 and 231.

console/src/feature/ni/task/types/index.ts
```diff
@@ -7,119 +7,496 @@
...
+      .check(createPortValidator())
+      .check(Task.validateChannelDevices),
+  })
+  .check(Task.validateStreamRate);
+
+const ZERO_ANALOG_READ_CONFIG = ni.analogReadConfigZ.parse({});
+
+const analogReadStatusDataZ = z
+  .object({ errors: z.array(z.object({ message: z.string(), path: z.string() })) })
+  .nullish()
```
sy-nico: `.nullish()` already implies `.optional()`, so the trailing call can probably be removed.

sy-nico (changes requested):

console/src/feature/ni/task/AIChannelForm.tsx
emilbon99: I ended up implementing forms and driver side handling for all of these.

console/src/feature/ni/task/AIChannelForm.tsx
sy-nico, line 372: `ai_accel` carries `units: accelUnitsZ` too, but only the two new accel forms offer `AccelUnitsField`. Add it here so a plain accelerometer is not locked to g.

driver/ni/channel/channels.h
```diff
@@ -1188,10 +1187,9 @@ struct AIFrequencyVoltage final : AICustomScale {
         TaskHandle task_handle,
         const char *scale_key
     ) const override {
-        const auto port = this->dev_loc + "ctr" + std::to_string(this->port);
         return dmx->CreateAIFreqVoltageChan(
             task_handle,
-            port.c_str(),
+            this->loc().c_str(),
```
sy-nico, line 1192: Shouldn't this be in part 3? `apply(const AIFreqVoltageChannel &)` there still builds `ctx.dev_loc + "ctr" + std::to_string(ch.port)`, without even a leading slash, so the merge needs to carry this over

sy-nico (changes requested):

driver/ni/channel/channels.h
emilbon99: I implemented the corresponding fix on part 3 a few commits ago, so this will all get squashed together whne we merge.

sy-nico (approved):

PR #2436, opened 2026-06-03
SY-4316: Bugfix - Make Schematic Symbols Follow the Console Theme

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4316](https://linear.app/synnax/issue/SY-4316/schematic-symbol-colors-not-responding-to-theme-change)

## Description

Schematic symbols baked a concrete color into their config when placed, so they did not re-color when the console light or dark theme was toggled. A symbol placed in light mode rendered with a near-black stroke that became nearly invisible on a dark background, and the reverse happened for symbols placed in dark mode. 

https://github.com/user-attachments/assets/139535ce-49bd-4b12-869f-63144aacaf96

Symbols now store the `color.ZERO` follow-theme sentinel by default and resolve it to the live theme color at render time, so they re-color on theme toggles while a symbol the user explicitly colors keeps its color. This matches the behavior the value and gauge visualizations and edges already use.

Key changes:
- Resolve the color in the shared `Primitive.SVG` keystone, which covers most symbols, plus the primitives that draw color outside the SVG inheritance path (junctions, several valves, safety, value, state indicator, tank, cylinder, gauge, and the form control symbols).
- Add a shared `resolveColor(c, theme)` and `useColor(c)` helper so the resolution lives in one place instead of being duplicated per primitive.
- Change every symbol `defaultConfig` color from `gray.l11` to `color.ZERO`.
- Preview the resolved theme color in the color picker swatch so a default symbol shows its real rendered color instead of a transparent swatch. Background color fields keep their transparent default.

Already placed symbols keep their captured color, since their persisted config holds an explicit value. Only newly placed symbols follow the theme.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes schematic symbols baking a concrete theme color into their config at placement time, which caused symbols placed in light mode to become nearly invisible in dark mode and vice versa. Symbols now default to `color.ZERO` as a "follow-theme" sentinel and resolve it to the live theme color at render time.

- A shared `resolveColor(c, theme)` / `useColor(c)` helper is introduced in `common/color.ts`, and every primitive that draws color outside the SVG inheritance path (junctions, valves, gauge, value, state indicator, tank, cylinder, etc.) is updated to call it before rendering.
- Every `defaultConfig` factory that previously captured `t.colors.gray.l11` at placement time now stores `color.ZERO`; the `ColorField` form swatch is updated to display the resolved visual color so the picker shows the actual rendered color rather than a transparent swatch.

Safe to merge — the change is additive and backward-compatible: existing symbols with an explicit color in their persisted config are unaffected, and new symbols resolve correctly to the active theme color at render time.

The sentinel value color.ZERO is cleanly identified by color.isZero, resolution is centralized in one helper, and the update is consistently applied across all 51 files. Unit tests cover the sentinel, undefined, explicit-color, and theme-divergence cases. Backward compatibility is preserved because previously placed symbols carry gray.l11 explicitly (which is non-zero) and pass through unchanged.

| Filename | Overview |
|----------|----------|
| pluto/src/schematic/node/common/color.ts | New shared color resolution helper — resolveColor and useColor are clean, minimal, and correctly use color.isZero to detect the follow-theme sentinel. |
| pluto/src/schematic/node/common/color.spec.ts | New unit tests for resolveColor covering the ZERO sentinel, undefined, theme-divergence, and explicit-color cases — thorough coverage. |
| pluto/src/schematic/node/common/primitive/SVG.tsx | Now always resolves color before rendering and always sets --pluto-symbol-color / --pluto-symbol-color-contrast CSS variables, eliminating the conditional guard that left them unset for missing colors. |
| pluto/src/schematic/node/common/form/Color.tsx | ColorField now resolves the color before passing it to the swatch so the picker preview shows the actual rendered color instead of a transparent swatch. |
| pluto/src/schematic/node/general/gauge/Primitive.tsx | Gauge arc stroke now resolves via useColor; replaces the old c ?? 'var(--pluto-primary-z)' fallback, aligning gauge default color with all other symbols. |
| pluto/src/schematic/node/general/value/Primitive.tsx | Value border and text color now derive from resolveColor; the old 'var(--pluto-gray-l0)' CSS-variable fallback for null color is replaced with pickByContrast against the resolved theme color. |
| pluto/src/schematic/node/vessels/crossJunction/Primitive.tsx | Refactored from arrow-function to block body to hoist useColor call; both the SVG and the inline path fill use the pre-resolved color correctly. |
| pluto/src/schematic/node/vessels/tJunction/Primitive.tsx | Same refactor as crossJunction — pre-resolves color before rendering, eliminating direct color.cssString(colorVal) calls with potentially-zero input. |
| pluto/src/schematic/node/general/polygon/Primitive.tsx | Default stroke color changes from gray.l9 (old ad-hoc fallback) to gray.l11 (via resolveColor) to align with all other symbols. |
| pluto/src/schematic/node/valves/symbols.ts | Solenoid valve defaultConfig updated to color.ZERO; _t unused-parameter convention applied consistently. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["Symbol defaultConfig\n(color: color.ZERO)"] --> B["Symbol placed on schematic\n(config persisted with color.ZERO)"]
    B --> C{Render time}
    C --> D["resolveColor / useColor called"]
    D --> E{color == null\nor isZero?}
    E -- Yes --> F["theme.colors.gray.l11\n(follows active theme)"]
    E -- No --> G["Explicit user color\n(unchanged)"]
    F --> H["Primitive.SVG / inline fills\nrendered with theme color"]
    G --> H
    H --> I["Theme toggled\n(light ↔ dark)"]
    I --> C

    J["Legacy symbol\n(config has explicit gray.l11)"] --> C
    J -.-> K["color.isZero = false\n→ keeps captured color"]
```

<sub>Reviews (3): Last reviewed commit: ["SY-4316: Inline single-use useColor call..."](https://github.com/synnaxlabs/synnax/commit/5a970eb237b75bce7dd655cf4117e6e8b88dfc68) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35479667)</sub>

### DIFF
--- a/pluto/src/schematic/node/common/color.spec.ts
+++ b/pluto/src/schematic/node/common/color.spec.ts
@@ -1,0 +1,38 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { color } from "@synnaxlabs/x";
+import { describe, expect, it } from "vitest";
+
+import { resolveColor } from "@/schematic/node/common/color";
+import { Theming } from "@/theming";
+
+describe("resolveColor", () => {
+  const light = Theming.themeZ.parse(Theming.SYNNAX_LIGHT);
+  const dark = Theming.themeZ.parse(Theming.SYNNAX_DARK);
+
+  it("should resolve the ZERO follow-theme sentinel to the theme's default color", () => {
+    expect(resolveColor(color.ZERO, light)).toEqual(light.colors.gray.l11);
+    expect(resolveColor(color.ZERO, dark)).toEqual(dark.colors.gray.l11);
+  });
+
+  it("should resolve an unset color to the theme's default color", () => {
+    expect(resolveColor(undefined, light)).toEqual(light.colors.gray.l11);
+    expect(resolveColor(undefined, dark)).toEqual(dark.colors.gray.l11);
+  });
+
+  it("should re-color a default symbol when the theme changes", () => {
+    expect(resolveColor(color.ZERO, light)).not.toEqual(resolveColor(color.ZERO, dark));
+  });
+
+  it("should honor an explicit color regardless of theme", () => {
+    expect(resolveColor("#ff0000", light)).toEqual("#ff0000");
+    expect(resolveColor("#ff0000", dark)).toEqual("#ff0000");
+  });
+});
--- a/pluto/src/schematic/node/common/color.ts
+++ b/pluto/src/schematic/node/common/color.ts
@@ -1,0 +1,24 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { color } from "@synnaxlabs/x";
+
+import { Theming } from "@/theming";
+
+/// resolveColor returns the theme's default symbol color when c is unset or the ZERO
+/// "follow-theme" sentinel, and otherwise honors the explicit color.
+export const resolveColor = (
+  c: color.Crude | undefined,
+  theme: Theming.Theme,
+): color.Crude => (c == null || color.isZero(c) ? theme.colors.gray.l11 : c);
+
+/// useColor resolves a symbol's configured color against the active theme; primitives
+/// that already hold a theme should call resolveColor directly.
+export const useColor = (c?: color.Crude): color.Crude =>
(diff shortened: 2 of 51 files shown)


sy-nico:
@greptile, review again. Regarding your concern: 

Going the other way here: the codebase inlines a single-use hook into the value expression rather than adding a named intermediate (e.g. `createErrorHandler(useAdder(ctx))` in `status/aether/aggregator.ts`, `Theming.use().colors.gray.l0` in `color/BaseSwatch.tsx`). ESLint doesn't flag it. The `const resolved = useColor(...)` form was the inconsistency, so
I've inlined the single-use sites (`circle`, `gauge`, `textBox`, `Label`). Multi-use sites keep the variable since a hook can't be called repeatedly.

emilbon99 (changes requested):
Not the sustainable fix IMO as discussed in slack: 

1. Does not match the pre-regression behavior
2. Doesn't help with all of our customers existing schematics, which all have explicit colors set

Why is the solution to make it so we only change when the color is not set? Is this the right approach? Why? Why aren't we changing based on contrast like we used to.

PR #2438, opened 2026-06-03
SY-4320: Modify createDispatch to Automatically Handle Empty Actions

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4320](https://linear.app/synnax/issue/SY-4320)

## Description

`createDispatch` now short-circuits when given an empty actions array, returning early instead of dispatching a no-op. This removes the need for callers to guard against empty actions themselves, so the redundant `if (actions.length === 0) return` checks are dropped from every call site (schematic/table clipboards, schematic node/edge change handlers, and the table toolbar variant/cell forms).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR centralises empty-action handling inside `createDispatch` by adding an early `return true` when `actions.length === 0`, then removes the per-call-site `if (actions.length === 0)` guards that were previously needed at each dispatch call site.

- **`pluto/src/flux/dispatch.ts`**: New early-exit path (`return true`) makes the empty-actions case a silent no-op across all consumers without any caller changes.
- **`pluto/src/schematic/Schematic.tsx`, `console/src/table/Toolbar.tsx`**: Removed redundant guards; the simplification is safe because the change to `createDispatch` covers the same cases.
- **`pluto/src/schematic/clipboard.ts` and `pluto/src/table/clipboard.ts`**: Guards removed, but the `onPaste` callback that follows is no longer protected — in the schematic case this can fire with an empty array when nothing was actually pasted.

Mostly safe — the core dispatch change and most call-site clean-ups are correct, but the schematic clipboard removal unguards the `onPaste` side-effect and can fire it with an empty selection after a no-op paste.

The schematic clipboard now calls `onPaste?.([])` in a reachable path (copy only edges, paste) where it previously returned silently. Any consumer that uses `onPaste` to update the visible selection could inadvertently clear or reset it when the user intended nothing to happen.

`pluto/src/schematic/clipboard.ts` — the removed guard also removed protection for the `onPaste` callback, not just the `dispatch` call.

| Filename | Overview |
|----------|----------|
| pluto/src/flux/dispatch.ts | Adds `if (actions.length === 0) return true` early exit in `dispatchAsync`; returning `true` correctly signals "trivially successful" so callers using the boolean result are not misled by a `false`/failure signal. |
| pluto/src/schematic/Schematic.tsx | Removed redundant `actions.length > 0` guards from `handleNodesChange` and `handleEdgesChange`; these changes are safe because `createDispatch` now handles the empty-actions case centrally. |
| pluto/src/schematic/clipboard.ts | Removed the `actions.length === 0` guard, but this also unguards the `onPaste` call — meaning `onPaste?.([])` can now fire after a no-op paste (e.g., edges-only clipboard), which was not possible before. |
| pluto/src/table/clipboard.ts | Removed the `actions.length === 0` guard; the existing `payload.cells.length === 0` check at line 115 and subsequent `addCol`/`addRow`/`setCell` logic make an empty-actions path virtually unreachable in practice. |
| console/src/table/Toolbar.tsx | Removed `actions.length` guards from `CellForm.handleVariantChange`, `MultiCellForm.applyPropPatch`, and `MultiCellForm.handleVariantChange`; all changes are safe given the central fix in `createDispatch`. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Caller: dispatch / dispatchAsync] --> B{client == null?}
    B -- yes --> C[return false]
    B -- no --> D[Normalize actions array]
    D --> E{actions.length === 0?}
    E -- yes --> F[return true ✓ NEW early exit]
    E -- no --> G[apply: replay + registerOutstandingDispatch]
    G --> H{replay == null?}
    H -- yes --> I[return false]
    H -- no --> J[commitStack / recordEntry]
    J --> K[send to server]
    K -- success --> L[return true]
    K -- failure --> M[stackRollback + rollback + addStatus]
    M --> N[return false]

    style F fill:#c8f7c5,stroke:#2ecc71
```

<sub>Reviews (1): Last reviewed commit: ["SY-4320: Modify createDispatch to Automa..."](https://github.com/synnaxlabs/synnax/commit/289bb083bc1a06018504a4dd39721ce95ce8163a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35486348)</sub>

### DIFF
--- a/pluto/src/flux/dispatch.ts
+++ b/pluto/src/flux/dispatch.ts
@@ -110,6 +110,7 @@ export const createDispatch = <
       async (input: DispatchInput<Key, Action>): Promise<boolean> => {
         if (client == null) return false;
         const actions = Array.isArray(input.actions) ? input.actions : [input.actions];
+        if (actions.length === 0) return true;
         return await apply(
           store,
           client,
--- a/pluto/src/schematic/clipboard.ts
+++ b/pluto/src/schematic/clipboard.ts
@@ -142,7 +142,6 @@ export const useClipboard = ({
         if (cfg != null)
           actions.push(schematic.setConfig({ key: newKey, config: cfg }));
       }
-      if (actions.length === 0) return;
       dispatch({ key, actions });
       onPaste?.(Object.values(remap));
     },
--- a/pluto/src/schematic/clipboard.ts
+++ b/pluto/src/schematic/clipboard.ts
@@ -143,7 +143,7 @@ export const useClipboard = ({
           actions.push(schematic.setConfig({ key: newKey, config: cfg }));
       }
       dispatch({ key, actions });
-      onPaste?.(Object.values(remap));
+      if (actions.length > 0) onPaste?.(Object.values(remap));
     },
     [key, dispatch, onPaste],
   );
--- a/console/src/table/Toolbar.tsx
+++ b/console/src/table/Toolbar.tsx
@@ -149,9 +149,11 @@ const CellForm = ({ tableKey, cellKey }: CellFormProps): ReactElement | null => 
 
   const handleVariantChange = useCallback(
     (variant: Base.Cell.Variant) => {
-      if (cell == null) return;
-      const actions = buildVariantSwapActions([[cellKey, cell]], variant, theme);
-      if (actions.length > 0) dispatch({ key: tableKey, actions });
+      if (cell != null)
+        dispatch({
+          key: tableKey,
+          actions: buildVariantSwapActions([[cellKey, cell]], variant, theme),
+        });
     },
     [cell, cellKey, dispatch, tableKey, theme],
   );
@@ -270,7 +272,6 @@ const MultiCellForm = ({ tableKey, cellKeys }: MultiCellFormProps): ReactElement
           }),
         );
       }
-      if (actions.length === 0) return;
       dispatch({ key: tableKey, actions });
     },
     [cellsByKey, dispatch, tableKey],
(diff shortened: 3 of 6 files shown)


pluto/src/schematic/clipboard.ts
```diff
@@ -142,9 +142,8 @@ export const useClipboard = ({
         if (cfg != null)
           actions.push(schematic.setConfig({ key: newKey, config: cfg }));
       }
-      if (actions.length === 0) return;
       dispatch({ key, actions });
-      onPaste?.(Object.values(remap));
+      if (actions.length > 0) onPaste?.(Object.values(remap));
```
pjdotson, line 146: why do we have to guard here?

pluto/src/flux/dispatch.ts
```diff
@@ -110,6 +110,7 @@ export const createDispatch = <
       async (input: DispatchInput<Key, Action>): Promise<boolean> => {
         if (client == null) return false;
         const actions = Array.isArray(input.actions) ? input.actions : [input.actions];
+        if (actions.length === 0) return true;
```
pjdotson, line 113: why do we even need to check here? e.g. who calls this with an empty actions array

pjdotson (changes requested):

pluto/src/flux/dispatch.ts
emilbon99: For example `onNodesChange` in `Schematic.tsx` will call it with zero actions because certain node changes are selection only (i.e. session state). 

Since this is a duplicated pattern, I decided it was best to consolidate.

pluto/src/schematic/clipboard.ts
emilbon99: If nothing was pasted, then don't call `onPaste`.

pjdotson (approved):

PR #2440, opened 2026-06-04
SY-4316: Bugfix - Make Schematic Symbols Follow the Console Theme

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4316](https://linear.app/synnax/issue/SY-4316)

## Description

Schematic symbol and edge colors stopped responding to console light/dark theme changes after an earlier schematic refactor. A symbol placed in light mode kept a near-black stroke that became nearly invisible in dark mode, and the reverse going the other way.

This restores theme-responsive colors as a UI-only transformation. The color a user picks is still stored unchanged, so live editing across multiple consoles and browsers stays in sync. Each console independently derives a legible display color from the stored color and its own active theme, entirely in CSS, so there is no per-symbol theming hook on the render path.

The transform lives in one place: a `.pluto-symbol-colored` utility class that computes `--pluto-symbol-display` in OKLCH. Achromatic colors (near black, white, or grey) snap to the theme's legible extreme, while chromatic colors keep their hue and only clamp lightness into a band, blended by chroma so there is no hard edge. A companion `--pluto-symbol-contrast` provides legible text on a filled color. Every symbol primitive (svg shapes, div-border symbols, toggle states, gauge) and edge (the line plus the hydraulic, pneumatic, and data decorations) sets its source color as a CSS variable and carries the class, so they share the single definition.

Symbol default colors now use the `color.ZERO` follow-theme sentinel, matching edges, so a newly placed symbol adapts to the theme instead of freezing the color of whichever theme it was created in. The properties color picker shows the resolved default in its swatch while leaving the stored source unchanged.

This also fixes a separate pre-existing regression where a linked off-page reference no longer rendered its filled state; that class was dropped in the same reorganization and is reapplied.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR restores theme-responsive colors to schematic symbols and edges that broke during an earlier refactor. The fix is a CSS-only transform: a new `.pluto-symbol-colored` utility class computes a `--pluto-symbol-display` OKLCH color from the stored source `--pluto-symbol-color`, snapping achromatic values to the theme's legible extreme while only clamping lightness for chromatic colors.

- Every symbol primitive (SVG-based, div-based, gauge, toggle) and edge now sets `--pluto-symbol-color` and carries the `.pluto-symbol-colored` class; the single CSS rule in `theme.css` derives the display and contrast colors automatically per theme.
- All `defaultConfig` functions are updated to store `color.ZERO` instead of `t.colors.gray.l11`, letting each console independently resolve the correct default from its own active theme at render time.
- The off-page reference linked-fill state that was dropped in a prior reorganization is restored by wiring the `linked` boolean through the symbol chain.

Safe to merge — the change is a CSS-only, render-path-only transform that does not touch stored data or any server-side logic.

The color transform is entirely in CSS custom properties and never mutates the stored color value, so there is no risk of data corruption or cross-session inconsistency. The color.ZERO sentinel change in every defaultConfig is the only behavioral change to persisted values, and newly placed symbols correctly fall back to the theme default. New tests cover the ZERO sentinel, the explicit-color, and the CSS-variable bypass paths for both edges and node primitives.

| Filename | Overview |
|----------|----------|
| pluto/src/theming/theme.css | Adds theme-specific CSS custom properties for OKLCH lightness clamping and the `.pluto-symbol-colored` class that derives `--pluto-symbol-display` and `--pluto-symbol-contrast`; also adds `--pluto-gray-l11-rgb` fallback. |
| pluto/src/schematic/node/common/primitive/primitive.css | Migrates all `rgb(var(--pluto-symbol-color), opacity)` usages to `oklch(from var(--pluto-symbol-display) l c h / opacity%)` and adds `stroke: var(--pluto-symbol-display)` to all SVGs inside `.pluto-symbol-primitive`. |
| pluto/src/schematic/node/common/primitive/SVG.tsx | Removes inline fill/stroke attributes and the Theming hook; adds pluto-symbol-colored class and conditionally sets --pluto-symbol-color only when the color is non-zero. |
| pluto/src/schematic/edge/common/base/Base.tsx | Replaces direct stroke: color.cssString(stroke) with the --pluto-symbol-color variable approach; adds a bypass for non-parseable CSS-variable strings (connection-line previews). |
| pluto/src/schematic/node/common/create.ts | Switches createStatic, createToggle, and createDummyToggle default config from t.colors.gray.l11 to color.ZERO, making all standard symbol variants follow-theme by default. |
| pluto/src/schematic/node/general/offPageReference/Symbol.tsx | Restores the linked-fill state by passing linked={page != null && page.length > 0} to the Primitive, fixing the pre-existing regression where a linked reference no longer rendered its filled state. |
| pluto/src/schematic/node/common/form/Color.tsx | Color picker swatch now resolves color.ZERO to the current theme's gray.l11 for display, keeping stored value unchanged. |
| pluto/src/schematic/edge/data/spec.tsx | Wraps data edge rendering in a <g className='pluto-symbol-colored'> so decoration circles inherit --pluto-symbol-display without needing an explicit color prop. |
| pluto/src/schematic/node/general/value/Primitive.tsx | Removes Theming hook and JS contrast calculation; uses var(--pluto-symbol-display) for border/units background and var(--pluto-symbol-contrast) for units text. |
| pluto/src/schematic/node/general/gauge/Primitive.tsx | Adds pluto-symbol-colored to the gauge container div and conditionally sets --pluto-symbol-color; the value arc path now strokes var(--pluto-symbol-display) instead of a computed fallback. |
| pluto/src/theming/css.ts | Adds --pluto-gray-l11-rgb CSS custom property so the .pluto-symbol-colored rule can fall back to the theme's legible default when no source color is set. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["Symbol/Edge config\nstored color (source)"] --> B{color.isZero?}
    B -- yes --> C["--pluto-symbol-color NOT set\n(inherit from theme)"]
    B -- no --> D["--pluto-symbol-color = rgbString\n(inline style)"]
    C --> E[".pluto-symbol-colored class"]
    D --> E
    E --> F["--pluto-symbol-display =\noklch from rgb(var(--pluto-symbol-color,\n  var(--pluto-gray-l11-rgb)))"]
    G["Theme vars:\n--pluto-symbol-l-extreme\n--pluto-symbol-l-min/max\n--pluto-symbol-c-thresh"] --> F
    F --> H{chroma < c-thresh?}
    H -- yes --> I["Snap lightness to l-extreme\n(pure black/white)"]
    H -- no --> J["Clamp lightness to\n[l-min, l-max]"]
    I --> K["--pluto-symbol-display\n(final display color)"]
    J --> K
    K --> L["stroke / fill / borderColor\nin CSS or inline style"]
    K --> M["--pluto-symbol-contrast\nfor text on filled areas"]
```

<sub>Reviews (2): Last reviewed commit: ["\[formatting\]"](https://github.com/synnaxlabs/synnax/commit/78938ee838312bbe62c65ad06535ac719e96fb7a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35531281)</sub>


sy-nico:
@greptile, update your review.

pluto/src/schematic/edge/data/spec.tsx
```diff
@@ -18,29 +19,39 @@ import { NAME, VARIANT } from "@/schematic/edge/data/config";
...
 
-const DataSymbol = ({ color: colorVal, position }: SymbolProps): ReactElement => (
+const DataSymbol = ({ position }: SymbolProps): ReactElement => (
   <circle
     cx={position.x}
     cy={position.y}
     r={3}
     fill="var(--pluto-gray-l0)"
-    stroke={color.cssString(colorVal)}
+    stroke="var(--pluto-symbol-display)"
```
emilbon99: Do we need this? Can we do this with CSS inheritance?

pluto/src/schematic/edge/hydraulic/spec.tsx
```diff
@@ -19,23 +20,18 @@ const SYMBOL_INTERVAL = 40;
...
-}: SymbolProps): ReactElement => {
+const HydraulicSymbol = ({ position, direction }: SymbolProps): ReactElement => {
   const pos = { ...position };
   if (direction === "x") pos.y += SYMBOL_SIZE / 2;
   else pos.x += SYMBOL_SIZE / 2;
   return (
     <path
       d={`M0,0 L0,-${SYMBOL_SIZE} L${SYMBOL_SIZE},-${SYMBOL_SIZE}`}
-      stroke={color.cssString(colorVal)}
+      stroke="var(--pluto-symbol-display)"
```
emilbon99: Same comment as above.

pluto/src/schematic/edge/pneumatic/spec.tsx
```diff
@@ -38,20 +34,19 @@ const PneumaticSymbol = ({
     pointTwo = xy.translateY(pos, -10);
     rotate += 90;
   }
-  const stroke = color.cssString(colorVal);
   return (
     <>
       <path
         d={`M0,0 L0,-${SYMBOL_SIZE}`}
-        stroke={stroke}
+        stroke="var(--pluto-symbol-display)"
```
emilbon99: +1

pluto/src/schematic/edge/pneumatic/spec.tsx
```diff
@@ -38,20 +34,19 @@ const PneumaticSymbol = ({
...
         fill="none"
         strokeWidth={2}
         transform={`translate(${pos.x},${pos.y}) rotate(${rotate})`}
         strokeLinecap="round"
       />
       <path
         d={`M0,0 L0,-${SYMBOL_SIZE}`}
-        stroke={stroke}
+        stroke="var(--pluto-symbol-display)"
```
emilbon99: +2

pluto/src/schematic/node/common/create.ts
```diff
@@ -48,9 +48,9 @@ export const createStatic = <V extends string>({
     scale: z.number().optional(),
   });
   type Config = z.infer<typeof configZ>;
-  const defaultConfig = (t: Theming.Theme): Config => ({
+  const defaultConfig = (_t: Theming.Theme): Config => ({
```
emilbon99: Are there any remaining places where the theme is actually necessary? Would be good to be able to get rid of passing a theme here entirely.

pluto/src/schematic/node/general/button/Primitive.tsx
```diff
@@ -21,6 +23,26 @@ interface ButtonProps extends Omit<Config, "variant"> {
   onMouseUp?: MouseEventHandler<HTMLButtonElement>;
 }
 
+// The button drives its background, border, and text off the theme-transformed
+// symbol vars instead of the base button's concrete color, so it follows the
+// console theme like every other symbol. The color is not forwarded to the base
+// button, whose JS path would otherwise overwrite the contrast text color.
+const themedStyle = (value?: colors.Crude): CSSProperties => ({
```
emilbon99: This is going to be doing a lot of heap allocations. An entirely new object will be allocated every single time the button is re-rendered, which will be a lot. Is there a way to simplify this code? seems like you could define a lot of this directly inside CSS instead of JS.

pluto/src/schematic/node/general/circle/Primitive.tsx
```diff
@@ -62,12 +62,11 @@ export const Circle = ({
           id="4"
         />
       </Handle.Boundary>
-      <Primitive.SVG dimensions={{ width, height }}>
+      <Primitive.SVG dimensions={{ width, height }} color={colorVal}>
```
emilbon99: Should figure out a way to memoize width/height here and/or pass as a stable object.

emilbon99 (changes requested):
Overall implementation architecture is good, I think there's a lot of places where we're running excessive javascript that could be dramatically reduced and memoized properly.

pluto/src/schematic/node/common/create.ts
sy-nico: Removed it from the 11 `defaultConfigs` that no longer needed it. Left the signature since ~10 still use the theme (button primary, form-control grays, shape background colors, value staleness).

pluto/src/schematic/node/general/circle/Primitive.tsx
sy-nico: Memoized `{ width, height }` with useMemo so it's a stable reference now. Applied the same to the other primitives and folded it into Primitive.SVG so it covers every svg symbol in one place.

pluto/src/schematic/node/general/button/Primitive.tsx
sy-nico: Moved all the static bg/border/text vars into button.css (.pluto-symbol-button), so JS now only sets the one dynamic value, `--pluto-symbol-color`. That lone style object is memoized with useMemo, so nothing new is allocated per re-render.

pluto/src/schematic/edge/data/spec.tsx
```diff
@@ -18,29 +19,38 @@ import { NAME, VARIANT } from "@/schematic/edge/data/config";
...
-  </>
-));
+export const spec = Segmented.createSpec(
+  VARIANT,
+  NAME,
+  ({ points, color: colorVal }) => (
+    <g
+      className={CSS.B("symbol-colored")}
+      style={{
+        [CSS.var("symbol-color")]: color.isZero(colorVal)
```
emilbon99: This expression seems to be duplicated in multiple places.

pluto/src/schematic/edge/hydraulic/spec.tsx
```diff
@@ -44,18 +39,24 @@ const HydraulicSymbol = ({
...
-  </>
-));
+export const spec = Segmented.createSpec(
+  VARIANT,
+  NAME,
+  ({ points, color: colorVal }) => (
+    <g
+      className={CSS.B("symbol-colored")}
+      style={{
+        [CSS.var("symbol-color")]: color.isZero(colorVal)
```
emilbon99: Repeated same pattern

pluto/src/schematic/edge/pneumatic/spec.tsx
```diff
@@ -61,18 +54,24 @@ const PneumaticSymbol = ({
...
-  </>
-));
+export const spec = Segmented.createSpec(
+  VARIANT,
+  NAME,
+  ({ points, color: colorVal }) => (
+    <g
+      className={CSS.B("symbol-colored")}
+      style={{
+        [CSS.var("symbol-color")]: color.isZero(colorVal)
```
emilbon99: Repeated same pattern

pluto/src/schematic/symbolColor.ts
```diff
@@ -0,0 +1,18 @@
...
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { color } from "@synnaxlabs/x";
+
+/// symbolColorVar returns the value for the --pluto-symbol-color custom property: the
+/// rgba channels of a symbol's color, or undefined for an unset or ZERO color so the
+/// display var falls back to the theme default.
+export const symbolColorVar = (c?: color.Crude): string | undefined =>
```
emilbon99, line 15: Seems like this is the helper?

emilbon99 (changes requested):

pluto/src/schematic/symbolColor.ts
sy-nico: Yes, moved this to a location where it could be accessed by the nodes and edges.

emilbon99 (approved):

PR #2421, opened 2026-06-03
SY-4297: Arc Params Refactor (Part 1) - Oracle Changes

sy-nico:
# Remove Config from Arc Schemas
## Linear Issue
[SY-4297](https://linear.app/synnax/issue/SY-4297)
## What this is
Phase 1 of the Arc function-call unification (RFC 0039), stacked on the base branch. It establishes the new generated type shape that the rest of the stack is written against: a single `Inputs` parameter list, with `Config` gone everywhere.
## The change
- **Schema (hand-authored):** remove `config` from `FunctionProperties` in `schemas/arc/types.oracle`, and from `Function` and `Node` in `schemas/arc/ir.oracle`.
- **Codegen (generated):** `oracle sync` regenerates every Go/C++/TS binding, codec, protobuf, and codec test fixture. `Param` now carries `Name` / `Type` / `Value` only; no `Config` field survives in `FunctionProperties`, `Function`, or `Node` across any of the three languages.
## Heads-up
**This branch does not build on its own, by design.** Removing `Config` from the generated types breaks every `.Config` reader until later phases convert them. The first green build is the final phase; intermediate PRs are reviewed on their diff, not on a green CI run (see plan §5).
## Review note
The only hand-written diff is the two `.oracle` schema files. Everything else is regenerated output; review the schema edit, then spot-check that the regenerated `Param` shape and the absence of `Config` look right.

<h3>Greptile Summary</h3>

This PR removes the `config` field from `FunctionProperties`, `Function`, and `Node` across the Arc schema and all generated bindings (Go, C++, TypeScript, protobuf, codec tests). It is Phase 1 of the Arc function-call unification (RFC 0039), intentionally breaking the build as intermediate work.

- **Hand-authored changes** are confined to two oracle schema files (`schemas/arc/types.oracle`, `schemas/arc/ir.oracle`), which drop the `config` field and update associated doc-strings.
- **Generated output** (`oracle sync`) consistently propagates the removal through every Go/C++/TS type, codec, protobuf definition, and codec test fixture; proto field numbers are renumbered accordingly, which is safe here because persistence uses name-keyed msgpack and the proto codec is same-version only.
- The graph-level `graph.Node.Config` (`msgpack.EncodedJSON`) is unrelated to the removed IR/types `Config` (`types.Params`) and is correctly preserved in tests and runtime code.

<h3>Confidence Score: 5/5</h3>

[...]

PR #2448, opened 2026-06-05
SY-4297: Arc Params Refactor (Part 3) - Collapse the Analyzer Call Paths

sy-nico:
# Issue Pull Request

## Linear Issue

## Description

Part 3 of the Arc Function-Call Unification stack ([RFC 0039](docs/tech/rfc/0039-260513-arc-function-call-unification.md)). Stacked on **Part 2 (#2447)**; the schema/codegen that removed the `Config` field landed in **Part 1 (#2421)**.

This is the analyzer collapse; the core of the refactor and its design-evaluation gate (RFC §4.2, §4.4).

**Unified call validation.** A new `analyzer/call` package provides `call.Analyze`, one routine that validates a call's arguments against the unified `Inputs` list for both surface forms. The two parallel validators are removed: `validateFunctionCall` (parens) and `validateFuncConfig` (brace) now adapt their arguments to `[]symbol.Argument` via `inputArguments` and delegate to `call.Analyze`, which also runs the unified `AnalyzeArguments` hook (replacing the old `AnalyzeCall` + `AnalyzeFlowConfig` pair).

**Explicit trigger consult.** The flow analyzer's `upstreamIsTrigger` heuristic and its three-branch upstream-handling block are replaced by `consultTrigger`, which keys off the symbol's explicit `Trigger` (added in Part 2) instead of guessing from `Exec` + `len(Config)`. User-defined functions are assigned `Trigger` in `CollectDeclarations` (the first parens-block param, else `TriggerOnly`).

**Cleanup.** The now-dangling analyzer `.Config` readers are dropped (the substitution
loop in `resolve.go`, the function-type clone in `constraints.go`).

### Reviewer notes

- **Does not compile yet, by design.** Downstream consumers (ir, graph, text, compiler) still read `.Config` and are converted in Parts 4–8; first green build is Part 9. CI red is expected — review on the diff. `gofmt -s` passes and gopls reports the changed files clean; `golangci-lint` is blocked only by the `ir` import.
- **No tests in this part** — the fixture sweep and new analyzer tests are Part 7.
- **Behavior decisions to scrutinize at the gate:**
  - Type-checking is unified on `atypes.Check`, stricter than the old parens `AddCompatible` (drops implicit concrete numeric promotion; literals still adapt).
  - Calls match the full unified `Inputs` (config-first); coverage is per-param.
  - A channel argument is dereferenced only when the target param is not itself channel-typed (reconciles the old parens-deref vs brace-channel-param split).
  - New diagnostic: a param bound by both a call-site argument and an upstream wire.
  - Dropped the `>1 parameter` func-to-func error and the standalone "use in a flow statement" hint (both subsumed by coverage + `consultTrigger`).

## Basic Readiness

<h3>Greptile Summary</h3>

Part 3 of the Arc function-call unification stack: collapses the two parallel call validators (`validateFunctionCall` for parens and `validateFuncConfig` for brace) into a single `call.Analyze` routine in the new `analyzer/call` package, and replaces the `upstreamIsTrigger` heuristic in the flow analyzer with an explicit `consultTrigger` dispatch keyed off the symbol's `Trigger` field added in Part 2.

- **`analyzer/call`**: New `Analyze[T]` function handles both surface forms; `matchParam` resolves named or positional args, `checkArgType` dereferences channel arguments only when the target param is not channel-typed, and `externallySatisfied` suppresses \"missing required argument\" for wire-fed trigger params.
- **`flow/flow.go`**: `parseFunction` and `analyzeRoutingTargetWithParam` now adapt brace-block values via `inputArguments` and delegate to `call.Analyze`; upstream wiring is validated in `consultTrigger` which distinguishes TriggerOnly (pure activation, no type check) from TriggerInput (value-typed, reports conflicts with call-site args).
- **`function/function.go` + `constraints/constraints.go` + `resolve.go`**: Config params are merged into `Inputs`, `Config` substitution loop removed, and user-defined functions gain an explicit `Trigger` set in `CollectDeclarations`.

The PR does not compile by design; the changed files are analyzer logic only and the diff is internally consistent with the RFC.

All changed files follow the same contract: brace-block and parens-block args are adapted into []symbol.Argument and routed through the single call.Analyze path. The consultTrigger replacement for upstreamIsTrigger reads explicit symbol metadata instead of inferring from Exec+len(Config). No logic regressions are visible in the changed code.

arc/go/analyzer/function/function.go has two function-level doc comments that still reference the old (config, inputs, outputs) signature pattern.

| Filename | Overview |
|----------|----------|
| arc/go/analyzer/call/call.go | New unified Analyze function; clean design with well-scoped matchParam/checkArgType helpers and correct externallySatisfied variadic for trigger params. |
| arc/go/analyzer/flow/flow.go | Replaces upstreamIsTrigger heuristic with consultTrigger; extracting resolveUpstreamType and hasRoutingTableBetween into helpers significantly improves clarity with no logic regressions visible. |
| arc/go/analyzer/function/function.go | Package-level doc comment updated; two function-level doc comments still say (config, inputs, outputs) which is stale after the Config to Inputs merge. |
| arc/go/analyzer/expression/expression.go | validateFunctionCall removed; delegates to call.Analyze via inputArguments adapter, unifying the parens-call path with the brace path cleanly. |
| arc/go/analyzer/resolve.go | Config substitution loop removed from ResolveNodeTypes; doc comment updated but has a minor awkward line break. |
| arc/go/analyzer/flow/expression.go | Single-line fix: constant value param now appended to Inputs instead of the removed Config field. |
| arc/go/analyzer/constraints/constraints.go | One-line removal of Config substitution from applySubstitutions; correct given Config is gone from the function type. |
| docs/tech/rfc/0039-260513-arc-function-call-unification.md | Appendix A added with Phase 3 running notes on brace/parens syntax and grammar terminology; informational only. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["Call site: parens form"] -->|inputArguments| C["[]symbol.Argument"]
    B["Call site: brace form"] -->|inputArguments| C
    C --> D["call.Analyze"]
    D --> E["matchParam: by name or index"]
    D --> F["checkArgType: UnwrapChan unless param is chan-typed"]
    D --> G["Coverage check: missing required params"]
    D --> H["ArgumentsHook: optional symbol hook"]
    G -->|externallySatisfied| I["Trigger param fed by upstream wire"]
    J["parseFunction"] -->|"prevNode != nil AND Trigger.Target set"| I
    J --> D
    K["resolveUpstreamType"] -->|"identifier / expr / func"| L["upstreamType"]
    L --> M["consultTrigger"]
    M -->|"target is empty"| N["TriggerOnly: pure activation, skip type check"]
    M -->|"target in suppliedAtCallSite"| O["Error: param bound by both call-site arg and wire"]
    M -->|"default"| P["atypes.Check: targetParam.Type vs upstreamType"]
```

1. `arc/go/analyzer/function/function.go`, line 10-15 ([link](https://github.com/synnaxlabs/synnax/blob/16601ddeab177fc140beff2c0b7d04effe2bd8b2/arc/go/analyzer/function/function.go#L10-L15)) 

   <a href="#">P2</a> The package-level doc comment still describes a three-category parameter model (Config / Inputs / Outputs) that no longer reflects the type system after this refactor. Config params are now merged into `Inputs`, so the listed "Config" bullet describes a concept that has been removed. A reader new to this code would be misled into expecting a separate `Config` field on the function type.

   **Rule Used:** Flag misleading doc comments that promise fallback... ([source](https://app.greptile.com/synnax/github/synnaxlabs/synnax/-/custom-context?memory=16df2708-f8e1-4fe6-a509-cee5a869fa6e))
   
</details>

<sub>Reviews (2): Last reviewed commit: ["SY-4297: Fix function doc string"](https://github.com/synnaxlabs/synnax/commit/8f45ece687e35ec2db6fadf187831680435e2757) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35754638)</sub>

### DIFF
--- a/arc/go/analyzer/expression/call.go
+++ b/arc/go/analyzer/expression/call.go
@@ -7,12 +7,7 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-// Package call centralizes argument validation shared by the expression analyzer
-// (parens form, positional args) and the flow analyzer (brace form, named or
-// anonymous input values). Both surface forms erase to a single Inputs list, so one
-// routine validates them; the only per-argument difference is whether it binds by
-// name (Argument.Name != "") or by position.
-package call
+package expression
 
 import (
 	"fmt"
@@ -28,11 +23,11 @@ import (
 	"github.com/synnaxlabs/x/set"
 )
 
-// Analyze validates a call's arguments against fnType.Inputs (count, type, name, and
+// AnalyzeCall validates a call's arguments against fnType.Inputs (count, type, name, and
 // required-param coverage) and runs the symbol's optional hook. args bind by name or
 // position. externallySatisfied names params filled by something other than a
 // call-site argument (in flow, the wire-fed trigger), so they aren't reported missing.
-func Analyze[T antlr.ParserRuleContext](
+func AnalyzeCall[T antlr.ParserRuleContext](
 	ctx acontext.Context[T],
 	fnName string,
 	fnType types.Type,
--- a/arc/go/analyzer/function/function.go
+++ b/arc/go/analyzer/function/function.go
@@ -9,16 +9,14 @@
 
 // Package function implements semantic analysis for Arc function declarations.
 //
-// Functions in Arc have three parameter categories:
-//   - Config: Compile-time configuration (in curly braces)
-//   - Inputs: Runtime parameters (in parentheses)
-//   - Outputs: Return values (after the parameter list)
+// A function has a list of inputs and a list of outputs. One input may be the
+// trigger: the param an upstream wire feeds in flow context. The analyzer collects
+// the declared params into Inputs and records the trigger.
 //
 // The analyzer validates:
 //   - Function names are unique
-//   - Parameter names are unique within their category
-//   - Input types are valid
-//   - Output types are valid
+//   - Parameter names are unique
+//   - Input and output types are valid
 //   - Optional parameters come after required parameters
 //   - Functions with return types return on all code paths
 //   - Named outputs are assigned in the function body
--- a/arc/go/analyzer/function/function.go
+++ b/arc/go/analyzer/function/function.go
@@ -47,22 +47,29 @@ func CollectDeclarations(ctx acontext.Context[parser.IProgramContext]) {
 		if fn := item.FunctionDeclaration(); fn != nil {
 			name := fn.IDENTIFIER().GetText()
 
-			// Collect signature (config, inputs, outputs) without adding params to scope
-			var config, inputs, outputs types.Params
-			collectConfig(ctx, fn.ConfigBlock(), &config)
+			// The brace block holds inputs and the parens block holds the trigger;
+			// they concatenate into one Inputs list, inputs first.
+			var inputs, outputs types.Params
+			collectConfig(ctx, fn.ConfigBlock(), &inputs)
+			parensStart := len(inputs)
 			collectInputs(acontext.Child(ctx, fn.InputList()), &inputs)
 			collectOutputs(ctx, fn.OutputType(), &outputs)
+
+			trigger := symbol.TriggerOnly
+			if len(inputs) > parensStart {
+				trigger = symbol.TriggerInput(inputs[parensStart].Name)
+			}
 
 			if _, err := ctx.Scope.Add(ctx, symbol.Symbol{
 				Name: name,
 				Kind: symbol.KindFunction,
 				Exec: symbol.ExecBoth,
 				Type: types.Function(types.FunctionProperties{
-					Config:  config,
 					Inputs:  inputs,
 					Outputs: outputs,
 				}),
-				AST: fn,
+				Trigger: trigger,
+				AST:     fn,
 			}); err != nil {
 				ctx.Diagnostics.Add(diagnostics.Error(err, fn))
 			}
@@ -388,7 +395,7 @@ func IfStmtAlwaysReturns(ifStmt parser.IIfStatementContext) bool {
 }
 
 // addConfigToScope adds config parameters to the function's scope.
-// The config types are already collected in fn.Type.Config by CollectDeclarations.
+// The config types are already collected in fn.Type.Inputs by CollectDeclarations.
 func addConfigToScope[T antlr.ParserRuleContext](
 	ctx acontext.Context[T],
 	configBlock parser.IConfigBlockContext,
--- a/arc/go/analyzer/call/call.go
+++ b/arc/go/analyzer/call/call.go
@@ -1,0 +1,122 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+// Package call centralizes argument validation shared by the expression analyzer
+// (parens form, positional args) and the flow analyzer (brace form, named or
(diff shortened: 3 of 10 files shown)


arc/go/analyzer/expression/call.go
```diff
@@ -0,0 +1,122 @@
...
+				WithCode(codes.FuncArgCount).WithNote("signature: " + signature))
+		}
+		return types.Param{}, false
+	}
+	return fnType.Inputs[arg.Index], true
+}
+
+// checkArgType checks arg's type against param, dereferencing a channel argument to
+// its value type unless param is itself channel-typed.
+func checkArgType[T antlr.ParserRuleContext](
```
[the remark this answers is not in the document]
sy-nico: Intentional. The loop emits exactly one `too many arguments: expected at most N` on the first out-of-range index, so the over-long call is flagged once. Args past it return `(Param{}, false)` and are skipped because there's no parameter to bind or type-check them against, and re-emitting the same count error per extra arg would just be duplicate noise. The old `validateFunctionCall` had the same effect (early return at the count check). The single count diagnostic already covers everything beyond N, so the extras aren't treated as valid.

sy-nico:
@greptile, update your review. I fixed the stale doc comment and replied to your non-issue comment.

arc/go/analyzer/function/function.go
```diff
@@ -47,22 +45,29 @@ func CollectDeclarations(ctx acontext.Context[parser.IProgramContext]) {
 		if fn := item.FunctionDeclaration(); fn != nil {
 			name := fn.IDENTIFIER().GetText()
 
-			// Collect signature (config, inputs, outputs) without adding params to scope
-			var config, inputs, outputs types.Params
-			collectConfig(ctx, fn.ConfigBlock(), &config)
+			// The brace block holds inputs and the parens block holds the trigger;
+			// they concatenate into one Inputs list, inputs first.
```
sy-nico, line 49: I’m considering changing the order and make Trigger first. That’s how it reads semantically in a flow.

arc/go/analyzer/expression/call.go
emilbon99, line 1: Why does this deserve a new package? Are there things that can be called that are not functions? Why isn't this in the `function` package?  If it's a circular dep issue, then why doesn't this still live in the `expression` package? Don't really see an arugment for a separate package.

emilbon99: Also, where are the unit tests for this?

emilbon99 (changes requested):

arc/go/analyzer/expression/call.go
sy-nico: It can't go in `function` because `expression` has to call this, and `function` already imports it transitively (`function -> statement -> expression`). Putting the logic in function means that `expression` would need to import `function` and close the loop (`function -> statement -> expression -> function`).

As for why not in expression, my reasoning was that a small shared package that `expression` and `flow` could import felt reasonable. To me, it didn't make sense for `expression` to own the primitive and have `flow` reach into it. Extracting the leaf seemed cleaner and consistent with how `analyzer/` is split (codes, context, statement, constraints, etc). I don't feel strongly about it though.

On things that can be called: yes, casts like `float64(x)` and builtins like `len` are call-shaped, but those parse and resolve on their own paths (typeCast in the grammar, and len is special-cased at the call site). `call.Analyze` is only reached once the callee resolves to a function symbol (`KindFunction`), so the thing it validates is always a function. It's unifying the two call syntaxes for functions (parens positional and brace named), not abstracting over different callable kinds.

Since the point is unification, I see the logic for folding it into expression though. I'll make the change.

sy-nico: I deferred tests to Part 7. However, since this is new code I can add coverage for it here without any issues.

emilbon99: Ok, LGTM

emilbon99 (approved):

PR #2451, opened 2026-06-05
SY-4238: Core Structure Refactor RFC

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4238](https://linear.app/synnax/issue/SY-4238)

## Description

Adds RFC 0041, which restructures `core/pkg` along five axes that have accumulated inconsistency as the Core has grown:

1. **Layer realignment** — `ontology`, `group`, `search`, and `signals` move from distribution to service. Distribution keeps only the topology-aware core (channels, frames, lease routing). The blocking dependency (`distribution/channel` and `distribution/node` import the substrate) is removed by hoisting ontology/group/search/CDC wiring into service-layer wrappers (`service/channel`, a new `service/node`).
2. **One type per entity, with resolved fields** — collapses the duplicate service/API types for `range`/`task`/`device`/`rack` to a single service-layer type. Resolved fields (`Labels`, `Parent`, `Status`) are declared in the schema, excluded from storage by Oracle, and filled by the API layer.
3. **Uniform versioned type layout** — standardizes on `<resource>/types/vN/` with one Go package per version (current included). Top-level `types/` re-exports current and owns the version-dispatch `Decode`. Supersedes the `migrations/vN/` layout in RFC 0033/0034.
4. **Peek-based import** — replaces RFC 0034's decode-to-`map[string]any` with a peek of `{version, type}` followed by a single typed decode into the frozen `types/vN/` struct. A unified `imex.Envelope` serves both import and export paths.
5. **Validation chokepoint** — generated `Validate()` is enforced once, in `gorp.Writer.set`, so no write path (service, import, CDC) can store unvalidated data. Oracle's `validate` domain gains numeric bounds and enum-variant enforcement.

Implementation is sequenced into phases starting with the pure-Go layer realignment, which unblocks the rest.

## Basic Readiness

<h3>Greptile Summary</h3>

RFC 0041 proposes a comprehensive structural refactor of `core/pkg` across five axes: relocating non-topology-aware packages (`ontology`, `group`, `search`, `signals`) from distribution to the service layer; collapsing duplicate service/API types into a single service type with schema-declared resolved fields; standardising on a `types/vN/` per-version layout; replacing the decode-to-`map[string]any` import path with a peek-based two-stage decode; and adding a `Validate()` chokepoint at `gorp.Writer.set`. Phase 5 now correctly matches §4.5.3/§5.4 (boot abort on failure, no silent drop).

- **§4.3.0 file tree vs §4.3.2 contradiction**: the tree lists both `Migrate` and `MigrateFromLegacy` under the first modern version's `migrate.go`, while §4.3.2 explicitly states the first modern version has no ordinary `Migrate`; an implementer following the tree alone will add an invalid function referencing a non-existent modern predecessor.
- **§4.5.0 CDC write-path claim**: CDC (`signals`) publishes Gorp changes as telemetry via `framer` (Cesium), not via KV writes, so listing CDC alongside service writers as a path that \"funnels through `Writer.set`\" appears imprecise; the RFC should clarify whether CDC is covered by the chokepoint directly or only indirectly (because its inputs were already validated on write).

The RFC is a draft design document — safe to iterate on, but it contains an internal contradiction that would produce incorrect implementation code.

The file tree in §4.3.0 shows both `Migrate` and `MigrateFromLegacy` under the first modern version's `migrate.go`, while §4.3.2 explicitly states the first modern version has no ordinary `Migrate`. An implementer who follows the tree notation will write an uncompilable or silently wrong migration function. This contradiction must be resolved before Phase 2 implementation begins.

docs/tech/rfc/0041-260527-core-structure-refactor.md needs the §4.3.0 file tree corrected to remove `Migrate` from the first-modern-version block before it can serve as an authoritative implementation guide.

| Filename | Overview |
|----------|----------|
| docs/tech/rfc/0041-260527-core-structure-refactor.md | New RFC 0041 specifying a five-axis Core restructure; one internal contradiction between the §4.3.0 file tree and §4.3.2's statement about first-modern-version `migrate.go`, and an imprecise claim in §4.5.0 about CDC traversing the Gorp write seam. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph After["After RFC 0041"]
        S[storage] --> D["distribution\nchannel keys + Cesium lifecycle\nframer · node · proxy · transport"]
        D --> SVC["service\nontology · group · search · signals\nservice/channel · service/node\nranger · task · device · rack"]
        SVC --> API[api]
        API --> T[transport]
    end

    subgraph Import["Peek-Based Import Flow"]
        ENV["Envelope\nunmarshal peek"] --> GUARD{v > LatestVersion?}
        GUARD -- yes --> ERR[ErrUnsupportedVersion]
        GUARD -- no --> DISPATCH["types/decode.go Dispatch"]
        DISPATCH -- "v <= MaxVersion" --> LEG["legacy.Decode + MigrateFromLegacy"]
        DISPATCH -- modern --> MODERN["imex.Decode T + walkFromVN"]
        LEG --> CURRENT[Current Resource]
        MODERN --> CURRENT
        CURRENT --> WRITER["gorp.Writer.set Validate + encode + store"]
    end
```

<sub>Reviews (3): Last reviewed commit: ["fix formatting"](https://github.com/synnaxlabs/synnax/commit/9e64b4bb6a4c52f952cea1a2ede5b3b560db0e68) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35922308)</sub>


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

docs/tech/rfc/0041-260527-core-structure-refactor.md
```diff
@@ -0,0 +1,994 @@
...
+```
+// schemas/ranger.oracle (illustrative syntax)
+struct Range {
+  key       uuid
+  name      string
+  time_range telem.TimeRange
+  color     string
+
+  labels []label.Label {
+    domain resolved { from label via ontology }
```
emilbon99: Should update this to use the actual oracle `@` syntax. How carefully have you thought about the design and syntax of this `from label via ontology` thing? My take is that this sort of relationship management piece should be out of scope for this RFC and initial implementations. Let's do the other refactors first and consider relationship management in a separate RFC, as it's very complex.  For now just use the `orc` encoding skip rules and then we come back to this.

docs/tech/rfc/0041-260527-core-structure-refactor.md
```diff
@@ -0,0 +1,994 @@
...
+distribution package imports `ontology`, `group`, `search`, or `signals`, so they
+relocate freely to the service layer.
+
+### 4.1.2 - Wiring Order
+
+`distribution/layer.go` sheds the four packages from its aggregate; `service`'s layer
+opens them in dependency order:
+
+```
+ontology  →  search  →  group  →  service/node  →  service/channel  →  signals  →  (entities)
```
emilbon99: So `signals` will now depend on `service/channel` instead of `distribution/channel`? If so, can you just add a sentence to this RFC explaining how we wire in the channel signals themselves (`sy_channel_set/delete`)? Separate package? `service/channel/signals?`

docs/tech/rfc/0041-260527-core-structure-refactor.md
```diff
@@ -0,0 +1,994 @@
...
+`Include<Field>` flags. Resolution is **batched across the result set** — one labels
+query, one status query, one parent traversal per page — not one round trip per entity.
+Oracle generates the batched resolver from the `resolved` domain (§4.6.1).
+
+The change from today is structural: one service-layer type carries the resolved fields
+as zero values, instead of a duplicate API type embedding the service type and adding
+them. The write path is unaffected — resolved fields are excluded from storage and the
+validation seam (§4.5) ignores them.
+
+## 4.3 - The `types/vN/` Layout
```
emilbon99, line 366: Have you answered the question of the incompatibility of the packages tied to the core version vs. the exported version numbers? Are we going with `v56.x` or are we going with incrementing versions? If we're using a hybrid how does that fit into the fodler naming convention?

docs/tech/rfc/0041-260527-core-structure-refactor.md
```diff
@@ -0,0 +1,994 @@
...
+four entities share one pattern.
+
+### 4.2.1 - Storage Exclusion
+
+Oracle excludes a `resolved` field from the generated `EncodeOrc`/`DecodeOrc` codec, so
+it is never persisted and never read back from storage. The same field _is_ serialized
+in the API/transport (JSON/proto) output, because clients need it. This solves the
+persisted-vs-derived split named in RFC 0026 §1.1.12 at the schema level.
+
+### 4.2.2 - Resolution Stays in the API Layer
```
emilbon99: What drove this decision? Circular dependencies? Reducing code churn? 

IMO resolution should eventually go in the service layer. This is also tied to the `from label via ontology` syntax you introduced up there. My take is that eventually relationship management is something that should be oracle handled entirely as well. 

The thought is: you define the relationship structure you want in oracle and it generates code to do the resolution/retrieval and it manages integrity (things like cascading deletes, etc.).

docs/tech/rfc/0041-260527-core-structure-refactor.md
```diff
@@ -0,0 +1,994 @@
...
+— including current — is its own package beneath it. Only the exported surface is listed
+below; unexported helpers (the private `importer`/`exporter` structs in `imex.go`, the
+`validate` helper on `Writer`, the change translators in `ontology.go`, etc.) live in
+the same files but are implementation detail and not part of the canonical contract.
+
+```
+core/pkg/service/<resource>/
+├── resource.go              # public surface — re-exports the current version
+│   ├── type Key = types.Key
+│   ├── type Resource = types.Resource
```
emilbon99: Are we proposing renaming every thing to `Resource` i.e instead of `channel.Channel` it's `channel.Resource`? I can sort of infer the context that this is not the case, but would appreciate clarity.

docs/tech/rfc/0041-260527-core-structure-refactor.md
```diff
@@ -0,0 +1,994 @@
...
+        │                            # — FIRST MODERN ONLY; absent on every other vN.
+        └── helpers.go               #  hand-written method receivers on Resource
+                                     #  (e.g. func (Resource) OntologyID() ontology.ID;
+                                     #   ontology.go's free OntologyID(Key) wraps it).
+                                     #  Oracle copies this file forward at each version bump (§4.6.0)
+```
+
+A few file-level rules fall out of this layout:
+
+**`resource.go` is intentionally tiny.** Its only job is to be the import surface for
```
emilbon99: Is this a literal `resource.go` or a template? i.e. `resource.go` is actually `channel.go` in the channel package? 

If not, should it be called `type`?

docs/tech/rfc/0041-260527-core-structure-refactor.md
```diff
@@ -0,0 +1,994 @@
...
+`Writer.set`, this is the single enforcement point — no path can store an entry without
+validating it. Per-type logic lives in the generated `Validate`; the call site is
+centralized so it cannot be forgotten.
+
+### 4.5.1 - Extended Constraints
+
+Oracle's `validate` domain (RFC 0027 §3.6) is extended so the generated `Validate`
+method covers the gaps from §2.4:
+
+- **Numeric bounds**: `domain validate { min 1; max 64 }` on numeric fields, generating
```
emilbon99: Should update this stuff to use actual oracle syntax.

docs/tech/rfc/0041-260527-core-structure-refactor.md
```diff
@@ -0,0 +1,994 @@
...
+
+Tests are co-located but unlisted above — `<resource>_test.go`,
+`<resource>_suite_test.go`, `codec_gen_test.go`, `retrieve_test.go`, `writer_test.go`,
+`migration_test.go`. These sit next to the file they test.
+
+Version bumps are a freeze operation owned by Oracle — see §4.6.0 for the steps. This
+supersedes RFC 0033 §4.3.0's rule that the current type lives in the service package;
+per RFC 0033 §3.6, each `vN/` still imports nothing from the parent.
+
+### 4.3.1 - Each Version Is Self-Contained
```
emilbon99, line 556: What about dependencies across packages i.e. nested fields? Will those all be hoisted into the parent? This is a crucial question that versioning along with the core cleanly solved that disparate version i.e. everything `v55` was guaranteed to be compatible.

This is a non-problem for resolved fields because they will be queried dynamically, but what about for fields that aren't? Take for example, schematics dependencies on `spatial.oracle`? What if we suddely decide that `spatial.Direction` is now `horizontal` and `vertical` as opposed to `x` and `y`? 

How would your system detect that migration is necessary? By versioning using the core you easily solve this problem cause you know exactly which data structures are compatible with which.

Under your design `v5` schematic is ONLY compatible with `v3` spatial? How are you tracking those dependency chains and generating migrations appropriately?

emilbon99 (changes requested):
Overall looks good. Collection of higher level thoughts:

1. Stored migration dependencies haven't really been solved. For example a `v5` schematic could depend on a `v3` `spatial` package in x? How are we tracking that? 
2. What are we doing with the oracle snapshots folder? 
3. We should eventually get rid of the entire `ontology.Resource` type and *not* allow data retrieval through the ontology. The `ontology` is only for managing relationships, and my take is that long term it should be consolidated into gorp.
4. Your resolved fields design seems immature and should be out of scope for this RFC. I saw this in the opened questions, but will leave this point here. 
5. I think implementing `zyn` was a mistake and it should be completely removed from the cdoebase and not used.

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: updated this to use the existing `channel/signals` packages that already exist

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: no. updated to make more clear

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: The decision here is mostly with eliminating stuff like the separate api range and
service range types. For the initial changes I wanted to reduce code churn as most
resolution for retrievals is done in the API layer right now. In the future (by the time
we would generate the code via Oracle) it should probably belong in the service layer,
but that has implications for what the `Retrieve` type API looks like that I didn't want
to touch right now.

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: see comment above, using the brackets for the convention. Wanted to avoid `type` because
we have the `types` folder.

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: haven't thought too carefully, deferring those questions til later

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: I am not completely sure of the solution, but the current idea is to remove the v56.x
versions and do something like this:

schemas
  - schematic
      - v0.oracle
      - v1.oracle

where there is some type of `extends` syntax to reduce duplicated code, and when
spatial.Direction changes, you have to have a new schematic vN.oracle.

This gives us a few things:
- matching exported version numbers and folder / oracle schema numbers.
- oracle can generated code based off of importing spatial/types/v0 or spatial/types/v1
  for schematic v3 and schematic v4
- For the exported JSON config files, a v3 schematic has to use "x" | "y" and a v4 has to use
  "horizontal" | "vertical".
- It makes import / export + migration code better because a v3 schematic has to have
  one of the spatial enums that was valid FOR THAT VERSION.

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: yes, see other comment. The v5 schematic should ONLY be compatible with v3 spatial, that
way we know how to parse files in ImEx (e.g. what would schematic.direction be, we don't
need to store the spatial node with a version.

I think we will need to either edit `oracle sync` to regen new schemas after modifying
something like `spatial.oracle`, do it manually for the moment, or add another `oracle
X` command. This means
we probably end up deleting the `snapshots` folder in favor of first-class `v1.oracle`,
`v2.oracle` files.

This also resolves stuff like how does auto generated code know whether to import
spatial/types/v54 or spatial/types/v55? we probably will need to explicitly declare a
version in the Oracle schemas somehow, or schema files numbers.

pjdotson:
> Overall looks good. Collection of higher level thoughts:
> 
> 1. Stored migration dependencies haven't really been solved. For example a `v5` schematic could depend on a `v3` `spatial` package in x? How are we tracking that?

Added some comments and updated the RFC for this. Have a fairly strong belief that we should move to per-schema versions to help with things like import paths in the `types/vN` folders and not accumulating huge `snapshots` folder for what is effectively dead / not needed code. I _think_ it might make it easier to auto-generate migrations as well, not sure.

> 2. What are we doing with the oracle snapshots folder?

Probably end up deleting this

> 3. We should eventually get rid of the entire `ontology.Resource` type and _not_ allow data retrieval through the ontology. The `ontology` is only for managing relationships, and my take is that long term it should be consolidated into gorp.

Agreed.

> 4. Your resolved fields design seems immature and should be out of scope for this RFC. I saw this in the opened questions, but will leave this point here.

Was immature and removed some stuff on syntax.

> 5. I think implementing `zyn` was a mistake and it should be completely removed from the cdoebase and not used.

Yeah. I think Zyn does two things:

1) parse to a shape

should be handled either by the Go type system or by the codecs converting binary to a specific type

2) validates values are within certain bounds

Should be handled by something like `Validate()` methods directly on strongly-typed objects.

We also may want to end up rethinking how we communicate validation-errors better across the whole stack.

docs/tech/rfc/0041-260527-core-structure-refactor.md
emilbon99: One of the things to be wary of with a folder structure like this is the cognitive overload of having multiple versions of things all dumped in the same location. For me personally, my working area is so much better if the folder and/or set of files i'm in deals with the current version only. All of the previous version and migration logic lives elsewhere, and your brain is free to focus on the core business logic.

This is one of the reasons why instead of a data structure -> version hierarchy I opted for a version -> data structure hierarchy i.e. `v1/schematic.oracle` vs `schematic/v1.oracle`. I have all of the current versions of things in one place without needing to mess around or deal with tons of different version files intermingled with my current ones.

You're welcome to take. whatever path you like, but my brain would love you if you did a version -> structure hierarchy over a structure -> version.

docs/tech/rfc/0041-260527-core-structure-refactor.md
emilbon99: Didn't you want to split the label definitions between `core` and `x`? Per this issue that you asked me to make based on a PR review: https://linear.app/synnax/issue/SY-3922/consider-splitting-labels-definitions-across-core-and-x

I agree, fewer data structures is better. I'm just wondering how this fits into the issue above.

docs/tech/rfc/0041-260527-core-structure-refactor.md
emilbon99: "I think we will need to either edit oracle sync to regen new schemas after modifying
something like spatial.oracle, do it manually for the moment, or add another oracle X command. This means"

This problem is WAAAAAAAAAAAAAAAAAAY harder than you think it is. I spent a long time working on it and it was brutal. If you want to do it okay, but dependency tracking, syncing, etc. is an absolute pain of a problem. 

---

This also resolves stuff like how does auto generated code know whether to import
spatial/types/v54 or spatial/types/v55? 

This is a non issue with the current model, as solved by the snapshots directory right?

emilbon99:
I think it might make it easier to auto-generate migrations as well, not sure.

-> I think you're wrong here, and you're going to have a _brutal_ time doing dependency management with this model, but you're smarter than I am.

emilbon99 (changes requested):
Replied to comments.

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: Don't know why issue was named like that - ideally we move label definition only to the
Core, and split statuses into "status metadata" (name, labels, key) defined in the
Core, and "status data" (variant, message, description, data) defined in X, and maybe in
the future moved the statuses to being a JSON-type channel.

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: While I don't disagree, it is a little hard to do this in practice. E.g. every versioned
Go package will have `<resource>/vN/`. In practice the better way to solve it is
probably better Go-To definition and documentation popups on the Oracle LSP.

docs/tech/rfc/0041-260527-core-structure-refactor.md
pjdotson: Thinking about this more and decided to keep the per-Core version structures for the
moment. The biggest worries are 1) a lot of "useless" code everywhere (e.g. a full copy
exists for every resource for every version even if nothing changes), which makes
changes harder to grok (e.g. what version did the schematic _actually_ change), 2)
desync with the ImEx version numbers. As long as we don't lock in version numbers
everywhere this isn't a one-way door.

emilbon99: Ok, yeah I mean take the route you think is best. I think it's going to be hard to solve either way, and I didn't do the best job so.

emilbon99 (approved):

PR #2455, opened 2026-06-07
SY-4338: Per-Key Context Subscription Propagation in Aether

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4338](https://linear.app/synnax/issue/SY-4338)

## Description

Aether context propagation previously **broadcast every context change to the entire descendant subtree**. `childCtxChangedKeys` was only an all-or-nothing gate: if a provider changed any key in its `afterUpdate`, `updateChildContexts()` pushed the full merged context to every child, re-running `afterUpdate` on every descendant — whether or not that descendant actually read the changed key. A component like a button re-ran on any context change anywhere above it.

This replaces that with **dynamic per-key subscription tracking** (signals-style):

- Each component holds a `_parent` back-pointer instead of a flattened `parentCtxValues` copy. `ctx.get`/`getOptional`/`has` resolve a key to the **nearest ancestor that publishes it** and record the read as a subscription on that provider.
- Subscriptions are reconciled after **every** `afterUpdate` (which already runs on both state and context updates), so conditional reads subscribe and unsubscribe as they appear and disappear.
- When a provider changes keys, only the **subscribers of those keys** re-run, cascading through their own changed keys via a depth-ordered worklist (each component runs at most once; depth is a valid topological order because providers are always strict ancestors of their subscribers).
- Deleting a component unsubscribes it.

The consumer-facing `Context` API (`get`/`getOptional`/`has`/`set`/`wasSetPreviously`) is unchanged, so every context consumer and `use()` helper (theming, render, status, telem, synnax, alamos, flux, control-state, …) is untouched. No context crosses the worker/main boundary, so `store.ts`/`main.tsx`/`message.ts` are unaffected.

**Impact:** a new sparse-subscription benchmark (same 3,616-node tree, but only 225 leaves read the root's value) runs **~16× faster** than the full-subscription case (≈22,000 vs ≈1,400 ops/s). Under the old model the sparse case was identical to full — every node re-ran regardless. The existing full-subscription benchmark is retained as a no-regression guard.

### Testing

Beyond the rewritten unit suite (sparse propagation, conditional subscribe/unsubscribe, shadowing, same-instance re-set, delete-unsubscribe), this adds a **property/oracle test** (`aether.prop.spec.ts`): across 300 deterministic seeds it generates random trees and random update/delete sequences, drives them through the real propagation engine, and asserts every live node's resolved context equals an independent from-scratch full recompute. It was validated to fail (pinpointing the stale node) when propagation is disabled.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR replaces Aether's broadcast context propagation with dynamic per-key subscription tracking. Instead of re-running every descendant on any context change, each component now records which keys it read and from which ancestor, and only those subscribers are re-run when a provider changes a key — cascading depth-first through intermediate providers.

- **Core engine** (`aether.ts`): `parentCtxValues` flat-copy is replaced by a `_parent` back-pointer; `readCtx` resolves keys by walking up the parent chain and recording each read as a subscription (`ctxSubscriptions`/`ctxSubscribers`). `propagate()` runs a depth-ordered worklist so each subscriber runs at most once and always after its furthest-changed ancestor has settled.
- **Tests** (`aether.spec.ts`, `aether.prop.spec.ts`): New unit cases cover sparse propagation, conditional subscribe/unsubscribe, key shadowing, same-instance re-triggers, and delete-unsubscribe. A 300-seed property test drives random trees through the engine and compares every live node against an independent brute-force recompute.
- **Benchmarks** (`aether.bench.ts`): A sparse-subscription benchmark is added alongside the retained full-subscription case; per the PR description the sparse case runs ~16× faster under the new model.

Safe to merge — the propagation rewrite is logically sound and backed by a thorough unit suite plus a 300-seed property/oracle test that compares every live node against a brute-force full recompute.

The new subscription tracking correctly handles all edge cases: cascade through intermediate providers, conditional subscribe/unsubscribe, key shadowing, same-value re-triggers (canvas pattern), and cleanup on delete. No consumer-facing API changed, so downstream components are unaffected. The two non-blocking observations are about code clarity, not correctness.

No files require special attention; the only notes are documentation/style observations on aether.ts.

| Filename | Overview |
|----------|----------|
| pluto/src/aether/aether/aether.ts | Core propagation rewrite: replaces flat parentCtxValues broadcast with per-key subscription tracking (ctxSubscribers/ctxSubscriptions), depth-ordered cascade propagation, and parent back-pointer; logic is sound with well-handled edge cases |
| pluto/src/aether/aether/aether.spec.ts | Unit tests expanded with targeted cases for per-key propagation, conditional subscribe/unsubscribe, key shadowing, same-value re-set trigger, and delete-unsubscribe; all test factories updated to use parent back-pointer |
| pluto/src/aether/aether/aether.prop.spec.ts | New property/oracle test: 300 deterministic seeds, each running a random tree through real propagation and asserting agreement with a from-scratch full recompute; correctly handles delete subtree tracking |
| pluto/src/aether/aether/aether.bench.ts | Adds sparse-subscription benchmark (225 of 3616 nodes subscribe) alongside retained full-subscription benchmark; factory signatures updated to parent back-pointer |
| pluto/src/log/aether/Log.spec.ts | Updated to use ContextSeed composite as a test parent that publishes context values; correctly replaces the old flat-map approach, though ContextSeed's _children map remains empty so structural invariants are only partially satisfied |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Root as Root (provider)
    participant Propagate as propagate()
    participant Sub1 as Subscriber A (depth 2)
    participant Sub2 as Subscriber B (depth 3)
    participant Ignorer as Non-subscriber C

    Root->>Root: _updateState → runAfterUpdate()
    Root->>Root: "ctx.set(key, newVal) — childCtxChangedKeys += key"
    Root->>Root: reconcileSubscriptions()
    Root->>Propagate: propagate()
    Propagate->>Propagate: enqueue subscribers of key → [A, B] (ignores C)
    Propagate->>Sub1: runAfterUpdate() at depth 2
    Sub1->>Sub1: ctx.get(key) resolves to Root — reconcileSubscriptions()
    Sub1-->>Propagate: childCtxChangedKeys empty — no cascade
    Propagate->>Sub2: runAfterUpdate() at depth 3
    Sub2->>Sub2: ctx.get(key) resolves to Root — reconcileSubscriptions()
    Note over Ignorer: Never called — not subscribed to key
```

<sub>Reviews (2): Last reviewed commit: ["SY-4338: Simplify context internals and ..."](https://github.com/synnaxlabs/synnax/commit/37979e8a1ffebfe28ed21e2ca1e597832bd75a62) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36090383)</sub>


pluto/src/aether/aether/aether.ts
```diff
@@ -341,21 +398,90 @@ export abstract class Leaf<
...
+      for (const node of bucket) {
+        if (node._deleted) continue;
+        try {
+          node.runAfterUpdate();
+        } catch (e) {
+          throw newTreeError(e, `${node.toString()}.updateContext`);
+        }
+        if (node.childCtxChangedKeys.size > 0) enqueue(node, node.childCtxChangedKeys);
+      }
     }
```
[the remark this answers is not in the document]
emilbon99: Intentional — leaving as-is. The depth range here is just the tree depth (realistically <10–15 levels), so the integer loop runs that many times and each empty level is a single Map.get + continue. The sorted occupied-depths alternative can't snapshot the bucket keys because the worklist grows during iteration (cascades enqueue deeper buckets and the loop re-reads maxDepth to pick them up), so it'd require maintaining a sorted, deduped structure inside enqueue — more bookkeeping than the integer loop, to optimize a loop bounded by tree depth. Prefer to keep this simple.

emilbon99:
@greptile re-review and update PR description

pluto/src/aether/aether/aether.ts
```diff
@@ -197,13 +194,29 @@ export abstract class Leaf<
   private _state: z.infer<StateSchema> | undefined;
   private _prevState: z.infer<StateSchema> | undefined;
   private _deleted: boolean = false;
-  /** Context values inherited from this component's parent. Read-only. */
-  protected readonly parentCtxValues: Map<string, any>;
-  /** Context values this component has published to its descendants. */
+  /** The parent component, used to resolve inherited context. `null` only at the root. */
+  private _parent: Leaf<any, any, any> | null;
```
pjdotson: shouldn't use `any`s

pluto/src/aether/aether/aether.ts
```diff
@@ -197,13 +194,29 @@ export abstract class Leaf<
...
   protected readonly childCtxValues: Map<string, any>;
-  /** Keys in {@link childCtxValues} whose changes should trigger descendants to re-run
-   * {@link afterUpdate}. Cleared at the start of each update. */
+  /** Keys in {@link childCtxValues} this component changed during the current
+   * {@link afterUpdate}. Drives selective propagation; cleared at the start of each run. */
   protected readonly childCtxChangedKeys: Set<string>;
+  /** Provider side: for each key this component publishes, the descendants currently
+   * subscribed to it (i.e. those that resolved this component as their nearest provider
+   * of that key on their last {@link afterUpdate}). */
+  private readonly ctxSubscribers = new Map<string, Set<Leaf<any, any, any>>>();
```
pjdotson: shouldn't use `any`

pluto/src/aether/aether/aether.ts
```diff
@@ -197,13 +194,29 @@ export abstract class Leaf<
...
+  /** Consumer side: keys this component read on its last {@link afterUpdate}, mapped to
+   * the provider it resolved each to. Diffed every run to keep subscriptions current. */
+  private readonly ctxSubscriptions = new Map<string, Leaf<any, any, any>>();
```
pjdotson: shouldn't use `any`

pluto/src/aether/aether/aether.ts
```diff
@@ -197,13 +194,29 @@ export abstract class Leaf<
...
+  /** Context values this component has published to its descendants. A component is the
+   * provider of every key in this map for the descendants nearest to it. */
   protected readonly childCtxValues: Map<string, any>;
-  /** Keys in {@link childCtxValues} whose changes should trigger descendants to re-run
-   * {@link afterUpdate}. Cleared at the start of each update. */
+  /** Keys in {@link childCtxValues} this component changed during the current
+   * {@link afterUpdate}. Drives selective propagation; cleared at the start of each run. */
   protected readonly childCtxChangedKeys: Set<string>;
+  /** Provider side: for each key this component publishes, the descendants currently
+   * subscribed to it (i.e. those that resolved this component as their nearest provider
```
pjdotson: how could a `Leaf` have descendants?

pluto/src/aether/aether/aether.ts
```diff
@@ -321,6 +334,19 @@ export abstract class Leaf<
     };
   }
 
+  /** Resolves `key` to the nearest ancestor publishing it (self excluded) and records
+   * the read against the in-progress {@link afterUpdate} so a subscription is formed.
+   * Returns the provider, or `null` if unresolved. Self is excluded so a component
+   * overriding a parent's key still reads the parent's value via {@link Context.get}.
+   * The read is not recorded outside an {@link afterUpdate} (e.g. during
+   * {@link afterDelete}), where {@link ctxReads} is `null`. */
+  private readCtx(key: string): Leaf<any, any, any> | null {
```
pjdotson: shouldn't use `any`

pjdotson (changes requested):
`any`s are not acceptable in prod code

emilbon99:
@pjdotson restructured to be cleaner

pjdotson (changes requested):
approved after 2 comments addressed

pluto/src/aether/aether/aether.ts
```diff
@@ -157,6 +154,258 @@ export type CallersFromSchema<T> = {
...
+    type,
+    sender,
+    instrumentation,
+    parent,
+  }: ComponentConstructorProps) {
+    this.type = type;
+    this.path = path;
+    this.sender = sender;
+    this.instrumentation = instrumentation.child(this.toString());
+    this._parent = parent as Node | null;
```
pjdotson: this type assertion

pluto/src/aether/aether/aether.prop.spec.ts
pjdotson, line 1: this file should probably be part of `aether.spec.ts` to match pattern elsewhere in codebase

PR #2458, opened 2026-06-08
SY-4344: Add Log Action Codec and Dispatch Endpoint

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4344](https://linear.app/synnax/issue/SY-4344)

## Description

Adds the granular action codec and `Dispatch` endpoint for logs — the analog of the line plot (#2358), table, and schematic action systems, and the second leg of the logs Pluto-state migration (after #2325 strongly typed the body). Eight typed actions replace whole-document `SetData` writes for partial mutations:

`Rename`, `AddChannel`, `RemoveChannel`, `SetChannelEntry`, `SetChannels`, `SetTimestampPrecision`, `SetShowChannelNames`, `SetShowReceiptTimestamp`.

`SetChannelEntry` is an upsert keyed by `entry.channel` (the per-channel config workhorse, like line plot's `SetLine`); `SetChannels` replaces the whole ordered list for bulk selection and reordering. Channel mutations are key-based, not index-based, so they round-trip cleanly through dispatch.

**Oracle / codegen**
- `schemas/log.oracle` declares the 8 `action` blocks. Oracle generates `core/pkg/service/log/actions.gen.go` (Action union + `Reduce`) and `client/ts/src/log/actions.gen.ts` (zod payload schemas + `actionZ` + `createReduceAll` + `scopedActionZ`/`dispatchReqZ`).

**Server**
- Hand-written `actions.go` `Handle` methods. `AddChannel` materializes a default `ChannelEntry` (defaults duplicated from the schema, in sync with the TS side); `SetTimestampPrecision` rejects out-of-range values with `validate.ErrValidation`.
- `Writer.Dispatch` applies the sequence inside a single `gorp` update via `Reduce`, then notifies the service observer post-commit.
- `service.go` gains an optional `Signals` provider and wires `sy_log_set` (action broadcast) + `sy_log_delete` (gorp deletes, `DisableSet`), plus `OnAction`.
- `/log/dispatch` endpoint wired across all 5 transport sites (HTTP server + gRPC noop, since logs are HTTP/WS-only).

**Client**
- `client.logs.dispatch(key, dispatchKey, actions)`; hand-written `actions.ts` handlers with snapshot-safe inverse capture for undo.

**Test infrastructure**
- Extracts the action-recorder helper that was copy-pasted across the line plot, schematic, and (new) log writer tests into a shared, dot-imported `core/pkg/service/actions/testutil` package, and adopts it in all three.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds the granular action codec and `/log/dispatch` endpoint for the log service — the fourth service (after line plot, table, and schematic) to adopt the Oracle-generated action pattern. Eight typed actions replace whole-document `SetData` writes, and `Writer.Dispatch` applies them atomically via `gorp.ChangeErr` then broadcasts via the `actions.State` observer, mirroring the existing architecture exactly.

- **Oracle codegen** (`schemas/log.oracle` → `actions.gen.go` / `actions.gen.ts`): declares 8 action blocks; the generated `Reduce` loop, payload constructors, and zod schemas are consistent with the other services.
- **Server** (`actions.go`, `writer.go`, `service.go`): hand-written `Handle` methods with range validation for `SetTimestampPrecision`; `OpenService` wires `sy_log_set` and `sy_log_delete` signals via the `service.NewOpener` / `ok()` closer pattern; the shared `testutil.Recorder` is extracted from lineplot/schematic into a reusable generic helper.
- **Client** (`actions.ts`, `client.ts`): snapshot-safe inverse capture for undo; `dispatch()` wires to `/log/dispatch`; `SET_CHANNEL_NAME` follows the existing cross-service signal-channel naming convention.

The change faithfully replicates the established action-dispatch architecture across Go and TypeScript and is well tested; the two observations are minor gaps rather than incorrect behavior.

All eight action handlers are correct and tested end-to-end. The redundant `addChannel` step in the TypeScript `removeChannel` inverse produces the right result but emits an extra downstream action on undo. The writer test also omits the dispatch-to-non-existent-key case that the schematic test covers, leaving a small behavioral assumption unverified.

`client/ts/src/log/actions.ts` (removeChannel inverse) and `core/pkg/service/log/writer_test.go` (missing ErrNotFound case) are worth a second look.

| Filename | Overview |
|----------|----------|
| client/ts/src/log/actions.gen.ts | Oracle-generated zod schemas and action factory functions for all 8 log action types; correct discriminated-union shape and well-typed payload constructors. |
| client/ts/src/log/actions.spec.ts | Thorough unit tests for all action handlers and round-trip undo semantics; covers single-element and multi-action cases, wire-format parsing, and scoped action envelopes. |
| client/ts/src/log/actions.ts | Hand-written handler implementations; correct overall, but the removeChannel inverse includes a redundant addChannel action before setChannelEntry, emitting an extra downstream action on undo. |
| client/ts/src/log/client.ts | Adds dispatch() method wiring the /log/dispatch endpoint and exports SET_CHANNEL_NAME following established cross-service convention. |
| core/pkg/service/log/actions.go | Hand-written Handle methods for all 8 action payloads; SetTimestampPrecision correctly validates range [0,3] and returns zero-value Log on error per codebase convention, enabling gorp rollback. |
| core/pkg/service/log/actions.gen.go | Oracle-generated Action union, Reduce loop, and action constructors; correctly nil-checks each payload pointer before dispatching and returns on first error. |
| core/pkg/service/log/service.go | Wires action state, signal publishers (sy_log_set + sy_log_delete), and OnAction subscription; correctly registers all resources with io.MultiCloser via the service.NewOpener ok() pattern. |
| core/pkg/service/log/writer.go | Adds Dispatch method applying Reduce inside a gorp ChangeErr then calling dispatcher.Notify; consistent with the lineplot/schematic/table pattern. |
| core/pkg/service/log/writer_test.go | Extensive Dispatch test coverage for all action types, atomicity, observer notification, and monotonic Seq; missing the non-existent key / ErrNotFound case that the schematic test includes. |
| core/pkg/service/actions/testutil/testutil.go | New shared Recorder helper extracted from lineplot/schematic tests; thread-safe, generic, and correctly dot-imported in all three service test packages. |
| core/pkg/api/log/log.go | Adds Dispatch API handler with RBAC enforcement and WithTx wrapping; consistent with SetData and other mutating endpoints. |
| schemas/log.oracle | Declares all 8 action blocks with correct field types and doc annotations; drives Oracle codegen for both Go and TypeScript output files. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant C as Client (TS)
    participant API as api/log/log.go
    participant W as Writer.Dispatch
    participant G as gorp / DB
    participant S as actions.State (observer)
    participant SIG as signals.Provider

    C->>API: "POST /api/v1/log/dispatch {key, dispatchKey, actions}"
    API->>API: RBAC enforce (ActionUpdate)
    API->>G: WithTx begin
    G->>W: Writer(tx).Dispatch(ctx, key, dispatchKey, actions)
    W->>G: table.NewUpdate().ChangeErr(Reduce).Exec(tx)
    G->>G: "Reduce(log, actions...) — validates & applies in order"
    G-->>W: updated Log (or error → rollback)
    W->>S: dispatcher.Notify(ctx, key, dispatchKey, actions)
    S-->>SIG: PublishSignals broadcasts Scoped on sy_log_set
    W-->>G: nil
    G->>G: commit transaction
    API-->>C: 200 OK
```

1. `core/pkg/service/log/writer_test.go`, line 1462 ([link](https://github.com/synnaxlabs/synnax/blob/a26a6dd0a00e9b10a7255d75b9e802fe3e6d84a9/core/pkg/service/log/writer_test.go#L1462)) 

   <a href="#">P2</a> **Missing ErrNotFound coverage for non-existent key**

   The `Dispatch` test suite covers every action type and atomicity invariants, but has no case for dispatching to a log key that does not exist. The schematic writer test has this scenario (`"Should fail with query.ErrNotFound and not notify subscribers when the target schematic does not exist"`). Without it here, a regression where `gorp.MatchKeys` silently ignores a missing key instead of returning `ErrNotFound` — and `Notify` consequently fires on a no-op — would go undetected.

2. `client/ts/src/log/actions.ts`, line 551-563 ([link](https://github.com/synnaxlabs/synnax/blob/a26a6dd0a00e9b10a7255d75b9e802fe3e6d84a9/client/ts/src/log/actions.ts#L551-L563)) 

   <a href="#">P2</a> **Redundant `addChannel` step in `removeChannel` inverse**

   The undo sequence `[addChannel, setChannelEntry]` is functionally correct: `addChannel` appends a default entry, then `setChannelEntry` (an upsert) replaces it in place with `oldEntry`. However, `setChannelEntry` already handles the insert path when the channel is absent, so the leading `addChannel` is redundant — `[setChannelEntry({ entry: oldEntry })]` alone produces the same result. The extra action also means the undo of a `removeChannel` emits two downstream actions to subscribers instead of one, which could matter when replay ordering is observable.

</details>

<sub>Reviews (1): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/rc&#39;..."](https://github.com/synnaxlabs/synnax/commit/a26a6dd0a00e9b10a7255d75b9e802fe3e6d84a9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36291216)</sub>


core/pkg/service/actions/testutil/testutil.go
```diff
@@ -0,0 +1,50 @@
...
+// expected by Service.OnAction.
+func (r *Recorder[K, A]) Record(_ context.Context, sa actions.Scoped[K, A]) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.scopeds = append(r.scopeds, sa)
+}
+
+// Snapshot returns a copy of the scoped actions recorded so far in the order they
+// were received.
+func (r *Recorder[K, A]) Snapshot() []actions.Scoped[K, A] {
```
pjdotson, line 44: should add tests for this file

core/pkg/service/actions/testutil/testutil.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// Recorder collects the Scoped action notifications a service emits through its
+// action observer so tests can assert on them. The zero value is ready to use,
+// and Recorder is safe for concurrent use. Pass Record directly to
+// Service.OnAction:
+//
+//	rec := &Recorder[log.Key, log.Action]{}
+//	DeferCleanup(svc.OnAction(rec.Record))
+type Recorder[K comparable, A any] struct {
```
pjdotson, line 29: is there a reason this isn't used in the table tests?

core/pkg/service/log/writer.go
```diff
@@ -91,6 +93,28 @@ func (w Writer) SetData(ctx context.Context, key Key, data Log) error {
 		}).Exec(ctx, w.tx)
 }
 
+// Dispatch applies a sequence of actions atomically to the log with the given key.
+// After a successful update the actions are notified to the service-level observer so
+// subscribers (cluster signals) can broadcast them. dispatchKey is a client-generated
+// identifier carried verbatim onto the broadcast so the originating client can match
+// its own echo against the set of outstanding local replays and skip a redundant
+// reduce when no foreign action interleaved.
+func (w Writer) Dispatch(
```
pjdotson, line 92: shouldn't we be deleting `Rename` methods on visualizations to instead use `Dispatch`?

pjdotson (changes requested):

core/pkg/transport/grpc/grpc.go
```diff
@@ -125,8 +125,8 @@ func Bind(layer *api.Layer, channelSvc *distchannel.Service) []grpc.BindableTran
 	t.LogCreate = noop.UnaryServer[log.CreateRequest, log.CreateResponse]{}
 	t.LogRetrieve = noop.UnaryServer[log.RetrieveRequest, log.RetrieveResponse]{}
 	t.LogDelete = noop.UnaryServer[log.DeleteRequest, types.Nil]{}
-	t.LogRename = noop.UnaryServer[log.RenameRequest, types.Nil]{}
 	t.LogSetData = noop.UnaryServer[log.SetDataRequest, types.Nil]{}
+	t.LogDispatch = noop.UnaryServer[log.DispatchRequest, types.Nil]{}
```
pjdotson, line 129: make sure to delete line plot rename, log set data, and line plot set data in separate
PRs when this is merged

core/pkg/service/actions/testutil/testutil_test.go
```diff
@@ -0,0 +1,83 @@
...
+				}
+			}()
+		}
+		wg.Wait()
+		Expect(rec.Snapshot()).To(HaveLen(goroutines * perGoroutine))
+	})
+})
+
+var _ = Describe("Record", func() {
+	It("Should match the handler signature expected by Service.OnAction", func() {
```
pjdotson: don't like tests like this, the compiler will get this for us

pjdotson (changes requested):

PR #2464, opened 2026-06-09
SY-4329: Atomic API Handlers and Access-Check Hardening

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4329](https://linear.app/synnax/issue/SY-4329)

## Description

Makes every API **write** handler atomic, moves transaction wrapping out of the handlers themselves and into the transport binding layer, and tightens access-check semantics across the board. Pure read handlers are deliberately left outside any transaction (see "Why reads skip the tx" below).

### Transaction wrapping at the binding layer

The original iteration of this PR wrapped each handler in its own `s.db.WithTx` block. That pattern was correct but repetitive — every handler opened, committed, and rolled back its own tx, and there was no single place to audit "is this call atomic?". This has been replaced with two transport-level adapters:

- New `freighter/go/gorp` module providing:
  - `CreateUnaryHandler[RQ, RS](db, handler)` — opens a `gorp.Tx`, runs the wrapped handler inside it, commits on `nil`, rolls back on error. On any error a zero `RS` is returned. Used for every handler that writes.
  - `CreateUnaryReadHandler[RQ, RS](db, handler)` — passes the `*gorp.DB` directly to the handler as its `gorp.Tx` argument. No batch is opened, no commit happens. Used for every pure-read handler.
- New `freighter.UnaryHandler[RQ, RS]` named type in `freighter/go/unary.go`. The grpc, http, mock, and noop transports' `BindHandler` signatures were updated to use it in place of the ad-hoc `func(context.Context, RQ) (RS, error)`.
- Every API handler's signature changed from `func(ctx, req) (res, error)` to `func(ctx, tx gorp.Tx, req) (res, error)`. The handler body no longer opens a transaction — it just uses the `tx` it was handed.
- `core/pkg/api/layer.go` now wires every unary handler via `fgorp.CreateUnaryHandler(db, l.X.Y)` for writes or `fgorp.CreateUnaryReadHandler(db, l.X.Y)` for reads in `BindTo`. This is the single place transaction lifetimes are managed for the entire API surface.
- The streaming framer endpoints (`FrameWriter`, `FrameIterator`, `FrameStreamer`) and `ArcLSP` continue to be bound directly — their lifetimes outlive any single tx by design.

### Why reads skip the tx

`gorp.OpenTx` is backed by `pebble.NewIndexedBatch`, which is **not** a snapshot. `pebble.(*Batch).Get` calls `db.getInternal(key, batch, nil /* snapshot */)` — every read through the batch sees the live DB state merged with the batch's own pending writes. So wrapping a pure-read handler in `WithTx` provides **no** read isolation; two `tx.Get` calls inside the same handler can already see different states if a concurrent writer commits between them.

The wrapper still costs real work: every wrapped retrieve allocates an indexed-batch arena, opens a `kv.Tx`, and routes reads through `Batch.Get`/`Batch.NewIter` (more expensive than direct `DB.Get`/`DB.NewIter`). Profiled under the full Pluto vitest suite (~26k API calls), wrapping retrieves cost:
- +22% CPU samples (22.95s vs 18.75s)
- +39% allocated bytes (1.21 GB vs 0.87 GB)
- +65% allocated objects (11.1M vs 6.7M)
- +18% post-test resident heap (358 MB vs 304 MB)
- No measurable wall-clock difference at this load.

`CreateUnaryReadHandler` skips all of that and gives identical consistency. If we ever want true snapshot isolation for reads, the right primitive is `pebble.NewSnapshot()` (cheap per-call, but pins compaction floor for its lifetime) — that's a separate follow-up.

The audit: every API handler with a read-shaped name (26 total: `Retrieve`, `RetrieveGroup`, `RetrieveSymbol`, `RetrieveSymbolGroup`, `RetrievePolicy`, `RetrieveRole`, `Resolve`, `List`, `Get`, `Export`) was traced through the service and distribution layers to confirm no `tx.Set` / `tx.Delete` / `gorp.NewCreate` / `gorp.NewUpdate` / `gorp.NewDelete` ever fires on the read path, including the access enforcer (`rbac.Enforcer.Enforce` → `policy.ResolveSubjects`). All 26 are pure-read end-to-end.

### Access-check correctness

Real bugs fixed:

- `label.Retrieve` no longer leaks labels through the `RetrieveFor` early-return path.
- `ontology.Retrieve` no longer returns search results without enforcing.
- `schematic.CreateSymbol` now enforces `ActionUpdate` on the parent in addition to `ActionCreate` on the new symbols.
- `status.SetByKeyOrName` now picks `ActionUpdate` when the target already exists, `ActionCreate` only for the true-new path.
- `RenameSymbol` / `DeleteSymbol` / `RetrieveSymbol` request types corrected from `schematic.Key` to `symbol.Key` (now an alias for `uuid.UUID`).

Systemic cleanup:

- Removed the `(*rbac.Service).Enforce` nil-tx shorthand. Every caller is now forced to thread a `gorp.Tx` via `s.access.NewEnforcer(tx).Enforce(...)`.
- Every `Create` handler now uses an explicit type-level ontology ID (`[]ontology.ID{{Type: ontology.ResourceTypeX}}`) instead of an `OntologyIDsFromX(req.X)` against client-supplied pre-create (zero) keys. Same matcher behavior in practice but unambiguous intent.
- Every `Retrieve` handler now enforces against the actual retrieved entities, not the requested keys.
- For `Retrieve` handlers that previously did multiple `Exec(ctx, nil)` calls (channel, device, rack, ranger, status, task), all reads now run on a single live-DB view through the shared `gorp.Tx` (= DB) argument.
- `access.RetrieveRole` aligned with `RetrievePolicy` (post-retrieve enforcement). Added `role.OntologyIDs` and `role.OntologyIDsFromRoles` helpers.
- `access.UnassignRole` uses the `allowInternal` constant instead of literal `true`.
- `framer.openWriter`'s dead nil-tx channel-key existence check removed (`NewStreamWriter` already validates channel keys via `validateChannelKeys`).
- `framer.Delete` retrieves and enforces in one tx, then deletes time ranges outside it (by design).

### Other cleanups in scope

- `KeysFromOntologyIDs` across ranger, policy, workspace, label, task, view, rack, arc, and symbol converted to `lo.MapErr`. This included matching the behavior of `channel.KeysFromOntologyIDs` to that of all other resources.
- Named returns stripped from every API handler; every handler now returns a zero-valued first return when `err != nil`.
- Symbol package introduces `type Key = uuid.UUID`; `uuid.UUID` usages across `symbol/` and `api/schematic` switched to `symbol.Key`.

### Tests

- New `freighter/go/gorp/gorp_test.go` suite covers both adapters: `CreateUnaryHandler` commit / rollback / zero-return / request-propagation, and `CreateUnaryReadHandler` value pass-through, error pass-through (no zeroing), and identity of the `Tx` argument with the DB itself.
- Updated grants in `api/status/status_test.go` to match the new `SetByKeyOrName` action semantics.
- All affected API tests (`api/status`, `api/schematic`, `api/user`) updated to pass the `db` argument to the new handler signature.
- Deleted the now-obsolete `Describe("Enforce")` block in `rbac_test.go` (its tested method was the nil-tx shorthand we removed).

## Basic Readiness


pjdotson:
@greptile /rerun

pjdotson:
@greptile rerun. Note that the change for `channel.KeysFromOntologyIDs` is intended and correct. The status access control check issue is deferred.

emilbon99 (changes requested):
If you're going to wrap almost everything in a `WithTx` it's probably worth creating a higher order function like `createHandlerWithTx(handlerWithtx) freighterHandler) that injects the tx as a third parameter, that way it's always available and we reduce explicit closure count a lot.

pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

emilbon99 (changes requested):
Overall looks good. My only concern is that it might be worth profiling the server under pluto tests before/after this PR to see what the implications are for opening a `tx` under retrieve requests.

It's likely not that expensive, but it does (a) allocate extra pebble machinery which we don't know what the cost is and (b) it increases GC pressure.

I think the idea of standardizing everything under a Tx is good, but *if* it does have performance implications, then it might be worth reconsidering for handlers that don't actually need txes.

Approved once you've done that.

PR #2466, opened 2026-06-09
SY-4341: Refactor Server-Side ImEx

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4341](https://linear.app/synnax/issue/SY-4341)

## Description

Implements the imex-peek direction of RFC 0041 against `core/pkg/service/imex`.

**Envelope.** Reshapes the existing `Envelope` with a private `raw []byte` / `body map[string]any` / bound `encoding.Codec`. `UnmarshalJSON` peeks `{version, type, name}` once (with `UseNumber` so the `Version` peek preserves int64 precision) and stashes the raw bytes plus the JSON codec for a later typed decode. Missing or empty `type` / `name` is rejected at the boundary so the failure surfaces before routing. The legacy `"5.0.0"`-style semver `Version` is still accepted at the wire boundary.

**`Decode[T](ctx, env)`** materializes the body straight into the requested type via the bound codec — no `map[string]any` intermediate. Envelopes constructed by `Encode` (export path) or by hand have no codec bound; calling `Decode` on one returns a descriptive error (rather than panicking on a nil-interface dispatch). Callers that need to decode them must `Marshal` and unmarshal back through one of the `UnmarshalX` methods first.

**`Encode[T](env *Envelope, data T) error`** is the symmetric inverse. The caller supplies an envelope carrying the desired `Version` / `Type` / `Name`; Encode mutates it in place rather than returning a new value, and leaves the envelope untouched on error. The typed value is reduced to a codec-independent map via a strict json-tag-only `structToMap` — fields without a `json:"…"` tag, with `json:"-"`, or with an empty tag name (e.g. `json:",omitempty"`) are dropped from the wire body; tag options after the name are ignored; pointers and non-structs are rejected. `Type` and `Name` are resolved symmetrically: if the body carries a `type` (or `name`) entry, it must be a string and overrides the envelope's value; otherwise the envelope's value is kept. After merging, both must be non-empty.

**`Service`** is now a parameterless registry — `NewService()` no longer requires a `gorp.DB`, and registration/lookup are guarded by a `sync.RWMutex`. `Service.Import` is single-resource and takes a caller-provided `gorp.Tx`; multi-envelope atomicity is the caller's responsibility via `db.WithTx`. `Service.Export` does **not** thread a `gorp.Tx` — for now the `Exporter` interface reads directly from its own storage handle (so the API `Export` endpoint and concrete exporters can't accidentally dereference a nil `tx`). Transactional reads on the export path will return in a follow-up once the wider API surface is settled. The API layer (`core/pkg/api/imex`) runs access enforcement immediately before each dispatch.

**`x/go/encoding/json`.** `Codec.Decode` and `DecodeStream` now go through `json.NewDecoder(...).UseNumber()` so `map[string]any` targets keep int64 precision. Typed targets are unaffected.

**Doc cleanup** across the package: envelope shape, codec binding semantics, single-resource Import/Export contract, interface contracts, and a `notFoundError` flag-arg cleanup.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR refactors the server-side import/export system (`core/pkg/service/imex`) per RFC 0041: the `Envelope` type gains private `raw []byte` / `body map[string]any` / bound `encoding.Codec` fields, with `Decode[T]` and `Encode[T]` generics replacing the old `Data map[string]any` field. `Service` is simplified to a parameterless, concurrent-safe registry, and `Import` now takes a caller-supplied transaction while `Export` reads directly from each exporter's storage handle.

- **Envelope**: `UnmarshalJSON` stashes raw bytes and binds `xjson.Codec`; `Decode[T]` materialises the payload directly into the target type; `Encode[T]` reduces a typed value via `structToMap` and stamps headers; `MarshalJSON` errors loudly on a nil body rather than emitting `null`.
- **Service registry**: `NewService()` no longer requires a `gorp.DB`; registration and lookup are guarded by `sync.RWMutex`; multi-envelope atomicity is now the caller's responsibility via `db.WithTx`.
- **Test coverage**: Extensive new specs for all `Encode`/`Decode` paths, embedded-struct promotion variants, round-trips, and a concurrency stress test for the registry.

The change is well-scoped and the core paths are thoroughly tested; no correctness regressions are apparent.

The refactored Encode/Decode paths, mutex guards, and envelope preservation-on-error contracts are all correctly implemented and backed by a comprehensive test suite. The two observations (structToMap embedding semantics vs. encoding/json for conflicting promotions, and the nil tx passed to the export RBAC enforcer) are non-blocking quality notes rather than defects in the changed paths.

The asymmetric nil-tx RBAC enforcer call in core/pkg/api/imex/imex.go (export path) and the 'mirroring encoding/json' claim in the structToMap doc comment in core/pkg/service/imex/imex.go are worth a second look before merging.

| Filename | Overview |
|----------|----------|
| core/pkg/service/imex/imex.go | Core envelope refactor: adds private raw/body/codec fields, Decode[T] and Encode[T] generics, structToMap with flattenStruct for embedded struct promotion. Logic is sound; the 'mirroring encoding/json' claim in the structToMap doc slightly overstates fidelity for conflicting embedded-field promotions. |
| core/pkg/service/imex/service.go | Service refactored to parameterless registry with sync.RWMutex; Import now takes a caller-supplied tx (single-envelope); Export drops tx by design. Concurrency guards are correct and consistent across all registration/lookup methods. |
| core/pkg/api/imex/imex.go | API layer updated to single-envelope Import/Export; access enforcement is run before dispatch. Export calls NewEnforcer(nil) unlike Import which passes the active tx — asymmetry worth confirming is intentional. |
| core/pkg/service/imex/imex_test.go | Comprehensive new test suite covering Encode/Decode round-trips, all flattenStruct embedding variants, error paths, and precision preservation. Tests are well-structured and clearly document intent. |
| core/pkg/service/imex/service_test.go | Service tests reorganised to use per-test svc+ts setup; adds concurrency test exercising concurrent RegisterImporter + ImporterType. Rollback test correctly models the new caller-owned tx atomicity contract. |
| core/pkg/service/imex/imex_suite_test.go | Suite setup simplified to bare DB fixture; all service/type fixtures moved to service_test.go. Clean and correct. |
| core/pkg/service/layer.go | NewService() call updated to match new parameterless signature; error-check wrapper removed appropriately since NewService no longer returns an error. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client
    participant API as api/imex.Service
    participant Svc as service/imex.Service
    participant Handler as Importer/Exporter

    rect rgb(220, 240, 255)
        Note over Client,Handler: Import path
        Client->>API: Import(ctx, tx, Envelope)
        API->>Svc: ImporterType(env.Type)
        Svc-->>API: ontology.ResourceType
        API->>API: RBAC enforce(tx)
        API->>Svc: Import(ctx, tx, Envelope)
        Svc->>Handler: Import(ctx, tx, Envelope)
        Handler->>Handler: Decode[T](ctx, env)
        Handler->>Handler: persist via tx
        Handler-->>Svc: Key
        Svc-->>API: Key
        API-->>Client: "ImportResponse{Key}"
    end

    rect rgb(240, 255, 220)
        Note over Client,Handler: Export path
        Client->>API: Export(ctx, ontology.ID)
        API->>API: RBAC enforce(nil tx)
        API->>Svc: Export(ctx, ontology.ID)
        Svc->>Handler: Export(ctx, Key)
        Handler->>Handler: read from own db handle
        Handler->>Handler: "Encode[T](&env, data)"
        Handler-->>Svc: "Envelope{body set}"
        Svc-->>API: Envelope
        API-->>Client: "ExportResponse (MarshalJSON -> body map)"
    end
```

<sub>Reviews (5): Last reviewed commit: ["Fix imex test"](https://github.com/synnaxlabs/synnax/commit/fcf1e6139993402a8a1222111e23e7b936f7b98f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36541587)</sub>


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

core/pkg/service/imex/imex.go
```diff
@@ -9,31 +9,39 @@
...
+	xjson "github.com/synnaxlabs/x/encoding/json"
 	"github.com/synnaxlabs/x/errors"
 	"github.com/synnaxlabs/x/gorp"
 	"github.com/synnaxlabs/x/validate"
 )
 
+// Key is the string-form identifier of a resource within its ontology type — i.e. the
+// Key field of an ontology.ID. Importers return one on persist, Exporters look one up
+// on retrieve.
+type Key = string
```
emilbon99: Why do we need a new key type? Why don't we pass around ontology IDs instead?

core/pkg/service/imex/imex.go
```diff
@@ -55,78 +63,173 @@ func NewErrUnsupportedVersion(typ string, given, supported Version) error {
...
 	if v, ok := m["type"]; ok {
 		s, ok := v.(string)
 		if !ok {
 			return errors.Newf("type must be a string, got %T", v)
 		}
 		e.Type = s
-		delete(m, "type")
+	}
+	if e.Type == "" {
+		return errors.New("type must be a non-empty string")
```
emilbon99: Should we be returning specific `validate.FieldErrors` for these?

core/pkg/service/imex/imex.go
```diff
@@ -55,78 +63,173 @@ func NewErrUnsupportedVersion(typ string, given, supported Version) error {
...
+func Encode[T any](env *Envelope, data T) error {
+	body, err := structToMap(data)
+	if err != nil {
+		return errors.Wrap(err, "encode envelope")
+	}
+	typ := env.Type
+	if v, ok := body["type"]; ok {
+		s, ok := v.(string)
+		if !ok {
+			return errors.Newf("encode envelope: type must be a string, got %T", v)
```
emilbon99: Same thing here, should we be using validate field errors?

emilbon99 (changes requested):
Overall looks good. See my comments.

PR #2467, opened 2026-06-09
SY-4351: Strongly-Typed Visualization Cleanup and DX Improvements

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4351](https://linear.app/synnax/issue/SY-4351)

## Description

A cleanup and developer-experience pass over the visualization stack (line plot, log,
schematic, table) that finishes migrating these resources onto the dispatch-based,
Oracle-defaulted creation path and removes the legacy machinery left behind by that
migration.

**Oracle: `= create` field default replaces the `@key generate` annotation.** Key
generation is now expressed directly in the schema grammar as a field default
(`key = create`) rather than a separate `@key generate` domain. The generator resolves
the field's underlying primitive: `uuid` keys generate a real UUID (`uuid.create()` /
`default_factory=uuid4`), and `string` keys generate a short id (`id.create()` /
`str(uuid4())`). The dead `key.HasGenerate` helper and its TS-plugin branch are removed.

**Schema cleanup.** The `New` structs for line plot, log, schematic, and table now use
explicit struct-level defaults (`key = create`, `rows = []`, `cells = {}`, etc.) instead
of hard-optional `?`/`??` fields, so the generated `newZ` schemas carry their own
defaults and creation no longer needs to backfill them by hand. The unused
`remote_created` field is dropped from `Log`.

**Removed superseded server endpoints.** The action-based `Dispatch` endpoint now covers
everything the old mutation endpoints did, so `LinePlot.SetData`, `LinePlot.Rename`, and
`Log.SetData` are removed end to end: service writers, API handlers, the `Transport`
struct, and all three transport wirings (`layer.go`, `http.go`, `grpc.go`). The TS
client's `lineplot.rename` is reimplemented on top of `dispatch`, and the `setData`
client methods are deleted.

**Console: one generic auto-upload hook.** The four near-identical per-visualization
`useUpload.ts` hooks (line plot, log, schematic, table) are consolidated into a single
`vis/useAutoUpload.ts` factory (`createUseAutoUpload`).

**Pluto / x.** Create hooks drop manual `data.key ??= uuid.create()` backfilling in favor
of parsing through the generated `newZ` defaults. `record.NullishToEmpty` now returns a
`z.ZodDefault` so `= []` / `= {}` defaults round-trip through the schema.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR is a large cleanup and DX pass over the visualization stack (line plot, log, schematic, table) that finishes migrating to the dispatch-based Oracle-defaulted creation path and removes superseded legacy code.

- **Schema and code-gen**: `New` structs now carry explicit defaults (`key = create`, `rows = []`, `cells = {}`, etc.) via `newZ`, replacing hand-rolled backfilling. `@key generate` is deleted in favor of `= create` field defaults. `RemoteCreated` is dropped from `Log` end-to-end.
- **Server cleanup**: `LinePlot.SetData`, `LinePlot.Rename`, and `Log.SetData` are removed from service, API, and transport layers; `lineplot.rename` on the TS client is reimplemented over `dispatch`.
- **Console consolidation**: Four per-visualization `useUpload.ts` hooks are replaced by a single `createUseAutoUpload` factory, and four per-visualization `useCreate.ts` implementations are replaced by `createUseCreate`.

Safe to merge for the TypeScript/console/Oracle changes; the Go binary codec for Log remains a backward-compatibility risk for any deployed instance with stored Log data.

The Oracle code-gen, console consolidation, and server endpoint removal are clean and well-tested. The one outstanding concern — flagged in the previous review cycle — is that the positional Oracle binary codec for Log removes RemoteCreated without a format-version guard, so any stored Log binary written by an older server will misalign TimestampPrecision and the two booleans when decoded by the new codec.

core/pkg/service/log/codec.gen.go — the positional binary codec change and its migration story deserves a second look before deployment against a cluster with existing Log data.

| Filename | Overview |
|----------|----------|
| core/pkg/service/log/codec.gen.go | Removes RemoteCreated from the positional binary codec without a format-version bump; decoding pre-existing stored Log binaries will misalign all subsequent fields. |
| console/src/vis/useAutoUpload.ts | New factory replacing four near-identical per-vis upload hooks; correctly adds startedRef guard missing from old lineplot/schematic/table hooks. |
| console/src/workspace/createUseCreate.ts | New factory consolidating four per-vis useCreate hooks; reads defaults at call time via store.getState() rather than at render time. |
| oracle/plugin/ts/types/types.go | Removes the HasGenerate branch and consolidates key-default handling under applyValidation; correctly distinguishes uuid vs string primitives. |
| oracle/plugin/py/types/types.go | Mirrors the TS plugin: uuid keys now use default_factory=uuid4 instead of lambda: str(uuid4()), properly distinguishing the two primitive types. |
| console/src/log/types/v2.ts | pendingUploadZ derived from log.logZ which now lacks defaults for display fields; suggest switching to log.newZ to preserve graceful parsing of stale persisted state. |
| pluto/src/lineplot/queries.ts | Replaces manual key backfilling and raw cast with newZ.parse(data) for the optimistic write. |
| pluto/src/log/aether/Log.ts | Switches logStateZ base from log.logZ to log.newZ; both schemas carry the same defaults for the picked fields. |
| x/ts/src/record/record.ts | Narrows NullishToEmpty return type to z.ZodDefault so schema inference carries the default value through newZ chains. |
| console/src/lineplot/useCreate.ts | Collapsed to a createUseCreate call; defaults now reads Range.selectActiveKey at call time rather than render time. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["Oracle Schema\n(.oracle files)\nkey = create / rows = [] / cells = {}"] -->|code-gen| B["newZ schemas\nCarry own defaults"]
    B -->|parse at create time| C["Flux update handler\nnewZ.parse(data)"]
    C -->|optimistic write| D["Flux SubStore"]
    C -->|server call| E["API Server"]
    G["createUseCreate factory"] -->|returns| H["useCreate hook"]
    H -->|calls| C
    I["createUseAutoUpload factory"] -->|returns| J["useAutoUpload hook"]
    J -->|startedRef guard| C
    K["HasGenerate removed"] -->|replaced by| L["= create default"]
```

1. `core/pkg/service/log/codec.gen.go`, line 75-115 ([link](https://github.com/synnaxlabs/synnax/blob/066251d2e795843e378dd983116f31bdff6d8d4f/core/pkg/service/log/codec.gen.go#L75-L115)) 

   <a href="#">P1</a> **Positional codec breaks backward compatibility with existing stored data**

   The Oracle binary codec for `Log` is positional — fields are read back in the same order they were written. Removing `RemoteCreated` shifts every subsequent field by one slot: a reader using the new codec will try to decode `TimestampPrecision` (4-byte int32) starting from where the 1-byte `RemoteCreated` bool used to be, reading `[old_remote_created_byte | first 3 bytes of old_timestamp_precision]` as the new int32 value. `ShowChannelNames` then reads the 4th byte of the old `TimestampPrecision`, and `ShowReceiptTimestamp` reads the old `ShowChannelNames`. Any stored `Log` whose `timestamp_precision` was changed from its default of `0` (or whose bool settings differ from the packed byte pattern that happens to be self-consistent) will be silently mis-decoded on the first read after upgrade.
</details>

<sub>Reviews (2): Last reviewed commit: ["more refinements"](https://github.com/synnaxlabs/synnax/commit/65e5f4eef2b713b2e64a7cf8e9d7561c8537d295) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36542221)</sub>


emilbon99:
@greptile re-run

client/ts/src/lineplot/client.ts
```diff
@@ -74,11 +72,7 @@ export class Client {
   }
 
   async rename(key: Key, name: string): Promise<void> {
```
pjdotson, line 74: do we still call these `client.rename`s? should we be deleting this or no?

console/src/vis/useAutoUpload.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+    useEffect(() => {
+      if (pendingUpload == null || startedRef.current) return;
+      startedRef.current = true;
+      create({
+        ...pendingUpload,
+        key,
+        workspace: workspaceKey ?? undefined,
+        name,
+      } as unknown as Input);
```
pjdotson: why do you have to do this type assertion? usually a bad sign.

console/src/workspace/createUseCreate.ts
```diff
@@ -0,0 +1,83 @@
...
+      },
+    });
+    return useCallback(
+      (init) =>
+        update({
+          name: defaultName,
+          ...defaults?.(store),
+          ...init,
+          workspace: targetWorkspace,
+        } as Input),
```
pjdotson: why do you have to do the type assertion here?

pjdotson (changes requested):

client/ts/src/lineplot/client.ts
emilbon99: I have been leaving them in for API consistency since we don't yet know what the ultimate shape we want is.

pjdotson (approved):

PR #2469, opened 2026-06-10
SY-4354: Introduce Panel Data Structure

emilbon99:
**NOTE: The `panel.oracle` schema is likely to evolve, but this sets a lot of the boilerplate and scaffolding. **
**NOTE 2: Ontology relationships are not clearly defined right now, that's ok**

# Issue Pull Request

## Linear Issue

[SY-4354](https://linear.app/synnax/issue/SY-4354)

## Description

Introduces the **Panel** resource as the foundational data structure for the panels-as-tabs work, carved out of the larger `sy-4193-panels` branch. A panel is an owned tree of visualization tabs — interior `Split` nodes dividing area between two children, and `Leaf` nodes holding an ordered tab strip. Each `Tab` has a stable identity independent of its content, and its content is a union of either a backing **resource** (an `ontology.ID`, e.g. a line plot) or an inline **view** (a self-describing app-view such as docs); when neither is set the tab renders the visualization selector.

Mutation is action-based: `InsertTab`, `RemoveTab`, `MoveTab`, `SplitLeaf`, `ResizeSplit`, `SetTabResource`, `SetTabView`, and `Rename`, reduced server-side and broadcast to connected clients via the dispatch action channel.

This PR is intentionally scoped to the **core service + `client/ts` package + schemas/tests**. Console and Pluto integration land in follow-up PRs.

Notable decisions:

- **Parented by a generic `ontology.ID`.** Since this is the first PR in the stack and the `project` resource does not exist yet, panel creation takes a generic parent (`ontology.ID`); a zero parent makes the panel a draft owned by the creating user. The `project`-key convenience field will be reintroduced alongside the project resource.
- **No rename REST endpoint.** Matching the other visualizations (line plot, log, schematic, table), rename is applied through a dispatched `Rename` action rather than a dedicated endpoint, so the change broadcasts to other connected clients.
- The shared panel tree protocol (`ROOT_PATH`, `childPath`, `walkPath`, `findTab`, `firstTab`, `tabLeafPath`) is exported from the client package for reuse by the reducer and downstream consumers.

The `panel` ontology resource type was added to `schemas/ontology.oracle` and regenerated across Go/TS/Python/C++; `workspace` is left untouched.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR introduces the **Panel** resource — a tree of visualization tabs (interior `Split` nodes + `Leaf` nodes with tab strips) — as foundational scaffolding for the panels-as-tabs feature. It adds the full Go service (CRUD + action-based mutation via `Reduce`), a matching TypeScript client with Immer-based optimistic reducers, protobuf/msgpack serialization, ontology integration, and signal broadcasting.

- **Go service** (`core/pkg/service/panel/`): path-keyed binary tree traversal, eight action payloads each with a `Handle` method, post-reduce `validateTree` inside a gorp transaction, and a dual-publisher signal setup (action frames on `sy_panel_set`, deletes via gorp observer).
- **TypeScript client** (`client/ts/src/panel/`): generated Zod schemas and action constructors, Immer draft-based reducers that mirror Go semantics, shared tree-traversal helpers (`walkPath`, `findTab`, `tabLeafPath`) exported for downstream consumers.
- **Schema additions**: `panel.oracle` defining the full discriminated-union tree, `ontology.oracle` registering the `panel` resource type, and a new `spatial.Decimal` type propagated across Go/TS/C++/Python.

Safe to merge with awareness that the TS moveTab reducer and server diverge on out-of-range indices (flagged in a previous review thread); everything else is well-tested and structurally sound.

The moveTab handler in actions.ts mutates state (appends the tab) when the provided index exceeds the post-remove tab count, while the Go reducer returns errIndexOutOfRange and aborts. Because the TS state is mutated the dispatch substrate detects a diff and sends the action to the server, which then rejects it, leaving the client with a stranded optimistic update. All other action handlers are consistent between TS and Go, the tree operations are thoroughly tested, and the Go service layer is clean.

client/ts/src/panel/actions.ts — the moveTab handler's out-of-range index branch; and client/ts/src/panel/client.ts — the rename helper's empty dispatchKey.

| Filename | Overview |
|----------|----------|
| core/pkg/service/panel/tree.go | Core binary-tree operations (walk, updateAt, insert/removeTab, splitLeaf, collapseEmptyLeaves). Logic is correct and consistent with the TS mirror; well-documented sentinel errors. |
| core/pkg/service/panel/actions.go | Eight action Handle methods. Logic correct, but the misleading ownership-promotion comment on RenamePayload (flagged separately) reduces confidence slightly. |
| client/ts/src/panel/actions.ts | Immer-based reducers mirroring Go. The moveTab handler's out-of-range index branch silently appends (state mutation) instead of returning NO_OP, diverging from the server's errIndexOutOfRange — the dispatch substrate sees a diff and sends the action, which the server then rejects. |
| client/ts/src/panel/client.ts | CRUD + dispatch client. Previously flagged: parent field in retrieveReqZ is silently ignored by the server. Additionally, rename passes an empty dispatchKey, preventing echo suppression. |
| core/pkg/service/panel/writer.go | Clean atomic Dispatch (Reduce + validateTree inside ChangeErr); Create/Delete correctly manage ontology relationships. Dispatcher.Notify is called after tx commit. |
| core/pkg/service/panel/service.go | Service wiring is clean; dual-publisher setup (action frames on sy_panel_set, deletes via gorp observer with DisableSet) matches the described design intent. |
| core/pkg/api/panel/panel.go | API layer applies RBAC before every mutation; post-query access enforcement on Retrieve follows the established pattern for this codebase. |
| client/ts/src/panel/tree.ts | Exported tree-traversal helpers (walkPath, findTab, firstTab, tabLeafPath, firstLeafPath) are clean, well-tested, and correctly share path arithmetic with the Go implementation. |
| client/ts/src/panel/actions.spec.ts | Good coverage of moveTab, resizeSplit, insertTab, removeTab, setTabResource, setTabView — but the rename action handler has no test in this file. |
| core/pkg/service/panel/actions.gen.go | Generated Reduce dispatcher and action constructors. Unknown action types are silently skipped (continue) which is appropriate for extensibility. Same misleading ownership-promotion comment on RenamePayload as in actions.go. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant C as Console Client
    participant API as Go API Layer
    participant Svc as Panel Service
    participant DB as gorp / DB
    participant Sig as Signals Provider
    participant Peers as Other Clients

    C->>API: POST /panel/dispatch
    API->>API: RBAC enforce ActionUpdate
    API->>Svc: Writer.Dispatch(key, dispatchKey, acts)
    Svc->>DB: BEGIN TX ChangeErr
    DB->>DB: Reduce(panel, acts...)
    DB->>DB: validateTree(root)
    DB->>DB: COMMIT
    Svc->>Sig: state.Notify(key, dispatchKey, acts)
    Sig-->>Peers: sy_panel_set broadcast
    API-->>C: 200 OK

    C->>API: POST /panel/create
    API->>API: RBAC enforce ActionCreate
    API->>Svc: Writer.CreateMany(panels)
    Svc->>DB: Insert panel row
    Svc->>DB: DefineResource ontology
    Svc->>DB: DefineRelationship parent-of
    API-->>C: 200 panels with keys
```

1. `client/ts/src/panel/actions.ts`, line 121-123 ([link](https://github.com/synnaxlabs/synnax/blob/8b1e50c975414de885b755e52ab1ff9babcd09c4/client/ts/src/panel/actions.ts#L121-L123)) 

   <a href="#">P1</a> **Invalid index silently falls back to append; Go reducer returns an error**

   The Go `MoveTabPayload.Handle` returns `ErrIndexOutOfRange` when `idx < 0 || idx > len(targetLeaf.Tabs)`. Here the TS reducer appends the tab instead. If a caller sends a `moveTab` action with an out-of-range index, the optimistic client state will show the tab appended at the end while the server rejects the action entirely, producing a divergent state between client and server.
</details>

<sub>Reviews (2): Last reviewed commit: ["SY-4354: Merge SY-3999 Member-Named Inli..."](https://github.com/synnaxlabs/synnax/commit/fbdf4ed7718792c76040e1b70f9e15ac1a9a40dc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36745958)</sub>


core/pkg/api/layer.go
```diff
@@ -477,6 +490,12 @@ func (l *Layer) BindTo(t Transport) {
 	t.LinePlotRetrieve.BindHandler(l.LinePlot.Retrieve)
 	t.LinePlotDelete.BindHandler(fgorp.CreateWriteUnaryHandler(db, l.LinePlot.Delete))
 
+	// PANEL
+	t.PanelCreate.BindHandler(l.Panel.Create)
+	t.PanelRetrieve.BindHandler(l.Panel.Retrieve)
+	t.PanelDelete.BindHandler(l.Panel.Delete)
+	t.PanelDispatch.BindHandler(l.Panel.Dispatch)
```
pjdotson: Create/Delete/Dispatch should use the `fgorp.CreateWriteUnaryHandler`

schemas/panel.oracle
```diff
@@ -0,0 +1,236 @@
...
+
+Tab struct {
+    key      uuid          {
+        @doc value "is the stable unique identifier of this tab within the panel. It is independent of the tab's content, so a tab's content may be swapped without changing the tab's identity or position."
+    }
+    resource ontology.ID?? {
+        @doc value "is the visualization resource displayed by this tab, set via SetTabResource. Exactly one of resource or view is set; when both are null the tab renders the visualization selector at render time. Setting one clears the other."
+    }
+    view     TabView??     {
+        @doc value "is the inline view displayed by this tab, set via SetTabView. Exactly one of resource or view is set; when both are null the tab renders the visualization selector at render time. Setting one clears the other."
```
pjdotson: line lengths please

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,125 @@
...
+ * (a backing core document, e.g. a line plot) or a view (an inline,
+ * self-describing app-view, e.g. docs). Exactly one of resource or
+ * view is set; when both are null the tab renders the visualization
+ * selector. Display attributes (name, icon, closability) are resolved
+ * at render time from the content. The same content may be referenced
+ * by multiple tabs in the same or other panels.
+ */
+export const tabZ = z.object({
+  /** key is the stable unique identifier of this tab within the panel. It is independent of the tab's content, so a tab's content may be swapped without changing the tab's identity or position. */
+  key: z.uuid(),
```
pjdotson: does this need to be a full UUID?

pjdotson: why not a counter for this for example

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,125 @@
...
+ * TabView is a self-describing, inline view displayed by a tab. Unlike a
+ * resource, a view has no backing core document: it carries its own
+ * type and opaque args. Used for app-views and tools (docs, explorers,
+ * about, the visualization picker).
+ */
+export const tabViewZ = z.object({
+  /** type is the Console-owned view type identifier (e.g., 'docs', 'about') used to select a renderer. */
+  type: z.string(),
+  /** name is the human-readable tab name for the view. A view has no backing resource to derive a name from, so it carries its own. May be renamed via SetTabView; when empty the Console falls back to a type-derived default. */
+  name: z.string().optional(),
```
pjdotson, line 61: what are example of when this is empty?

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,125 @@
...
+ */
+export const tabZ = z.object({
+  /** key is the stable unique identifier of this tab within the panel. It is independent of the tab's content, so a tab's content may be swapped without changing the tab's identity or position. */
+  key: z.uuid(),
+  /** resource is the visualization resource displayed by this tab, set via SetTabResource. Exactly one of resource or view is set; when both are null the tab renders the visualization selector at render time. Setting one clears the other. */
+  resource: ontology.idZ.optional(),
+  /** view is the inline view displayed by this tab, set via SetTabView. Exactly one of resource or view is set; when both are null the tab renders the visualization selector at render time. Setting one clears the other. */
+  view: tabViewZ.optional(),
+});
+export interface Tab extends z.infer<typeof tabZ> {}
```
pjdotson: TypeScript type should be `{key: string; resource: ontology.ID} | {key: string; view TabView}`

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,125 @@
...
+});
+export interface Leaf extends z.infer<typeof leafZ> {}
+
+/** Split is an interior split node dividing its area between two children. */
+export interface Split {
+  direction: spatial.Direction;
+  size: number;
+  first?: Node;
+  last?: Node;
+}
```
pjdotson: shouldn't these be required in TS types then?

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,125 @@
...
+export const splitZ: z.ZodType<Split> = z.object({
+  /** direction is the axis along which this node is split. */
+  direction: spatial.directionZ,
+  /**
+   * size is the fraction in [0, 1] of the parent area allocated to first.
+   * The remainder is allocated to last.
+   */
+  size: z.number(),
```
pjdotson: this should be bounded

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,125 @@
...
+  resource: ontology.idZ.optional(),
+  /** view is the inline view displayed by this tab, set via SetTabView. Exactly one of resource or view is set; when both are null the tab renders the visualization selector at render time. Setting one clears the other. */
+  view: tabViewZ.optional(),
+});
+export interface Tab extends z.infer<typeof tabZ> {}
+
+/** Leaf is a leaf node in the panel tree displaying a tab strip. */
+export const leafZ = z.object({
+  /** tabs is the ordered list of tabs in this leaf. */
+  tabs: array.nullishToEmpty(tabZ),
```
pjdotson, line 111: shouldn't there be some type of requirement that there is at least one tab in a leaf?

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Node is a node in the panel tree. Exactly one of leaf or split is set
+ * (the other is null); nodes are identified by path-derived numeric
+ * keys during traversal (1 = root, 2k = first child, 2k+1 = last child).
+ */
+export interface Node {
+  leaf?: Leaf;
+  split?: Split;
+}
```
pjdotson: type needs to be `{leaf: Leaf} |{split: Split}`

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * size is the fraction in [0, 1] of the parent area allocated to first.
+   * The remainder is allocated to last.
+   */
+  size: z.number(),
+  /** first is the first child (left for x, top for y). Always present at runtime; double-optional only for Go pointer indirection over the recursive type. */
+  get first() {
+    return nodeZ.optional();
+  },
+  /** last is the second child (right for x, bottom for y). Always present at runtime; double-optional only for Go pointer indirection over the recursive type. */
```
pjdotson: are comments like "double-optional only for Go pointer indirection over the recursive
type" really that great documentation on the field?

client/ts/src/panel/client.ts
```diff
@@ -0,0 +1,103 @@
...
+const retrieveReqZ = z.object({
+  keys: keyZ.array().optional(),
+  searchTerm: z.string().optional(),
+  offset: z.int().optional(),
+  limit: z.int().optional(),
+});
+export interface RetrieveRequest extends z.infer<typeof retrieveReqZ> {}
+const createReqZ = z.object({
+  parent: ontology.idZ.optional(),
+  panels: newZ.array(),
```
pjdotson: parent should probably be a field on a panel in the TS client. Then create request would
just be an array of devices.

core/pkg/api/panel/panel.go
```diff
@@ -0,0 +1,167 @@
...
+		Subject: subject,
+		Action:  access.ActionCreate,
+		Objects: objects,
+	}); err != nil {
+		return res, err
+	}
+	// No parent -> draft, parented to the creating user.
+	parent := subject
+	if !req.Parent.IsZero() {
+		parent = req.Parent
```
pjdotson: Do we want to show these draft panels and stuff under specific users? does it make sense
to implicitly do this instead of explicitly letting the client calls specify user?

core/pkg/api/panel/panel.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	return res, s.db.WithTx(ctx, func(tx gorp.Tx) error {
+		w := s.internal.NewWriter(tx)
+		for i, p := range req.Panels {
+			if err := w.Create(ctx, &p, parent); err != nil {
```
pjdotson: should have a `CreateMany`

core/pkg/service/panel/tree.go
```diff
@@ -0,0 +1,287 @@
...
+// node at that path is a leaf.
+var ErrNotASplit = errors.New("node at path is not a split")
+
+// ErrTabNotFound is returned when an action references a tab key that does not exist
+// in the tree.
+var ErrTabNotFound = errors.New("tab not found in tree")
+
+// ErrIndexOutOfRange is returned when an InsertTab index exceeds the leaf's current
+// tab count.
+var ErrIndexOutOfRange = errors.New("index out of range")
```
pjdotson: are we testing against these errors in prod code or no?

client/ts/src/panel/tree.ts
```diff
@@ -0,0 +1,104 @@
...
+
+// Nodes in the panel tree are identified by path-derived numeric keys: the root
+// is ROOT_PATH, and a split's children are childPath(key, "first" | "last").
+// The scheme is shared with the Go and TypeScript reducers; every consumer that
+// addresses nodes (dispatching InsertTab/SplitLeaf, adapting the tree for
+// rendering) must use these helpers rather than re-deriving the math.
+export const ROOT_PATH = 1;
+
+/** childPath returns the path key of a split's child on the given side. */
+export const childPath = (pathKey: number, side: "first" | "last"): number =>
```
pjdotson: first/last should be a type

client/ts/src/panel/tree.ts
```diff
@@ -0,0 +1,104 @@
...
+  root: N | undefined | null,
+  pathKey: number,
+): N | null => {
+  if (root == null) return null;
+  let n: N = root;
+  for (const isLast of pathDirections(pathKey)) {
+    if (n.split == null) return null;
+    const next = isLast ? n.split.last : n.split.first;
+    if (next == null) return null;
+    n = next as N;
```
pjdotson: type assertion is bad

client/ts/src/panel/actions.ts
```diff
@@ -0,0 +1,222 @@
...
+  if (survivor == null) return;
+  n.leaf = survivor.leaf;
+  n.split = survivor.split;
+};
+
+// directionAndSideForLocation maps a spatial.Location onto the (direction, side)
+// pair that places a new empty leaf on that side of the original. Returns null
+// for locations that do not divide the area in two (e.g., "center").
+const directionAndSideForLocation = (
+  loc: spatial.Location,
```
pjdotson, line 84: will we ever pass in center to this function?

client/ts/src/panel/actions.ts
```diff
@@ -0,0 +1,222 @@
...
+};
+
+export const reduceAll = createReduceAll(handlers);
+
+// All current panel actions are eligible for the undo stack from the substrate's
+// perspective. Inverse vectors are returned empty for Phase 1 so the substrate
+// records targets and remote-touched timestamps correctly without effecting
+// undoable replay; richer inverses follow once tree-collapse round-tripping is
+// implemented.
+export const isUndoable = (_action: Action): boolean => true;
```
pjdotson: seems pointless for right now

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,125 @@
...
+    return nodeZ.optional();
+  },
+});
+
+/**
+ * Node is a node in the panel tree. Exactly one of leaf or split is set
+ * (the other is null); nodes are identified by path-derived numeric
+ * keys during traversal (1 = root, 2k = first child, 2k+1 = last child).
+ */
+export interface Node {
```
pjdotson: is there a reason this isn't `Node = Leaf | Split`?

pjdotson (changes requested):

client/ts/src/panel/types.gen.ts
emilbon99: Converted Tab to an oracle discriminated union: variant resource | view | empty, each carrying the stable key. The empty variant covers the no-content state that previously relied on both fields being null (it renders the visualization selector). Generated TS is now a z.discriminatedUnion with per-variant types.

client/ts/src/panel/types.gen.ts
emilbon99: Fixed via the Node union conversion: split children are required (first Node / last Node) in the schema, TS, and Go. The Go side no longer needs pointer indirection since recursion goes through the variant interface, so the double-optionals and their apology comments are gone.

client/ts/src/panel/types.gen.ts
emilbon99: Yeah implemented

client/ts/src/panel/types.gen.ts
emilbon99: Done: Node is a discriminated union now ({variant: "leaf", ...} | {variant: "split", ...}). The tree reducers in Go and TS were rewritten over the variant types, which also eliminated the type assertion in tree.ts. Required merging the SY-3999 union branch into this one, plus a fix to the union TS codegen for recursive unions (pushed to the SY-3999 branch).

client/ts/src/panel/types.gen.ts
emilbon99: We can't use counters because of optimistic client side insertion. Two clients can optimistically insert a tab with key 5 at the same time. Different tabs, different intention, same key. UUIDs don't have that problem. We can shorten to a string with `id.create()` but I felt it was better to use a UUID

client/ts/src/panel/types.gen.ts
emilbon99: No, because the root node can have an empty leaf to display an empty mosaic. I made sure that all  conditions where we shouldn't have an empty array have been guarded at runtime. In many cases if a leaf ends up with no tabs we collapse it with its parent.

core/pkg/api/panel/panel.go
emilbon99: I'm intentionally omitting draft related and a lot of parenting logic in this PR, as that hasn't been fully refined yet.

client/ts/src/panel/types.gen.ts
emilbon99: I made a decision for certain layout renders to define their names statically. For example, there's no reason that the documentation tab needs to store it's name server side. Reduces sync mechanisms needed.

That said, there are currently some data structures (like tasks) that can be renamed before they are configured. To support that we need to allow dynamic rename and we need to store the name somewhere. Again, this data structure is not super fixed (per PR description) and may evolved as I learn more about the right way to do things.

client/ts/src/panel/actions.ts
emilbon99: Yes, we will. I always lean towards moving more logic to the reducer, it's recommended by redux and I think it's really good because you can test against it. We could theoretically add this guard at every call side, but it's better and can be unit tested her.

pjdotson: If we will realistically be passing in `"center"` then you can keep the logic but I am
not sure that this is always the right move. e.g. you could have two types for this
function:

1) `(loc: spatial.Location): { direction: spatial.Direction; side: spatial.Order } | null`
2) `(loc: spatial.OuterLocation): {direction: spatial.Direction; side: spatial.Order }`

I would argue that the second is "better" as if the caller just has an outer location,
they get a smaller type (and therefore more information from the function). Whereas a
lot of times with the first one, they don't have to do a check of `if (loc == "center")`
_before_ calling, but then have to do a check `if (res == null)` after calling, so they
still have do a check, but if the caller knows they have a `location.OuterLocation`,
they can just pass that in and never have to do a check at all.

See this: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/.

Anyways it's not a big deal and this is fine.

client/ts/src/panel/types.gen.ts
pjdotson: long-term then, would we need `name` for a `View`, if we can move things like tasks to
`ResourceTab`? since theoretically the docs and explorers should all have a fixed name.
anyways fine for this PR

schemas/panel.oracle
```diff
@@ -0,0 +1,310 @@
...
+
+ViewTab struct {
+    key  uuid {
+        @doc value """
+            is the stable unique identifier of this tab within the panel. It is
+            independent of the tab's content, so a tab's content may be swapped
+            without changing the tab's identity or position.
+        """
+    }
+    view View {
```
pjdotson: why do `View` and `ViewTab` need to be separate types? e.g. why can't the fields name,
type, args be at the same level as key

schemas/panel.oracle
```diff
@@ -0,0 +1,310 @@
...
+            Cross-panel moves are RemoveTab on the source plus InsertTab on
+            the destination (two dispatches; not atomic).
+        """
+    }
+
+    action SplitLeaf {
+        leaf     int32
+        location spatial.Location
+        size     float64??        {
+            @validate {
```
pjdotson: would be nice to have some type of `Size` type that is (0, 1) so we don't have to redo
the validation a bunch of different places

client/ts/src/panel/types.gen.ts
```diff
@@ -0,0 +1,224 @@
...
+
+export const tabResourceZ = resourceTabZ.extend({
+  variant: z.literal("resource"),
+});
+export interface TabResource extends z.infer<typeof tabResourceZ> {}
+
+export const tabViewZ = viewTabZ.extend({
+  variant: z.literal("view"),
+});
+export interface TabView extends z.infer<typeof tabViewZ> {}
```
pjdotson, line 68: should we really be generating different `ViewTab` and `TabView` in TypeScript? get that
we kinda had to do it in Go but a little odd here.

pjdotson (changes requested):

client/ts/src/panel/types.gen.ts
emilbon99: Solved by implementing inline union definitions in oracle.

schemas/panel.oracle
emilbon99: Solved

pjdotson (approved):

PR #2471, opened 2026-06-10
SY-4359: Task and Task Configuration Split RFC

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4359](https://linear.app/synnax/issue/SY-4359)

## Description

Adds RFC 0042: splitting the task record into a general-purpose envelope and per-type
strongly typed configuration records owned by dedicated core services (starting with
`core/pkg/service/ni`), linked via the standard `parent` ontology relationship.

The motivating problem is that task configs are opaque JSON blobs the core never
parses, so every client migrates config shapes independently (or not at all): the
Console carries permanent zod union-transforms, the Python client silently misreads old
shapes, and the driver fails to parse them. The split moves config schema evolution to
a single server-side migration chain that runs once per cluster at startup.

Key design points:

- The wire format does not change. `Task.Config` becomes a resolved field (RFC 0041
  §4.2): composed into the payload by the API layer, so Console, clients, and driver
  are unaffected in every phase.
- Config services follow the standard service anatomy and the `types/vN/` version
  layout from RFC 0041; old-shaped wire input is normalized with the same per-payload
  decode-and-walk runner used by imex.
- Task types without a registered config provider keep today's embedded-blob behavior,
  preserving driver pluggability.
- Rollout is copy-then-cut across three releases (inert plumbing, bootstrap copy +
  dual-write, cut), so every release before the last is downgrade-safe.
- Implementation phases call out the hard sequencing constraints: union support in the
  migrate plugin's auto-copy generator, and PRs #2416/#2433 merging before the NI
  service phase.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds RFC 0042, which defines splitting the task record into a general-purpose envelope and per-integration typed configuration records owned by dedicated core services, with the wire format preserved unchanged via API-layer composition. The RFC is thorough and well-motivated, but two design gaps in §4.2 and §4.6 need resolution before implementation begins.

- The `ConfigProvider.Load` method takes a task `Key` but configs have independent keys (§4.1), leaving the task→config lookup mechanism unspecified — a gap that could yield divergent and potentially slow implementations across providers.
- §4.6 Stage 1 simultaneously requires "one transaction" for the bootstrap pass and "a malformed config must never block startup," which are contradictory without an explicit partial-success model.
- The `ConfigProvider` interface lacks a batch-load method, and with config always composed on every retrieve (Decision #8), list operations over many tasks will yield N serial loads.

The RFC is safe to publish as a draft but has two design contradictions that will cause implementation ambiguity if not resolved before Phase 1 coding begins.

The bootstrap migration section simultaneously mandates a single atomic transaction and promises that malformed configs never block startup — these constraints cannot both hold without an explicit partial-success model. Separately, the Load method lookup mechanism is entirely undefined despite config records having independent keys, which is the core retrieval path every task fetch exercises. Both gaps are in the design document and need to be addressed before implementation phases start.

docs/tech/rfc/0042-260610-task-config-split.md — specifically §4.2 (ConfigProvider.Load) and §4.6 Stage 1 (bootstrap transaction model)

| Filename | Overview |
|----------|----------|
| docs/tech/rfc/0042-260610-task-config-split.md | New RFC defining the task/config split architecture; contains an internal contradiction in the bootstrap migration design (atomicity vs. skip-on-failure) and leaves the Load lookup mechanism unspecified despite config records having independent keys. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client
    participant APILayer
    participant TaskWriter
    participant ProviderRegistry
    participant ConfigProvider
    participant GorpTx
    participant OntologyService

    Note over Client,OntologyService: Task Create Flow
    Client->>APILayer: CreateTask(task + embedded config)
    APILayer->>APILayer: Strip config from payload
    APILayer->>TaskWriter: Write(tx, task)
    TaskWriter->>ProviderRegistry: Lookup(task.Type prefix)
    alt Provider registered
        ProviderRegistry-->>TaskWriter: ConfigProvider
        TaskWriter->>ConfigProvider: Create(ctx, tx, task.Key, task.Type, cfg)
        ConfigProvider->>GorpTx: Insert typed config record
        ConfigProvider-->>TaskWriter: ontology.ID
        TaskWriter->>OntologyService: DefineRelationship(task → parent → config)
    else No provider (fallback)
        TaskWriter->>GorpTx: Store embedded blob on task record
    end
    TaskWriter-->>APILayer: OK
    APILayer-->>Client: Task (wire format unchanged)

    Note over Client,OntologyService: Task Retrieve Flow
    Client->>APILayer: RetrieveTask(key)
    APILayer->>GorpTx: Fetch task record
    GorpTx-->>APILayer: Task (no stored config)
    APILayer->>ProviderRegistry: Lookup(task.Type prefix)
    alt Provider registered
        ProviderRegistry-->>APILayer: ConfigProvider
        APILayer->>ConfigProvider: Load(ctx, tx, task.Key)
        ConfigProvider-->>APILayer: msgpack.EncodedJSON
    else Fallback
        APILayer->>GorpTx: Read legacy embedded blob
    end
    APILayer->>APILayer: Compose config into Task.Config resolved field
    APILayer-->>Client: Task with embedded config (wire-compatible)
```

<sub>Reviews (1): Last reviewed commit: ["SY-4359: Clarify wire compatibility as t..."](https://github.com/synnaxlabs/synnax/commit/853814ff979726439d822e3488a4f98a5373095e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36753027)</sub>

> Greptile also left **4 inline comments** on this PR.

### DIFF
--- a/docs/tech/rfc/0042-260610-task-config-split.md
+++ b/docs/tech/rfc/0042-260610-task-config-split.md
@@ -32,10 +32,6 @@ server-side migration chain that runs once per cluster at startup.
   provider keep. Identical to today's behavior.
 - **Resolved Field** - A field present on a type and on the wire but excluded from the
   storage codec, filled by the API layer (RFC 0041 §4.2). `Task.Config` becomes one.
-- **Bootstrap Migration** - The one-time startup migration that copies embedded config
-  blobs into typed config tables.
-- **Dual-Write** - The transitional release in which config writes update both the typed
-  record and the legacy embedded blob, keeping server downgrades safe.
 
 # 2 - Motivation
 
@@ -125,9 +121,6 @@ Config records have their own keys, independent of the task key. Independent ide
 keeps config lifecycle decoupled from task lifecycle: snapshots, history, and templated
 configs become representable later without schema surgery.
 
-Config writes validate against the generated schema at the write seam, so a malformed
-config is rejected at create time instead of surfacing as a driver error.
-
 ## 4.2 - The Provider Registry
 
 The task service exposes a registry mapping task-type prefixes to config providers:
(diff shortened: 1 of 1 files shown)


emilbon99:
Closing per the RFC reconciliation (written by Claude on Emiliano's behalf). This RFC and RFC 0051 (#2660) target the same split; we reconciled the two and every decision landed on 0051's design: config record with its own UUID and a `config parent_of task` edge, closed task-type set, internal tasks included, composition in the task service, and generated version chains. RFC 0051 was amended on its branch to carry the two decisions that went the other way: stored configs migrate through boot-time rewrite passes (not on decode), and the Oracle groundwork phase is complete (cross-file union bases verified on rc in e07ca75d6f).

PR #2479, opened 2026-06-11
SY-4356: Migrate Existing Project Layouts to Panels

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4356](https://linear.app/synnax/issue/SY-4356)

## Description

Migrates existing users' saved workspace arrangements into the new Panel resource. Before this PR, a project's tiling lived only in its legacy `layout` blob (the Console's persisted layout slice), which the panel-based Console will not read — an upgrading user would open their project to an empty canvas.

A new `v56_migrate_project_layouts_to_panels` migration runs when the panel table opens (after the workspace-to-project lift, since the project service opens first). For each project it parses the blob's `mosaics` and `layouts` records and converts every window mosaic into a panel parented under both the project and the root Panels group:

- Mosaic splits map onto panel `Split` nodes; directions are preserved and sizes outside the `(0, 1)` fraction range fall back to an even split (legacy mosaics persisted sizes in several units over time).
- Mosaic tabs whose layout type is a server-backed visualization (`arc`, `lineplot`, `log`, `schematic`, `table` — where the layout key is the document key) become resource tabs with fresh tab identities. Tabs with no layout entry, with a non-document layout type (inline app views), or whose backing document no longer exists are dropped, and splits that lose a side collapse into the surviving child.
- Panels are named from the window's layout entry (`Main` for the main window), falling back to the window key.
- Blobs that fail to parse are logged and skipped, so a corrupt layout cannot block an upgrade. The whole migration commits atomically with the table open, matching the v56 pattern.

The Panels group is now created before the panel table opens so the migration can close over its ontology ID.

**Intentionally additive:** the `layout` field stays on `Project` because the Console at this point in the stack still renders the Redux mosaic and syncs the blob. The field's removal belongs to the Console panels slice; landing both in the same release train means the blob is converted exactly once at upgrade.

Stacked on `sy-3298` (require active project).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR introduces a one-shot `v56_migrate_project_layouts_to_panels` migration that runs when the panel table opens and converts each project's legacy `layout` blob into first-class `Panel` entities.

- **`migrate.go`** — Parses each project's legacy mosaic JSON, recursively converts split/leaf nodes into the new `Node` tree, drops non-document tabs and tabs whose backing resource no longer exists, collapses splits that lose one side, and writes each window mosaic as a `Panel` with correct ontology resource and parent relationships. Blobs that fail to parse are logged and skipped so a corrupt layout cannot block the upgrade.
- **`service.go`** — Group creation is moved before `gorp.OpenTable` so the `Panels` group ID is available when the migration closure is captured at table-open time.
- **`migrate_test.go`** — Three scenarios are covered: multi-window mosaics with stale/non-document tabs, split collapse, and corrupt/empty/unmigrable layouts, plus a no-projects baseline.

The migration is additive, atomic with the table open, and skips unparseable blobs — an upgrading cluster cannot be bricked by corrupt legacy data.

The migration logic, error handling, and test coverage are thorough. The only issue is that the sortedWindowKeys comparator returns -1 instead of 0 when both arguments equal mainWindowKey, technically violating the slices.SortFunc contract. This is harmless today (map keys are always unique) but is a latent trap if the comparator is ever reused outside that guarantee.

core/pkg/service/panel/migrate.go — specifically the sortedWindowKeys comparator

| Filename | Overview |
|----------|----------|
| core/pkg/service/panel/migrate.go | New migration that converts legacy project layout blobs into Panel entities; logic is correct and well-commented, with one minor comparator contract violation in sortedWindowKeys. |
| core/pkg/service/panel/migrate_test.go | Comprehensive test suite covering the happy path, split collapse, corrupt/empty layouts, and the no-projects baseline; test design is sound. |
| core/pkg/service/panel/service.go | Group creation moved before table open so the migration can close over the group ID; change is minimal and the ordering comment clearly explains the intent. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[OpenService] --> B[CreateOrRetrieve 'Panels' group]
    B --> C[OpenTable with migrations]
    C --> D{v56 migration already run?}
    D -- yes --> E[Skip]
    D -- no --> F[collectProjects]
    F --> G{Layout empty?}
    G -- yes --> H[Skip project]
    G -- no --> I[Unmarshal legacySlice]
    I --> J{Parse error?}
    J -- yes --> K[Log warn, skip project]
    J -- no --> L[sortedWindowKeys - main first]
    L --> M[convertNode recursively]
    M --> N{Node has children?}
    N -- yes --> O[Recurse First + Last]
    O --> P{Both sides nil?}
    P -- yes --> Q[Return nil - drop mosaic]
    P -- one nil --> R[Collapse to surviving child]
    P -- both alive --> S[Return Split node]
    N -- no --> T[Filter tabs]
    T --> U{Tab in migratableLayoutTypes?}
    U -- no --> V[Drop tab]
    U -- yes --> W{Resource exists in ontology?}
    W -- no --> X[Drop tab]
    W -- yes --> Y[Add Tab with fresh key]
    Y --> Z{tabs empty?}
    Z -- yes --> AA[Return nil]
    Z -- no --> AB[Return Leaf node]
    S --> AC[Create Panel]
    AB --> AC
    AC --> AD[Write Panel to KV]
    AD --> AE[Write OntologyResource]
    AE --> AF[Write parent rels: group + project]
```

<sub>Reviews (1): Last reviewed commit: ["SY-4356: Migrate Existing Project Layout..."](https://github.com/synnaxlabs/synnax/commit/71a9c3db8cbff3d558335e30fc26570763434a3e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36875450)</sub>


core/pkg/service/panel/service.go
```diff
@@ -81,16 +81,25 @@ func OpenService(ctx context.Context, configs ...ServiceConfig) (s *Service, err
 	s = &Service{cfg: cfg, state: actions.NewState[Key, Action]()}
 	cleanup, ok := service.NewOpener(ctx, &s.closer)
 	defer func() { err = cleanup(err) }()
+	if s.group, err = cfg.Group.CreateOrRetrieve(ctx, "Panels", ontology.RootID); !ok(err, nil) {
```
pjdotson: why do we have a Panels group - what is that used for?

core/pkg/service/panel/migrate.go
```diff
@@ -0,0 +1,309 @@
...
+		}
+		return nil
+	}
+}
+
+// scanStagedLayouts gathers every staged layout before any are converted, since
+// writing panels while iterating the staging prefix would be unsafe. Blobs that fail
+// to parse are skipped with a warning.
+func scanStagedLayouts(
+	ctx context.Context,
```
pjdotson: unused parameters

pjdotson (changes requested):

core/pkg/service/panel/migrate.go
```diff
@@ -0,0 +1,306 @@
...
+	"github.com/synnaxlabs/x/kv"
+	"github.com/synnaxlabs/x/spatial"
+	"go.uber.org/zap"
+)
+
+// migratableLayoutTypes maps the Console layout types whose layout key is the key of
+// a backing core document to that document's ontology resource type. Tabs of any
+// other layout type (inline app views, modals) have no server-side document to
+// reference and are dropped by the migration.
+var migratableLayoutTypes = map[string]ontology.ResourceType{
```
pjdotson, line 35: should we make a note to add tasks to this after they get completed?

pjdotson (approved):
approved after comment addressed

PR #2481, opened 2026-06-11
SY-4340: LabVIEW Client MVP

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4340](https://linear.app/synnax/issue/SY-4340/add-labview-writer-integration)

## Description

Adds an MVP LabVIEW client for Synnax, built in two layers:

- `client/clib/`: a flat C-ABI shared library (`synnax_clib`) that wraps the C++ client and exposes a stable, LabVIEW-friendly surface (client open/close/version, channel retrieve, and a writer). Built as a Win64 DLL.
- `client/lv/`: a LabVIEW client packaged as a PPL, organized as `Synnax Client` and `Synnax Writer` LVOOP classes that call the DLL through Call Library Function Nodes.

Highlights:

- Polymorphic `Write` supporting 12 data types across 4 shapes (1Ch/1S, 1Ch/Ns, NCh/1S, NCh/NS).
- Both write paths validate their buffers before reading. `synnax_writer_write` takes a `data_size` and rejects any buffer that is not exactly `channel_count * sample_count * density` bytes; `synnax_writer_write_strings` takes a uint32 little-endian length-prefixed buffer plus its byte length, avoiding delimiter ambiguity and guarding against out-of-bounds reads.
- Channel retrieve returns each channel's key, index-channel key, and data type in one call.
- Example VIs for each write shape.

Testing:

- A full C-ABI test suite under `client/clib/` following the C++ client's gtest/xtest conventions: `synnax_test.cpp` (client open/close/version), `framer/framer_test.cpp` (writer null guards, fixed-width and string round-trips, multi-channel packing, undersized and trailing-byte buffer rejection), and `channel/channel_test.cpp` (key/index/dtype resolution, not-found handling, output-buffer bounds).
- Server-free cases run without a cluster; round-trip cases write through the DLL and read back over a live local cluster.

CI:

- `client/clib/**` is wired into `test.client.yaml`: the `cpp` job runs `clang-format` over `client/clib` and runs the clib tests under `--config=asan-leak`.

Limitations (MVP):

- Write and retrieve only. Streaming is not yet supported.
- The DLL is built locally and CLFN paths are filename-only for portability.

image

## Basic Readiness

<h3>Greptile Summary</h3>

Adds a new LabVIEW client MVP for Synnax, implemented as a flat C-ABI shared library (`client/clib/`) and a LabVIEW PPL (`client/lv/`). The C layer wraps the existing C++ client with consistent null-guard/try-catch/error-filling patterns and is wired into the existing `cpp` CI job for formatting checks and ASAN-leak tests.

- **C-ABI library** (`synnax.cpp`, `framer/framer.cpp`, `channel/channel.cpp`): exposes client open/close/version, channel name-to-key resolution, and a writer supporting 12 fixed-width types plus strings, with buffer size validation on both write paths.
- **Test suite**: covers null-guard validation, buffer size mismatch and trailing-byte rejection, fixed-width and string round-trips, and multi-channel packing; server-free cases run without a cluster.
- **LabVIEW shim** (`synnax_labview.h`): provides a re-spelled header for the Import Shared Library wizard that spells `size_t` as `uint64_t` and opaque handles as `void*`; `SynnaxError` is intentionally passed as NULL for the MVP (error detail parsing is deferred).

Safe to merge; the C-ABI layer is internally consistent and the previously identified defects (missing data_size guard on the fixed-width write path, handle leak on client open failure) have both been addressed.

All fallible entry points validate arguments, wrap their bodies in try/catch(...), and report through the same clear_err/set_err helpers. Both write paths now enforce exact buffer-size matching. The client open path uses unique_ptr to prevent leaks on connectivity failure. Tests cover null guards, buffer-size rejection, fixed-width and string round-trips, multi-channel packing, and auto-index mode. No logic bugs were found in this pass.

No files require special attention. The synnax_labview.h shim's SynnaxError-as-NULL limitation is explicitly documented as deferred work.

| Filename | Overview |
|----------|----------|
| client/clib/framer/framer.cpp | Implements writer open/write/commit/close with consistent null guards, buffer-size validation on both write paths, and exception-safe cleanup. Fixed-width and string paths are both well-guarded. |
| client/clib/channel/channel.cpp | Channel name-to-key resolution with '\n'-delimited input; handles missing channels with aligned empty slots and aggregates all misses into a single not-found error. |
| client/clib/synnax.cpp | Client open/close/version with unique_ptr-guarded allocation, connectivity check, and TLS validation; no handle leaks on error path. |
| client/clib/synnax_labview.h | LabVIEW wizard shim that maps opaque types to void* and size_t to uint64_t; SynnaxError is void* (always NULL from LabVIEW callers) — error detail inspection deferred as an acknowledged MVP limitation. |
| client/clib/framer/framer_test.cpp | Comprehensive test suite covering null guards, fixed-width/string round-trips, multi-channel packing, buffer-size mismatch, trailing-byte rejection, and auto-index writer. |
| client/clib/channel/channel_test.cpp | Tests cover null client, name_count mismatch, key/index/dtype resolution, missing-channel alignment, buffer-too-small, and null optional outputs. |
| client/clib/BUILD.bazel | Defines cc_library (clib), cc_shared_library (synnax_clib with .def), and three cc_test targets; SYNNAX_BUILDING_DLL is a local_define so test headers see no annotation on Linux, which is correct for the Ubuntu-only CI. |
| .github/workflows/test.client.yaml | Adds client/clib to path triggers, cache key, clang-format check, and ASAN-leak test step in the existing cpp job. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant LV as LabVIEW VI
    participant DLL as synnax_clib (C ABI)
    participant CPP as C++ Client
    participant SRV as Synnax Server

    LV->>DLL: synnax_client_open(host, port, ...)
    DLL->>CPP: synnax::Synnax(config)
    CPP->>SRV: connectivity check
    SRV-->>CPP: CONNECTED
    DLL-->>LV: "SynnaxClient* handle + CODE_OK"

    LV->>DLL: synnax_channel_retrieve_keys(client, names, ...)
    DLL->>CPP: channels.retrieve(names)
    CPP->>SRV: gRPC retrieve
    SRV-->>CPP: channel list
    DLL-->>LV: out_keys[], out_index_keys[], out_dtypes + CODE_OK

    LV->>DLL: synnax_writer_open(client, start, channels, ...)
    DLL->>CPP: telem.open_writer(config)
    CPP->>SRV: open writer stream
    DLL-->>LV: "SynnaxWriter* handle + CODE_OK"

    LV->>DLL: synnax_writer_write(writer, channels, data, data_size, ...)
    Note over DLL: validates data_size == ch_count*samp_count*density
    DLL->>CPP: writer.write(frame)
    CPP-->>DLL: ok

    LV->>DLL: synnax_writer_commit(writer, ...)
    DLL->>CPP: writer.commit()
    CPP->>SRV: commit frame
    SRV-->>CPP: end_ts
    DLL-->>LV: CODE_OK + end_ts

    LV->>DLL: synnax_writer_close(writer, ...)
    DLL->>CPP: writer.close() + delete
    LV->>DLL: synnax_client_close(client)
    DLL->>CPP: delete SynnaxClient
```

<sub>Reviews (2): Last reviewed commit: ["SY-4340: Fix client handle leak on faile..."](https://github.com/synnaxlabs/synnax/commit/875ccd0d27b19a1b8d3cdf70c7ff2cd3071bbcec) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37023169)</sub>


client/lv/lvproj/lvlib/classes/Writer/Private/Writes/U8/Write NCh 1Samp (U8).vi
pjdotson, line 1: Can you look into Git LFS for all of these LabVIEW non-text file types and set it up? This will clobber our Git performance, especially if we end up using more of these things.

client/clib/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_cc//cc:cc_shared_library.bzl", "cc_shared_library")
+
+cc_library(
+    name = "clib",
+    srcs = [
+        "channel/channel.cpp",
```
pjdotson, line 8: Should probably use `.c` extension for pure C files

client/clib/synnax_test.cpp
```diff
@@ -0,0 +1,32 @@
...
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+#include <cstring>
+
+#include "gtest/gtest.h"
+
+#include "client/clib/synnax.h"
+
+TEST(ClibVersion, ReturnsNonEmptyVersion) {
```
pjdotson: we should have actual tests for the C client in CI

client/clib/synnax.cpp
```diff
@@ -0,0 +1,95 @@
...
+            }
+            c.ca_cert_file = ca;
+        }
+        const std::string client_cert = str_or(client_cert_file, "");
+        if (!client_cert.empty()) c.client_cert_file = client_cert;
+        const std::string client_key = str_or(client_key_file, "");
+        if (!client_key.empty()) c.client_key_file = client_key;
+        if (max_retries != 0) c.max_retries = max_retries;
+        if (clock_skew_threshold > 0)
+            c.clock_skew_threshold = x::telem::TimeSpan(clock_skew_threshold);
```
pjdotson, line 64: importing x will cause errors won't it?

pjdotson (changes requested):
C code is all super jank. Would like to see if you can set up a CI so this is at least formatted and somewhat tested. Fine for right now but will need to clean this up a lot if we want to actually publish this.

client/clib/BUILD.bazel
sy-nico: These wrap the C++ client and are genuine `.cpp` files (`new SynnaxClient`, `std::move`, `std::vector`), so they have to compile as C++. 

It's the ABI and headers that have to be in C. Renaming them would fail to compile.

client/clib/synnax.cpp
sy-nico: No, `x::telem` is the C++ client's own namespace. Its public header (`client/cpp/synnax.h`) uses `x::telem::TimeSpan` identically; it's pulled in transitively via `internal.h`. 

I've also compiled and cleanly linked it into the DLL multiple times without any issues.

client/clib/synnax_test.cpp
sy-nico: Added tests, but I don't want to add this into CI yet. It's still an MVP and not worth the upkeep yet unless we determine this is worth pursuing. Part of the CI infrastructure will involve publishing to VIPM.

sy-nico: Never mind. I'm just going to piggy back off of thje cpp job in `test.client.yaml`.

client/clib/channel/channel.cpp
```diff
@@ -0,0 +1,116 @@
...
+        for (size_t i = 0; i < name_count; i++) {
+            if (i > 0) types += '\n';
+            const auto it = by_name.find(req[i]);
+            if (it == by_name.end()) {
+                out_keys[i] = 0;
+                if (out_index_keys != nullptr) out_index_keys[i] = 0;
+                if (!missing.empty()) missing += ", ";
+                missing += req[i];
+            } else {
+                out_keys[i] = it->second->key;
```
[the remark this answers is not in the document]
sy-nico: This one's a false positive. The `\n` separator is prepended once per index (`if (i > 0) types += '\n'`), independent of whether a dtype name is appended, so there are always exactly `name_count - 1` separators and therefore `name_count` slots. A missing channel leaves an *empty* slot rather than dropping one, so the array stays aligned with the requested names.

The example bears this out: `typeA\n\ntypeC` splits into `["typeA", "", "typeC"]`, i.e. three requested names, three slots, with the missing one empty. Behavior is covered by `testReportsMissingChannelWithAlignedDataTypes`.

client/clib/synnax_test.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+TEST(ClibVersion, TruncatesToBuffer) {
+    char buf[4] = {'x', 'x', 'x', 'x'};
+    const int32_t n = synnax_version(buf, sizeof(buf));
+    EXPECT_EQ('\0', buf[sizeof(buf) - 1]);
+    EXPECT_GT(static_cast<size_t>(n), sizeof(buf) - 1);
+}
+
+TEST(ClibVersion, ReturnsLengthWithNullBuffer) {
+    EXPECT_GT(synnax_version(nullptr, 0), 0);
```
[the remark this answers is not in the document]
sy-nico: Updated tests.

client/clib/framer/framer.cpp
```diff
@@ -0,0 +1,291 @@
...
+        const auto *bytes = static_cast<const uint8_t *>(data);
+        for (size_t i = 0; i < channel_count; i++)
+            frame.emplace(
+                channels[i],
+                x::telem::Series(
+                    bytes + i * sample_count * dt.density(),
+                    sample_count,
+                    dt
+                )
+            );
```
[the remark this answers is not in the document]
sy-nico: Fixed. `synnax_writer_write` now takes a `data_size` parameter and validates the buffer before any read, bringing the fixed-width path in line with how `synnax_writer_write_strings` already guards.

Since every fixed-width record is exactly `density()` bytes, the required size is closed-form, so the per-record bounds checks the string path needs collapse into a single comparison:

    `data_size != channel_count * sample_count * density`

Using `!=` (not `<`) matches the string path's strictness: it rejects both undersized buffers (the over-read flagged here) and trailing bytes (the string path's `offset != data_size` case). Covered by `testRejectsDataBufferSizeMismatch` and `testRejectsDataBufferWithTrailingBytes`.

sy-nico:
@greptile, your concerns were addressed. Please update the review score on the PR.

@pjdotson , I decided to go ahead and implement minor CI infrastructure. It was easy enough for minor CI infrastructure (linting + tests). Builds and VIPM publishing are held off until we officially pursue this

On "all super jank": I'd push back on this a bit. The layer is built to one deliberate pattern (every fallible entry point validates args, wraps the body in `try/catch(...)`, and reports through a consistent `clear_err`/`set_err`), so I think it reads as foreign because it's a flat C ABI, not because it's sloppy. That said, the self-audit did turn up two real defects, both now fixed: a missing `data_size` bounds guard on `synnax_writer_write`, and a handle leak in `synnax_client_open` (now `unique_ptr`, with a leak-sanitized test). Happy to keep hardening before publish.

PR #2483, opened 2026-06-12
SY-4297: Arc Params Refactor (Parts 4-7) - Consumers, Stdlib, Runtimes, and Tests

sy-nico:
# Issue Pull Request

## Linear Issue

## Description

Parts 4, 5, 6, and 7 of the Arc Function-Call Unification stack ([RFC 0039](docs/tech/rfc/0039-260513-arc-function-call-unification.md)), combined into a single PR against the base branch. Earlier parts: schema/codegen `Config` removal (Part 1, #2421), symbol hooks + `Trigger` (Part 2, #2447), analyzer collapse (Part 3, #2448).

> **Production conversion plus the test sweep.** Parts 4-6 (the production conversion) and Part 7 (the test sweep, #2490) are now combined in this PR. Only the data/codec migration remains out of scope, deferred to Part 8. The tree still does not fully build on its own by design while that migration is pending; migration snags are flagged at the bottom and handled separately.

This PR carries the refactor from the analyzer (Parts 1-3) through the consumers, the standard library, the LSP, the service layer, and both the Go and C++ runtimes, so every producer and reader of IR now speaks one parameter list with an explicit `Trigger`.

### Part 4 - Downstream consumers

- **Compiler & IR** ([compiler/compiler.go](arc/go/compiler/compiler.go), [ir/function.go](arc/go/ir/function.go), [ir/node.go](arc/go/ir/node.go)): `slices.Concat(Config, Inputs)` becomes plain `Inputs`; `Config` dropped from `Type()` and the tree renderers.
- **Analyzers** ([graph/analyze.go](arc/go/graph/analyze.go), [text/analyze.go](arc/go/text/analyze.go)): node/function builders fold `Config:` fields into `Inputs:`; the text analyzer's `upstreamIsTrigger`/`firstInputParam` heuristic is replaced by a direct `sym.Trigger` consult.
- **`KindConfig` narrowing:** the graph analyzer binds all params as `KindInput`, and `KindInput` absorbs `KindConfig`'s channel behavior. Full enum removal is deferred post-Part-9 (logged in RFC Appendix A).

### Part 5 - Stdlib, LSP, service layer, runtime decoupling

- **Stdlib symbols:** every STL package merges its `Config:` list into `Inputs:` and declares an explicit `Trigger` per the RFC §4.1 audit table (time, math, stable, constant, channels, control, op, errors, selector, series).
- **Go runtime decoupling:** the config/input split is retired at runtime. [runtime/node/state.go](arc/go/runtime/node/state.go) gains name-based input access (`InputNamed`/`InputTimeNamed`/`InitInputNamed`); the WASM path ([stl/wasm](arc/go/stl/wasm)) replaces the `configCount` prefix layout with a per-input `edgeFed` mask (literal-fed inputs set once, wire-fed inputs streamed); host nodes read inputs by name rather than position.
- **LSP** ([lsp/hover.go](arc/go/lsp/hover.go), [lsp/completion.go](arc/go/lsp/completion.go)): signatures render the unified `Inputs`, the function-kind label keys off `Exec` instead of the removed `Config`, and the `Trigger` target is surfaced in hover.
- **Service layer** ([service/arc/status/status.go](core/pkg/service/arc/status/status.go)): the `Config` mirror is dropped in favor of a single `AnalyzeArguments` hook plus an explicit `Trigger`; factory reads `Inputs.ValueMap()`.

### Part 6 - C++ runtime

- **Name-based input access** in [runtime/state](arc/cpp/runtime/state) (`input_named`/`input_time_named`/`init_input_named`).
- **Host nodes** (math, stable, selector, channels) read inputs by name. This fixes the math aggregator's reset input, whose hardcoded positional index shifted under the merged list.
- **WASM decoupling** ([runtime/wasm](arc/cpp/runtime/wasm)): the `Function` `config_count` layout is replaced by a per-input `edge_fed` mask, matching the Go runtime. The node loop now skips literal-fed inputs when computing sample length and timestamps (the old "stream everything" path could zero out output timestamps).
- **`config` to `inputs`** in the constant/control/channels/time factories, and the `ir::Node`/`ir::Function` string formatters drop their `config` section.

A small Go follow-through is also included: the `ir` msgpack legacy decoder ([ir/msgpack.go](arc/go/ir/msgpack.go)) drops its `Config` field, the last non-migration `.Config` reader.

### Part 7 - Test sweep

The deferred fixture sweep and new coverage, merged in from #2490.

- **Fixtures onto the unified `Inputs` list:** node and function literals across the analyzer, graph, text, stdlib, and runtime tests drop their `Config:` fields for a single `Inputs:` list (literal/brace params before wire-fed/parens params); the C++ IR builders move from `node.config` to `node.inputs` to match.
- **Construction-time input resolution:** per-factory tests assert a stdlib node naming an undeclared input fails at load with `NOT_FOUND` (previously `VALIDATION`); both runtimes gain a `ResolveInput`/`resolve_input` unit test.
- **Unified hook signature:** the `AnalyzeArguments` hook tests adopt the Parts 4-6 signature, receiving the diagnostics sink directly.
- **New coverage:** cross-runtime IR constant parity (`ir/cross_runtime_test.go`), `Trigger` binding (`symbol/trigger_test.go`), and WASM node tests exercising the per-input `edge_fed` mask.

### Migrations: deferred, with a known snag

The codec/data migration for the `Config` removal is not in this PR. Generating it ran into trouble worth flagging: an initial `oracle migrate` ran against a stale snapshot baseline and produced a colliding version; after back-merging `rc` to refresh the baseline, re-running it produced broken output across several unrelated packages (missing `migrate_auto.gen.go` for some service packages, custom migrations copied into frozen version dirs without their dependencies, and a type-conversion error in the regenerated IR migration). We discarded that output rather than patch generated files reactively. The migration will be done deliberately in a scoped branch once we have aligned synchronously on the correct oracle workflow for this field removal.

### Reviewer notes

- **Does not build fully on its own, by design.** The only remaining `.Config` readers live in the deferred migration packages. Review on the diff.
- **Test sweep included (Part 7).** The fixture migration and new coverage are now part of this PR; see the Part 7 section above.

## Basic Readiness

_Tests now included via the Part 7 sweep (#2490). Docs readiness left unchecked: the RFC/plan docs landed earlier in the stack._

<h3>Greptile Summary</h3>

This PR completes the Arc `Config`→`Inputs` unification (Parts 4-6) across downstream consumers, the stdlib, LSP, service layer, and both the Go and C++ runtimes. Every producer and reader of IR now uses a single merged `Inputs` list with an explicit `Trigger` field replacing the `Config`/`Inputs` split.

- **Part 4** drops `Config` from the compiler, IR string formatters, graph analyzer, and text analyzer; the text analyzer replaces its `upstreamIsTrigger` heuristic with a direct `sym.Trigger` consult.
- **Part 5** merges `Config` into `Inputs` for all stdlib packages, replaces the WASM `configCount` layout with a per-input `edgeFed` boolean mask, adds name-based `ResolveInput`/`input_named` at construction time in both Go and C++ runtimes, and updates the LSP to render unified signatures.
- **Part 6** applies the same name-based input access to all C++ host nodes and aligns the C++ WASM `Function` with the Go side via `edge_fed` — fixing a timestamp-zeroing bug where the old path could zero output timestamps for literal-fed inputs.

The PR is safe to merge as the prod-side completion of the Config→Inputs unification; no new logic defects were found in the changed paths.

Every changed file follows the same mechanical pattern — Config fields removed, Inputs extended, positional indices replaced with name-based ResolveInput/ir::resolve_input called at construction time so naming mistakes fail at load rather than silently at runtime. The WASM edge_fed mask correctly handles literal vs wire-fed inputs in both Go and C++, and the timestamp-zeroing bug for literal-only nodes is fixed by the longestInputIdx=-1 sentinel. The selector consistently uses ir.DefaultOutputParam as both the declared input name and the resolution key — an unusual but self-consistent convention. No regressions were introduced in the analyzer, LSP, or service-layer paths.

No files require special attention. The intentionally incomplete build (deferred migration packages still reference Config) is called out explicitly and is not a concern for this diff.

| Filename | Overview |
|----------|----------|
| arc/go/runtime/node/state.go | Adds ResolveInput for name-based index lookup at construction time, isReference tracking for channel-type inputs that skip data gating, and moves the no-inputs guard to no-data-inputs. |
| arc/go/stl/wasm/wasm.go | Replaces configCount prefix layout with a per-input edgeFed mask; literal inputs set once at construction, edge-fed inputs streamed per sample in Next(). |
| arc/go/stl/wasm/node.go | longestInputIdx initialised to -1 and skips non-edge-fed slots; fixes timestamp zeroing for literal-only nodes. |
| arc/cpp/runtime/wasm/module.h | Mirrors Go WASM changes: config_count removed, edge_fed vector added, Function constructor size-checks edge_fed against inputs. |
| arc/cpp/stl/math/math.h | All aggregator/derivative/arithmetic nodes switch to ir::resolve_input at construction; reset_idx now optional<size_t> instead of sentinel -1. |
| arc/go/graph/analyze.go | bindParams call for Config removed; config-value loop rewritten to iterate freshType.Inputs. |
| arc/go/text/analyze.go | analyzeFunctionNode drops the upstreamIsTrigger heuristic and uses sym.Trigger.Target directly. |
| arc/go/analyzer/flow/flow.go | Adds isFedByRoutingTable and TriggerOnly early-return for upstream type-checking. |
| core/pkg/service/arc/status/status.go | Replaces split AnalyzeCall+AnalyzeFlowConfig hooks with a single unified AnalyzeArguments. |
| arc/go/ir/msgpack.go | Drops Config from both Function and Node legacy decoder structs. |
| arc/cpp/ir/ir.h | Adds ir::resolve_input; renames lhs/rhs_input_param to a/b to align with Go constants. |
| x/go/telem/series_factory.go | Adds TimeSpan handling in castNumeric, castToString, and castToBytes. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph IR["IR Layer"]
        F["Function: Inputs only, no Config"]
        N["Node: Inputs only, no Config"]
    end
    subgraph Analyzers["Part 4 - Analyzers and Compiler"]
        GA["graph/analyze.go\nbindParams Config removed"]
        TA["text/analyze.go\nsym.Trigger.Target replaces heuristic"]
        CC["compiler.go\nparams = Inputs"]
    end
    subgraph STL["Part 5 - Stdlib Symbols"]
        CH["channels: TriggerOnly / TriggerInput"]
        MA["math: TriggerInput default_input"]
        ST["stable: TriggerInput default_input"]
        TM["time: TriggerOnly"]
        OP["op: TriggerInput lhs"]
    end
    subgraph Runtime["Part 5-6 - Go and C++ Runtimes"]
        RS["ResolveInput / ir::resolve_input\nname-to-index at construction"]
        EF["edgeFed mask\nreplaces configCount prefix"]
        IS["isReference flag\nskips channel params in data gating"]
    end
    subgraph LSP["Part 5 - LSP"]
        HC["hover.go: unified Inputs signature"]
    end
    F --> GA
    F --> TA
    N --> GA
    GA --> Runtime
    TA --> Runtime
    STL --> Runtime
    STL --> LSP
    Runtime --> RS
    Runtime --> EF
    Runtime --> IS
```

1. `arc/go/text/analyze.go`, line 466-479 ([link](https://github.com/synnaxlabs/synnax/blob/67e2a8a33eebaadc85c27b2e96c541f343543a9b/arc/go/text/analyze.go#L466-L479)) 

   <a href="#">P2</a> **Inverted input param ordering vs. symbol definition**

   `buildChannelWriteNode` places `"channel"` at index 0 and `ir.DefaultInputParam` at index 1, but `channels.go`'s `NewSymbols` declares the write symbol with `ir.DefaultInputParam` at index 0 and `"channel"` at index 1. Both the Go and C++ runtimes now build an `inputIndex` map from the position in `n.Inputs`, so name-based lookup works correctly regardless of order. However, for any code path that still creates a WASM `Function` from the node's `Inputs` slice (the `edgeFed` mask is built from index position in `wasm.go`/`factory.h`), the two orderings would produce different WASM argument slot assignments. Because the write node is currently a host node this doesn't cause a runtime failure today, but the inconsistency could silently break things if the node is ever migrated to WASM, and it makes the IR harder to reason about during debugging.
</details>

<sub>Reviews (3): Last reviewed commit: ["SY-4297: Fix comments"](https://github.com/synnaxlabs/synnax/commit/be167ae6d486e0bc74659e6b4888f151db0f3f70) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37142914)</sub>


sy-nico:
@greptile, please update the review. Regarding your concern:

Removed the name-based accessors and moved the check to construction instead of guarding the runtime lookup.

Each node resolves its input names once via `State.ResolveInput(name) (int, error)`. The factory returns that error, so an undeclared name fails at program load with a clear message, never silently at runtime. `Next()` reads through the pre-resolved index, so it cannot miss. Mirrored in C++ with `resolve_input`. Covered by a construction-validation test per factory, so a typo fails CI.

arc/cpp/runtime/state/state.cpp
```diff
@@ -274,6 +280,16 @@ void Node::write_series(
 const Series &Node::input_time(const size_t param_index) const {
     return this->aligned_time[param_index];
 }
+
+std::pair<size_t, x::errors::Error> Node::resolve_input(const std::string &name) const {
+    const auto it = this->input_index.find(name);
+    if (it == this->input_index.end())
+        return {
+            0,
+            x::errors::Error(x::errors::VALIDATION, "node has no input named " + name)
```
emilbon99: Should use a not found error here, not a validation error.

arc/cpp/runtime/state/state.h
```diff
@@ -80,13 +82,15 @@ class Node {
         std::vector<size_t> output_idx,
         std::vector<InputEntry> accumulated,
         std::vector<Series> aligned_data,
-        std::vector<Series> aligned_time
+        std::vector<Series> aligned_time,
+        std::unordered_map<std::string, size_t> input_index
```
emilbon99: Adding this extra data structure does not seem like it's worth the cost. 

1) Adds extra fields, heap allocations, and state to manage.
2) For small sets, a linear scan on a slice is often faster than a map lookup. It's very likely that a linear scan will be much more performant here.

arc/cpp/runtime/state/state.h
```diff
@@ -105,6 +109,11 @@ class Node {
 
     [[nodiscard]] const Series &input_time(size_t param_index) const;
 
+    /// @brief returns the position of the named input, or an error if the node
+    /// has no such param. Resolve at construction so wiring mistakes fail at load.
+    [[nodiscard]] std::pair<size_t, x::errors::Error>
```
emilbon99: Why is this a function on state and not the node configuration itself?

arc/cpp/runtime/wasm/module.h
```diff
@@ -399,6 +398,11 @@ class Module {
 
             return x::errors::NIL;
         }
+
+        /// @brief one flag per input, true when wire-fed (streamed per sample).
+        [[nodiscard]] const std::vector<bool> &input_edge_fed() const {
```
emilbon99: Why isn't the member variable just public? Why do we need this getter?

arc/cpp/runtime/wasm/module.h
```diff
@@ -411,12 +415,20 @@ class Module {
...
 
-    /// @brief Returns a WASM function wrapper for the given function name.
-    /// @param name The function name.
-    /// @param node_config The node's config params with values. If empty, uses the
-    /// function's config.
-    std::pair<Function, x::errors::Error>
-    func(const std::string &name, const types::Params &node_config = {}) {
+    /// @brief Wraps the WASM export `name` using its declared inputs, all wire-fed.
+    /// Use the three-arg overload for nodes with literal/config inputs.
+    std::pair<Function, x::errors::Error> func(const std::string &name) {
```
emilbon99, line 414: What does wire fed mean? Connected by an edge? We shouldn't mangle the term 'edge' and 'wire'. Pick one and stick with it.

core/pkg/service/arc/status/status.go
```diff
@@ -202,36 +201,19 @@ func dispatchSet(
...
-}
-
-func analyzeStatusSetFlowConfig(
-	diags *diagnostics.Diagnostics,
-	config parser.IConfigValuesContext,
-) {
-	if config == nil {
+// analyzeStatusSetArguments validates the variant argument across both call
+// forms: it binds by name ("variant") or, when positional, by index.
+func analyzeStatusSetArguments(ctx any, args []symbol.Argument) {
```
emilbon99: Why is `ctx` any? This code is nasty, and `any` should not be used here. Use the proper type.

core/pkg/service/arc/status/status.go
```diff
@@ -202,36 +201,19 @@ func dispatchSet(
...
+	db, ok := ctx.(interface {
```
emilbon99: Bad type assertion and inline interface declaration.  AFAIK We don't do this anywhere else in the codebase, and definitely shouldn't start doing it here.

arc/go/analyzer/context/context.go
```diff
@@ -105,6 +105,10 @@ type Context[AST antlr.ParserRuleContext] struct {
 	InTypeInferenceMode bool
 }
 
+// Diags returns the diagnostics sink, letting a holder of an untyped context
+// (e.g. an ArgumentsHook's ctx any) reach diagnostics without the AST type param.
+func (c Context[AST]) Diags() *diagnostics.Diagnostics { return c.Diagnostics }
```
emilbon99: This is definitely bad, and this method should not exist.

docs/tech/rfc/0039-260513-arc-function-call-unification.md
```diff
@@ -1134,3 +1134,55 @@ it lands.
...
+### Part 5 - Runtime inputs are addressed by name, retiring the config/input split
+
+The plan scoped the runtime touch as a one-liner (`len(Config)` to `len(Inputs)`), but
+the merge broke more than that: the runtime read inputs by position and split params via
+a `configCount` prefix (`params[configCount+j]`). With the lists unified there is no
+contiguous boundary to recover (`math.avg`'s wire-fed `reset` carries a default value
+while its wire-fed `input` does not), and a host node reading `Input(0)` for its data
+now hits a former config param.
+
+So the split is retired from the runtime, not patched. `node.State` already materializes
```
emilbon99, line 1175: Having `ctx any` is a bad design. I'm not sure what the optimal solution is here, but I am sure there is a better one.

emilbon99 (changes requested):

core/pkg/service/arc/status/status.go
sy-nico: Fixed. Retyped `ArgumentsHook` to take `*diagnostics.Diagnostics` directly, so the hook no longer needs `any`.

core/pkg/service/arc/status/status.go
sy-nico: Gone. With the hook receiving the diagnostics sink directly, the type assertion and inline interface are removed.

arc/go/analyzer/context/context.go
sy-nico: Removed

docs/tech/rfc/0039-260513-arc-function-call-unification.md
sy-nico: I reworked the `AnalyzeArguments` hook to now receive the diagnostics sink directly, so there's no `ctx any` or type switch. RFC updated to match

arc/cpp/runtime/state/state.h
sy-nico: Resolution is now a linear scan over the node's input params, done once at construction, so there's no map, no heap allocation, and no extra member to manage.

arc/cpp/runtime/state/state.h
sy-nico: Changed to be a free function `ir::resolve_input(node, name)` over the IR node's params, since name-to-index depends only on the static IR, not runtime state. Factories resolve against `cfg.node`.

arc/cpp/runtime/wasm/module.h
sy-nico: The member's mutable (it's `std::move`'d in and the mask has to stay parallel to args/inputs, which `call()` relies on), so the getter is there to hand it out read-only as a `const&`. A public member would let a holder reassign it and desync the function; a const public member can't work either since `Function` is copied by value into `wasm::Node`.

sy-nico:
@greptile, Update your review. Concerns have been addressed and feedback from reviewers has been implemented.

arc/cpp/runtime/state/state.h
emilbon99: I really don't understand why this isn't a member function on the `Node` struct? Can you please explain why?

emilbon99 (changes requested):
I don't understand this whole `edge_fed` thing. A particular input param is `edge_fed` if an edge is connected to it? Ok, makes sense.

But what are the other options. If it's not edge fed, is it config fed? Is there a way for an input param to not be edge fed AND not be config fed?

My reason for asking this is: you have all of this weird tracking of `edge_fed` where you're copying vectors and initializing the `edge_fed` vectors in weird places. If it's true that all input params must be either edge fed or config fed, then can't edge fed be the set of all inputs that are not config fed?

If that's true and the config values are defined statically, then aren't the edge fed edges the ones where the param has a statically defined value in the IR? i.e. edge_fed can be a method on the ir node itself htat doesn't need to do all of this nasty `edge_to` stuff to create a vector of bools.

sy-nico:
Quick framing note, because I think it's the crux: there's no "config-fed" category anymore. That was the old split this refactor removed. There's one `Inputs` list, and each input's value comes from one of two sources at runtime: an edge (streamed per sample) or its static default (set once). "Has a default" and "is edge-fed" aren't two halves of one partition. They're independent axes. A param can carry a default AND be wired, in which case the wire overrides the default.

That's why `edge_fed` can't be `value == nil`. `math.avg.reset` declares a default of `0` AND gets wired in `math_test.go`. Under `edge_fed = (value == nil)` it would be treated as a constant `0` and never stream the wire, so the reset would silently never fire. Edge-fed is a property of `(node, edges)`, not of the node alone, which is also why it can't be a node-local method on `ir.Node`.

On the "weird tracking" itself: the mask is one bool per input, built once at construction and read by index after that. Nothing mutates or re-derives at runtime. It was computed inline in two places (the C++ factory and the Go `Create`), plus a duplicated test helper. I pulled that into a single `EdgeFedMask` helper so there's one definition; `Function` and `Node` just carry the precomputed mask.

arc/cpp/runtime/state/state.h
sy-nico: `resolve_input` is behavior, not data. Oracle generates `Node` from the cross-language schema as a pure data type: its fields plus serialization (`parse`/`to_proto`/etc.). It doesn't generate domain logic, and a param lookup is domain logic, not a field on the node.

Making it a method would mean either hand-writing into a generated struct, or pushing a consumer-side runtime lookup into the schema so it's emitted into Go, C++, Python, and TS identically. Neither is right: the lookup isn't a property of the IR data, it's an algorithm a consumer runs over it.

So it's a free function in the hand-written `ir.h`, alongside `node_member`, `scope_member`, and the `Handle`/`Edge` `operator==` overloads, which are all free functions over generated IR types for the same reason. `resolve_input(node, name)` reads the same as `node.resolve_input(name)` would; the difference is just that behavior over generated types lives next to them, not on them.

emilbon99: What do you mean a lookup isn't a property of the IR data? It's a resolved field that's directly derivable from what's stored in the IR. There is a `@cpp methods` rule on `oracle` struct that let's you attach and define methods on a generated schema.

All this is is getting a derived property from a static type. It should be a method. If you still disagree let's talk verbally.

arc/cpp/ir/ir.h
```diff
@@ -17,8 +17,29 @@
...
+        if (node.inputs[i].name == name) return {i, x::errors::NIL};
+    return {
+        0,
+        x::errors::Error(x::errors::NOT_FOUND, "node has no input named " + name)
+    };
+}
+
+/// @brief one flag per input of node: true when an edge feeds it (edge-fed, the edge
+/// wins over any default), false otherwise (literal-fed, the default holds).
+[[nodiscard]] inline std::vector<bool> edge_fed_mask(const IR &prog, const Node &node) {
```
emilbon99: This should also be a method on the IR. Use `@cpp methods`

arc/cpp/runtime/state/state.cpp
```diff
@@ -23,8 +23,11 @@ Series parse_default_value(
 ) {
     auto data_type = type.telem();
     if (value.has_value()) {
-        auto casted = data_type.cast(*value);
-        return x::mem::make_local_shared<x::telem::Series>(casted);
+        // Non-telem kinds (e.g. a channel handle) carry their concrete value
+        // already; only numeric/string kinds need a cast to their telem type.
+        if (data_type == x::telem::UNKNOWN_T)
```
emilbon99: This seems like weird one-off logic that should be consolidated somehow? Why isn't `type.telem()` for a channel a `uint32`? and then this code can just be left as is.

arc/cpp/runtime/wasm/module.h
```diff
@@ -307,33 +308,31 @@ class Module {
...
+            // Literal inputs are set once here; edge-fed inputs stream in call().
+            for (size_t i = 0; i < inputs.size(); i++) {
+                if (this->edge_fed[i] || inputs[i].value.is_null()) continue;
+                if (inputs[i].value.is_string()) {
+                    // A literal string gets a stable handle, not cleared by flush().
+                    // strings is null only in tests without host bindings.
+                    const auto &s = inputs[i].value.get<std::string>();
                     if (module.cfg.strings != nullptr)
                         this->args[i] = wasmtime::Val(
                             static_cast<int32_t>(module.cfg.strings->create_config(s))
```
emilbon99: Why is this method still called `create_config` if we don't do config anymore?

arc/cpp/runtime/wasm/module.h
```diff
@@ -307,33 +308,31 @@ class Module {
...
-                    // String config params get a stable handle created once at
-                    // configure time — not cleared by flush(), no per-call refresh.
-                    // bindings is always non-null in production; the null branch is
-                    // only reachable in tests that construct a Module without WASM
-                    // host bindings, where args[i] stays 0 (unused).
-                    const auto &s = config[i].value.get<std::string>();
+            this->args.resize(inputs.size(), wasmtime::Val(0));
+            // Literal inputs are set once here; edge-fed inputs stream in call().
+            for (size_t i = 0; i < inputs.size(); i++) {
+                if (this->edge_fed[i] || inputs[i].value.is_null()) continue;
```
emilbon99: What happens if the node is not edge fed and the input value is null? Isn't that a programming error?

emilbon99 (changes requested):
The `IR` isn't stored on disk so it doesn't need to be migrated. Delete all migration functionality related to migrating the IR.

emilbon99:
> Quick framing note, because I think it's the crux: there's no "config-fed" category anymore. That was the old split this refactor removed. There's one `Inputs` list, and each input's value comes from one of two sources at runtime: an edge (streamed per sample) or its static default (set once). "Has a default" and "is edge-fed" aren't two halves of one partition. They're independent axes. A param can carry a default AND be wired, in which case the wire overrides the default.
> 
> That's why `edge_fed` can't be `value == nil`. `math.avg.reset` declares a default of `0` AND gets wired in `math_test.go`. Under `edge_fed = (value == nil)` it would be treated as a constant `0` and never stream the wire, so the reset would silently never fire. Edge-fed is a property of `(node, edges)`, not of the node alone, which is also why it can't be a node-local method on `ir.Node`.
> 
> On the "weird tracking" itself: the mask is one bool per input, built once at construction and read by index after that. Nothing mutates or re-derives at runtime. It was computed inline in two places (the C++ factory and the Go `Create`), plus a duplicated test helper. I pulled that into a single `EdgeFedMask` helper so there's one definition; `Function` and `Node` just carry the precomputed mask.

Ok, so why doesn't the compiler purge default values when they aren't used? i.e. if `math.avg.reset` has an edge connected, then the compiler purges its default value back to `nil`. Then both runtimes can completely get rid of the `edge_fed` mask mechanics and simply rely on `value == nil`. This is an IR structure and compiler concern after all, not a runtime concern.

arc/cpp/runtime/state/state.cpp
sy-nico: Parity miss on my end. The channel-as-reference handling landed in the Go runtime this stack but never made it to C++, so a defaulted channel input still flowed through `parse_default_value` and that special case was the band-aid.

Fixed properly: C++ now treats channel inputs as references like Go, so the special case is gone.

arc/cpp/ir/ir.h
sy-nico: Done. `edge_fed_mask` is now a method on `IR` via `@cpp methods`, right next to `edge_to`/`edges_to`, since it needs the edge list off the IR. Did the same for `resolve_input` on `Node`. Bodies live in `ir.cpp` like `IR::edge_to`.

Still getting familiar with the oracle codebase, so I took the round trip to untangle the right way to do this rather than guess. Appreciate the pointer to `@cpp methods`.

arc/cpp/runtime/wasm/module.h
sy-nico: Just a miss. Renamed to `literal`

arc/cpp/runtime/wasm/module.h
sy-nico: Yes, it's a programming error and it's caught at analysis time. `ResolveNodeTypes` rejects any input with no default value and no incoming edge as "missing required input", so it never reaches the runtime. By the time we're here, a null value means the input is edge-fed, so this branch was just defensive.

That branch is gone now anyway: with the default-purge, the runtime keys off `value == nil` directly.

sy-nico:
> Ok, so why doesn't the compiler purge default values when they aren't used? i.e. if `math.avg.reset` has an edge connected, then the compiler purges its default value back to `nil`. Then both runtimes can completely get rid of the `edge_fed` mask mechanics and simply rely on `value == nil`. This is an IR structure and compiler concern after all, not a runtime concern. 

Good suggestion, and done. The compiler now purges an input's default when an edge targets it (`ResolveNodeTypes`), so `value == nil` is the one signal of edge-fed and both runtimes drop the mask and key off the value directly. Agreed it belongs in the IR/compiler, not the runtime.

One meta note since we'll keep iterating on this stack: when you already have the fix in mind, stating it directly ("I think we should purge defaults at compile time") is easier to act on than the "why not X" framing. The question form reads as a challenge and usually costs a round-trip before the real idea surfaces. Good call on this one regardless.

sy-nico:
> The `IR` isn't stored on disk so it doesn't need to be migrated. Delete all migration functionality related to migrating the IR.

Removed the IR migration logic. Stopped before the generated packages since the graph still persists IR function definitions; want those dropped too? That would require a schema change to stop persisting the IR, then a regen.

PR #2485, opened 2026-06-12
M. SY-4299: Strongly-Typed NI Task Types (Part 3 — Driver)

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Makes the C++ driver consume the Oracle-generated NI types, completing the third leg
of the NI strong-typing work (Part 1: schema + generated clients, Part 2: console).
The driver's hand-rolled JSON parsing and its ~50-class virtual-inheritance channel
hierarchy are replaced entirely: the generated wire schemas are now the driver's
channel representation, so the driver, Console, and all clients parse NI task configs
from a single source of truth.

### Architecture

The old `channels.h` fused three concerns per channel class: JSON parsing,
converted-field storage, and the DAQmx `apply()` call. After this PR there are exactly
two kinds of code left:

- **Two generic wrappers** (`Input`/`Output`) holding the parsed generated union
  (`AIChannel`/`CIChannel`/`DIChannel`, `AOChannel`/`DOChannel`) plus the runtime
  state the task binds (resolved Synnax channel, device location).
- **One free `apply` function per channel type** (47 total) that converts strings to
  DAQmx constants inline and makes the hardware call. Dispatch is `std::visit` over
  the generated unions, so adding a schema variant without a driver `apply` is a
  compile error instead of a runtime failure.

`parse_input`/`parse_output` dispatch on the task type through the generated union
parsers, so each task accepts exactly its own channel family at parse time with proper
field errors.

### Polarity

The generated schemas follow the RFC 0043 zero-default boolean rule: channels carry
`disabled` and configs carry `data_saving_disabled`. The driver keeps positive
internal semantics (`Input::enabled`, writer-mode selection) and derives them from the
generated fields. The parsers read only the schema keys: configs stored with the old
`enabled`/`data_saving` keys migrate once at Core boot (the storage cutover PR later
in this chain), so there is no legacy-key handling in the driver.

### Testing

Adds a generated recording `FakeAPI` for the full 397-method DAQmx surface. Every
channel test parses a wire fixture and asserts the **actual DAQmx call and its
arguments** (physical channel string, converted unit/enum constants, scalars).

### Plumbing and newly working

- Adds `//client/cpp/ni`, `//client/cpp/task/common`, and `//driver/ni/daqmx:fake`
  Bazel targets; `//driver/ni/channel` is the first driver consumer of generated
  client types.
- Eight channel types that existed but had no factory mapping (`ai_voltage_rms`,
  `ai_current_rms`, `ai_voltage_with_excit`, `ai_accel_charge`, `ai_thermistor_iex`/
  `_vex`, `ai_rosette_strain_gage`, `ai_freq_voltage`) are now configurable, and
  `CreateAIThrmstrChanVex` (declared but never implemented) now links.
- Table scales now parse (the legacy parse read `pre_scaled`/`scaled` while consoles
  always wrote `pre_scaled_vals`/`scaled_vals`); polynomial scales use the schema's
  `reverse_coeffs` directly; rosette strain gauges pass the full measurement-type
  array to DAQmx instead of only the first entry.
- Counter read tasks support per-channel devices (cross-device), covered by a new
  config test.
- `x::telem::Rate` gains `to_json`/`from_json` so generated configs round-trip rates.

## Basic Readiness


driver/ni/channel/channels.h
```diff
@@ -119,112 +99,93 @@ static int32_t get_ci_z_index_phase(const std::string &s) {
...
-
-    ~PolynomialConfig() {
-        delete[] forward_coeffs;
-        delete[] reverse_coeffs;
+    PolynomialConfig(
+        const ::synnax::ni::BridgePolynomial &b,
+        const std::string &physical_units_str
+    ):
+        forward_coeffs(b.forward_coeffs), reverse_coeffs(b.reverse_coeffs) {
+        const auto eu = channel::UNITS_MAP.find(b.electrical_units);
```
sy-nico: `PolynomialConfig` hand-rolls `units_or_volts`. Call it here too.

driver/ni/channel/scale.h
```diff
@@ -215,20 +184,49 @@ struct TableScale final : BaseScale {
...
+                        return {
+                            nullptr,
+                            x::errors::Error(
+                                x::errors::VALIDATION,
+                                "pre_scaled and scaled values must be the same size"
+                            )
+                        };
+                    return {std::make_unique<TableScale>(v, units), x::errors::NIL};
+                } else {
+                    if (v.reverse_coeffs.empty())
```
sy-nico, line 214: Polynomial scales now fail when `reverse_coeffs` is empty, but the old code derived them with `CalculateReversePolyCoeff`. The Console does not offer polynomial scales, so any config from a client that leaves the field at its default breaks.

driver/ni/channel/channels.h
```diff
@@ -263,2226 +225,1975 @@ inline std::string format_cjc_port(const std::string &path, const std::int32_t p
...
+};
+
+inline x::errors::Error apply(
+    const ::synnax::ni::AIAccelChannel &ch,
+    const ApplyContext &ctx,
+    const std::shared_ptr<daqmx::SugaredAPI> &dmx,
+    const TaskHandle task_handle
+) {
+    const auto loc = ctx.dev_loc + "/ai" + std::to_string(ch.port);
+    auto [scale, scale_err] = make_scale(ch.custom_scale);
```
sy-nico, line 287: Bad units and scales are now caught in `apply` at task start instead of at parse, so they surface as generic errors with no field path. Is this intentional?

sy-nico (changes requested):

driver/ni/channel/scale.h
emilbon99: I'm going to defer this for now and build the console UI alongside this fix once this stack is merged. https://linear.app/synnax/issue/SY-4575/fix-polynomial-scales

driver/common/read_task.h
```diff
@@ -21,39 +23,28 @@
...
     ):
-        BaseTaskConfig(cfg),
-        sample_rate(x::telem::Rate(cfg.field<float>("sample_rate", 0))),
-        stream_rate(x::telem::Rate(cfg.field<float>("stream_rate", 0))),
+        ::synnax::task::common::BaseReadConfig(
+            ::synnax::task::common::BaseReadConfig::parse(cfg)
+        ),
         timing(timing_cfg) {
-        if (sample_rate <= x::telem::Rate(0))
+        if (this->sample_rate <= x::telem::Rate(0))
```
sy-nico, line 43: Absent `sample_rate` and `stream_rate` no longer fail here, since the generated config defaults them to 10 Hz and 5 Hz, so a task with no rates silently runs instead of being rejected. `BaseWriteConfig` does the same to `device`, defaulting it to an empty string where the driver used to require it.

sy-nico (changes requested):
Just ne validation gap left. The generated base configs default `sample_rate`, `stream_rate`, and `device`, so configs the driver used to reject now start with silent defaults.

driver/common/read_task.h
emilbon99: What do you want me to do here? Having a default for the rate values generally means that if you don't supply a rate it will run at its defaults? Do we want the rates to have defaults or not? 

My take is that this isn't a huge issue. We can validate against device being an empty string though.

emilbon99: I implemented validation for the device.

sy-nico (approved):

PR #2486, opened 2026-06-12
SY-4289: Strongly Typed Table Cell Configs

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4289](https://linear.app/synnax/issue/SY-4289)

## Description

Applies the schematic typed-config model (#2474) to tables, end to end. The schema now declares a `CellConfig` union over both cell variants: text cells get typed `level`, `weight`, `align`, and `background_color` fields, and value cells trade their stored telem pipeline spec for the semantic args the reworked `TelemForm` already edits (`channel`, `rolling_average`, `precision`, `notation`) plus a typed `Redline`, colors, and staleness fields. The `cells` map flips from `map<string, Cell>` with opaque props to `map<string, CellConfig>`, so cell configs are validated on every write path instead of trusted; `Cell` survives only as the `{key, config}` pair carried by actions, and `CellTemplate` dissolves into `CellConfig`. Unlike schematic, tables have no partial-merge action, so the live write path is typed wire-to-storage with no server-side key normalization.

Pluto cell components consume flat typed configs and build their pipelines from semantic args via `stringSource`, which moves from `schematic/node/common/telem` into `vis/value` now that schematic symbols and table cells share it. The cell registry resolves schemas from the generated schema map with a `satisfies` guard against registry/variant drift, and the Console toolbar's cell form runs on the generated union schema directly, mirroring the schematic properties form.

Text cell `align` moves from the x-location values the old schema declared (`left`/`right`/`center`) to the flex alignments the alignment control has always written (`start`/`center`/`end`); the old schema and the control disagreed, and stored data contains both. The migration maps x-locations onto flex alignments and named CSS font weights onto the numeric weights the weight control writes.

Storage migrates losslessly: the migration chain re-anchors on the v56 snapshot, the v55 blob lift moves into the v56 mirror package, and a new `v56_typed_cell_configs` step normalizes stored camelCase props, extracts legacy telem pipeline specs into semantic args, and decodes through the union. Cells that decode under no variant degrade to an empty text cell rather than being dropped, because rows reference cells by key and a dropped entry would hole the grid. The Console's v0 state migration and table import path run the same conversion, and imports now reject configs that match no variant instead of storing them.

Stacked on #2474, which supplies the generator capabilities and shared pieces this branch uses: union support, the semantic-args `TelemForm`, `color.Stop`, and `spatial.Alignment`-adjacent enums. Note for reviewers: `oracle migrate` also wants to re-anchor lineplot and log because their v56 snapshots predate their typed lifts; those retargets are reverted here as out of scope.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR applies the oracle-generated strong-typing model to tables end-to-end: `Cell.props` (an opaque JSON blob) is replaced by a validated `CellConfig` discriminated union (`TextCellConfig` / `ValueCellConfig`), the `cells` map type changes from `map<string, Cell>` to `map<string, CellConfig>`, and the migration chain re-anchors on a new v56 snapshot that bridges between the old opaque props format and the current typed one.

- **Schema / types**: `CellTemplate` dissolves into `CellConfig`; value cells trade their stored telem-pipeline spec for semantic args (`channel`, `rollingAverage`, `precision`, `notation`); text cells get typed `level`, `weight`, `align`, and `background_color` fields. Generated code spans Go, TypeScript, and protobuf.
- **Migration**: Two new steps — `v55_lift_typed_table` (v55 → v56, opaque props) and `v56_typed_cell_configs` (v56 → current, typed union) — normalise camelCase keys, extract legacy telem args, and degrade unknown-variant cells to an empty text cell rather than dropping them (preserving row references). The same conversion runs in the TypeScript Console state migration and the import path.
- **Pluto**: `stringSource` is promoted from `schematic/node/common/telem` to `vis/value/telem` so both schematic symbols and table cells share it; cell components now consume flat typed configs instead of nested `props` spreads.

The migration and schema changes are well-structured and thoroughly tested for the happy paths, but two issues on the data-migration path need resolution before merging to avoid silent data loss or storage corruption in production.

The `backgroundColor` migration gap means any user who had colored text cells could have their cell content silently replaced with an empty text cell on first load after upgrade — no error, no warning, no recovery. The `CellConfigFromPB` nil-Variant path means a malformed or empty protobuf cell config received from the wire permanently corrupts the table in storage the next time it is re-encoded. Both are on the write/migration paths that run automatically on existing data.

console/src/table/types/v1.ts (missing backgroundColor coercion before Zod parse), core/pkg/service/table/pb/translator.gen.go (nil-Variant fallback in CellConfigFromPB)

| Filename | Overview |
|----------|----------|
| core/pkg/service/table/migrations/v56/migrate.go | New v56 migration snapshot: correctly moves the v55→v56 step into its own package, but contains a misleading doc comment that inaccurately describes v56 as "strongly-typed" |
| core/pkg/service/table/migrate.go | New v56→current migration step: correctly normalizes camelCase keys, extracts legacy telem args, and degrades unknown variants to empty text cells |
| core/pkg/service/table/convert.go | New camelCase→snake_case normalizer and telem arg extractor; in-place mutation of slice values in normalizeConfigValue could surprise callers who retain a reference to the original slice |
| console/src/table/types/v1.ts | New TypeScript migration logic mirrors Go convert.go; align/weight conversions are correct; backgroundColor hex-string case is not explicitly converted before Zod parse |
| pluto/src/table/cells/Cells.tsx | Cell components updated to consume flat typed configs; backgroundTelem memoization correctly includes t as a dependency since the source segment uses it |
| pluto/src/table/cells/registry.ts | Registry updated with typed Spec, satisfies guard, and schema map sourced from generated types; defaultConfig renamed from defaultProps |
| core/pkg/service/table/types.gen.go | Generated typed CellConfig union with custom JSON marshaling/unmarshaling; nil-Variant CellConfig (from CellConfigFromPB) would later fail EncodeOrc |
| schemas/table.oracle | Schema updated from opaque Cell.props to typed CellConfig union; CellTemplate dissolved into CellConfig; cells map switches to map<string, CellConfig> |
| console/src/table/Toolbar.tsx | Toolbar updated to work with flat typed CellConfig; variant swap logic correctly delegates to deep.overrideValidItems with the target variant schema |
| core/pkg/service/table/pb/translator.gen.go | Expanded PB translator; CellConfigFromPB with a non-nil pb but nil Variant falls through without setting Variant, returning a nil-Variant CellConfig that will cause EncodeOrc to fail |
| pluto/src/vis/value/telem.ts | stringSource extracted from schematic/node/common/telem into vis/value so both schematic and table cells can share it; re-exported from the old location for backward compatibility |
| console/src/table/types/v0.ts | ZERO_TEXT_CELL_PROPS frozen as a literal record using the legacy camelCase/hex format; backgroundColor still uses a raw hex string which may fail color.colorZ parse |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph Storage["Storage Migration Chain"]
        V55["v55.Table\n(opaque Data blob)"]
        V56["v56.Table\nCells: map[string]Cell\nCell.Props: msgpack.EncodedJSON\n(camelCase, unvalidated)"]
        CURRENT["table.Table\nCells: map[string]CellConfig\nCellConfig: typed union\n(text | value)"]

        V55 -->|"v55_lift_typed_table\n(v56.MigrateTable)\n• legacy.MigrateData\n• flatten rows/cols/cells"| V56
        V56 -->|"v56_typed_cell_configs\n(table.MigrateTable)\n• normalizeConfigKeys (camel→snake)\n• extractLegacyArgs (telem→semantic)\n• decodeCellConfig\n• fallback: empty text cell"| CURRENT
    end

    subgraph Console["Console State Migration (v0 → v1)"]
        V0S["v0.State\ncells: CellState[]\nprops: opaque unknown"]
        V1S["v1 PendingUpload\ncells: map[string]CellConfig"]

        V0S -->|"migrateCell\n• extractLegacyArgs\n• cellConfigZ.safeParse\n• fallback: EMPTY_TEXT_CONFIG"| V1S
    end

    subgraph Wire["Wire / PB Layer"]
        PB_IN["Incoming CellConfig PB\nCellConfig_Text | CellConfig_Value"]
        TABLE_CELL["table.CellConfig\nVariant: CellConfigText|Value"]

        PB_IN -->|"CellConfigFromPB"| TABLE_CELL
        TABLE_CELL -->|"CellConfigToPB"| PB_IN
    end
```

1. `core/pkg/service/table/pb/translator.gen.go`, line 545-562 ([link](https://github.com/synnaxlabs/synnax/blob/26aa7f8eaf1fc1d1d03077ae87bef17029fe3660/core/pkg/service/table/pb/translator.gen.go#L545-L562)) 

   <a href="#">P1</a> **`CellConfigFromPB` silently returns nil-Variant config when `pb.Variant` is nil**

   When `pb` is non-nil but `pb.Variant` is nil (an empty `CellConfig` protobuf message), the switch falls through and returns `table.CellConfig{Variant: nil}`. That value then lives in `r.Cells`, and any subsequent `EncodeOrc` call on it will fail with "nil or unknown variant." A malformed message from the wire — or a future extension that sets no variant — can write a nil-Variant entry that silently corrupts the table on next serialization. Returning an error (or degrading to an empty text cell, matching the migration strategy) would be safer.

</details>

<sub>Reviews (1): Last reviewed commit: ["SY-4289: Strongly Typed Table Cell Confi..."](https://github.com/synnaxlabs/synnax/commit/26aa7f8eaf1fc1d1d03077ae87bef17029fe3660) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37324135)</sub>

**Context used:**

- Rule used - Flag misleading doc comments that promise fallback... ([source](https://app.greptile.com/synnax/github/synnaxlabs/synnax/-/custom-context?memory=16df2708-f8e1-4fe6-a509-cee5a869fa6e))


core/pkg/service/table/migrations/v56/migrate.go
```diff
@@ -0,0 +1,73 @@
...
+// MigrateTable transforms the previous Table snapshot (v55) into the current
+// strongly-typed Table. AutoMigrateTable handles the trivially-copyable
+// gorp-entry fields (Key, Name); the structural fields (Rows, Columns, Cells)
+// are sourced from the opaque blob the console used to persist alongside those
+// fields, after legacy.MigrateData decodes it as v0.Data. v55 is the last
+// snapshot in which Table.Data is untyped; future migrations transform one
+// typed snapshot into another and never need this blob handling.
+func MigrateTable(ctx context.Context, old v55.Table) (Table, error) {
+	out, err := AutoMigrateTable(ctx, old)
+	if err != nil {
```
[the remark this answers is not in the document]
emilbon99: *Emiliano + Claude: drafted in our review session.*

Good catch. The file moved to `versions/v1/migrate.go` since you flagged it, and the comment moved with it unchanged. Fixed in 772498dd: dropped "strongly-typed" and the claim that future migrations never need the blob handling, since the v2 migration in this PR does exactly that.

console/src/table/types/v1.ts
```diff
@@ -66,6 +66,80 @@ export const ZERO_SLICE_STATE: SliceState = {
...
+        segProp(cfg.telem, "rollingAverage", "windowSize"),
+      );
+      setIfPresent("precision", segProp(cfg.telem, "stringifier", "precision"));
+      setIfPresent("notation", segProp(cfg.telem, "stringifier", "notation"));
+      delete cfg.telem;
+      break;
+    case "text":
+      if (typeof cfg.align === "string" && cfg.align in LEGACY_ALIGNS)
+        cfg.align = LEGACY_ALIGNS[cfg.align];
+      if (typeof cfg.weight === "string" && cfg.weight in LEGACY_WEIGHTS)
```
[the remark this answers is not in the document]
emilbon99: *Emiliano + Claude: drafted in our review session.*

This one does not reproduce. `decodeCellConfig` round-trips through `json.Marshal` and `json.Unmarshal`, and `TextCellConfig.BackgroundColor` is a `*color.Color` from `x/go/color/versions/v0`, whose `UnmarshalJSON` takes a hex string as its first case. Migrating a text cell with `#00000000`, `#112233`, or `#ff000080` produces the right color with `value` intact, no fallback to an empty text cell.

The TypeScript half is moot now: `console/src/table/types/v1.ts` is gone and the migration is Go only.

You are right that `migrate_test.go:82` asserts only the key rename and not the decoded value. Leaving that as is for now.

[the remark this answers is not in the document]
emilbon99: *Emiliano + Claude: drafted in our review session.*

Coming back on this one: you were closer to right than my last reply gave you credit for. The hex decode works, so the failure mode you described does not happen, but looking at the stored value was the correct instinct.

The real problem is that `#00000000` in pre-v2 Console data means "nobody chose a color", not "transparent". v2 reads a stored color as a deliberate choice, and the schema says an absent `background_color` lets the theme pick the fill while a transparent one paints nothing. So the migration was turning "theme picks" into "paint nothing" on every legacy cell, and our own fixtures in `import_console_v0_data.json` are full of them.

Fixed in e19c84ed with `stripZeroColors`, matching what the schematic v9 migration already does. Gradient stops are excluded since `color.Stop.color` is required. Added migration tests for `background_color`, `color`, and `staleness_color`, plus one that a chosen color survives and one that a transparent gradient stop is left alone.

schemas/synnax/versions/table/v2.oracle
```diff
@@ -0,0 +1,167 @@
...
+Key = v0.Key
+
+FlexAlignment enum {
+    start   = "start"
+    center  = "center"
+    end     = "end"
+    stretch = "stretch"
+
+    @doc value "is a cross-axis flex alignment for laid-out content."
+}
```
pjdotson: is this table-specific or should it be in X?

schemas/synnax/versions/table/v2.oracle
```diff
@@ -0,0 +1,167 @@
...
+Redline struct {
+    bounds   spatial.Bounds {
+        @doc value "is the numeric range mapped onto the gradient."
+    }
+    gradient color.Stop[]   {
+        @doc value "is the color gradient applied across the bounds."
+    }
+
+    @doc value  "maps a numeric range to a color gradient for limit visualization."
+    @go marshal
```
pjdotson: shouldn't this be in X?

schemas/synnax/versions/table/v2.oracle
```diff
@@ -0,0 +1,167 @@
...
+            """
+        }
+
+        @doc value "is the configuration for static text cells."
+    }
+    value {
+        channel           channel.Key = 0 {
+            @doc value "is the channel whose value the cell displays."
+        }
+        rolling_average   int32 = 1 {
```
pjdotson: shouldn't this be `uint` something?

schemas/synnax/versions/table/v2.oracle
```diff
@@ -0,0 +1,167 @@
...
+            @doc value "is the sample window for rolling-average smoothing."
+        }
+        precision         int32?            {
```
pjdotson: shouldn't this also be uint?

schemas/synnax/versions/table/v2.oracle
```diff
@@ -0,0 +1,167 @@
...
+        color             color.Color?      {
+            @doc value """
+                is the color of the displayed text. When absent the value renders with a
+                theme-derived legible color.
+            """
+        }
+        units             string = "" {
+            @doc value "is the unit suffix displayed after the value."
+        }
+        staleness_timeout float64 = 5 {
```
pjdotson: should this instead be some type of telem.TimeSpan? We also may want to rethink how we do time duration inputs across Synnax, they aren't really standardized

oracle/plugin/go/internal/schemadiff/structural.go
```diff
@@ -18,10 +19,10 @@ import (
...
 // StructurallyEqual reports whether two declarations share a persisted shape, resolving
 // each side's references through its own table. It is schemadiff's persisted-shape
-// equality with two tightenings: enums must match member-for-member, and the declared
-// field list — omitted fields and their marshal values included — must agree.
-// Wire-compatible enum additions need no migration, but they are still a shape change
-// the history must record.
+// equality with three tightenings: enums must match member-for-member, the declared
+// field list — omitted fields and their marshal values included — must agree, and the
+// declared defaults must agree. Wire-compatible enum additions need no migration, but
+// they are still a shape change the history must record.
```
pjdotson: why do we make this a new version instead of just going backwards in the old file and editing the default?

emilbon99:
@greptile re-review

PR #2492, opened 2026-06-15
SY-4379: Oracle Default and Validation Refactor RFC

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4379](https://linear.app/synnax/issue/SY-4379)

## Description

Adds RFC 0043, proposing a unified model for field **optionality**, **default application**, and the **input/output (`New`) type projection** in Oracle.

These three concerns are currently entangled and handled inconsistently across Go, TS, Python, C++, and the codecs — the gap that let the SY-4289 table migration write an invalid `level: ""` enum straight to storage (Go neither applies defaults nor validates on decode). The RFC settles:

- One optionality marker (`?` = nullable); `??` removed. Optionality states domain meaning, not serialization convenience.
- Static defaults on required fields; derived defaults are just nullable fields.
- Value-based default *overlay* at boundaries — codec-agnostic, no presence detection (which protobuf and ORC cannot provide).
- A generator-enforced invariant (`default == zero` OR `zero` is invalid) that makes value-based defaulting provably correct, with compile-time errors and concrete remediations.
- `New` as a derived projection of the base, keyed by one `@create` marker, with `@output` dropping server-owned fields.
- Storage as a trusted invariant: writes and migrations default+validate; reads decode + presence-fill.

Docs-only — the RFC itself, no code changes. The implementation is sketched in §8 and the remaining minor questions are in §9. The `action`/mutation system (RFC 0040) is explicitly out of scope.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adds RFC 0043, a docs-only proposal that unifies Oracle's field optionality, default application, and `New`/input-output type projection across Go, TS, Python, C++, and all codecs — directly motivated by the SY-4289 bug where a Go migration wrote `level: ""` (an invalid enum zero) to storage unchecked.

- Collapses `?`/`??` into a single `?` = nullable marker, routes misused `??` to required+default, and enforces a generator-level invariant (`default == zero` OR `zero` is invalid) that makes value-based overlay provably safe across codecs that lack scalar presence (protobuf, ORC).
- Introduces `@create` and `@output` markers to derive `New` input projections automatically in TS/Python, and specifies write, migration, and read as the three default+validate boundaries, with Go getting a generated `ApplyDefaults` and server-side `IsValid` validation.
- The \"mutation\" suffix in the filename is misleading since mutation (RFC 0040) is explicitly out of scope; §5.0–5.1 use \"no presence detection\" and \"presence-based fill\" in apparently contradictory ways without clarifying that both paths rely on the same value-based mechanism; and the migration error path for invalid non-zero values (which defaults cannot overwrite) is left unspecified.

Docs-only RFC with no code changes; safe to merge as a draft, but the §5.0–5.1 framing contradiction should be resolved before the implementation phase begins to avoid misdirecting the Go/C++ implementers.

The contradictory framing between "no presence detection" in §5.0 and "presence-based fill" in §5.1 is the sharpest edge — an implementer reading §5.1 in isolation could attempt presence-based logic for ORC that is both incorrect and architecturally unsupported. The @key resolution inconsistency and the missing migration error path for invalid non-zero values are contained gaps in the draft that don't block merging but should be addressed before implementation begins.

The single changed file, docs/tech/rfc/0043-260615-oracle-optionality-defaults-mutation.md, warrants a second look on §5.0–5.1 for the presence-detection framing and on §5.1 for the migration validation-failure path.

| Filename | Overview |
|----------|----------|
| docs/tech/rfc/0043-260615-oracle-optionality-defaults-mutation.md | New RFC introducing a unified optionality, defaults, and I/O projection model for Oracle; contains a terminological contradiction between §5.0 ("no presence detection") and §5.1 ("presence-based fill") that could mislead implementers, an unresolved @key assignment presented as settled, and a gap in the migration error path for invalid non-zero values. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph Write["Write Boundary (create/dispatch)"]
        W1[Client sends input] --> W2[Server strips @output fields]
        W2 --> W3[ApplyDefaults]
        W3 --> W4[IsValid validation]
        W4 --> W5[Store to ORC/protobuf]
    end

    subgraph Migration["Migration Boundary (RFC 0033)"]
        M1[Legacy record] --> M2[Lenient decode — no validation]
        M2 --> M3[ApplyDefaults]
        M3 --> M4[IsValid validation]
        M4 --> M5[Store updated record]
        M4 -->|invalid non-zero value?| M6[⚠️ Unspecified behavior]
    end

    subgraph Read["Read Boundary (retrieve)"]
        R1[Stored record] --> R2[Decode]
        R2 --> R3["Value-based fill\n(zero == default OR zero invalid\n→ invariant from §4 makes this safe)"]
        R3 --> R4[Return complete output type]
    end

    subgraph Invariant["§4 Generator Invariant (compile-time)"]
        I1["For every required field with default:\ndefault == zero(type)\nOR zero(type) is invalid"]
    end

    Invariant --> Read
    Invariant --> Write
    Invariant --> Migration
```

<sub>Reviews (1): Last reviewed commit: ["SY-4379: Oracle Default and Validation R..."](https://github.com/synnaxlabs/synnax/commit/4c41765aa437c37214e5be3a10277c7b5a13aa26) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37693724)</sub>

> Greptile also left **4 inline comments** on this PR.

### DIFF
--- a/docs/tech/rfc/0043-260615-oracle-optionality-defaults-mutation.md
+++ b/docs/tech/rfc/0043-260615-oracle-optionality-defaults-mutation.md
@@ -1,0 +1,355 @@
+# 0043 - Oracle Optionality, Defaults, and Input/Output Types
+
+**Feature Name:** Field Optionality, Defaults, and Input/Output Projection in Oracle
+
+**Status:** Draft (core model settled)
+
+**Related:** [RFC 0027](./0027-251229-oracle-schema-system.md),
+[RFC 0033](./0033-260320-oracle-migrations.md),
+[RFC 0040](./0040-260508-action-based-undo-redo.md),
+[RFC 0042](./0042-260331-oracle-struct-unions.md)
+
+---
+
+# 0 - Summary
+
+Three things in Oracle have never had one clear definition: whether a field is optional,
+where its default value gets filled in, and how the type you send to create a record
+relates to the type you read back. They are really one question, answered differently in
+each language: what does it mean for a field to be missing, or to hold its zero value?
+
+This RFC settles that question:
+
+1. There is one optionality marker, `?`, meaning the field can be null. The old `??` is
+   removed. A marker says what a value means, not how it serializes.
+2. Required fields can carry a static default. A "default" that depends on context, like
+   a color that follows the theme, is just a nullable field the consumer fills in.
+3. Defaults are filled the same way everywhere: take a baseline and overlay what the
+   caller actually sent. This needs no record of which fields were present, which
+   matters because protobuf and the ORC codec cannot provide it.
+4. A build-time rule guarantees that filling defaults this way is always safe.
+5. The create type `New` is derived from the base type, switched on by one `@create`
+   marker. An `@output` marker hides the fields only the server may set.
+6. Stored data is always complete and valid: writes and migrations fill defaults and
+   validate, and reads just decode.
+
+The `action` system (RFC 0040) is unchanged. Mutations touch this RFC only because
+dispatching an action is a write, so its payload is defaulted and validated like any
+other.
+
+---
+
+# 1 - Motivation
+
+The problem surfaced while giving table cell configs real types (SY-4289). The server
+migrates old table data into the new typed form, in Go. A typed enum field that had no
+value came out as the empty string, which is not a valid member, and that empty string
+went straight to storage. When the Console read the table back, it could not parse it,
+so the table failed to load.
+
+Teaching Go to fill defaults would fix the symptom. The real gap is that Oracle never
+decided, in one place, when a field may be missing, what its absence means, and who
+fills it in. Each language answers differently: TypeScript (zod) and Python (pydantic)
+can see which fields a caller sent, while Go and the ORC codec cannot. zod already
+follows a sensible model where the data on the wire is complete and defaults are filled
+at the edges. This RFC writes that model down and makes every language follow it.
+
+---
+
+# 2 - Optionality
+
+A field is optional to say something about the domain, not about serialization. A text
+cell always has a level, so `level` is required. If a language has trouble producing a
+non-zero default for a required field, that is a tooling problem we fix in the tooling.
+We do not relabel the field nullable to avoid it.
+
+We audited all 33 schemas. The soft marker `?` never meant more than "required, with a
+default of zero, empty string, or false." For instance, `alias string?` behaves exactly
+like `alias string = ""`. The hard marker `??` is the only one that carries weight, but
+it is overused: of schematic's 106 uses, about 85 just stand in for a fixed default.
+Where `??` genuinely matters, it falls into three cases:
+
+- a value that does not exist yet, like a task's status before it runs;
+- a reference that may point to nothing, like a parent ID, where an all-zero ID is
+  meaningless;
+- a value the consumer derives later, like a color left null so the renderer can match
+  the theme.
+
+So we keep one marker and drop the other. `?` means the field can be null, and what null
+means is the consumer's call. The roughly 85 misused `??` fields become required fields
+with a default; the genuinely nullable ones keep a single `?`.
+
+```oracle
+level text.Level = text.LevelH5   // required, with a fixed default
+color color.Color?                // nullable; the renderer picks a color
+```
+
+---
+
+# 3 - Defaults
+
+There are two kinds of default, one for each field shape.
+
+- **Static:** a fixed value on a required field, like `precision = 2`. It is stored as
+  written and is the same in every language.
+- **Derived:** a nullable field whose value the consumer computes when needed. A cell's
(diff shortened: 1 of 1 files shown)


docs/tech/rfc/0043-260615-oracle-optionality-defaults-mutation.md
```diff
@@ -0,0 +1,355 @@
...
+where its default value gets filled in, and how the type you send to create a record
+relates to the type you read back. They are really one question, answered differently in
+each language: what does it mean for a field to be missing, or to hold its zero value?
+
+This RFC settles that question:
+
+1. There is one optionality marker, `?`, meaning the field can be null. The old `??` is
+   removed. A marker says what a value means, not how it serializes.
+2. Required fields can carry a static default. A "default" that depends on context, like
+   a color that follows the theme, is just a nullable field the consumer fills in.
```
pjdotson, line 26: have you thought about how this interacts with forms? e.g. what should the color value
be on the form of a lineplot?

pjdotson (changes requested):

docs/tech/rfc/0043-260615-oracle-optionality-defaults-mutation.md
emilbon99: Two mechanisms:

1) We do the same auto-selection mechanism i.e. we select the default at form render time from the theme. So it would look the same as if it was stored.
2) We eventually work improving the 'auto mode' UX for inputs i.e. better indicators to know that the particular input is in auto/default mode and ways to revert a manually set value to auto mode.

pjdotson (approved):
lgtm

docs/tech/rfc/0043-260615-oracle-optionality-defaults-mutation.md
```diff
@@ -0,0 +1,355 @@
...
+The model rests on one promise: every stored record is complete and valid. Three places
+keep it.
+
+- **Create and update** (a write): the server fills defaults, validates, and stores. It
+  also drops any `@output` fields the client tried to send and sets them itself, so a
+  client cannot forge the `author`. The client defaults locally first so its own view
+  updates right away, and the server repeats it as the authority. For example, a client
+  may send `{ "variant": "value" }` for a cell, and the server fills in `precision`,
+  `notation`, and the rest before storing.
+- **Migration** (also a write): it reads old data loosely, without validating, since the
```
sy-nico, line 163: Defaults only overwrite zero values, so a legacy record holding an invalid non-zero value (an enum stored as bogus, not "") survives the overlay and then fails output validation. What does the migration do at that point: abort, skip the record, or quarantine it?

docs/tech/rfc/0043-260615-oracle-optionality-defaults-mutation.md
```diff
@@ -0,0 +1,355 @@
...
+   removed. A marker says what a value means, not how it serializes.
+2. Required fields can carry a static default. A "default" that depends on context, like
+   a color that follows the theme, is just a nullable field the consumer fills in.
+3. Defaults are filled the same way everywhere: take a baseline and overlay what the
+   caller actually sent. This needs no record of which fields were present, which
+   matters because protobuf and the ORC codec cannot provide it.
+4. A build-time rule guarantees that filling defaults this way is always safe.
+5. The create type `New` is derived from the base type, switched on by one `@create`
+   marker. An `@output` marker hides the fields only the server may set.
+6. Stored data is always complete and valid: writes and migrations fill defaults and
```
sy-nico, line 33: Item 6 says reads "just decode," but 5.1 has the read path fill defaults as a rolling-upgrade backstop. Worth reconciling the summary with 5.1.

docs/tech/rfc/0043-260615-oracle-optionality-defaults-mutation.md
```diff
@@ -0,0 +1,355 @@
...
+| `precision int = 2`, where `0` (whole numbers) is a real choice | Make it nullable: `precision int?`, where null means "formatter default" and `0` is a choice. |
+| an int enum whose default is not its first member               | Reorder so the default is first, or make it nullable.                                         |
+
+---
+
+# 5 - Filling and Validating Defaults
+
+## 5.0 - One way to fill
+
+Wherever a default is filled, it works the same: start from a baseline that already
```
sy-nico, line 137: This says protobuf and ORC cannot tell whether a scalar was on the wire, yet nullable fields with no default like `precision int?` need exactly that null vs zero distinction to round-trip. How is a stored null kept distinct from `0` on those codecs, given the section 4 invariant only protects required fields with a default?

sy-nico (commented):
Mostly good. I just have some clarification nits.

sy-nico (approved):

PR #2495, opened 2026-06-15
SY-4247: Standardize service and API layer resolved status retrieval

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4247](https://linear.app/synnax/issue/SY-4247)

## Description

Devices, racks, and tasks are re-created on every driver scan cycle. The writers for
all three resolved a status from the entity and **unconditionally** wrote it, so any
no-op re-configure overwrote a live, driver-reported status (e.g. "running",
"connected") with the default "unknown" status. Conversely, if a status row went
missing out-of-band, retrieval that joined entities against their status rows could
misalign.

This PR makes status writes self-healing across all three writers:

- **Writers (device, rack, task)**: on `Create`, when no status is provided, the default
  status is written only if no row already exists (`healStatus`), preserving a live one;
  when a status is explicitly provided, it is written as before. The device writer
  already had this behavior; rack and task now match it.
- **Channel retrieval API**: adds an `IncludeStatus` option that attaches each channel's
  status row via a key-based map join. Channels legitimately have sparse statuses, so a
  channel with no row is returned with a nil status.
- **Task/rack/device retrieval API**: the writer self-heal guarantees every one of these
  entities has a status row, so status attachment continues to use the existing
  index-based assignment.

Regression tests added for both the heal-on-missing and don't-clobber-live paths across
device, rack, and task.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR standardizes status retrieval and write semantics across the device, rack, and task service layers to prevent scan-cycle churn from clobbering live driver-reported statuses. It also adds a new `IncludeStatus` option to the channel retrieval API with a correct map-based join for sparse status rows.

- **Writers (device, rack, task)**: `Create` now calls `healStatus` when no explicit status is provided, writing the default "unknown" row only if no row already exists. When an explicit status is provided, it is always written unconditionally. Regression tests cover both the heal-on-missing and don't-clobber-live paths for all three entity types.
- **Channel API**: New `IncludeStatus bool` field on `RetrieveRequest` attaches each channel's status via a key-indexed map join, correctly returning `nil` for channels with no status row.
- **Rack API cleanup**: Removes an unused `keys` slice built before the status query and drops an unnecessary `(*rack.Status)` type cast.

The change is safe to merge. All three writers now share consistent heal-on-missing semantics and the channel API uses a key-indexed map join rather than a positional zip.

The writer changes are narrowly scoped: an extra KV existence check followed by a conditional write. Both the heal-on-missing and don't-clobber-live paths are directly exercised by new regression tests for every entity type. The channel IncludeStatus implementation correctly uses a map to handle sparse status rows. No pre-existing logic was restructured in a way that could silently regress other callers.

| Filename | Overview |
|----------|----------|
| core/pkg/api/channel/channel.go | Adds IncludeStatus field to RetrieveRequest; uses a correct map-based join keyed on ontology IDs to attach statuses, handling sparse channel statuses properly (nil for missing rows). |
| core/pkg/api/rack/rack.go | Removes a dead keys-slice that was never used; changes unnecessary type cast (*rack.Status)(&stat) to plain &stat. Positional status assignment loop unchanged but heal-on-write guarantee makes it safe in the common path. |
| core/pkg/api/task/task.go | Removes the TODO(SY-4247) comment; no logic changes; positional status assignment remains, relying on heal-on-write to ensure every task has a status row. |
| core/pkg/service/device/writer.go | Adds healStatus helper; the existing-device/same-rack/no-name-change path now calls healStatus instead of returning nil, restoring missing rows without clobbering live ones. |
| core/pkg/service/rack/writer.go | Adds healStatus helper; Create now conditionally writes status (immediate write for explicit status, healStatus for nil-status path), matching the device writer pattern. |
| core/pkg/service/task/writer.go | Adds healStatus helper; Create now conditionally writes status based on whether providedStatus is non-nil, matching the rack and device writer patterns. |
| core/pkg/service/device/device_test.go | Adds two regression tests: heal-on-missing (deletes row then re-creates) and don't-clobber-live (creates with live status then re-creates without). Both correctly exercise the new code paths. |
| core/pkg/service/rack/rack_test.go | Adds identical heal-on-missing and don't-clobber-live tests for racks, matching the device test structure. |
| core/pkg/service/task/task_test.go | Adds identical heal-on-missing and don't-clobber-live tests for tasks, matching the device and rack test structure. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["Writer.Create(ctx, entity)"] --> B{Entity already exists\nwith same rack/key?}
    B -- No --> C["resolveStatus(entity, provided)"]
    C --> D["status.Set(ctx, stat)\nunconditional write"]
    B -- Yes --> E{Explicit status\nprovided?}
    E -- Yes --> F["resolveStatus(entity, provided)"]
    F --> G["status.Set(ctx, stat)\nunconditional write"]
    E -- No --> H["resolveStatus(entity, nil)\ndefault unknown status"]
    H --> I["healStatus(ctx, stat)"]
    I --> J{Status row\nalready exists?}
    J -- Yes --> K["return nil\nlive status preserved ✓"]
    J -- No --> L["status.Set(ctx, stat)\nrestore missing row ✓"]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    A["Writer.Create(ctx, entity)"] --> B{Entity already exists\nwith same rack/key?}
    B -- No --> C["resolveStatus(entity, provided)"]
    C --> D["status.Set(ctx, stat)\nunconditional write"]
    B -- Yes --> E{Explicit status\nprovided?}
    E -- Yes --> F["resolveStatus(entity, provided)"]
    F --> G["status.Set(ctx, stat)\nunconditional write"]
    E -- No --> H["resolveStatus(entity, nil)\ndefault unknown status"]
    H --> I["healStatus(ctx, stat)"]
    I --> J{Status row\nalready exists?}
    J -- Yes --> K["return nil\nlive status preserved ✓"]
    J -- No --> L["status.Set(ctx, stat)\nrestore missing row ✓"]
```

</a>

<sub>Reviews (3): Last reviewed commit: ["Keep index-based status join for task, r..."](https://github.com/synnaxlabs/synnax/commit/3b230f95fdfdaab9ffff815ec71cf25aae56fff9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37745129)</sub>


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

core/pkg/service/device/writer.go
```diff
@@ -54,6 +54,27 @@ func resolveStatus(d *Device, provided *Status) *status.Status[StatusDetails] {
...
 
+// healStatus restores a device's status row if it has gone missing (e.g. deleted
+// out-of-band) without clobbering a live one. Devices are re-created on every scan
+// cycle, so on a no-op update the default "unknown" status must not overwrite a status
+// the driver has already reported; it is only written when no row exists.
+func (w Writer) healStatus(
+	ctx context.Context,
+	stat *status.Status[StatusDetails],
+) error {
+	// If the status already exists, do nothing.
```
emilbon99: Useless comment. Can also shorten this to:

```
if exists, err := gorp.NewRetrieve[string, status.Status[StatusDetails]]().
		Where(gorp.MatchKeys[string, status.Status[StatusDetails]](stat.Key)).
		Exists(ctx, w.tx); err != nil || exists {
		return err
	}
```

Can apply same principle for other code locations

emilbon99 (changes requested):
Generally looks fine. Left a control flow nit.

Just want to make sure we're under the same understanding though: **this is a temporary fix to a deeper problem**

1) There should be better guards against statuses for things like devices getting deleted.
and
2) We should move the materialization of resolved fields into the service layer.

Merge when ready.

pjdotson:
> Generally looks fine. Left a control flow nit.
> 
> Just want to make sure we're under the same understanding though: **this is a temporary fix to a deeper problem**
> 
> 1. There should be better guards against statuses for things like devices getting deleted.
>    and
> 2. We should move the materialization of resolved fields into the service layer.
> 
> Merge when ready.

Yes this is just temporary fix to a bug that was sometimes popping up + some standardization. Eventually the API layer should be a very thing wrapper around the service layer that just handles converting the request/response shapes into code to be executed by service layer + auth

PR #2505, opened 2026-06-17
SY-4379: Migrate Schemas to Defaults and Single Optionality

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4379](https://linear.app/synnax/issue/SY-4379)

## Description

Stacked on #2504 (the engine). Migrates every schema to the new defaults/single-optionality model and adapts all consumers.

- Migrates all `schemas/*.oracle`: explicit field defaults (`= 0` / `= false` / `= ""` / `= []`), `??` collapsed to a single `?`, and `@create`-derived `New` types (ranger/status keep a hand-written `New`; label uses `@create`).
- Invariant-driven boolean flips: log `show*` -> `hide*`, lineplot legend `visible` -> `hidden`, axis `autoBounds` -> `manualBounds`, with consumer logic inverted (user-facing toggles keep their "Show" semantics at the store/schema boundary).
- Adapts client/pluto/console for now-required-with-default fields (`description`, `cmd`, `internal`, `snapshot`, ...), routing `Status` construction through `status.create`.
- Removes the `project.author` field entirely (schema, writer, API layer, client, tests).
- Includes the `rc` backmerge of the Arc Params Refactor + Runtime Split, re-applying the migration to the new arc schemas.

Go (`core`/`arc`/`x`) and TypeScript (`x`/`client`/`pluto`/`console`) build clean.

> **Note:** `core/pkg/service/project/migrate_auto.gen.go` is a generated migration whose `author` copy was hand-removed as a stopgap to unblock the build. It must be regenerated via `oracle migrate` before merge.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR is a large, systematic schema migration that renames boolean fields from positive semantics (`visible`, `showX`, `autoBounds`) to negative semantics (`hidden`, `hideX`, `manualBounds`) across all schema layers (Oracle schemas → Go types → TypeScript generated types → Console/Pluto consumers), and also removes the `project.author` field entirely.

- All boolean field renames apply consistent inversion logic: `legend.visible → hidden` (with `!`), `showChannelNames → hideChannelNames` (with `!`), `autoBounds → manualBounds` (with `!`), across server-side Go migrations, client-side TypeScript Redux migrations, and every consumer site.
- `StickyXY.root` and `StickyXY.units` are promoted from optional to required-with-default; Go's `ApplyDefaults()` in the lineplot/log writers ensures incoming data with missing fields is filled before storage, and the server-side gorp migrations handle all existing persisted records.
- The previously flagged P1 (`TYPED_EXPORT.legend` still using `visible: false` in the import spec) is fixed — the fixture now uses `hidden: true` with a matching assertion added.

Safe to merge after regenerating migrate_auto.gen.go via oracle migrate as noted in the PR description.

Every boolean field inversion applied consistently across all layers from Oracle schema through Go and TypeScript generated types; the stale import spec fixture is fixed; Go ApplyDefaults() in the lineplot and log writers handles missing StickyXY fields from client payloads; server-side gorp migrations correctly invert all field values for existing DB records.

core/pkg/service/project/migrate_auto.gen.go — hand-edited generated file flagged by the author for regeneration before merge.

| Filename | Overview |
|----------|----------|
| console/src/lineplot/services/import.spec.ts | Previously-flagged stale visible: false in TYPED_EXPORT fixture is now fixed to hidden: true; a new assertion validates the legend-hidden state is preserved through parseImport. |
| console/src/lineplot/types/v5.ts | New v5 migration: correctly inverts autoBounds to manualBounds and visible to hidden in buildPendingUpload. Go writer's ApplyDefaults() fills any missing root/units before persisting, making the lean client payload safe. |
| core/pkg/service/lineplot/migrate.go | Correctly inverts legend.Visible to Hidden and AutoBounds to ManualBounds in all migration helper functions; migrateStickyRoot now returns a concrete value type with a proper left/top default for nil input. |
| core/pkg/service/log/migrate.go | logFromV1 correctly inverts ShowChannelNames to HideChannelNames and ShowReceiptTimestamp to HideReceiptTimestamp; migration test asserts both inversions. |
| x/go/spatial/types.gen.go | StickyXY.Root and .Units promoted from optional pointers to value types; new ApplyDefaults() fills empty-string fields with canonical defaults (left/top, px/px). |
| console/src/schematic/types/v6.ts | v6 migration adds root default to the null-fallback legend position and defaults zIndex to 0 via ?? operator; test fixtures updated. |
| core/pkg/api/project/project.go | Clean removal of author field: user key extraction, author assignment, author-based retrieve filter, and the parent-of ontology relationship all removed consistently. |
| core/pkg/service/project/migrate_auto.gen.go | Hand-edited generated file (author copy removed as stopgap); noted by PR author as requiring regeneration via oracle migrate before merge. |
| x/ts/src/spatial/sticky/sticky.ts | toCSS and toDecimal now use direct property access (pos.root.x) instead of optional chaining; safe because stickyXYZ now uses prefault for root/units. |
| pluto/src/log/aether/Log.ts | All showX to hideX renames applied; rendering logic correctly uses !hideReceiptTimestamp and !hideChannelNames in all four prefix branches to preserve existing behavior. |

</details>

<details><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `console/src/lineplot/services/import.spec.ts`, line 56 ([link](https://github.com/synnaxlabs/synnax/blob/996b8ee344a50d2041af82f36f07ba6b3b40366e/console/src/lineplot/services/import.spec.ts#L56)) 

   <a href="#">P1</a> **Stale legend field in TYPED_EXPORT fixture**

   The `legend` entry still uses the old `visible: false` field that was renamed to `hidden: true` in this PR. When `parseImport` falls through to `lineplot.linePlotZ.parse(data)`, Zod strips the unrecognised `visible` key and defaults `hidden` to `false`, so any user-exported file where the legend was hidden would be silently re-imported with the legend shown. The test assertions only check `lines` and `channels`, so this data-loss path is undetected. The fixture should be updated to `hidden: true` and a legend-visibility assertion added.

2. `core/pkg/service/project/migrate_auto.gen.go`, line 19-24 ([link](https://github.com/synnaxlabs/synnax/blob/996b8ee344a50d2041af82f36f07ba6b3b40366e/core/pkg/service/project/migrate_auto.gen.go#L19-L24)) 

   <a href="#">P2</a> **Hand-edited generated file needs regeneration before merge**

   The PR description explicitly flags this: `migrate_auto.gen.go` was hand-modified to remove the `Author` copy as a stopgap, and must be regenerated via `oracle migrate` before merging. The current hand-edit is functionally correct (the `Author` field is gone from the destination struct), but the file may diverge from what the generator produces in other respects, and generated-file drift can mislead future contributors.

3. `console/src/lineplot/services/import.spec.ts`, line 56 ([link](https://github.com/synnaxlabs/synnax/blob/3f369cc508cc7ff93c5edf65cd4892a048567348/console/src/lineplot/services/import.spec.ts#L56)) 

   <a href="#">P1</a> **Stale `visible` key in TYPED_EXPORT fixture silently loses legend-hidden state**

   `TYPED_EXPORT.legend` still uses `{ visible: false, … }`, which is the old field name that was renamed to `hidden` in this PR. When `parseImport` falls through to `linePlotZ.parse(data)`, Zod strips the unrecognised `visible` key and defaults `hidden` to `false`, so the imported legend will be shown even though the exported plot had it hidden. The test assertions only check `lines` and `channels`, so this data-loss path goes undetected. The fixture should use `{ hidden: true, … }` and a legend-visibility assertion should be added.
</details>

<sub>Reviews (3): Last reviewed commit: ["Fix stale legend visible key in TYPED\_EX..."](https://github.com/synnaxlabs/synnax/commit/d7c321ad904c9d6b71c13e9fea8d6fbcab45605f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38461152)</sub>


schemas/synnax/policy.oracle
```diff
@@ -33,7 +33,7 @@ Policy struct {
     actions  access.Action[] {
         @doc value "is the list of actions this policy permits."
     }
-    internal bool? = false {
+    internal bool = false {
```
pjdotson, line 36: does the analyzer make sure that defaults and optionals are mutually exclusive?

schemas/synnax/ranger.oracle
```diff
@@ -45,18 +45,20 @@ Range struct {
...
-    parent     Range??         {
-        @doc value  """
+    parent     Range?          {
+        @doc value     """
             is an optional parent range for hierarchical organization. Ranges
             can be nested within other ranges.
         """
-        @go marshal omit
+        @go marshal    omit
+        @validate skip
```
pjdotson, line 61: what does `@validate skip` do?

emilbon99:
@greptile please re-review

schemas/synnax/status.oracle
```diff
@@ -70,16 +70,19 @@ Status struct<Details?, V extends Variant = Variant> {
 }
 
 New struct<Details?, V extends Variant = Variant> extends Status<Details, V> {
```
pjdotson: why is there still a New for status?

emilbon99:
@greptile please re-review

x/py/x/dicts.py
```diff
@@ -0,0 +1,13 @@
...
+#
+#  Use of this software is governed by the Business Source License included in the file
+#  licenses/BSL.txt.
+#
+#  As of the Change Date specified in that file, in accordance with the Business Source
+#  License, use of this software will be governed by the Apache License, Version 2.0,
+#  included in the file licenses/APL.txt.
+
+
+def none_to_empty[K, V](v: dict[K, V] | None) -> dict[K, V]:
```
pjdotson: add test

x/py/x/lists.py
```diff
@@ -7,10 +7,14 @@
 #  License, use of this software will be governed by the Apache License, Version 2.0,
 #  included in the file licenses/APL.txt.
 
-
 from typing import Any
 
 
+def none_to_empty[T](v: list[T] | None) -> list[T]:
```
pjdotson: add test

x/py/x/control/types_gen.py
```diff
@@ -88,4 +90,6 @@ class Update(BaseModel, Generic[R]):
         transfers: Is the list of control transfers that occurred in this update.
     """
 
-    transfers: list[Transfer[R]]
+    transfers: Annotated[list[Transfer[R]], BeforeValidator(lists.none_to_empty)] = (
```
pjdotson: why is `transfers` ever seeing `None` now if the schema doesn't mark it optional? Why do we need `none_to_empty` if we already have the `default_factory=list`?

client/py/synnax/device/types_gen.py
```diff
@@ -75,6 +75,11 @@ def __hash__(self) -> int:
         return hash(self.key)
 
 
+class New(Device):
```
pjdotson: how is the `New` type helpful here?

client/py/synnax/status/types_gen.py
```diff
@@ -59,7 +59,7 @@ class Status(BaseModel, Generic[Details]):
     name: str = Field(default="")
     variant: Variant
     message: str
-    description: str | None = None
+    description: str = Field(default="")
```
pjdotson: is there a reason not to just do `description: str = ""`

x/go/gorp/migrate.go
pjdotson, line 1: add test

core/pkg/distribution/channel/lease_proxy.go
```diff
@@ -60,6 +61,11 @@ func (s *Service) renameHandler(ctx context.Context, msg RenameRequest) (types.N
 
 func (s *Service) create(ctx context.Context, tx gorp.Tx, _channels *[]Channel, opts CreateOptions) error {
 	channels := *_channels
+	for i := range channels {
+		if err := channels[i].Validate(); err != nil {
+			return validate.PathedError(err, strconv.Itoa(i))
+		}
+	}
```
pjdotson, line 68: add test

core/pkg/distribution/ontology/ontology_suite_test.go
```diff
@@ -39,7 +39,7 @@ type sampleService struct {
 
 var _ ontology.Service = (*sampleService)(nil)
 
-const sampleOntologyType ontology.ResourceType = "sample"
+const sampleOntologyType ontology.ResourceType = "channel"
```
pjdotson: kinda pointless to not just just `ontology.ChannelResourceType` now

core/pkg/service/user/types.gen.go
```diff
@@ -33,3 +34,9 @@ type User struct {
 	// cannot be deleted.
 	RootUser bool `json:"root_user" msgpack:"root_user"`
 }
+
+func (u User) Validate() error {
```
pjdotson, line 38: should add validation tests to writers everywhere you have added this functionality

client/ts/src/view/types.gen.ts
```diff
@@ -24,7 +24,7 @@ export type Key = z.infer<typeof keyZ>;
  */
 export const viewZ = z.object({
   /** key is the unique identifier for this view. */
-  key: keyZ,
+  key: keyZ.default(() => uuid.create()),
```
pjdotson: `.default(uuid.create)` would look better

client/ts/src/table/types.gen.ts
```diff
@@ -83,14 +83,7 @@ export const tableZ = z.object({
...
 
-export const newZ = tableZ
-  .omit({ key: true, rows: true, columns: true, cells: true })
-  .extend({
-    key: keyZ.default(() => uuid.create()),
-    rows: array.nullishToEmpty(rowZ),
-    columns: array.nullishToEmpty(columnZ),
-    cells: caseconv.preserveCase(record.nullishToEmpty(z.string(), cellZ)),
-  });
+export const newZ = tableZ;
```
pjdotson: whats the point of not just using `tableZ`?

client/ts/src/schematic/schematic.spec.ts
```diff
@@ -140,7 +140,9 @@ describe("Schematic", () => {
         });
         await expect(
           client.schematics.dispatch(schem2.key, "sess-1", [
-            schematic.setNode({ node: { key: "n1", position: { x: 0, y: 0 } } }),
+            schematic.setNode({
+              node: { key: "n1", position: { x: 0, y: 0 }, zIndex: 0 },
```
pjdotson: a little odd that defaults aren't applied to the actions here - no reason `setNode` couldn't handle changing `zIndex` to one internally.

client/py/synnax/task/types_gen.py
```diff
@@ -50,7 +51,9 @@ class Command(BaseModel):
     task: Key = Field(ge=0, le=18446744073709551615)
     type: str
     key: str
-    args: dict[str, Any] | None = None
+    args: Annotated[dict[str, Any], BeforeValidator(dicts.none_to_empty)] = Field(
```
pjdotson: similar question here - why are we ever getting an explicit `None` for args, why do we need the default factory and the `BeforeValidator`

client/ts/src/device/types.gen.ts
```diff
@@ -90,12 +90,12 @@ export const newZ = <
   Make extends z.ZodType<string> = z.ZodString,
   Model extends z.ZodType<string> = z.ZodString,
 >({ properties, make, model }: Partial<NewSchemas<Properties, Make, Model>> = {}) =>
-  deviceZ({ properties, make, model }).partial({ key: true, configured: true });
+  deviceZ({ properties, make, model }).partial({ configured: true });
```
pjdotson: should we just be getting rid of all `newZ` everywhere they appear, since they seem to just be the same as the other schemas?

pluto/src/access/role/queries.ts
```diff
@@ -226,7 +229,7 @@ export const useForm = Flux.createForm<
   },
   update: async ({ client, value, store, set, rollbacks }) => {
     const v = value();
-    let r: access.role.Role = { key: uuid.create(), ...v };
+    let r: access.role.Role = { ...v, key: v.key || uuid.create() };
```
pjdotson: too cute

pluto/src/vis/diagram/Diagram.tsx
```diff
@@ -389,8 +394,31 @@ export const create = ({
     );
 
     const handleNodesChange = useCallback(
-      (changes: RFNodeChange[]) =>
-        processChanges(changes, translateNodeChangeForward, onNodesChange),
+      (changes: RFNodeChange[]) => {
+        const passthrough: RFNodeChange[] = [];
```
pjdotson: can you add tests somehow?

pjdotson (changes requested):

schemas/synnax/policy.oracle
emilbon99: Yes

schemas/synnax/ranger.oracle
emilbon99: Makes it not run validation on that field. Otherwise it would validate all of the fields on `Parent`. including name.

client/ts/src/device/types.gen.ts
emilbon99: I've removed the vast majority of `newZ` generation from the codebase. There are a few left that still have handwritten `New` oracle structs. I'd like to leave them that way for now, and make decisions further down the line to remove as we refine the data structures.

client/py/synnax/task/types_gen.py
emilbon99: The `core` sends 'None' for empty arrays that have not been allocated. To completely get rid of that, we have to make it so array fields returned by the core are always allocated i.e. they encode to `json` as `[]`. Either we manually make sure every slice gets the empty array, we generate custom `MarshalJSON` methods for every generated type sent over the wire that has a slice field, or we implement our own custom JSON encoder.

This seemed like the best alternative to those options.

client/ts/src/table/types.gen.ts
emilbon99: Removed.

client/ts/src/schematic/schematic.spec.ts
emilbon99: Implemented.

client/py/synnax/device/types_gen.py
emilbon99: Removed.

x/py/x/control/types_gen.py
emilbon99: The core sends 'None' for empty arrays that have not been allocated. To completely get rid of that, we have to make it so array fields returned by the core are always allocated i.e. they encode to json as []. Either we manually make sure every slice gets the empty array, we generate custom MarshalJSON methods for every generated type sent over the wire that has a slice field, or we implement our own custom JSON encoder.

This seemed like the best alternative to those options.

schemas/synnax/status.oracle
emilbon99: Revmoed.

client/py/synnax/status/types_gen.py
emilbon99: Implemented.

x/py/x/control/types_gen.py
pjdotson: Couldn't you get rid of it by adding `omitzero` tags to array / record types? Then the Python code won't receive anything for the field at all, so it will use the default here.

pluto/src/flux/dispatch.spec.tsx
```diff
@@ -135,23 +135,6 @@ describe("Flux.createDispatch", () => {
...
-          key,
-          actions: schematic.setNodeMeasured({
-            key: "n1",
-            measured: { width: 1, height: 1 },
-          }),
-        });
-      });
-      expect(result.current.undo.canUndo).toBe(false);
-    });
-
```
pjdotson, line 154: why was this test removed?

core/pkg/service/lineplot/writer.go
```diff
@@ -42,6 +42,10 @@ func (w Writer) Create(ctx context.Context, projectKey project.Key, lp *LinePlot
 	// Materialize lines for any channel/range bindings supplied at creation so a plot
 	// created with channels and ranges but no lines is fully populated.
 	lp.Lines = reconcileLines(*lp)
+	*lp = lp.ApplyDefaults()
```
pjdotson: is there a reason to return an entirely new struct instead of just modifying in-place?

core/pkg/service/project/writer.go
```diff
@@ -48,14 +50,6 @@ func (w Writer) Create(
...
 		return err
 	}
-	if err := w.otg.DefineRelationship(
-		ctx,
-		user.OntologyID(p.Author),
-		ontology.RelationshipTypeParentOf,
-		otgID,
-	); err != nil {
-		return err
-	}
```
pjdotson, line 58: do you need a migration for deleting this ontology relationships as well?

core/pkg/service/schematic/migrate.go
```diff
@@ -78,7 +78,8 @@ func migrateNode(n v0.Node) Node {
 		Position: spatial.XY{X: n.Position.X, Y: n.Position.Y},
 	}
 	if n.ZIndex != nil {
-		out.ZIndex = int16(*n.ZIndex)
+		z := int16(*n.ZIndex)
+		out.ZIndex = z
```
pjdotson: weird change

pjdotson (changes requested):

core/pkg/service/lineplot/writer.go
emilbon99: I prefer value semantics on methods like that, as it makes the caller explicitly modify the value by pointer if they want.  Also follows the config override pattern. Want me to refactor?

pluto/src/flux/dispatch.spec.tsx
emilbon99: We deleted the `setNodeMeasured` action and I realized that this test was pretty low value, as the logic was already covered by other tests in the suite and related ones. Decided it was best left removed.

core/pkg/service/project/writer.go
emilbon99: Implemented.

x/py/x/control/types_gen.py
emilbon99: Good suggestion. Implemented.

emilbon99:
@pjdotson I implemented the `omitzero` change for array encoding. That meant I had to switch all remaining codecs (python streamer) to use `JSONCodec`. We should (relatively soon) completely remove the msgpack codec from the codebase. 

Issue link: https://linear.app/synnax/issue/SY-4412/completely-remove-msgpack-an-an-encoding-mechanism-and-transport

core/pkg/service/lineplot/writer.go
pjdotson: would rather have it modify in-place if every use case will look like `*v = v.ApplyDefaults()`, plus we are already may be implicitly modifying things like maps and slices in the value anyways.

emilbon99: Implemented.

oracle/resolution/type.go
```diff
@@ -328,6 +328,49 @@ func primitiveBase(ref TypeRef, table *Table, visited set.Set[string]) string {
 	return ""
 }
 
+// CollectionKind reports the collection kind that ref ultimately resolves to:
+// "Array", "Map", or "record". It follows type-parameter constraints and defaults,
+// alias targets, and distinct bases. ok is false when ref does not resolve to a
+// collection. Generators use it to decide which fields are nilable collections (for
+// omitzero tagging and empty defaults) even when the field is typed through an alias
+// or a constrained type parameter.
+func CollectionKind(ref TypeRef, table *Table) (kind string, ok bool) {
```
pjdotson, line 337: need tests

core/pkg/service/project/migrate.go
```diff
@@ -202,6 +202,34 @@ func rewriteWorkspaceRelationships(ctx context.Context, tx gorp.Tx) error {
...
 }
 
+// RemoveAuthorRelationships deletes the parent-of relationships that pointed from a
+// project's author user to the project. The author field is no longer part of a
+// project, so its relationships are orphaned on existing clusters; a project's only
+// parent is now the Projects group. This must run after the workspace-to-project
+// migration, which re-types the legacy user-to-workspace relationships to
+// user-to-project before they can be matched here.
+func RemoveAuthorRelationships(ctx context.Context, tx gorp.Tx, _ alamos.Instrumentation) error {
+	stale, err := collectEntries(
```
pjdotson, line 212: why can't this use the normal ontology writer / reader

pjdotson (changes requested):

core/pkg/service/project/migrate.go
emilbon99: Implemented.

oracle/resolution/type.go
emilbon99: One sec implementing this, sorry accidentally re-requrested review too early

emilbon99: Ok good to go.

emilbon99:
@pjdotson ready to go

pjdotson (approved):

PR #2515, opened 2026-06-22
SY-4415: Remove Theme State from the Console Layout Slice

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4415](https://linear.app/synnax/issue/SY-4415)

## Description

Theming is now owned entirely by Pluto's built-in `Theming.Provider` rather than the console's layout Redux slice. Pluto's provider already defaults the theme from the OS `prefers-color-scheme`, subscribes to OS changes, applies CSS vars and the worker theme, and exposes `toggleTheme`/`setTheme` via `Theming.useContext()` — so the parallel Redux machinery was redundant.

This removes from the layout slice:

- The `activeTheme` and `themes` state fields (dropped at their `v0`/`v5` schema declaration sites — no new migration; old persisted/exported theme blobs are ignored on rehydrate and stripped by zod on import parse).
- The `setActiveTheme`/`toggleActiveTheme` reducers and the `SetActiveThemePayload` type.
- The `selectActiveThemeKey`/`selectTheme`/`selectRawTheme`/`useSelectTheme` selectors.
- The `useThemeProvider` bridge hook (deleted) and the `themes` entry in `PERSIST_EXCLUDE`.

`Console.tsx` now renders `Pluto.Provider` with no `theming` prop, the command-palette "Toggle color theme" command calls `Theming.useContext().toggleTheme`, and the project importer no longer needs its stale-theme defense.

**Behavior change:** theme selection no longer persists to disk or syncs across windows via Drift. Each window follows the OS preference plus an in-session toggle, matching Pluto's built-in behavior.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR removes the parallel Redux-based theme management from the console's layout slice, delegating theming entirely to Pluto's built-in `Theming.Provider`. The `useThemeProvider` bridge hook, `setActiveTheme`/`toggleActiveTheme` reducers, `selectActiveTheme*` selectors, and the `themes`/`activeTheme` state fields are deleted; `Pluto.Provider` now renders without a `theming` prop and `ToggleCommand` calls `Theming.useContext().toggleTheme` directly.

- Removes `activeTheme` and `themes` from all layout slice schema versions (v0 and v5) so old exported blobs with stale theme shapes are silently stripped by Zod rather than validated; `stripThemes` in `import.ts` and its test are correctly retired as a result.
- Intentional behavior change: the selected theme is no longer persisted to disk or synced across windows via Drift — each window follows the OS preference plus an in-session toggle.

Safe to merge — the change is a clean removal of redundant Redux theming machinery with no new logic introduced.

Every deleted symbol (reducers, selectors, the bridge hook) has been traced to zero remaining call sites. The Zod-schema changes correctly shed themes/activeTheme from all version schemas, so old exports with those fields are silently stripped rather than rejected. The stripThemes guard in the importer was the last defensive layer that referenced the removed fields; its removal is safe because Zod's default strip mode makes it redundant. The intentional behavior change (theme no longer persists across sessions) is clearly documented in the PR description.

| Filename | Overview |
|----------|----------|
| console/src/Console.tsx | Removes useThemeProvider call and theming prop from Pluto.Provider, delegating theme management to Pluto's built-in provider. |
| console/src/layout/external.ts | Removes useThemeProvider export, consistent with the file deletion. |
| console/src/layout/selectors.ts | Deletes selectActiveThemeKey, selectTheme, selectRawTheme, and useSelectTheme; drops unused UnexpectedError and Theming imports. |
| console/src/layout/slice.ts | Removes setActiveTheme, toggleActiveTheme reducers, SetActiveThemePayload type, and themes from PERSIST_EXCLUDE; cleans up the themes/activeTheme copy in setProject/clearAll. |
| console/src/layout/types/v0.ts | Removes activeTheme and themes fields from the v0 schema; all later versions (v1-v4) inherit from v0 so they shed these fields automatically. |
| console/src/layout/types/v5.ts | Removes the explicit themes override added at v5 and the initialization of activeTheme/themes in the v4->v5 migration; the schema now simply bumps the version string. |
| console/src/layout/useThemeProvider.ts | File deleted; the functionality it provided (OS-theme sync, Redux dispatch) is now handled natively by Pluto's Theming.Provider. |
| console/src/project/services/import.ts | Removes stripThemes helper and its call site; safe because Zod's default strip mode silently drops unknown themes/activeTheme keys from old exports. |
| console/src/project/services/import.spec.tsx | Removes staleThemesSlice helper and the stale themes import test; the test scenario is no longer possible since the schema does not validate theme color fields. |
| console/src/theme/palette.tsx | Switches ToggleCommand from dispatching a Redux action to calling Theming.useContext().toggleTheme; store prop is no longer destructured but is still passed and stripped by CommandListItem. |
| console/src/layout/slice.spec.ts | Removes tests for setActiveTheme and toggleActiveTheme, keeping coverage aligned with the remaining slice behavior. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph Before["Before (Redux-managed theming)"]
        OS1[OS prefers-color-scheme] -->|MediaQueryList event| UTP[useThemeProvider hook]
        UTP -->|dispatch setActiveTheme| RS[(Redux layout slice\nthemes / activeTheme)]
        RS -->|useSelectTheme| PP1[Pluto.Provider theming prop]
        PP1 --> UI1[UI renders with theme]
        TC1[ToggleCommand] -->|store.dispatch setActiveTheme| RS
    end

    subgraph After["After (Pluto-native theming)"]
        OS2[OS prefers-color-scheme] -->|built-in subscription| TP[Pluto Theming.Provider\ninside Pluto.Provider]
        TP --> UI2[UI renders with theme]
        TC2[ToggleCommand] -->|Theming.useContext.toggleTheme| TP
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
flowchart TD
    subgraph Before["Before (Redux-managed theming)"]
        OS1[OS prefers-color-scheme] -->|MediaQueryList event| UTP[useThemeProvider hook]
        UTP -->|dispatch setActiveTheme| RS[(Redux layout slice\nthemes / activeTheme)]
        RS -->|useSelectTheme| PP1[Pluto.Provider theming prop]
        PP1 --> UI1[UI renders with theme]
        TC1[ToggleCommand] -->|store.dispatch setActiveTheme| RS
    end

    subgraph After["After (Pluto-native theming)"]
        OS2[OS prefers-color-scheme] -->|built-in subscription| TP[Pluto Theming.Provider\ninside Pluto.Provider]
        TP --> UI2[UI renders with theme]
        TC2[ToggleCommand] -->|Theming.useContext.toggleTheme| TP
    end
```

</a>

<sub>Reviews (1): Last reviewed commit: ["SY-4415: Remove theme state from the con..."](https://github.com/synnaxlabs/synnax/commit/8ece7e0d4e7a10bbc05637d4b1a15fe639199d49) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39135316)</sub>


pjdotson (changes requested):
Now if my OS is in light mode, and I open a secondary window, I have no way to switch the secondary window to dark mode without changing the OS's color theme as the palette command only changes the main window's color theme. Seems like odd behavior.

console/src/layered/session/theme/slice.ts
```diff
@@ -0,0 +1,46 @@
...
+  })
+  .prefault({});
+
+export interface SliceState extends z.output<typeof sliceStateZ> {}
+
+export interface StoreState {
+  [SLICE_NAME]: SliceState;
+}
+
+export const ZERO_SLICE_STATE: SliceState = sliceStateZ.parse({});
```
pjdotson, line 28: is there a reason to have all of these `defaults`/`prefaults` if they are not actually needed in the code? e.g. we could just define `ZERO_SLICE_STATE = {version: 0, selected: "synnaxLight"}` right here and remove all the default/prefault as those aren't being used.

pjdotson (changes requested):

console/src/layered/session/theme/slice.ts
emilbon99: I thought you wanted to move away from having hardcoded `ZERO` structs and rely more on `zod` schema parsing? That's why I went the defaulting and pre-faulting route. I think it's good for a few reasons:

1) It makes the system more resilient i.e. we'd rather things default out than fail for our users. Downside is that it's harder to find cases where an actual value should have been propagated but defaulted over. 

2) I assume that eventually all of our migrations will rely on schema parsing, and if we do in fact want to have defaults for things in the schema we probably want the defaults defined in one place.

pjdotson: 1. why are things failing for our users anyways? we should be using `safeParse` for parsing the persisted state and if it fails goes to the zero slice. The only time we should have to use the Zod schemas for session state is on bootup, so we should call `sliceStateZ.safeParse(stored)`, and if that fails, `console.error` and change to the `ZERO_SLICE_STATE`. If we call `sliceStateZ.parse(stored)` and hope our defaults work fine then there is some risk of erroring, so we still need this `ZERO_SLICE_STATE` struct
2. Good point. can keep

emilbon99: Copy, I'm going to ship as is. I think the decision on the parsing side will come when I refactor state persistence logic.

PR #2522, opened 2026-06-23
SY-3961: Arc Ranges (Create & End)

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-3961](https://linear.app/synnax/issue/SY-3961)

## Description

Adds a `ranges` Arc stdlib module so automation programs can create and close Synnax ranges directly, on both the Go server runtime and the C++ driver.

Members:

* `ranges.create(name, color, parent)` creates a range that starts now and stays open (end at `TimeStampMax`), returning its key. `color` is optional and accepts hex (`#3bc454`), `rgb(r, g, b)`, or `rgba(r, g, b, a)`; an invalid color literal is flagged during analysis. `parent` is an optional parent range key that nests the new range under an existing one.
* `ranges.end(key)` sets the end bound of an existing range to now and returns the key, closing an open range.

Both members work in func form (`key := ranges.create("Burn")`) and flow form (`trigger -> ranges.create{name="Burn"}`).

Implementation notes:

* Registered in the Arc root, so the module resolves without extra wiring.
* `ranges.end` is atomic via `ranger.SetEnd` (retrieve, set end, write in one transaction) to avoid a read-modify-write race.
* Mirrored on the C++ driver runtime so range programs run on external racks the same as on the embedded server.
* `ranges.create` leaves the range color unset when none is given, kept consistent across Go (`*color.Color` nil) and C++ (`std::optional` empty) so a colorless range round-trips as `null`.

Scope decisions:

* `ranges.end` takes only `key` and always evaluates at now. An earlier attempt to allow computed expressions in flow brace inputs (e.g. `ranges.end{key, time.now() + 1h}`) was reverted because it added significant complexity and open design questions for marginal value; the supporting mechanism (synthetic expression nodes, `node.ResolveInputs`, `telem.Series.AtAny`) was removed.
* Dropped an earlier `ranges.start` member and the `create` time/label params to keep the surface minimal.

Tests:

* Go unit tests for the module, symbol signatures, WASM host functions, and analyzer color validation.
* C++ unit tests for `dispatch_create` / `dispatch_end` and node construction.
* Integration test `arc/stl_ranges` exercising both func and flow forms on the embedded driver and an external rack, asserting each range's start, end (now vs `TimeStampMax`), color (hex/rgb/rgba/none), and parent relationship.

## Basic Readiness

### Greptile Summary

This PR adds a `ranges` Arc stdlib module so automation programs can create and close Synnax ranges directly, implemented on both the Go server runtime (with a new atomic `ranger.SetEnd`) and the C++ driver (for external racks). A new `color.FromCSS` utility is added to both Go and C++ to validate and parse hex/rgb/rgba color strings at both analysis and runtime.

* `ranges.create(name, color?, parent?)` — creates an open range starting now (`end = TimeStampMax`), with optional color and parent-range nesting, returning the new key in both func and flow form.
* `ranges.end(key)` — closes the range by setting its end to now, with the Go server path wrapping the read-modify-write in a DB transaction (`SetEnd`) for atomicity; the C++ driver path uses a retrieve-modify-upsert over gRPC (explicitly noted in the doc comment as non-atomic, since the client API has no server-side transaction support).
* `color.FromCSS` — mirrors the same parsing logic and error messages across Go and C++, with analyzer integration that flags invalid color literals at analysis time before the program runs.

### Confidence Score: 4/5

Safe to merge; the only non-documentation issue is a bare `print()` call in the integration test framework that adds an unintended blank line to stdout on every log dump.

The implementation is well-structured and comprehensively tested — Go unit tests, C++ unit tests, and an integration test covering all color formats, parent nesting, and both func/flow execution forms. The `ranger.SetEnd` atomic transaction and the color parsing utilities are correct. The sole finding is a bare `print()` in `log_client.py`'s `dump()` method that sits outside the documented contract and appears to be leftover debug output.

integration/framework/log_client.py — the bare `print()` in `dump()` should be removed or documented if intentional.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| core/pkg/service/arc/ranges/ranges.go | New Arc stdlib `ranges` module: implements `create` (open range) and `end` (close range) in both func and flow forms, with WASM host bindings, color/UUID validation, and analyzer color-literal hook. |
| core/pkg/service/ranger/service.go | Adds `SetEnd` — an atomic read-modify-write wrapped in a DB transaction that sets only the end timestamp of a range, preserving all other fields. |
| driver/arc/ranges/ranges.h | C++ mirror of the Go ranges module; `dispatch_end` uses a retrieve-modify-upsert (non-atomic, acknowledged in comment) since the gRPC client API has no server-side transaction support. |
| x/go/color/color.go | Adds `FromCSS` to the Go color package: parses '#'-prefixed hex or rgb()/rgba() strings, with channel/alpha range validation and clean error messages. |
| x/cpp/color/color.h | C++ counterpart of `FromCSS`/`from_hex`, implementing the same parsing logic and error messages as the Go version. |
| integration/tests/arc/stl_ranges.py | Integration test exercising all color formats (hex/rgb/rgba/none), parent relationship, and both func and flow forms for ranges.create/end. |
| integration/framework/log_client.py | Single-line change adds a bare `print()` to `dump()` that is not part of the documented contract and appears to be leftover debug output. |
| core/pkg/service/arc/task/factory.go | Wires the `ranger.Service` dependency into `FactoryConfig` with appropriate validation so the ranges module can be instantiated per task. |
| core/pkg/service/arc/task/task.go | Instantiates the ranges module during task startup and adds it to the node compound factory, consistent with how other modules (status, math) are registered. |

### Sequence Diagram

```mermaid
sequenceDiagram
    participant Arc as Arc Program
    participant Node as ranges Node (Go/C++)
    participant Dispatch as dispatch_create / dispatch_end
    participant RangerSvc as ranger.Service (Go)
    participant Client as synnax.Client (C++)
    participant DB as Database

    Note over Arc,DB: ranges.create flow
    Arc->>Node: trigger fires → createNode.Next()
    Node->>Dispatch: dispatchCreate(name, color, parent)
    Dispatch->>Dispatch: parse color (FromCSS)
    Dispatch->>Dispatch: parse parent UUID
    alt Go server runtime
        Dispatch->>RangerSvc: "NewWriter(nil).Create(ctx, &r)"
        RangerSvc->>DB: INSERT/UPSERT range
    else C++ driver runtime
        Dispatch->>Client: ranges.create(r)
        Client-->>DB: RPC → server upserts
    end
    Dispatch-->>Node: range key string
    Node-->>Arc: "Output(0) = key"

    Note over Arc,DB: ranges.end flow
    Arc->>Node: trigger fires → endNode.Next()
    Node->>Dispatch: dispatchEnd(key)
    Dispatch->>Dispatch: parse UUID
    alt Go server runtime (atomic)
        Dispatch->>RangerSvc: SetEnd(ctx, uid, Now())
        RangerSvc->>DB: BEGIN TX
        RangerSvc->>DB: SELECT range by key
        RangerSvc->>DB: "UPDATE range.end = now"
        RangerSvc->>DB: COMMIT
    else C++ driver runtime (RMW, not atomic)
        Dispatch->>Client: ranges.retrieve_by_key(uid)
        Client-->>Dispatch: range r
        Dispatch->>Dispatch: "r.time_range.end = now"
        Dispatch->>Client: ranges.create(r)
        Client-->>DB: RPC → server upserts
    end
    Dispatch-->>Node: key string
    Node-->>Arc: "Output(0) = key"
```

```mermaid
sequenceDiagram
    participant Arc as Arc Program
    participant Node as ranges Node (Go/C++)
    participant Dispatch as dispatch_create / dispatch_end
    participant RangerSvc as ranger.Service (Go)
    participant Client as synnax.Client (C++)
    participant DB as Database

    Note over Arc,DB: ranges.create flow
    Arc->>Node: trigger fires → createNode.Next()
    Node->>Dispatch: dispatchCreate(name, color, parent)
    Dispatch->>Dispatch: parse color (FromCSS)
    Dispatch->>Dispatch: parse parent UUID
    alt Go server runtime
        Dispatch->>RangerSvc: "NewWriter(nil).Create(ctx, &r)"
        RangerSvc->>DB: INSERT/UPSERT range
    else C++ driver runtime
        Dispatch->>Client: ranges.create(r)
        Client-->>DB: RPC → server upserts
    end
    Dispatch-->>Node: range key string
    Node-->>Arc: "Output(0) = key"

    Note over Arc,DB: ranges.end flow
    Arc->>Node: trigger fires → endNode.Next()
    Node->>Dispatch: dispatchEnd(key)
    Dispatch->>Dispatch: parse UUID
    alt Go server runtime (atomic)
        Dispatch->>RangerSvc: SetEnd(ctx, uid, Now())
        RangerSvc->>DB: BEGIN TX
        RangerSvc->>DB: SELECT range by key
        RangerSvc->>DB: "UPDATE range.end = now"
        RangerSvc->>DB: COMMIT
    else C++ driver runtime (RMW, not atomic)
        Dispatch->>Client: ranges.retrieve_by_key(uid)
        Client-->>Dispatch: range r
        Dispatch->>Dispatch: "r.time_range.end = now"
        Dispatch->>Client: ranges.create(r)
        Client-->>DB: RPC → server upserts
    end
    Dispatch-->>Node: key string
    Node-->>Arc: "Output(0) = key"
```

Reviews (1): Last reviewed commit: ["Merge remote-tracking branch < />origin/rc< />..."](<https://github.com/synnaxlabs/synnax/commit/c960b235c647dcd66efd510a41fc6aa55d271b83>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=39397705>)

<h3>Greptile Summary</h3>

This PR adds a `ranges` Arc stdlib module that lets automation programs create and close Synnax ranges directly, implemented symmetrically on the Go server runtime and the C++ driver. A new `color.FromCSS` utility (Go + C++) parses hex/rgb/rgba color strings, with analyzer-time validation so invalid color literals are flagged before the program runs.

- **`ranges.create(name, color?, parent?)`** — creates an open range starting now (`end = TimeStampMax`), with optional color and parent-range nesting, returning the new key in both func and flow form.
- **`ranges.end(key)`** — closes a range by setting its end to now; the Go server path wraps the read-modify-write in a gorp `ChangeErr` (DB-level atomic update), and the C++ driver path calls the new `RangeSetEndService` gRPC endpoint which runs the same atomic `SetEnd` on the server.
- **`color.FromCSS`** — mirrors the same parsing logic across Go and C++ with analyzer integration that flags invalid color literals at analysis time; both runtimes consistently reject invalid inputs, though `std::stod` and `strconv.ParseFloat` diverge on malformed multi-dot alpha strings (both still reject the input, but with different error messages).

Safe to merge; all new code paths are well-tested with Go unit tests, C++ unit tests, and an integration test covering every color format, parent nesting, and both func/flow forms.

The ranges module, atomic SetEnd, and color parsing are all implemented correctly. The new gRPC SetEnd endpoint is wired symmetrically on both the Go and C++ sides, the analyzer color-validation hook matches runtime behavior, and the ChangeErr-based update preserves start and other range fields. No logic errors or missing guards were found in the changed paths.

x/cpp/color/color.cpp — std::stod accepts a malformed alpha like '1.2.3' as 1.2 where Go's strconv.ParseFloat rejects the whole string; a minor message divergence between runtimes on an already-invalid input.

| Filename | Overview |
|----------|----------|
| core/pkg/service/arc/ranges/ranges.go | New Arc stdlib ranges module — implements create and end in func/flow forms with WASM host bindings, color/UUID validation, and analyzer color-literal hook; logic is correct and well-tested. |
| core/pkg/service/ranger/writer.go | Adds SetEnd — atomic read-modify-write using gorp's ChangeErr with start/end validation; mirrors the existing Rename pattern and is thoroughly tested. |
| x/go/color/color.go | Adds FromCSS with hex and rgb/rgba parsing; regex correctly limits channel digits to 1-3, ParseUint with bitSize=8 enforces 0-255, and alpha 0-1 is checked after float parse. |
| x/cpp/color/color.cpp | C++ counterpart of FromCSS; logic mirrors Go closely except std::stod parses '1.2.3' as 1.2 (triggering 'alpha must be 0-1') while Go's ParseFloat rejects the string entirely ('invalid rgb alpha') — a minor behavioral divergence on an already-invalid input. |
| driver/arc/ranges/ranges.h | C++ mirror of the Go ranges module; dispatch_end correctly calls the new atomic set_end gRPC endpoint rather than doing a manual retrieve-modify-upsert. |
| core/pkg/transport/grpc/ranger/handler.go | Adds setEndServer and setEndRequestTranslator using the same pattern as existing rename/delete translators; correctly wires the new gRPC service into the compound transport. |
| integration/tests/arc/stl_ranges.py | Integration test exercising all color formats, parent nesting, and func/flow forms; uses toleranced timestamp comparison for the ranges.end case and verifies parent relationship via ontology children. |
| core/pkg/service/arc/task/factory.go | Adds Ranger *ranger.Service to FactoryConfig with override and validation, consistent with existing config fields. |
| docs/site/src/pages/reference/control/arc/reference/standard-library/ranges.mdx | New documentation page for the ranges module; parameters, types, and code examples are accurate and navigation links are correctly updated in math.mdx and status.mdx. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Arc as Arc Program
    participant Node as ranges Node (Go/C++)
    participant RangerSvc as ranger.Service (Go)
    participant SetEndGRPC as RangeSetEndService (gRPC)
    participant DB as Database

    Note over Arc,DB: ranges.create
    Arc->>Node: trigger → createNode.Next()
    Node->>Node: color.FromCSS / x::color::from_css()
    Node->>Node: uuid.Parse(parent)
    Node->>RangerSvc: "NewWriter(nil).Create(ctx, &r)"
    RangerSvc->>DB: "INSERT range (start=now, end=TimeStampMax)"
    DB-->>Node: r.Key assigned
    Node-->>Arc: "Output(0) = key string"

    Note over Arc,DB: ranges.end (Go server runtime)
    Arc->>Node: trigger → endNode.Next()
    Node->>Node: uuid.Parse(key)
    Node->>RangerSvc: NewWriter(nil).SetEnd(ctx, uid, now)
    RangerSvc->>DB: UPDATE via ChangeErr (atomic read-modify-write)
    DB-->>Node: ok
    Node-->>Arc: "Output(0) = key string"

    Note over Arc,DB: ranges.end (C++ driver runtime)
    Arc->>Node: trigger → EndRange::next()
    Node->>Node: x::uuid::UUID::parse(key)
    Node->>SetEndGRPC: ranges.set_end(uid, now)
    SetEndGRPC->>RangerSvc: SetEnd(ctx, key, end) [atomic on server]
    RangerSvc->>DB: UPDATE via ChangeErr (atomic read-modify-write)
    DB-->>Node: ok
    Node-->>Arc: "Output(0) = key string"
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
sequenceDiagram
    participant Arc as Arc Program
    participant Node as ranges Node (Go/C++)
    participant RangerSvc as ranger.Service (Go)
    participant SetEndGRPC as RangeSetEndService (gRPC)
    participant DB as Database

    Note over Arc,DB: ranges.create
    Arc->>Node: trigger → createNode.Next()
    Node->>Node: color.FromCSS / x::color::from_css()
    Node->>Node: uuid.Parse(parent)
    Node->>RangerSvc: "NewWriter(nil).Create(ctx, &r)"
    RangerSvc->>DB: "INSERT range (start=now, end=TimeStampMax)"
    DB-->>Node: r.Key assigned
    Node-->>Arc: "Output(0) = key string"

    Note over Arc,DB: ranges.end (Go server runtime)
    Arc->>Node: trigger → endNode.Next()
    Node->>Node: uuid.Parse(key)
    Node->>RangerSvc: NewWriter(nil).SetEnd(ctx, uid, now)
    RangerSvc->>DB: UPDATE via ChangeErr (atomic read-modify-write)
    DB-->>Node: ok
    Node-->>Arc: "Output(0) = key string"

    Note over Arc,DB: ranges.end (C++ driver runtime)
    Arc->>Node: trigger → EndRange::next()
    Node->>Node: x::uuid::UUID::parse(key)
    Node->>SetEndGRPC: ranges.set_end(uid, now)
    SetEndGRPC->>RangerSvc: SetEnd(ctx, key, end) [atomic on server]
    RangerSvc->>DB: UPDATE via ChangeErr (atomic read-modify-write)
    DB-->>Node: ok
    Node-->>Arc: "Output(0) = key string"
```

</a>

<sub>Reviews (3): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/rc&#39;..."](https://github.com/synnaxlabs/synnax/commit/edf8afa7addea0d13826084aca0f8ebb02f3e245) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39397705)</sub>


sy-nico:
Color selector works without any additional changes:
image

integration/framework/log_client.py
```diff
@@ -165,6 +165,7 @@ def dump(self) -> None:
         """Emit all buffered entries to sinks, then clear the buffer."""
         for entry in self._buffer:
             self._emit(entry)
+        print()
```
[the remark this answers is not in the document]
sy-nico: Added on purpose as a way to add spacing between the dump and the next output set. Light addition that requires no comment and should not impact review.

sy-nico:
@greptile, rerun. Comment addressed as non-issue.

core/pkg/service/ranger/service.go
```diff
@@ -195,3 +196,16 @@ func (s *Service) RetrieveParentKey(
 	}
 	return KeyFromOntologyID(resources[0].ID)
 }
+
+// SetEnd atomically sets the end bound of the range with the given key, preserving all
+// other fields. Returns query.ErrNotFound if no range with the key exists.
+func (s *Service) SetEnd(ctx context.Context, key Key, end telem.TimeStamp) error {
```
pjdotson: this should be a method on the `Writer` like `Rename`

core/pkg/service/ranger/set_end_test.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"github.com/google/uuid"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/synnaxlabs/synnax/pkg/service/ranger"
+	"github.com/synnaxlabs/x/query"
+	"github.com/synnaxlabs/x/telem"
+)
+
+var _ = Describe("SetEnd", func() {
```
pjdotson: this should be in `writer_test.go`

x/go/color/color.go
```diff
@@ -73,6 +75,59 @@ func MustFromHex(s string) Color {
...
+// rgb()/rgba() function. Unlike FromHex, the leading '#' is required.
+func FromCSS(s string) (Color, error) {
+	s = strings.TrimSpace(s)
+	if strings.HasPrefix(s, "#") {
+		return FromHex(s)
+	}
+	if m := rgbPattern.FindStringSubmatch(s); m != nil {
+		return fromRGBMatch(m, s)
+	}
+	return Color{}, errors.Newf(
```
pjdotson: these should all return things that wrap validate.error

x/go/color/color.go
```diff
@@ -73,6 +75,59 @@ func MustFromHex(s string) Color {
...
+
+func fromRGBMatch(m []string, s string) (Color, error) {
+	hasAlpha := m[5] != ""
+	if m[1] == "rgb" && hasAlpha {
+		return Color{}, errors.Newf("rgb() takes 3 channels; use rgba() for alpha: %q", s)
+	}
+	if m[1] == "rgba" && !hasAlpha {
+		return Color{}, errors.Newf("rgba() requires a 4th alpha channel: %q", s)
+	}
+	r, err := strconv.Atoi(m[2])
```
pjdotson: should be using `ParseUint` instead of `Atoi`

docs/site/src/pages/reference/control/arc/reference/standard-library/ranges.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+**Output** `str`: the new range key.
+
+</Details>
+
+```arc
+range_key := trigger -> ranges.create{name="Terminal Count", color="#3bc454"}
+
+func ranges_create() {
+    range_key = ranges.create("Terminal Count", "#3bc454")
```
pjdotson: how does the idea of "func context is things that immediately execute, no blocking
stuff" square away with the idea of having calls like `ranges.create` which will
inherently take time to execute due to the network round-trip? not saying this is an
issue with this PR in particular, but something you should think about.

docs/site/src/pages/reference/control/arc/reference/standard-library/ranges.mdx
```diff
@@ -0,0 +1,68 @@
...
+}
+```
+
+## `end`
+
+Sets the end bound of an existing range to now, closing an open range, and outputs the
+range key.
+
+<Details>
+<Text.Text slot="summary" level="p"><Icon.Details /> Parameters</Text.Text>
```
pjdotson, line 52: you should think about making some reusable component for Arc STL function documentation

integration/tests/arc/stl_ranges.py
```diff
@@ -0,0 +1,232 @@
...
+
+    A single trigger runs the func body and fires every flow create, producing
+    ranges with and without an explicit end, with rgb/rgba/hex colors, and under
+    a parent. Each created range is verified for start, end, color, and parent.
+    """
+
+    arc_source = ARC_STL_RANGES_SOURCE
+    arc_name_prefix = "ArcStlRanges"
+    start_cmd_channel = START_CMD
+    subscribe_channels = [DONE]
```
pjdotson, line 153: no point in defining these as constants above if they are only used in one place here

integration/tests/arc/stl_ranges.py
```diff
@@ -0,0 +1,232 @@
...
+    ranges.create("RangeFunc_RGBA_2", "rgba(0, 255, 0, 0.25)")
+    ranges.create("RangeFunc_RGBA_3", "rgba(0, 0, 255, 0.75)")
+
+    // Hex
+    ranges.create("RangeFunc_Hex_1", "#112233")
+    ranges.create("RangeFunc_Hex_2", "#44aa66")
+    ranges.create("RangeFunc_Hex_3", "#ddeeff")
+
+    // Parent (key injected by the test)
+    ranges.create("RangeFunc_Child", "rgb(10, 20, 30)", "__PARENT_KEY__")
```
pjdotson: what if i want to create a range with a parent and without a color? there should be
tests for that.

integration/tests/arc/stl_ranges.py
```diff
@@ -0,0 +1,232 @@
...
+class Case(NamedTuple):
+    """One created range and what it should look like afterward.
+
+    color is the expected (r, g, b, a) or None when the range has no color.
+    ends_now is True when ranges.end closed the range (end ~= start), False when
+    the range is left open (end == TimeStampMax).
+    """
+
+    name: str
+    color: tuple[int, int, int, float] | None
```
pjdotson: why isn't this the x.Color?

core/pkg/service/arc/task/factory.go
```diff
@@ -87,6 +93,7 @@ func (c FactoryConfig) Validate() error {
 	validate.NotNil(v, "framer", c.Framer)
 	validate.NotNil(v, "status", c.Status)
 	validate.NotNil(v, "get_program", c.GetProgram)
+	validate.NotNil(v, "ranger", c.Ranger)
```
pjdotson, line 96: should have Override/Validate tests

core/pkg/service/arc/ranges/ranges_suite_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+var _ = ShouldNotLeakGoroutinesPerSpec()
+
+var (
+	db       *gorp.DB
+	otg      *ontology.Ontology
+	groupSvc *group.Service
+	labelSvc *label.Service
+	rangeSvc *ranger.Service
+)
```
pjdotson: if these variables are only ever used in `BeforeSuite` they shouldn't be declared at the
package level like this

core/pkg/service/arc/ranges/ranges.go
```diff
@@ -0,0 +1,343 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+package ranges
```
pjdotson, line 10: now we have an inconsistency in Arc library naming: `ranges.create` vs `status.set`.
fine to keep for this PR but we should address before release.

core/pkg/service/arc/ranges/ranges.go
```diff
@@ -0,0 +1,343 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/synnaxlabs/arc/ir"
+	"github.com/synnaxlabs/arc/literal"
+	"github.com/synnaxlabs/arc/parser"
+	"github.com/synnaxlabs/arc/runtime/node"
+	stlstrings "github.com/synnaxlabs/arc/stl/strings"
```
pjdotson: just import as strings if it doesn't conflict with anything

core/pkg/service/arc/ranges/ranges.go
```diff
@@ -0,0 +1,343 @@
...
+			}
+			return heap.Create(dispatchEnd(ctx, m.rng, m.report, key))
+		}).Export(endMemberName)
+	if _, err := builder.Instantiate(ctx); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+func (m *Module) ModuleName() string { return moduleName }
```
pjdotson: why is `ModuleName()` part of the API here? shouldn't the public API just be a
`node.Factory` instead of having a separate `ranges.Module`?

core/pkg/service/arc/ranges/ranges.go
```diff
@@ -0,0 +1,343 @@
...
+
+func (m *Module) Create(_ context.Context, cfg node.Config) (node.Node, error) {
+	switch cfg.Node.Type {
+	case createMemberName:
+		var in createInputs
+		if err := createInputsSchema.Parse(cfg.Node.Inputs.ValueMap(), &in); err != nil {
```
pjdotson, line 181: why are we validating the types of the arguments in one place, but whether the string is
a valid CSS in another place? seems like that should happen at the same time?

x/cpp/color/color.h
```diff
@@ -9,6 +9,61 @@
...
+    }
+    return {
+        Color{},
+        x::errors::Error(x::errors::VALIDATION, "invalid hex color length: " + input)
+    };
+}
+
+/// @brief from_css parses a CSS color: a '#'-prefixed hex or an rgb()/rgba() value.
+/// Mirrors the Go color.FromCSS.
+std::pair<Color, x::errors::Error> from_css(const std::string &input);
```
pjdotson, line 24: weird to have some implementation in `color.h` and some in `color.cpp`

driver/arc/ranges/ranges.h
```diff
@@ -0,0 +1,286 @@
...
+                "ranges.create: invalid color \"" + color_hex + "\": " + err.message()
+            );
+        else
+            c = parsed;
+    }
+    synnax::ranger::Range r;
+    r.name = name;
+    r.time_range = x::telem::TimeRange{
+        x::telem::TimeStamp::now(),
+        x::telem::TimeStamp::max()
```
pjdotson: max should be a constant and not a function

driver/arc/ranges/ranges.h
```diff
@@ -0,0 +1,286 @@
...
+) {
+    auto [uid, parse_err] = x::uuid::UUID::parse(key);
+    if (parse_err) {
+        report(
+            synnax::status::VARIANT_WARNING,
+            "ranges.end: invalid range key \"" + key + "\""
+        );
+        return "";
+    }
+    auto [r, retrieve_err] = client->ranges.retrieve_by_key(uid);
```
pjdotson: Right now you have two network trips:
ranges.retrieve_by_key
ranges.create

would be much better to add a new API method ranges.set_end that mirrors ranges.rename
so this is only one network trip.

pjdotson (commented):
mostly good, see comments

pjdotson (changes requested):
wrong submission type on my end, my bad

docs/site/src/pages/reference/control/arc/reference/standard-library/ranges.mdx
sy-nico: [SY-4426](https://linear.app/synnax/issue/SY-4426/make-reusable-component-for-arc-stl-function-documentation)

core/pkg/service/ranger/service.go
sy-nico: Moved to `writer.go`

core/pkg/service/ranger/set_end_test.go
sy-nico: moved to `writer_test.go`

driver/arc/ranges/ranges.h
sy-nico: Added a `set_end` API method (gRPC + HTTP) backed by `Writer.SetEnd`. `dispatch_end` now closes the range in a single request via `client->ranges.set_end(uid, now())`.

x/cpp/color/color.h
sy-nico: Consolidated all implementation into `color.cpp`

core/pkg/service/arc/ranges/ranges.go
sy-nico: `NewModule` now returns `node.Factory` and `module` is unexported. I applied the same fix to `status` for consistency.

core/pkg/service/arc/ranges/ranges.go
sy-nico: They do happen together, the type check comes from the symbol's `color: String()` signature, and `analyzeCreateArguments` runs `color.FromCSS` on the literal at analysis time too, so a bad color is a compile-time error. Line 173's `Parse` is just runtime input deserialization, not the user-facing validation. CSS validity needs its own hook since the type system can't express "valid CSS" vs "string", and folding it into `Parse` would lose the compile-time error.

driver/arc/ranges/ranges.h
sy-nico: Added `x::telem::TIME_STAMP_MAX` as an `inline constexpr` constant (and made `TimeStamp`'s ctor + `max()`/`min()` `constexpr` to back it). `ranges.h` now uses the constant.

integration/tests/arc/stl_ranges.py
sy-nico: Added a parent-without-color case in the flow context (`RangeFlow_Child_NoColor`). The func/paren call form doesn't support named args (yet), so `ranges.create(name=..., parent=...)` is a syntax error.

I will add the equivalent test when we enable that feature.

core/pkg/service/arc/ranges/ranges.go
sy-nico: This is one where @emilbon99 decided in person that we would leave it as `ranges` and `status` and kick the can down the road. I don't feel strongly about this either way. Just say the word and I'll make the change.

core/pkg/service/arc/ranges/ranges_suite_test.go
sy-nico: Made `db`, `otg`, `groupSvc`, and `labelSvc` locals inside `BeforeSuite`; only `rangeSvc` (used by the specs) stays package-level.

docs/site/src/pages/reference/control/arc/reference/standard-library/ranges.mdx
sy-nico: I've been chewing on this a lot. Particularly on the Func/Sequence distinction and similarities. 

IMO the line isn't "takes zero time," it's atomic with respect to the tick: `ranges.create` does block on the round-trip, but it runs to completion inside a single scheduler tick and never yields control back to the graph, so nothing else advances or can interrupt it mid-call. That's what keeps it a compute node (a leaf), versus something like `time.wait` that yields and spans ticks. Latency ≠ yielding. There's a fuller model behind this (leaf vs. control-region, the yield litmus) that I am working on the side to capture in a doc.

core/pkg/service/arc/ranges/ranges.go
pjdotson: Copy

docs/site/src/pages/reference/control/arc/reference/standard-library/ranges.mdx
pjdotson: If it never yields control back to the graph - is there some worry in a scenario where
we have say a completely local Arc automation, it loses connection to the Core, it can
still execute reads and writes on local channels to that Driver bypass, but now it
stalls on `ranges.create`?

Just trying to throw some ideas to you, not particularly relevant for this PR but more
for bigger-picture Arc decisions.

core/pkg/service/arc/ranges/ranges.go
pjdotson: Copy. Should also consider if you allow computed expressions for something like this how
that relates to validating when it is a valid string or not, e.g `"#FF00" + my_var`
where the value of `my_var` is not known at compile time.

This is good for this PR.

core/pkg/service/ranger/writer.go
```diff
@@ -136,6 +137,21 @@ func (w Writer) Rename(ctx context.Context, key Key, name string) error {
...
+	var r Range
+	if err := w.table.
+		NewRetrieve().
+		Where(gorp.MatchKeys[Key, Range](key)).
+		Entry(&r).
+		Exec(ctx, w.tx); err != nil {
+		return err
+	}
+	r.TimeRange.End = end
+	return w.table.NewCreate().Entry(&r).Exec(ctx, w.tx)
```
pjdotson: should 1) use `NewUpdate` like in `Rename`, 2) use some guard to make sure that the end
they are setting is NOT later than that ranges start time.

x/cpp/color/color.h
```diff
@@ -9,6 +9,21 @@
...
 #include "x/cpp/color/json.gen.h"
 #include "x/cpp/color/proto.gen.h"
 #include "x/cpp/color/types.gen.h"
+#include "x/cpp/errors/errors.h"
+
+namespace x::color {
+
+/// @brief from_hex parses a 6- or 8-character hex color, with or without a leading
+/// '#'. Mirrors the Go color.FromHex.
+std::pair<Color, x::errors::Error> from_hex(const std::string &input);
```
pjdotson: `from_hex` isn't used anywhere outside this module, should just remove

x/cpp/color/BUILD.bazel
```diff
@@ -1,13 +1,21 @@
...
     hdrs = [
         "color.h",
         "json.gen.h",
         "proto.gen.h",
         "types.gen.h",
     ],
+    # GCC's libstdc++ <regex> trips a false-positive maybe-uninitialized warning when
+    # compiling from_css; only GCC on Linux is affected.
+    copts = select({
+        "@platforms//os:linux": ["-Wno-error=maybe-uninitialized"],
```
pjdotson, line 16: should just push and pop right between the imports (see `driver/http/mock/server.cpp`)

core/pkg/service/arc/ranges/ranges.go
```diff
@@ -0,0 +1,343 @@
...
+	report taskreporter.Reporter
+}
+
+// ModuleConfig wires the Arc ranges module into a wazero runtime.
+type ModuleConfig struct {
+	Ranger   *ranger.Service
+	Strings  *strings.ProgramState
+	Runtime  wazero.Runtime
+	Reporter taskreporter.Reporter
+}
```
pjdotson, line 131: is there a reason for this not to implement `config.Config`?

core/pkg/service/arc/ranges/ranges.go
```diff
@@ -0,0 +1,343 @@
...
+			}
+			return heap.Create(dispatchEnd(ctx, m.rng, m.report, key))
+		}).Export(endMemberName)
+	if _, err := builder.Instantiate(ctx); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+func (m *module) ModuleName() string { return moduleName }
```
pjdotson: dead code

core/pkg/service/arc/ranges/ranges_test.go
```diff
@@ -0,0 +1,574 @@
...
+
+	"github.com/google/uuid"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/synnaxlabs/arc/ir"
+	"github.com/synnaxlabs/arc/runtime/node"
+	stlstrings "github.com/synnaxlabs/arc/stl/strings"
+	arctest "github.com/synnaxlabs/arc/stl/testutil"
+	"github.com/synnaxlabs/arc/symbol"
+	symboltestutil "github.com/synnaxlabs/arc/symbol/testutil"
```
pjdotson: should dot import `testutil` libraries

driver/arc/ranges/ranges.h
```diff
@@ -0,0 +1,277 @@
...
+#include "arc/cpp/runtime/state/state.h"
+#include "arc/cpp/stl/stl.h"
+#include "arc/cpp/stl/strings/state.h"
+#include "arc/cpp/types/types.h"
+
+namespace driver::arc::ranges {
+
+/// @brief Reporter surfaces a stdlib-originated failure as a task-level status,
+/// mirroring the Go-side taskreporter.Reporter so failures land as warnings.
+using Reporter = std::function<
```
pjdotson: why is this type defined in `ranges.h`, surely this is more general?

driver/arc/ranges/ranges_test.cpp
```diff
@@ -0,0 +1,373 @@
...
+        .tolerance = x::telem::TimeSpan(0),
+        .reason = ::arc::runtime::node::RunReason::TimerTick,
+        .mark_changed = [](size_t) {},
+        .mark_self_changed = [] {},
+        .set_deadline = [](x::telem::TimeSpan) {},
+        .report_error = [](const x::errors::Error &) {},
+    };
+}
+
+::arc::types::Param str_param(const std::string &name, const std::string &value) {
```
pjdotson, line 41: `make_context` and `str_param` seem like they should be in some test library instead of
ranges specific

driver/arc/ranges/ranges_test.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+// reads the string written to a node's first output via a fresh state handle, which
+// shares the same output buffers as the executed node.
+std::string output_key(::arc::runtime::state::State &s) {
+    auto check = ASSERT_NIL_P(s.node("ranges"));
+    return std::get<std::string>((*check.output(0)).at(0));
+}
+
+} // namespace
```
pjdotson: bad comment

core/pkg/service/arc/status/set_test.go
```diff
@@ -171,7 +171,7 @@ var _ = Describe("Module", func() {
 	Describe("Construction", func() {
 		It("Should construct without WASM wiring when rat is nil", func(ctx SpecContext) {
 			Expect(mod).ToNot(BeNil())
-			Expect(mod.ModuleName()).To(Equal("status"))
+			Expect(mod.(node.ModuleNamer).ModuleName()).To(Equal("status"))
```
pjdotson: this `ModuleNamer` assertion is a little odd. I am a little confused where the
`ModuleName` is used. Generally I would like to avoid interface assertions where
possible. Should the `node.Factory` interface also have a `ModuleName()` method?

The problem with using interface assertions is that you have some "magic" in the
codebase that is hard to track and see where things exactly link up with each other.

integration/tests/arc/stl_ranges.py
```diff
@@ -0,0 +1,234 @@
...
+        )
+        self.arc_source = ARC_STL_RANGES_SOURCE.replace(
+            PARENT_KEY_TOKEN, str(self._parent.key)
+        )
+        create_virtual_channel(self.client, DONE, sy.DataType.UINT8)
+        self._before = sy.TimeStamp.now()
+        super().setup()
+
+    def teardown(self) -> None:
+        with self._try_to("delete created ranges"):
```
pjdotson, line 168: shouldn't we fail the integration test if we can't tear down properly?

pjdotson (changes requested):
See comments

driver/arc/ranges/ranges_test.cpp
sy-nico: This matches our convention for anonymous-namespace closers (see `x/cpp/mem/indirect_test.cpp:206`). `FixNamespaceComments` is off in `.clang-format`, so it's intentional, not formatter-generated. Can drop it if you'd rather, but it'd diverge from the sibling tests.

integration/tests/arc/stl_ranges.py
sy-nico: Teardown is best-effort by convention. `_try_to` (`framework/test_case.py`) logs and swallows so a cleanup failure can't abort the rest of teardown. The run() assertions are the pass/fail signal, not cleanup.

x/cpp/color/BUILD.bazel
sy-nico: The `server.cpp` pattern works because `httplib.h` trips the warning in its own code, right at the include line. Here it's different: the false positive comes from instantiating `std::regex` inside `from_css`, so GCC emits it at the `from_css` body, not at `#include <regex>`. Since `#pragma diagnostic ignored` only applies between the push/pop, wrapping the include wouldn't cover where the warning actually fires. `copts` demotes it across the whole TU, which does.

driver/arc/ranges/ranges.h
sy-nico: Moved to a shared `driver::arc::Reporter` in `driver/arc/reporter.h`, used by both `ranges` and `status`.

driver/arc/ranges/ranges_test.cpp
sy-nico: `make_context` is already duplicated across ~10 arc test files (`arc/cpp/runtime/wasm` + all of `arc/cpp/stl/*` + `driver/arc/status`), so hoisting it and `str_param` into `arc/cpp/runtime/testutil` and updating every call site is the right fix, not a ranges-only change. Filing a follow-up to do it across the suite: [SY-4442](https://linear.app/synnax/issue/SY-4442/hoist-shared-arc-node-test-helpers-into-testutil).

core/pkg/service/arc/ranges/ranges.go
sy-nico: These `ModuleConfig`s are built once in `task.go` with no defaults and no second override source, so a full `config.Config` `Override` would be dead boilerplate. Added the part with actual value: `validate.NotNil` on the required deps (`ranger`/`status`, `reporter`, and `strings` when a `Runtime` is provided) inline in `NewModule`. Can go full `config.Config` if you'd rather match the convention exactly, but it's a lot of ceremony.

core/pkg/service/arc/status/set_test.go
sy-nico: Dropped the assertions. `ModuleName` is covered in `node_test.go`.

core/pkg/service/arc/ranges/ranges.go
sy-nico: removed

driver/arc/ranges/ranges_test.cpp
pjdotson: When we've discussed this previously we have wanted to avoid these types of comments in C++ code. I only see 2 existing uses in the whole code base (`driver/arc/status/status_test.cpp` and `x/cpp/mem/indirect_test.cpp`), whereas we have dozens of anonymous namespaces. Should drop and change the `.clang-format` to remove these.

integration/tests/arc/stl_ranges.py
pjdotson: Right, but conceptually failing a cleanup should cause the test case to fail because that means something went wrong, correct?

core/pkg/service/arc/ranges/ranges_test.go
```diff
@@ -0,0 +1,573 @@
...
+	})
+
+	Describe("Construction", func() {
+		It("Should construct without WASM wiring when the runtime is nil", func() {
+			Expect(mod).ToNot(BeNil())
+		})
+
+		It("Should wire host functions when a wazero runtime is provided", func(ctx SpecContext) {
+			rt := wazero.NewRuntimeWithConfig(ctx, wazero.NewRuntimeConfigCompiler())
+			defer func() { Expect(rt.Close(ctx)).To(Succeed()) }()
```
pjdotson: use `DeferClose`.

core/pkg/service/arc/ranges/ranges_test.go
```diff
@@ -0,0 +1,573 @@
...
+				Strings:  stlstrings.NewProgramState(),
+				Runtime:  rt,
+				Reporter: rep.report,
+			}))
+			Expect(wired).ToNot(BeNil())
+		})
+
+		It("Should error when the runtime can't re-instantiate the host module", func(ctx SpecContext) {
+			rt := wazero.NewRuntimeWithConfig(ctx, wazero.NewRuntimeConfigCompiler())
+			defer func() { Expect(rt.Close(ctx)).To(Succeed()) }()
```
pjdotson: use `DeferClose`

core/pkg/service/arc/ranges/ranges_test.go
```diff
@@ -0,0 +1,573 @@
...
+		name := "create_rgb_" + uuid.NewString()
+		n, state := build(ctx, name, "rgb(223, 109, 56)", "")
+		n.Next(nodeCtx(ctx))
+		newKey := telem.UnmarshalSeries[string](*state.Output(0))[0]
+		r := MustSucceed(retrieveRange(ctx, newKey))
+		Expect(*r.Color).To(Equal(MustSucceed(color.FromCSS("rgb(223, 109, 56)"))))
+		Expect(rep.get()).To(BeEmpty())
+	})
+
+	It("Should warn but still create the range when the color is invalid", func(ctx SpecContext) {
```
pjdotson, line 344: why should we warn and create the range when the color is invalid instead of failing?

driver/arc/ranges/ranges.h
```diff
@@ -0,0 +1,272 @@
...
+    const std::shared_ptr<synnax::Synnax> &client,
+    const Reporter &report,
+    const std::string &name,
+    const std::string &color_hex,
+    const std::string &parent
+) {
+    std::optional<x::color::Color> c;
+    if (!color_hex.empty()) {
+        auto [parsed, err] = x::color::from_css(color_hex);
+        if (err)
```
pjdotson, line 48: we report an error but still try to create the range?

driver/arc/ranges/ranges_test.cpp
```diff
@@ -0,0 +1,372 @@
...
+        module.create(::arc::runtime::node::Config(ir, ir.nodes[0], std::move(st)))
+    );
+
+    auto ctx = make_context();
+    ASSERT_NIL(created->next(ctx));
+
+    const auto key = output_key(s);
+    const auto uid = ASSERT_NIL_P(x::uuid::UUID::parse(key));
+    auto [r, err] = client->ranges.retrieve_by_key(uid);
+    ASSERT_NIL(err);
```
pjdotson: should be using test helpers like `ASSERT_NIL_P` throughout all of these tests

pjdotson (changes requested):

core/pkg/service/arc/ranges/ranges_test.go
sy-nico: Done — switched to `DeferCleanup(rt.Close)`. `rt.Close` takes a context so it isn't a plain `io.Closer`, but Ginkgo auto-supplies the context to the cleanup and asserts the returned error.

core/pkg/service/arc/ranges/ranges_test.go
sy-nico: Switched to `DeferCleanup(rt.Close)` (same as above).

core/pkg/service/arc/ranges/ranges_test.go
sy-nico: Used `DeferCleanup` because `rt.Close` takes a context, so it doesn't satisfy `io.Closer` for `DeferClose`.

core/pkg/service/arc/ranges/ranges_test.go
sy-nico: Capturing the range with its start/end times is the important part, so we shouldn't fail on a cosmetic value and lose the range entirely. In a high-pressure setting, operators still want the range captured, and a bad color is easy to fix later (much easier than digging through logs to reconstruct the range after the fact).

Related to this comment: https://github.com/synnaxlabs/synnax/pull/2522#discussion_r3476235117

We validate as much as we can at analysis/compile time, but we don't want to penalize operators for a cosmetic mistake.
image

driver/arc/ranges/ranges.h
sy-nico: It already reports a _warning_, not an error. `err` is just the local name for `from_css`'s result; the emitted status is `VARIANT_WARNING`, and the range is still created without a color. Same reasoning as the (previous thread)[https://github.com/synnaxlabs/synnax/pull/2522#discussion_r3500325016].

(We don't want to lose the range over a cosmetic value, so we warn and continue)

driver/arc/ranges/ranges_test.cpp
sy-nico: Dropped. On `clang-format`, there's no option that removes existing namespace comments.
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

core/pkg/service/arc/service.go
```diff
@@ -129,16 +131,29 @@ type Service struct {
...
+	return s.cfg.Channel.NewArcSymbolResolver(tx)
+}
+
+// NewSymbolResolver is the dynamic resolver attached to a program root's
+// GlobalResolver. It resolves cluster channels by name or numeric key. Static prelude
+// symbols (STL, status, ranges) are attached to the ambient by NewRoot rather than
+// chained behind this resolver.
+func (s *Service) NewSymbolResolver(tx gorp.Tx) arc.SymbolResolver {
+	return s.NewChannelResolver(tx)
+}
```
pjdotson: what's the point of adding these two functions back in? they were deleted because they were only ever used in `NewRoot`

core/pkg/service/arc/ranges/ranges_test.go
pjdotson: But in our other client libraries calling `client.ranges.create` would fail if the color is invalid, right? If this is an automation that someone is writing, then assigning the color is part of the program correctness. This doesn't match how validation works for other things in other parts of the codebase - if something is wrong, we will just fail creation, rather than creating with some parts not what the operator intended.

How far would this logic work - what if an operator tried to set a status variant with an invalid variant string. We pick one for them? We would fail that but pass this, meaning different Arc client APIs have different behaviors, so users would have to dig through documentation to figure out what is the exact behavior for this exact function, instead of having the same modal for all `create` function calls?

And our other client libraries would all fail creating the ranges if someone passed in an invalid color - shouldn't Arc match Go,Python,TypeScript,and C++?

pjdotson (changes requested):

core/pkg/service/arc/service.go
sy-nico: This got added back in when I merged `rc`. The conflict in `service.go` kept my branch's side of `NewRoot`, which still carried these two helpers, instead of taking rc's deletion. Nothing references them outside `NewRoot`. Removing them and inlining the resolver into `NewRoot`.

core/pkg/service/arc/ranges/ranges_test.go
sy-nico: A few things, in order:

The status-variant case isn't equivalent. With a bad variant we'd have to pick a variant for the user, so we'd be inventing data and making a decision on their behalf (something I am staunchly against). With a bad color we aren't choosing anything for them: we just omit a cosmetic value. A range without a color is still that same range with all the values that actually matter (start, end, name, parent). A status with the wrong variant is genuinely wrong data. Those aren't the same failure.

On "shouldn't Arc match Go/Python/TS/C++": in the other client libraries, yes, `client.ranges.create` fails on an invalid color, but the user has a real, feasible error-handling path there. They can catch it, branch, retry, surface it, etc. In an Arc flow there's no equivalent recourse mid-run, so failing the create just drops the range on the floor. The cost of the same mistake is much higher here.

The principle is: don't punish a user for a cosmetic mistake by destroying the thing they actually cared about creating. We validate as much as we can at analysis/compile time and testing precisely so we can flag the issue early. Creating the range anyway on an invalid color is the last slice of swiss cheese: the data is captured, and the bad color is a trivial fix afterward.

sy-nico: The other libraries failing isn't a justification for copying that behavior here. The intent of this function is reliability. Refusing to create a range when its main parameters are valid, over a cosmetic color, is itself a reliability problem. Matching the other clients for consistency's sake would mean importing the less reliable behavior, which is the opposite of what we want this path to do

pjdotson: So we are adding a new modality and difference that users have to learn about (The [range].[color] in [Arc] is different than any other [metadata_type].[field] in [Arc/C++/Go/Python/TypeScript]) just to tackle this case.

Not having a color when a user intended a color is still data. The user was obviously trying to make a range with a color if they pass the color argument, so now we omit the color means we are inventing null data. What if the user intended some semantic meaning for that color? Now we remove that by not having the color. If the color field didn't "matter", why did the user pass the color field into the create call? Why do we decide that the color field doesn't matter but the parent does? If we add labels would they matter? Now we have to go case-by-case through every field on every metadata structure to decide if they "matter" instead of having a single consistent rule across the entire Synnax ecosystem.

And what is the real-life scenario that we are protecting against? A user dynamically creating a color string from some string channel that sometimes doesn't have valid color strings flowing through it? This is what we are introducing the extra complexity and mental model to handle?

Why can't a user retry based off of if the return from `ranges.create` is an empty string instead of a key? Isn't this more of an argument for having some type of actual error-handling in Arc rather than trying to handle all of the error behavior internally inside these functions? Maybe a user would like to fail creation if they get a validation error on the field color, maybe they would like to skip the color field and try re-creating it, but why introduce a new standard and new complexity just for this case?

The other libraries failing is justification for copying the behavior here unless there is a very clear reason why this behavior is warranted. But there isn't, so now we are introducing more complexity across Synnax's ecosystem. And I don't think failing to create when you pass in invalid data makes something less reliable - I would argue that failing to create forces our users to write valid code and fix the actual issue rather than just hiding the issue.

pjdotson (changes requested):

core/pkg/service/arc/ranges/ranges_test.go
sy-nico: Consider the big picture: this is a UX and quality-of-life call, not a match-the-client one.

Picture the scenario you're asking about: an operator in a high-pressure test loses a range because of a malformed color string. The thing that mattered, its start, end, name, and parent, is gone over a cosmetic field. Someone is going to be PISSED, at the engineer who fumbled a color or at us for throwing away their data. That's what we're protecting against.

On "inventing null data": omitting an invalid color invents nothing. A range with no color is complete and correct on every axis that matters; color is display metadata, trivially fixed later. We're recording the absence of a value, which is exactly what the user specified. If you want to talk about actually inventing data, that's the divide-by-0 conversation...

On matching the client: we already diverge, and the principle gets used both ways depending on the desired outcome. `ranges.end` has no client equivalent (the client takes an explicit timestamp; Arc's `end` auto-closes to now). Division cuts the other way and is where inventing data really happens: the Python client floats, so casting `/` to f64 (RFC 0038) matches it and gives correct IEEE results, while preserving the int type fabricates an answer for 0/0, and f64 is the option being resisted. "Be like the client" isn't a rule we hold, it's applied selectively.

On "why can't the user retry off the return value": Arc has no error-handling machinery to branch on a failed create, and doesn't need it here. Write that retry path out as a real Arc program yourself and watch how it plays out. It gets messy fast, and that mess is the argument against it.

This is real, beneficial behavior. Failing the create doesn't force better code, it just hides a trivially fixable mistake behind a lost range. It's a judgment call about what users need in the field, and leaning on "technical correctness" as the tiebreaker, applied as inconsistently as it is, chooses dogma over the person using the tool.

sy-nico:
@greptile, rerun

PR #2533, opened 2026-06-25
SY-4434: Collapse Clamp into Bounds and Make It a Closed Projection

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4434](https://linear.app/synnax/issue/SY-4434)

## Description

Removes the standalone `x/ts` `clamp` module and consolidates clamping onto `bounds.clamp`, while fixing a latent semantic bug in `bounds.clamp` and the bounds constants that had been authored around it.

**Closed clamp.** `bounds.clamp` previously behaved as a half-open *index* clamp: a target `>= upper` returned `upper - 1`. That assumption only holds for integer indices into `[0, length)`; for continuous magnitudes it is wrong (a half-open real interval has no maximum). It now projects onto the closed interval `[lower, upper]`. The half-open contract that keeps contiguous bounds non-overlapping (`[0,1)`, `[1,2)`) is a property of `contains`/`overlapsWith`, which are unchanged — `clamp` answers "nearest valid magnitude", not "which bucket". This fixes off-by-one behavior at the upper bound for every continuous caller (numeric inputs, gauge values, panel/pixel sizes, downsample factors) and an outright regression where dragging a gradient stop to the far right snapped it to the far left (`clamp([0,1), x>=1)` -> `0`).

**De-inflating compensated bounds.** Because the old clamp returned `upper - 1`, several bounds constants had inflated their `upper` by 1 so the clamp would land on the true inclusive max. Under the closed clamp those now clamp one too high, and three feed a stricter schema that would then reject the over-high value. Each is restored to its true inclusive maximum:

| Constant | Was | Now | Why |
| --- | --- | --- | --- |
| `AUTHORITY_BOUNDS` | `256` | `255` | control authority is `uint8` (max 255) |
| `PRECISION_BOUNDS` (log) | `18` | `17` | log precision schema `.max(17)` |
| `RECEIPT_TIMESTAMP_PRECISION_BOUNDS` | `4` | `3` | schema `.max(3)` |
| console `SCALE_BOUNDS` | `1001` | `1000` | matches the pluto definition |
| `STROKE_WIDTH_BOUNDS` | `11` | `10` | matches `LINE_WIDTH_BOUNDS` |
| `DOWNSAMPLE_BOUNDS` | `1001` | `1000` | true max downsample |

Audited every `bounds.clamp` caller and every numeric-input `bounds` prop: all inflated constants are used only as clamp inputs (none in `contains`/partition), so de-inflation is safe. The `series.ts` `SubIterator` index clamp is updated from the compensating `construct(0, length + 1)` to `construct(0, length)`, which is exactly equivalent under the closed clamp (the iterator's `index >= end` guard prevents `at(length)`).

**Partial bounds.** `bounds.construct` now accepts a number-only partial bounds object via a dedicated overload: a missing lower defaults to `-Infinity`, a missing upper to `Infinity`. bigint partials are rejected at the type level. `Resize.Single` and the nav `Drawer` migrate from `minSize`/`maxSize` to a single `sizeBounds` prop.

**Circular dependency.** `math/round.ts` no longer imports `spatial/bounds` (which created a cycle); it depends on the base `Bounds` type and computes span inline.

## Basic Readiness


emilbon99:
@greptile re-review

pjdotson (approved):

emilbon99:
@greptile re-review

PR #2541, opened 2026-06-28
SY-4417: Separate Modal and Standard Layout Placing Functionality

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4417](https://linear.app/synnax/issue/SY-4417)

## Description

Splits the modal system out of the standard layout-placing path and into the layered
architecture, then removes the by-reference modal-opening machinery entirely.

**Modal package split** (`layered/{app,service,session}/modals`): modals are no longer
registered as layout renderers under string layout types. The old `@/modals` and
`@/layout/Modals` are removed; consumers migrate to `@/layered/service/modals`. Calculated
channels and device configuration move from layout-registered renderers to the
prompt/create modal API, and nav drawer items use `Service.Nav.Item`.

**SPEC removal**: deletes the by-reference opener (`SPEC` symbol, `Spec`, `renderer`, and
the `Modals.use()` `Placer`). A modal is now opened only by calling its hook directly.
- Device configure/connect modals drop the `CONFIGURE_MODALS` hook registry. Each
  integration keeps its own modal hook; the shared device components (`Select`, `Provider`,
  `ConfigureMenuItem`, `EditConnectionMenuItem`, `TaskContextMenuItems`) take a plain
  `onConfigure` callback instead of a modal-by-reference, preserving the plugin boundary
  with no circular dependency on the device aggregator. The one runtime-make dispatch
  (device notifications) goes through a new aggregator hook (`make.useConfigure`) that
  calls every integration hook unconditionally and dispatches by make.
- Palette commands and ontology context-menu handlers call their modal hooks directly;
  `placeModal` is removed from `CommandProps` and the ontology `BaseProps` /
  `TreeContextMenuProps`.

**Fixed size removed**: `create`/`prompt` no longer take a `size` option; the modal store
`Entry` and dialog frame drop the size plumbing. Modals size to their content.

Stacked on SY-4435 (deep-linking hook-based API).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR moves modal opening out of layout placement and into the layered modal system. The main changes are:

- Adds a session modal provider around the console layout window.
- Replaces layout-registered modals with hook-based modal openers.
- Migrates channel, cluster, CSV, hardware, project, range, status, user, and version modal call sites.
- Reworks device configuration flows to pass plain configure callbacks.
- Removes fixed modal size options so modal frames size from their content.

This looks safe to merge.

| Filename | Overview |
|----------|----------|
| console/src/Console.tsx | Adds the session modal provider around the main layout window. |
| console/src/layered/session/modals/Context.tsx | Introduces the session-level modal store and provider. |
| console/src/layered/service/modals/factory.ts | Adds hook-based modal creation and prompt helpers. |
| console/src/hardware/device/make.tsx | Adds a hardware-make dispatcher for device configure and connect modals. |

1. `console/src/layered/app/modals/Modal.tsx`, line 56 ([link](https://github.com/synnaxlabs/synnax/blob/d18f76354d219b4b760dcf6bb03d09b7c69ec209/console/src/layered/app/modals/Modal.tsx#L56)) 

   <a href="#">P2</a> **Visibility Callback Ignores State**

   `Dialog.Frame` calls `onVisibleChange` for visibility changes, but this handler dismisses the modal without checking whether the next value is `false`. If the dialog reports `true` when it mounts or becomes visible, the new modal immediately closes and prompt callers receive `null` instead of seeing the dialog.
</details>

<sub>Reviews (2): Last reviewed commit: ["fixed schematic symbol edit styles"](https://github.com/synnaxlabs/synnax/commit/30e3c4c6d47c8e566c4ec25904f79775b962866d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=40406549)</sub>

**Context used:**


emilbon99:
@greptile re-run

console/src/layered/session/modals/Context.tsx
```diff
@@ -0,0 +1,136 @@
...
+  /** Renders the modal's content with its bound params and close callback. */
+  render: () => ReactNode;
+  /** Dismisses the modal, resolving its caller with null. */
+  dismiss: () => void;
+}
+
+type Listener = () => void;
+
+/**
+ * An in-memory, per-window stack of open modals. Created once by {@link Provider} and
```
pjdotson, line 71: should we have a stack of open modals? When should we ever have more than one modal in a given window?

console/src/layered/session/modals/Context.tsx
```diff
@@ -0,0 +1,136 @@
...
+  type ReactNode,
+  useState,
+  useSyncExternalStore,
+} from "react";
+
+/**
+ * The props handed to every modal content component: the modal's typed params spread
+ * alongside a single close callback. Calling close with a result resolves the caller's
+ * promise (for prompts); calling it with no argument dismisses without a result.
+ */
```
pjdotson, line 38: Do / should modals have to interact with Drift multi-window at all - e.g. we would never want to move a modal from one window to another, right?

console/src/layered/service/modals/Header.tsx
```diff
@@ -0,0 +1,63 @@
...
+      location="top"
+      size="6rem"
+      bordered
+      className={CSS(CSS.BE("modal", "header"), className)}
+      {...rest}
+    >
+      <Nav.Bar.Start>
+        <Breadcrumb.Breadcrumb gap="tiny">
+          {icon != null && <Breadcrumb.Segment color={9}>{icon}</Breadcrumb.Segment>}
+          {children.split(".").map((segment) => (
```
pjdotson, line 49: should this stay as children being split via a string? e.g. what if you want the modal to have two icons. Should children just be a general component instead?

console/src/layered/service/modals/useConfirm.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <Triggers.SaveHelpText action={confirm.label ?? DEFAULT_CONFIRM_LABEL} />
+      <Nav.Bar.End x align="center">
+        <CancelButton {...cancel} onClick={() => close(false)} />
+        <ConfirmButton {...confirm} onClick={() => close(true)} />
+      </Nav.Bar.End>
+    </Footer>
+  </Frame>
+);
+
+export const useConfirm = createPrompt<boolean, ConfirmParams>(Confirm);
```
pjdotson, line 103: do things like `Confirm` and `Rename` belong in the `service/modals` folder? It feels like the rest of the `modals` folder is "building blocks for modals" and then these two are "common built-in modal types".

console/src/layered/service/log/Log.tsx
```diff
@@ -26,9 +26,11 @@ const Internal: Layout.Renderer = ({ visible }) => {
   const channelKeys = Base.useSelectChannelKeys();
   const hasChannels = channelKeys.some((k) => !primitive.isZero(k));
 
+  const modals = Session.Modals.useStore("Log");
   const enableTriggers = useCallback(
-    () => Layout.selectActiveMosaicTabKeyAndNotBlurred(store.getState()) === key,
-    [store, key],
+    () =>
+      Layout.selectActiveMosaicTabKeyAndNotBlurred(store.getState(), modals) === key,
```
pjdotson, line 32: seems a little weird to always have to call `Session.Modals.useStore` and then pass it into `Layout.selectActiveMosaicTabKeyAndNotBlurred`, should be some common hook somewhere.

console/src/layout/slice.ts
```diff
@@ -338,7 +338,7 @@ export const { actions, reducer } = createSlice({
     },
     setProject: (state, { payload: { slice } }: PayloadAction<SetProjectPayload>) => {
       // Mosaic.insertTab mutates tabs arrays in place; clone before
-      // reconciling so the helper does not fight frozen nested objects
+      // reconciling s the helper does not fight frozen nested objects
```
pjdotson: typo

console/src/project/purgeExcludedLayouts.ts
```diff
@@ -16,8 +16,7 @@ export const purgeExcludedLayouts = (
 ): Layout.SliceState => {
   const copy = deep.copy(sliceState);
   Object.entries(copy.layouts).forEach(([key, layout]) => {
-    if (layout.excludeFromProject == true || layout.location === "modal")
-      delete copy.layouts[key];
+    if (layout.excludeFromProject == true) delete copy.layouts[key];
```
pjdotson, line 19: what layouts are still getting excluded from projects?

console/src/status/Explorer.tsx
```diff
@@ -42,8 +42,8 @@ const Internal = () => {
   const listProps = Status.useList({
     initialQuery: View.useContext().getInitialView().query,
   });
-  const placeLayout = Layout.usePlacer();
-  const handleCreate = useCallback(() => placeLayout(CREATE_LAYOUT), [placeLayout]);
+  const openCreate = useCreateModal();
+  const handleCreate = useCallback(() => openCreate(), [openCreate]);
```
pjdotson, line 46: a lot of these callbacks are useless now and should be removed

console/src/status/useCreateModal.tsx
```diff
@@ -8,91 +8,76 @@
...
-import { Layout } from "@/layout";
-import { Modals } from "@/modals";
+import { Modals } from "@/layered/service/modals";
 import { Triggers } from "@/triggers";
 
-export type CreateLayoutArgs = Partial<z.infer<typeof Status.formSchema>>;
+export type CreateModalParams = Partial<z.infer<typeof Status.formSchema>>;
 
-export const CREATE_LAYOUT_TYPE = "createStatus";
+export const useCreateModal = Modals.create<CreateModalParams>(
```
pjdotson, line 22: files like these should probably be renamed to `useCreateModal.tsx` if their main export is a hook called `useCreateModal`

console/src/hardware/device/link.spec.ts
```diff
@@ -32,17 +31,18 @@ const createDevice = async (make: string, name: string) => {
...
 describe("Device.useLink", () => {
-  it("should place a configure layout for a device with a known make", async () => {
+  it("should open the configure modal for a device with a known make", async () => {
     const device = await createDevice(NI.Device.MAKE, "cDAQ Chassis");
-    const { handler, store } = renderLinkHook(Device.useLink);
+    const { handler, modals } = renderLinkHook(Device.useLink);
+    expect(modals.isAnyOpen()).toBe(false);
     await handler({ client, key: device.key });
-    expect(Layout.select(store.getState(), device.key)?.name).toBe("cDAQ Chassis");
+    expect(modals.getState()).toHaveLength(1);
```
pjdotson: last test asserting on name was better

console/src/channel/services/palette.tsx
```diff
@@ -9,26 +9,48 @@
...
+    <Palette.CommandListItem
+      {...listProps}
+      name={CREATE_COMMAND_NAME}
+      icon={<PChannel.CreateIcon />}
+      onSelect={handleSelect}
+    />
+  );
+};
+CreateCommand.key = "create_channel";
+CreateCommand.commandName = CREATE_COMMAND_NAME;
```
pjdotson: why do we need to both pass name into `CommandListItem` and assign `CreateCommand.commandName`

console/src/hardware/opc/device/palette.tsx
```diff
@@ -9,16 +9,28 @@
...
-  name: "Connect an OPC UA server",
-  icon: <Icon.Logo.OPC />,
-  layout: CONNECT_LAYOUT,
-  useVisible: () => Access.useCreateGranted(device.TYPE_ONTOLOGY_ID),
-});
+const COMMAND_NAME = "Connect an OPC UA server";
+
+const ConnectServerCommand: Palette.Command = (listProps) => {
+  const connect = useConnectModal();
+  const handleSelect = useCallback(() => connect(), [connect]);
```
pjdotson: should there be some palette command list item factory that takes in a hook that returns a function as a parameter and then just calls that hook for `onSelect`?

pjdotson (changes requested):
See comments. Also bad styling for:

1) the version info modal

Image

2) shrunken windows

Image

console/src/layered/session/modals/Context.tsx
emilbon99: Yes, we should have a stack, and we will have more than one modal in a given window. Think about the parent range selector.

console/src/layered/session/modals/Context.tsx
emilbon99: No, and if we do want to then we put the stat eback in redux.

console/src/layered/service/modals/Header.tsx
emilbon99: I feel like this is out of scope for this refactor.

console/src/layered/service/log/Log.tsx
emilbon99: This is all going to get removed with panels.

console/src/project/purgeExcludedLayouts.ts
emilbon99: I don't know but this is an artifact that will get removed.

console/src/status/Explorer.tsx
emilbon99: Yes, but this callback is not useless. `openCreate` is compatible with the `onclick` signature.

console/src/layered/service/modals/useConfirm.tsx
emilbon99: Maybe? There's no clear alternative place to put them right now, so I'd like to leave them there for now.

emilbon99: I mean I guess I could put them into dedicated `Cofirm.useModal` and `Rename.useModal` packages?

console/src/layered/session/modals/Context.tsx
emilbon99: I think that eventually some of this will be consolidated into pluto. Don't quite know what yet though.

console/src/channel/services/palette.tsx
emilbon99: There's probably a refactor here but I think it should be done separately.

console/src/layered/service/modals/useConfirm.tsx
pjdotson: Don't have a strong opinion of where they go, just wanted to see how this relates to the `layered` structure

console/src/palette/command.tsx
```diff
@@ -139,6 +139,40 @@ export const createSimpleCommand = ({
...
+  sortOrder?: number;
+}
+
+/**
+ * createCommand builds a Command whose onSelect is produced by a hook. On render the
+ * command calls useOnSelect and binds its returned callback to the list item, suiting
+ * commands that open a modal or otherwise act through a hook rather than placing a
+ * layout (see createSimpleCommand for the layout case).
+ */
+export const createCommand = ({
```
pjdotson, line 158: should have some test for this

console/src/layered/session/modals/Context.tsx
pjdotson: I guess we do use multiple modals now, but we should probably consider refactoring in the future to not have more than one modal open at the same time. Everything I've read online recommends against stacking modals and where we have to use them (like with the range creation having multiple windows pop up for the datetime selector, or how the label editing works) I feel like the UX is clunky.

pjdotson (changes requested):
Approved after comments addressed:

A couple of styling issues:

1) Core connect modal has Name to close to top:

Image

2) OPC UA server connect w/ small window

Image

3) Range create w/ small window:

Image

4) Calculated channel w/ small window:

Image

5) Download CSV data when small:

Image

6) When I right-click on a custom schematic symbol and click "Edit", instead of opening up a form for editing, it opens up a form for importing a new SVG.

PR #2569, opened 2026-07-06
SY-4454: Route All Writes Through Cesium

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4454](https://linear.app/synnax/issue/SY-4454/route-all-writes-through-cesium)

## Description

Implements RFC 0044 (included): free channels become **runtime-only virtual channels** hosted in each node's local Cesium, so every write in the cluster — stored, virtual, and free — now flows through the storage engine.

**Cesium**

- Virtual channels are never persisted: `virtual.DB` is a pure in-memory component, the `Virtual` field is excluded from meta serialization, and registrations vanish when the database closes. Legacy directories persisted for virtual channels by previous versions are removed during the boot scan.
- Virtual channels can now be indexed by virtual timestamp index channels. Writers allocate one alignment domain per index group per open from the index channel's DB, stamp every series in the group with the group's current alignment, and advance the sample position only on index writes — reproducing the free-write alignment semantics engine-side. Rekeying a virtual index updates the index references of its dependent channels, and deletes refuse to remove an index that other virtual channels depend on (single and batch delete alike).
- The stream writer validates every frame up front: series must target channels in the writer's channel set and carry structurally valid buffers, and data types are checked against Cesium's own authoritative channel records.
- Opening a streamer validates its initial channel set against the local registry; mid-stream subscription requests remain a pure filter and are drained before each frame is filtered, so a request doubles as a readiness barrier.

**Core**

- The channel service registers virtual channels eagerly: free channels register on the bootstrapper at create time, and every node re-registers the virtual channels it serves writes for — free channels and virtual channels leased to it — by scanning the channel table at boot.
- Free keys route down the gateway write branch, and the relay serves free demands from the gateway tap's storage streamer. The `SendOpenAck` happens-before barrier generalizes: the tapper applies a demand's keys to the gateway tap before the ack fires, and the Cesium streamer applies pending requests to every frame written afterward.
- Deleted end to end: `free.go` and its service-level alignment counters, the `FreeWrites` pipeline through the writer/framer/relay configs, the relay's `freeWriteTap` and its bespoke synchronization, and the distribution writer's entry-node validator — Cesium validates every write on the node that services the channel. The validator's sequence-number duty survives as a thin `sequencer` segment.

Behavioral notes: free writers now appear in control-state reporting (they open Cesium gates), and concurrent free writers at unequal authorities are gated by shared-concurrency control (all writers default to absolute authority, so the common case is unchanged).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR routes free and virtual writes through Cesium. The main changes are:

- Runtime-only virtual channel registration in Cesium.
- Free channel registration through local storage on create and startup.
- Free writes moved onto the gateway writer path.
- Distribution writer validation shifted to storage-side checks.
- Tests and RFC documentation for the new virtual-channel model.

This looks safe to merge.

| Filename | Overview |
|----------|----------|
| core/pkg/distribution/channel/lease_proxy.go | Registers free virtual channels in local Cesium during the bootstrapper create path. |
| core/pkg/distribution/channel/service.go | Re-registers free and locally served virtual channels when the channel service opens. |
| core/pkg/distribution/framer/writer/service.go | Routes free write keys through the gateway writer and replaces the validator segment with sequencing. |
| cesium/channel.go | Creates virtual channels in memory and validates indexed virtual channel relationships. |

<sub>Reviews (3): Last reviewed commit: ["Fix virtual channel batch deletes, index..."](https://github.com/synnaxlabs/synnax/commit/58e43e07d1b6fab3fa10c9acb3d36edaed9e5ada) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42223829)</sub>

**Context used:**


pjdotson:
@greptile /rerun

core/pkg/distribution/framer/writer/service.go
```diff
@@ -344,38 +330,37 @@ func (s *Service) NewStream(ctx context.Context, cfgs ...Config) (StreamWriter,
...
-		hasPeer           = len(batch.Peers) > 0
-		hasGateway        = len(batch.Gateway) > 0
-		hasFree           = len(batch.Free) > 0
 		receiverAddresses []address.Address
-		routeValidatorTo  address.Address
+		routeSequencerTo  address.Address
+	)
+	// Free channels are registered as virtual channels in the local storage engine, so
+	// their writes ride the gateway branch.
+	batch.Gateway = append(batch.Gateway, batch.Free...)
```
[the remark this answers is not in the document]
pjdotson: Ignored for now. Multi-node support is still in beta and not a current priority.

pjdotson:
@greptile /rerun noting comments

emilbon99 (changes requested):
Are you confident that this is really the right architecture? I'm concerned that we're loading responsibilities into Cesium that seem like things it shouldn't be responsible for. What do we gain by moving free channels into cesium? Consolidated validation logic? Would like to talk about this design verbally.

core/pkg/distribution/channel/service.go
```diff
@@ -170,6 +170,23 @@ func OpenService(ctx context.Context, cfgs ...ServiceConfig) (s *Service, err er
 		return nil, err
 	}
 	s.mu.externalNonVirtualSet = set.NewInteger(KeysFromChannels(externalNonVirtualChannels))
+	// Virtual channels are never persisted by the storage layer, so every boot
+	// re-registers the ones this node serves writes for: free channels and virtual
+	// channels leased to this node.
+	var virtualChannels []Channel
```
emilbon99, line 176: If you're going to take this approach, why is the meta file for any channel persisted? You're creating a weird architectural seam between the time series database and the meta-data store. You're also creating a much tighter dependency between cesium and basically the service layer now.

emilbon99:
I would actually argue that moving the other way is better. Make cesium responsible SOLELY for data storage, and move virtual and free virtual channels OUT of cesium. Also move the control logic OUT of cesium. Unfortunately not super realistic because of domain control, but I feel relatively strongly that the design laid out here is not the right one.

docs/tech/rfc/0044-260706-virtual-channels.md
```diff
@@ -0,0 +1,332 @@
...
+else. Instead, a `FreeWrites` pipeline is threaded from the writer service through the
+framer service config into the relay (`relay/relay.go:55`), terminating in a
+`freeWriteTap` (`relay/tap.go:256`) — a permanently-open pseudo-tap with its own
+atomically-published key set. The `SendOpenAck` happens-before guarantee, which for
+storage-backed channels falls out of the demand-acknowledgement flow, requires bespoke
+synchronization for the free tap: the relay must wait for the free tap to install a
+streamer's keys before acking the open. An entire `Ordered` test suite in
+`relay_test.go` exists to pin down that machinery.
+
+## 2.3 - Validation Is Duplicated, and the Duplicate Is the Only Coverage Free Gets
```
emilbon99, line 107: Something critical that hasn't been considered is with multi-node clusters you want to move validation UP the stack, not down. If you can validate as many errors on the gateway as possible, you reduce a lot of the possible synchronization errors resulting from storage layer writers on different nodes erroring out.

emilbon99:
The storage layer is responsible for *storage* and we've decided we wanna move more non-storage related infrastructure down to it.

PR #2572, opened 2026-07-07
SY-4455: Destructure Monolithic Console Ontology Services

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4455](https://linear.app/synnax/issue/SY-4455)

## Description

Removes the monolithic `Ontology.Service` interface from the Console and replaces it
with a tree `Item` registry plus per-feature mosaic-drop hooks.

- Renames `console/src/platform/ontology/` to `platform/tree/` and the `Ontology.*`
  namespace to `Tree.*` throughout the Console.
- Replaces the `Service`/`Services`/`NOOP_SERVICE` interface with `Tree.Item`, an
  FC-with-static-properties (`type`, `hasChildren`, `canDrop`, `haulItems`,
  `ContextMenu`, `visible`) built via `Tree.createItem(...)`.
- Each feature's `ontology.tsx` becomes `tree.tsx`, exporting a `TREE_ITEM` instead of
  registering into a shared service map; `onSelect` becomes an internal
  `useOnSelect` hook driven by tree double-click instead of an exported handler.
- The five mosaic-droppable features (schematic, table, line plot, task, log) each
  export a `useMosaicDrop()` hook, assembled by `app/Mosaic.tsx`'s `useMosaicDrops()`
  instead of dispatching through the old service registry.
- `app/services.tsx` becomes `app/treeItems.tsx` (`TREE_ITEMS: Tree.Items`), consumed
  by `<Tree.Provider items={TREE_ITEMS}>` in `App.tsx`.
- Continues the app-layer refactor on top of this split (further reorganization of
  `console/src/app`).

This branch is stacked on `sy-4457-refactor-console-notifications-to-use-compoinent-composition`
(SY-4457), which is why that branch is the merge base here.

## Basic Readiness


console/src/platform/tree/item.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { type Haul, Icon, List, Text, Tree as Base } from "@synnaxlabs/pluto";
+import { type FC, type ReactElement, useCallback } from "react";
+
+import { type ContextMenu } from "@/platform/tree/types";
+
+// ItemProps are the props the Tree passes to a resource type's Item. The Item owns its
+// own icon and double-click behavior, so neither is threaded through here.
+export interface ItemProps extends Omit<Base.ItemProps<string>, "id" | "resource"> {
+  id: ontology.ID;
+  resource: ontology.Resource;
```
pjdotson, line 19: isn't having the `id` and the `resource` both duplicated logic? since the resource already contains the ID?

console/src/platform/tree/item.tsx
```diff
@@ -0,0 +1,118 @@
...
+// Item renders a single resource row in the ontology tree. The component carries the
+// static, render-independent behavior for its resource type (children, drag, context
+// menu, visibility) as properties, read imperatively by the Tree.
+export interface Item extends FC<ItemProps> {
+  type: ontology.ResourceType;
+  hasChildren: boolean;
+  canDrop: Haul.CanDrop;
+  haulItems: (resource: ontology.Resource) => Haul.Item[];
+  ContextMenu?: ContextMenu;
+  visible?: (resource: ontology.Resource) => boolean;
```
pjdotson, line 41: should `visible` be a method on the item, or should it be determined the by the `Item` function component returning `null`?

console/src/platform/tree/item.tsx
```diff
@@ -0,0 +1,118 @@
...
+  item.canDrop = canDrop;
+  item.haulItems = haulItems;
+  item.ContextMenu = ContextMenu;
+  item.visible = visible;
+  return item;
+};
+
+// DEFAULT_ITEM is the fallback used for resource types with no registered Item. It
+// reproduces the previous NOOP service: an unadorned name row with expandable children
+// and no drag or drop behavior.
```
pjdotson: I don't really like having comments saying things like "the previous NOOP service". If someone was reading this code for the first time they don't know anything about the NOOP service. Should probably update `CLAUDE.md` to mention this too.

console/src/feature/ethercat/device/ContextMenuItems.tsx
pjdotson, line 1: should these files still be named `ontology.tsx`? is probably worth searching for any file names in `console/src` that use `ontology`.

console/src/feature/group/tree.spec.ts
```diff
@@ -8,17 +8,20 @@
...
-import { renderTreeContextMenu } from "@/platform/ontology/menuTestutil";
-import { createResource } from "@/platform/ontology/testutil";
+import { renderTreeContextMenu } from "@/platform/tree/menuTestutil";
+import { createResource } from "@/platform/tree/testutil";
 import { assertDefined, uniqueName } from "@/testutil";
 
 const client = createTestClient();
 
+const item = Group.TREE_ITEMS.group;
+assertDefined(item, "no group tree item");
```
pjdotson: is this `assertDefined` really needed, won't all of the tests fail if this doesn't exist?

console/src/feature/log/tree.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    await waitFor(() => expect(removeLayout).toHaveBeenCalledWith(l.key));
+    const logExists = async (): Promise<boolean> => {
+      try {
+        await client.logs.retrieve({ key: l.key });
+        return true;
+      } catch {
+        return false;
+      }
+    };
+    await waitFor(async () => expect(await logExists()).toBe(false));
```
pjdotson: should just assert an expecting it to be thrown with query error not found

console/src/feature/lineplot/useMosaicDrop.ts
```diff
@@ -0,0 +1,38 @@
...
+            name: linePlot.name,
+            location: "mosaic",
+            tab: { mosaicKey: nodeKey, location },
+          }),
+        );
+      }, "Failed to load line plot");
+    },
+    [client, placeLayout, handleError],
+  );
+};
```
pjdotson, line 38: shouldn't these hooks be using the client from flux instead of the raw `Synnax.use()` client?

console/src/app/notifications/Feed.tsx
```diff
@@ -0,0 +1,13 @@
...
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { NOTIFICATIONS } from "@/app/notifications/notifications";
+import { Notifications as Base } from "@/platform/notifications";
```
pjdotson: no need to alias here

console/src/app/notifications/notifications.ts
```diff
@@ -7,23 +7,17 @@
...
 import { Cluster } from "@/feature/cluster";
 import { Device } from "@/feature/device";
 import { Rack } from "@/feature/rack";
 import { Task } from "@/feature/task";
 import { Version } from "@/feature/version";
-import { Notifications as Base } from "@/platform/notifications";
+import { type Notifications } from "@/platform/notifications";
 
-const NOTIFICATIONS: Base.Notification[] = [
+export const NOTIFICATIONS: Notifications.Notification[] = [
```
pjdotson: why does this and `Feed.tsx` need to be two separate files?

console/src/app/layout/layouts.ts
```diff
@@ -0,0 +1,44 @@
...
+import { LinePlot } from "@/feature/lineplot";
+import { Log } from "@/feature/log";
+import { Range } from "@/feature/range";
+import { Schematic } from "@/feature/schematic";
+import { Status } from "@/feature/status";
+import { Table } from "@/feature/table";
+import { Task } from "@/feature/task";
+import { type Layout } from "@/platform/layout";
+
+export const LAYOUTS: Layout.Renderers = {
```
pjdotson: does `Context.tsx` and `layouts.ts` really need to be two separate files? also goes for pretty much everything else in this app layer.

console/src/platform/command/command.tsx
```diff
@@ -0,0 +1,90 @@
...
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { Flex, type Icon, type List, Text } from "@synnaxlabs/pluto";
+import { type FC, type ReactElement, useCallback } from "react";
+
+import { Palette } from "@/platform/palette";
+import { type Session } from "@/session";
+
+import { Layout } from "../layout";
```
pjdotson: shouldn't use relative path imports

console/src/platform/search/external.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+export * from "@/platform/search/ListItem";
```
pjdotson: don't need single-export `external.ts` files

console/src/platform/palette/List.tsx
```diff
@@ -0,0 +1,135 @@
...
+  onSelect,
+  itemKey,
+  ...rest
+}: Select.ListItemProps<string>) => {
+  const handleClick = useCallback(
+    (e: React.MouseEvent<HTMLDivElement>) => {
+      // Only trigger on the synthetic click, which means we won't accidentally call
+      // `onSelect` twice.
+      if (e.detail === SYNTHETIC_CLICK_DETAIL) onSelect?.(itemKey);
+    },
```
pjdotson, line 59: why would we be calling `onSelect` twice? isn't this indicative of a bug we should fix instead of hacky stuff like this?

console/src/feature/http/commands.spec.ts
```diff
@@ -11,15 +11,15 @@ import { createTestClient } from "@synnaxlabs/client";
...
 import { HTTP } from "@/feature/http";
-import { renderPalette } from "@/platform/palette/testutil";
 import { stubGeometry, waitForPlacedLayout } from "@/testutil";
 
 stubGeometry();
 
 const client = createTestClient();
 
-describe("HTTP palette commands", () => {
+describe("HTTP.Device Commands", () => {
```
pjdotson: Feel like this should be two separate test files, one in `http/device` and another in `http/task`

console/src/feature/log/commands.tsx
```diff
@@ -10,24 +10,17 @@
...
-      icon={<Log.CreateIcon />}
-      onSelect={create}
-    />
-  );
-};
-CreateCommand.key = "create_log";
-CreateCommand.commandName = COMMAND_NAME;
-CreateCommand.useVisible = () => Access.useCreateGranted(log.TYPE_ONTOLOGY_ID);
+const CreateCommand = Command.create({
+  key: "create_log",
```
pjdotson, line 19: is there a reason that `Command.create` can't just make the key a UUID if it is a factory function that is only used once? Somewhat pointless to be manually specifying keys for stuff like this.

console/src/app/toolbars/toolbars.ts
```diff
@@ -20,15 +20,15 @@ import { type Nav } from "@/platform/nav";
...
 
-export const LEFT: Nav.Item[] = [
+export const LEFT: Nav.Toolbar[] = [
   Channel.TOOLBAR,
   Range.TOOLBAR,
   Project.TOOLBAR,
-  ...Task.NAV_DRAWER_ITEMS,
-  ...Device.NAV_DRAWER_ITEMS,
+  ...Task.TOOLBARS,
+  ...Device.TOOLBARS,
```
pjdotson: isn't there only one `TOOLBAR` for both `Task` and `Device`?

console/src/platform/tree/testutil.ts
```diff
@@ -8,37 +8,27 @@
...
-  overrides: Partial<Record<ontology.ResourceType, Ontology.Service>> = {},
-): Ontology.Services =>
-  new Proxy({} as Ontology.Services, {
-    get: (_target, prop) => {
-      if (typeof prop !== "string") return undefined;
-      const type = prop as ontology.ResourceType;
-      return overrides[type] ?? { ...Ontology.NOOP_SERVICE, type };
-    },
-  });
+export const createTreeItems = (overrides: Tree.Items = {}): Tree.Items => overrides;
```
pjdotson: pointless function.

console/src/feature/search/external.ts
```diff
@@ -7,6 +7,5 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-export * from "@/app/App";
-export * from "@/app/links";
-export * from "@/app/nav";
+export * from "@/feature/search/List";
+export * from "@/platform/search/external";
```
pjdotson: we reexport from `platform/search`?

console/src/feature/search/List.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { type Search } from "@/platform/search";
+
+const emptyContent = (
+  <Text.Text status="disabled" center level="h4">
+    <Icon.Resources />
+    No resources found
+  </Text.Text>
+);
+
+export const createList = (items: Search.ListItems) => {
```
pjdotson: why does this need to be a factory function instead of a normal React component?

console/src/feature/project/search.tsx
```diff
@@ -0,0 +1,39 @@
...
+      store.dispatch(
+        Session.Layout.setProject({
+          slice: Session.Layout.migrateLayout(proj.layout),
+        }),
+      );
+    }, `Failed to select ${name}`);
+  };
+};
+
+export const SearchListItem = Search.createListItem({
```
pjdotson: weird to export both `SearchListItem` and `SEARCH_LIST_ITEMS`. Also should probably have `search.spec.tsx` tests for all `search.tsx` files.

pjdotson (changes requested):

console/src/platform/tree/item.tsx
emilbon99: You can't return `null` on virtualized lists as items need to have a consistent height.

console/src/feature/lineplot/useMosaicDrop.ts
emilbon99: These hooks are being heavily refactored in the panels PR. leaving as is for now.

console/src/platform/palette/List.tsx
emilbon99: Yes, but this code existed before, and it worked well. Fixing it requires another deep refactor of the list component. I want to do that at some point, but not now.

console/src/feature/log/commands.tsx
emilbon99: I think that having a stable, human readable key for statically defined items in a set like this is useful and nice. I don't think we should generate UUIDs. Reduces production debuggability for a very minor UX improvmeent.

console/src/app/toolbars/toolbars.ts
emilbon99: I did this to standardize the way we do registry composition for everything in the codebase.

console/src/feature/search/external.ts
emilbon99: I've been thinking about this sort of composition pattern in the console to avoid extra import aliasing. So basically each layer above will re-export the layer below.

It's a thought and it kind of works, but there are also potential future issues with re-exporting things of the same name. I don't know.

console/src/feature/search/List.tsx
emilbon99: The reasoning is that it enforces an invariant i.e. all search items *must* be provided statically instead of dynamically through a prop. It's especially relevant in the cases where you have a static list of hooks i..e the `useVisible` ones.

emilbon99: An alternative is to move the entire component into the app layer.

emilbon99: Changed my mind on this, and instead created a thin wrapper in the `app` layer that passes the static list.

console/src/app/toolbars/toolbars.ts
pjdotson: but this isn't standardized with `Channel.TOOLBAR`, `Range.TOOLBAR`.

pjdotson (changes requested):
approved after comment addressed and CI passes

console/src/app/toolbars/toolbars.ts
emilbon99: :)

PR #2577, opened 2026-07-09
SY-4293: Arc Flow Variables

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4293](https://linear.app/synnax/issue/SY-4293)

## Description

Arc programs can now declare lexically scoped variables and channel aliases inside sequences, stages, and at the top level. A variable names and reuses a value, and a program selects a variable's behavior through how it initializes that value. This change implements scoped variables in the Go runtime; support in the driver runtime follows in a separate change.

### Variable kinds

A variable's kind is inferred from its initializer:

```go
gain := 2 * 3         // constant: a held value, set by assignment
count $= 0            // stateful: persists across entries into its scope
scaled := raw * 2     // reactive: re-derives whenever raw updates
c := chan_long_name   // channel alias: reads and writes pass through to the channel
```

### The register model

Every reactive or reassigned variable lowers to a `variable` node: a register holding what the variable is mapped to; a never-reassigned constant inlines as its value. A constant's register holds its value, an alias's register holds the bound channel key, and a reactive variable's register holds the index of its active derivation, which a companion deref node demuxes. Reassignment in every kind is the same operation, a write to the register, and reads dereference it. Registers reset to their initial value on scope entry; `$=` registers skip the reset. Variables no longer borrow the channels module: the earlier channel-backed implementation, and its key-collision class, is gone. A variable used only inside one compiled unit still folds to direct value storage.

### Intended behavior

These are deliberate semantics. They are the decisions most worth understanding before reading the code.

1. A variable is usable only after its lexical declaration, including across asynchronous stages. This is a readability guarantee: whoever reads a variable has already read its declaration.

```go
threshold := 100
stage watch {
    high := reading > threshold   // ok: threshold is declared above
}
// a variable declared below this point is not visible inside watch
```

2. Constant variables reset to their declared value on each entry into their declaring scope and trigger once on entry. Stateful variables (`$=`) persist their value across entries, which is what makes accumulators and edge detection possible.
3. Sequence steps observe the effects of prior steps: a step that writes a variable settles before the next step runs, so `x = 42.0` followed by `x -> out` writes 42.
4. Reactive variables are read-derived. Writing to a reactive variable is rejected during analysis. A derivation emits on every fresh sample, even when the recomputed value is unchanged; a fresh sample is an event.
5. Reassigning a reactive variable rebinds its derivation and does not trigger downstream readers. A reassignment changes how future values are produced rather than producing a value, so values that predate the rebind never fire; readers respond to values arriving afterward.

```go
scaled := raw * 2
scaled -> output     // output tracks raw * 2

scaled = raw * 3     // rebinds the rule only. output does not fire here.
                     // output updates on the next raw sample, now as raw * 3.
```

6. A channel alias reads and writes through to its underlying channel, and reassigning the alias rebinds it to a different channel, including for reads inside expressions and format strings. Within a stage, flows run concurrently, so a rebind applies to values arriving after it, not to writes fired on the same entry.

### Follow-ups

Loosely tracked, to pick off later:

* [SY-4470: Deduplicate reactive-variable triggers into a single mux node](https://linear.app/synnax/issue/SY-4470/evaluate-only-the-selected-derivation-of-a-reactive-variable)
* [SY-4471: C++ Driver support](<https://linear.app/synnax/issue/SY-4471/arc-variables-c-support>)
* [SY-4472: Explicit variable kind declaration](<https://linear.app/synnax/issue/SY-4472/add-explicit-variable-kind-declaration>)
* [SY-4474: Enable const-expression initializers for variables](<https://linear.app/synnax/issue/SY-4474/enable-const-expressions-for-stateful-variables>)
* [SY-4506: Reuse series buffers in Arc runtime emit paths](https://linear.app/synnax/issue/SY-4506/reuse-series-buffers-in-arc-runtime-emit-paths)
* [SY-4508: Reuse Arc Tabs in user docs](https://linear.app/synnax/issue/SY-4508/reuse-arc-tabs-in-user-docs)

### Testing

Go unit tests cover declaration, scoping, initial values, reset on scope entry, sequence-step ordering across reassignment, reactive re-expression (re-point silence, re-firing on fresh and equal values, re-emission of repeated derivations), channel-alias rebinding for writes, reads, and expression reads across stages, stage re-entry semantics, triggered expression reads, graph-mode function-body parity, and the analysis-time rejections. An integration test (`integration/tests/arc/variables.py`) exercises the full set of kinds end to end against a running cluster, and `integration/tests/arc/stl_ranges.py` is expanded to capture a `ranges.create` result into a scoped variable and use it to parent child ranges, in both func-body and cross-stage flow forms.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR implements lexically scoped variables in the Arc Go runtime (`arc/go`). Variables can be constant (`:=`), stateful (`$=`), reactive (channel-derived), or channel aliases, each with a distinct lowering strategy. Constants fold to WASM literals when never reassigned; reassigned constants lower to a `variable` STL register node that resets to the initial declared value on scope re-entry; reactive variables lower to an `exprRead` deref node with a dispatcher; stateful variables persist their value across scope entries.

- **Register model**: A new `variable`/`stateful_variable` STL pair (`arc/go/stl/variable`) replaces the older channel-backed variable scheme. The register's unedged \"value\" input holds the initial literal, so `Reset()` always copies back the declared initial value correctly.
- **Scheduler settle loop**: `scheduler.Next` now re-walks the reactive graph until no new change lands on an already-visited node, bounded by `len(changedFlags)+1` iterations; sequential scopes append variable nodes as trailing strata members that run every pass.
- **Channel alias rebinding**: `channels.go` `source`/`sink` now accept an optional `channel` reference input that is read at runtime via `boundKey`, enabling `p = valve_cmd` style rebinds at runtime.

This PR is safe to merge. The register model is internally consistent, the settle loop is correctly bounded, and the rearm rule matrix is thoroughly tested.

The core invariants are all correct and covered by tests. The two observations raised are about diagnostic quality and a missing guard, neither of which affects correctness for well-formed Arc programs.

arc/go/runtime/scheduler/scheduler.go (settle-loop silent failure path) and arc/go/compiler/compiler.go (dispatcher last-branch fallthrough) are worth a second look, but neither blocks merging.

| Filename | Overview |
|----------|----------|
| arc/go/stl/variable/variable.go | New STL module implementing register (constant/alias) and exprRead (reactive) node types. Reset correctly restores the literal initial value via the unedged value input's pre-populated aligned data. |
| arc/go/runtime/node/state.go | Adds rearm rules, trigger-edge entries for nodes with no edge-fed inputs, and new API (ConsumeInput, InputFresh, LastChanged, RefSourced, RefInput). Rearm rule assignment for self-write cycles and exprRead nodes is well-reasoned. |
| arc/go/runtime/scheduler/scheduler.go | Adds the settled re-walk loop (up to N+1 passes) to ensure sequential step ordering across reactive variable writes. Convergence failure silently drops pending changes with no diagnostic. |
| arc/go/runtime/scheduler/builder.go | Sequential scopes now append strata variable nodes after steps, keeping step indices stable. resolveTransitions correctly limits to steps only. |
| arc/go/compiler/compiler.go | Adds compileDispatcherSynthetic for reactive variable dispatching (if/else chain keyed on $sel). The dispatcher's last branch runs unconditionally for any out-of-range selector. |
| arc/go/compiler/expression/identifier.go | Variable identifier compilation distinguishes never-reassigned (fold to constant), cross-function channel references, and function-body locals. castAndEmitConst handles all numeric and string kinds. |
| arc/go/analyzer/statement/statement.go | AnalyzeVariableDeclaration classifies variables as reactive/alias/literal via inferVarKind. Stateful variables are rejected at top-level. |
| arc/go/symbol/symbol.go | Adds IsValueVariable/IsChannelReadWrite/IsReactive/IsLiteral helpers, maxProgramLocalChannelKey bound (1<<20), and improved variable ID allocation split between root-counter and local-counter paths. |
| arc/go/stl/channels/channels.go | source and sink now support an optional channel reference input for alias binding. boundKey correctly falls back to the configured key when unbound. |
| x/go/telem/series.go | Adds CopyFrom, which reuses the destination buffer capacity via append(s.Data[:0], ...) to avoid allocation on hot emit paths. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
    subgraph Analysis["Analyzer"]
        VD["VariableDeclaration\n(:= / $=)"]
        IVK["inferVarKind"]
        IVK --> LIT["KindVariable\nDefaultValue = literal"]
        IVK --> REACT["KindVariable\nType = ReadChan"]
        IVK --> STAT["KindStatefulVariable"]
        VD --> IVK
    end
    subgraph Compiler["Compiler"]
        CI["compileIdentifier"]
        CI --> FOLD["castAndEmitConst\nWASM const"]
        CI --> VARLOAD["KindVarRef param\nload node output"]
        DISP["compileDispatcherSynthetic\n$sel if/else chain"]
    end
    subgraph STL["Runtime STL"]
        REG["register\nReset: CopyFrom Input0"]
        EXPR["exprRead\nReset: AbsorbInputs"]
    end
    subgraph Sched["Scheduler"]
        NEXT["Next() settle loop"]
        WALK["walkSequential"]
        SETTLED{"settled?"}
        NEXT --> WALK --> SETTLED
        SETTLED -->|no| NEXT
        SETTLED -->|yes| DONE["clear changedFlags"]
    end
    LIT --> REG
    REACT --> EXPR
    DISP --> EXPR
    REG --> VARLOAD
```

<sub>Reviews (5): Last reviewed commit: ["SY-4293: Fold literal timer reassignment..."](https://github.com/synnaxlabs/synnax/commit/d3c197ac3298d43cfd4e56cff36d11164418c46d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42912633)</sub>


sy-nico:
@greptile, rerun. All concerns addressed. On your concern for ` sink.Next runs output writes even for internal variable channels`:

Not an issue. Output buffers get allocated per declared output param in `ProgramState.New` (`node/state.go:40`), independent of wiring, and `buildChannelWriteNode` always declares the `u8` output. So `Output(0)` is always a live zero-length series that `Resize(1)` just grows, no nil to hit, and the write always emits `1` (a "fired" pulse, not a success flag). Only inputs consult the edge table; outputs are always materialized.

Verified manually, and the reassignment integration tests already cover internal-variable writes with unwired status outputs.

pjdotson:
@sy-nico Emiliano would be better to review this one, removing myself

emilbon99 (changes requested):
First big thing:

`VarKind` is completely unnecessary. Every usage of the enum can be derived from two properties:

1. The symbol kind (and symbol), so just a variable.
2. The data type. The rule is simple, there is already a `ChanDirection` enum that marks the channel direction on a symbol. So you have:

1. A variable symbol that has a data type of channel (with a direction) replaces both of the channel variable kinds.
2. Any other variable symbol that is not a channel data type is by definition a VarKindLiteral.

Completely remove `VarKind`, or call me if you disagree, and then I will give a second pass.

sy-nico:
@emilbon99 ,

Removed `VarKind`. Classification now derives from symbol kind + data type. Writable channel type = read/write alias, read-only = reactive, non-channel = literal. `ChanDirection` carries the distinction.

schemas/arc/ir.oracle
```diff
@@ -183,21 +186,27 @@ Functions Function[] {
...
     }
-    outputs  types.Params   {
+    outputs                types.Params   {
         @doc value "contains output parameter type signatures."
     }
-    channels types.Channels {
+    channels               types.Channels {
         @doc value "contains channel read/write mappings."
     }
+    is_literal             bool           {
```
emilbon99: Double check if this field is actually used.

sy-nico:
Still need to ensure global top-level variables are immutable

arc/go/stl/variable/variable.go
emilbon99, line 1: This file needs a dedicated unit test suite.

arc/go/runtime/scheduler/scheduler.go
emilbon99, line 1: All of these scheduler changes should get high quality unit tests.

arc/go/runtime/node/state.go
```diff
@@ -216,6 +293,117 @@ func (n *State) RefreshInputs() (recalculate bool) {
...
+// false when input i is a reference or has no new data.
+func (n *State) ConsumeInput(i int) (telem.Series, bool) {
+	if i < 0 || i >= len(n.ir.inputs) || n.isReference[i] {
+		return telem.Series{}, false
+	}
+	src := n.inputSources[i]
+	if src == nil || src.data.Len() == 0 {
+		return telem.Series{}, false
+	}
+	ts := telem.TimeStamp(0)
```
emilbon99: Nit: could just do `var ts telem.TimeStamp`

arc/go/runtime/node/state.go
```diff
@@ -216,6 +293,117 @@ func (n *State) RefreshInputs() (recalculate bool) {
...
+// absorbInput marks input i consumed at its current source timestamp.
+func (n *State) absorbInput(i int) {
+	if n.isReference[i] {
+		return
+	}
+	src := n.inputSources[i]
+	if src == nil {
+		return
+	}
+	ts := telem.TimeStamp(0)
```
emilbon99: Nit: could just do `var ts telem.TimeStamp`

arc/go/runtime/node/state.go
emilbon99, line 1: Every public method in tis file needs dedicated unit tests.

arc/go/symbol/symbol.go
```diff
@@ -152,6 +149,33 @@ const (
...
 )
 
+// IsValueVariable reports whether s is a `:=`/`$=` value binding.
+func (s *Symbol) IsValueVariable() bool {
+	return s.Kind == KindVariable || s.Kind == KindStatefulVariable
+}
+
+// IsChannelReadWrite reports whether s is a value variable aliasing a channel.
+func (s *Symbol) IsChannelReadWrite() bool {
+	return s.IsValueVariable() && s.Type.Kind == types.KindChan && s.Type.ChanDirection.IsWrite()
```
emilbon99, line 159: Is a write channel automatically `ReadWrite`? I don't think that I set up the runtime that way. What I mean is that if a channel is marked as `Write` it doesn't get added to the list of channels opened by the streamer.

arc/go/symbol/symbol.go
emilbon99, line 1: Make sure all of the public methods in this file have proper tests.

pluto/src/arc/queries.ts
```diff
@@ -286,7 +286,11 @@ export const { useUpdate: useDelete } = Flux.createUpdate<
   },
 });
 
-export const formSchema = arc.arcZ
+export interface FormValues extends Omit<arc.Arc, "key"> {
```
emilbon99: Use `optional.Optional<arc.Arc, "key">`

pluto/src/arc/language.ts
```diff
@@ -116,6 +116,16 @@ const TOKEN_CONFIG = {
     light: "#292929",
     scopes: ["variable.other.arc"],
   },
+  channelAlias: {
```
emilbon99: Would make sure that having this be a different color is strongly justified. Ending up with a 'rainbow syntax' just to identify minor differences means that the things that should really be differentiated don't, and it ends up being hard to follow what's important and what's not. We already use too many colors in our arc programs IMO, so be disciplined.

arc/go/lsp/hover.go
```diff
@@ -487,7 +499,7 @@ func (s *Server) getUserSymbolHover(
 		}
 	case symbol.KindVariable:
 		d = doc.New(doc.TitleWithKind(displayName, "Variable"))
-		d.Add(doc.Detail("Type", sym.Type.String(), true))
```
emilbon99, line 490: Is there a strong reason why sym.Type.String() doesn't add these details itself?

arc/go/parser/ast.go
emilbon99, line 1: Needs dedicated unit tests.

arc/go/runtime/node/state.go
```diff
@@ -216,6 +293,117 @@ func (n *State) RefreshInputs() (recalculate bool) {
...
+	best, bestTS, found := -1, telem.TimeStamp(0), false
+	for i := range n.ir.inputs {
+		if n.isReference[i] {
+			continue
+		}
+		src := n.inputSources[i]
+		if src == nil || src.data.Len() == 0 {
+			continue
+		}
+		ts := telem.TimeStamp(0)
```
emilbon99: Same here.

arc/go/analyzer/flow/expression.go
```diff
@@ -97,3 +104,76 @@ func AnalyzeSingleExpression(ctx acontext.Context[parser.IExpressionContext]) {
 	}
 	expression.Analyze(ctx.WithScope(blockScope))
 }
+
+// LiftFmtStrVarReads lifts reassigned-variable reads in a format string's
+// placeholders into input params. No-op for non-format literals.
+func LiftFmtStrVarReads[T antlr.ParserRuleContext](
```
emilbon99, line 114: These need dedicated unit tests.

emilbon99 (changes requested):
You need to make sure that all public functions and methods of modified packages have dedicated unit tests.

emilbon99:
The test suite also doesn't compile and run

arc/go/lsp/hover.go
sy-nico: The direction is on the type, so it could. But `Type.String()` renders every compile diagnostic ("cannot cast %s to %s", etc.) and signature hovers, and directioned chan types flow through all of them. Direction would show up in every chan error message, and `chan read/write f32` isn't Arc syntax. So I kept it as hover presentation.

arc/go/symbol/symbol.go
sy-nico: For variables it's by construction: an alias declaration stamps the type `Read|Write`, so a `Write`-only variable can't exist. The flag is analyzer-side only; what gets opened comes from the lowered nodes. Alias reads lower to an `on` node whose key lands in `Channels.Read`, writes to a `write` node in `Channels.Write`, with the alias resolved to its real channel key. Rebind candidates are covered too: each `c1 = other_chan` records the candidate on the alias, and `bind{Read,Write}ToAlias` copies them onto the nodes, so every channel the alias can point at is opened up front.

pluto/src/arc/language.ts
sy-nico: I changed the color to the same blue as channels. It removes an accent color and it standardizes to anything that can be used as a node in a flow.

sy-nico:
@greptile, rerun.

arc/go/lsp/hover.go
emilbon99: Copy

arc/go/symbol/symbol.go
emilbon99: Ok, just to confirm: we aren't needlessly opening a channel for writes when it never gets written to? And vice versa we never open a channel for streaming when it never gets read from?

arc/go/analyzer/statement/statement.go
```diff
@@ -63,6 +65,115 @@ func Analyze(ctx context.Context[parser.IStatementContext]) {
...
+	if err != nil {
+		return
+	}
+	if stateful {
+		if statefulRHSTracksChannel(ctx, expr) {
+			ctx.Diagnostics.Add(diagnostics.Errorf(ctx.AST,
+				"channels and channel-read expressions cannot be assigned to stateful variables"))
+			return
+		}
+		// SY-4474
```
emilbon99: Nit: If possible, should just inline any justified explanation here, makes it better to read for people who are not part of the Synnax linear.

arc/go/analyzer/analyzer_test.go
emilbon99, line 1: Have you made it so that top level variables remain constants? Is that something we are going to do?

arc/go/compiler/expression/identifier.go
```diff
@@ -65,6 +85,107 @@ func compileIdentifier[ASTNode antlr.ParserRuleContext](
...
+// sameFunction reports whether a and b compile into the same unit (same enclosing function, or both module-level).
+func sameFunction(a, b *symbol.Symbol) bool {
+	af, _ := a.ClosestAncestorOfKind(symbol.KindFunction)
+	bf, _ := b.ClosestAncestorOfKind(symbol.KindFunction)
+	return af == bf
+}
+
+// emitSeedConst compiles a read of a never-reassigned seeded variable as its
+// compile-time seed value.
+func emitSeedConst[ASTNode antlr.ParserRuleContext](
```
emilbon99: `seed` is a bad name for this function. if my suspicion is correct, claude chose this name and it's bad. I don't know why claude really loves using `seed` for stuff.

Anyway, this should really be `castAndEmitConst` or something. I added a ban for the word `seed` across the entire repo to `CLAUDE.md` because its a bad function name. 

Also, are you sure this casting logic isn't duplciated anywhere?

arc/go/ir/pb/ir.proto
```diff
@@ -101,6 +101,9 @@ message Scope {
   // transitions contains state-transition rules for sequential scopes. Empty for
   // parallel scopes.
   repeated Transition transitions = 7;
+  // reset_nodes contains keys of variable nodes re-seeded each time this scope
```
emilbon99: What does it means to `re-seed` something, I don't like the word seed.  Is it just resetting it to inits initial state? Then just say `reset to initial state`

arc/go/stl/channels/channels.go
```diff
@@ -211,9 +264,10 @@ func (s *source) Next(ctx node.Context) {
 
 type sink struct {
 	*node.State
-	state    *ProgramState
-	key      uint32
-	inputIdx int
+	state     *ProgramState
+	key       uint32
+	inputIdx  int
+	targetIdx int
```
emilbon99: What does `targetIdx` mean? (a) what does a write channels output mean? is it the truthy output? and (b) why do we need to know a specific `targetIdx`?

arc/go/stl/variable/variable.go
```diff
@@ -0,0 +1,142 @@
...
+		return &exprRead{State: cfg.State, selIdx: selIdx}, nil
+	}
+	return &register{
+		State:    cfg.State,
+		stateful: cfg.Node.Type == statefulSymbolName,
+	}, nil
+}
+
+// register holds what its variable is mapped to: a value, a channel key, or a
+// derivation index. Writes are last-wins; the unedged f0 carries the seed.
```
emilbon99: Again, what is a `seed` and what is an `unedged f0`. Very confusing to understand. If the `seed` is an initial value, just call it initial value.

arc/go/stl/variable/variable.go
```diff
@@ -0,0 +1,142 @@
...
+	*v.OutputTime(0) = telem.NewSeriesV[telem.TimeStamp](v.clock.Now())
+}
+
+func (v *register) Next(ctx node.Context) {
+	data, ok := v.LastChanged()
+	if !ok {
+		return
+	}
+	// Feeders reuse their output buffers in place; the register value must not alias them.
+	*v.Output(0) = data.DeepCopy()
```
emilbon99: Woooooooaaaaaaaaah really really really sus. Why do we need to deep copy this?

arc/go/stl/variable/variable.go
```diff
@@ -0,0 +1,142 @@
...
+		data, ok := v.ConsumeInput(i)
+		if !ok || i != v.active || repointed {
+			continue
+		}
+		// A derivation only changes when its value does; recomputes are not events.
+		if data.DataType == v.Output(0).DataType &&
+			bytes.Equal(data.Data, v.Output(0).Data) {
+			continue
+		}
+		*v.Output(0) = data.DeepCopy()
```
emilbon99: Why deep copy?

arc/go/stl/wasm/node.go
```diff
@@ -104,10 +105,22 @@ func (n *nodeImpl) Next(ctx node.Context) {
 		return
 	}
 
+	// A channel-reference param carries its bound key, resolved fresh each pass.
+	for i := range n.ir.Inputs {
```
emilbon99, line 132: Why is this here? And why does it need custom machinery?

arc/go/stl/variable/variable.go
emilbon99: A nodes outputs should be considered immutable by consumers, and, as a result, a deep copy should not be required.

arc/go/stratifier/stratifier.go
```diff
@@ -35,7 +35,21 @@ func Stratify(
 	if prog == nil || prog.Root.IsZero() {
 		return diag
 	}
-	return stratifyScope(&prog.Root, prog.Edges, diag)
+	// A variable holds its value across passes, so its edges impose no
+	// ordering, mirroring channel state.
```
emilbon99: Why is it important that we do this?

arc/go/symbol/resolver.go
```diff
@@ -28,3 +28,21 @@ type Resolver interface {
 	// and "did you mean" suggestions.
 	Search(ctx context.Context, term string) ([]*Symbol, error)
 }
+
+// scopeResolver adapts a *Symbol to the Resolver interface, delegating
+// lookups to that scope's lexical resolution.
+type scopeResolver struct{ scope *Symbol }
```
emilbon99, line 34: Why is this necessary? Looks like `Symbol` already implements this interface to me.

arc/go/symbol/scope_test.go
```diff
@@ -798,3 +853,32 @@ var _ = Describe("Scope", func() {
...
+	// A conflict message names the kind of the symbol already holding the name,
+	// exercising nounForKind for each kind.
+	DescribeTable("Should name the colliding symbol's kind",
+		func(bCtx SpecContext, kind symbol.Kind, noun string) {
+			ast := MustSucceed(parser.ParseStatement("x := 1"))
+			root := symbol.NewRoot(nil, nil)
+			MustSucceed(root.Add(bCtx, symbol.Symbol{
+				Name: "dup", Kind: kind, Type: types.I32(), AST: ast,
+			}))
+			_, err := root.Add(bCtx, symbol.Symbol{
```
emilbon99: Not following the assertion standard. Use `.Error().To(MatchError())` and do a pass through the rest of your codebase for similar violations.

arc/go/stl/wasm/node.go
```diff
@@ -147,7 +160,7 @@ func (n *nodeImpl) Next(ctx node.Context) {
 	var alignmentSum telem.Alignment
 	var timeRange telem.TimeRange
 	for i := range n.ir.Inputs {
-		if n.ir.Inputs[i].Value != nil {
+		if n.ir.Inputs[i].Value != nil || n.refInputs[i] {
```
emilbon99: If it's absolutely necessary to have this, will a `refInput` ever be something that's not a channel? if not, then should just call this `isChanInputs[i]` or something.

arc/go/symbol/symbol.go
```diff
@@ -152,6 +149,33 @@ const (
...
+}
+
+// IsLiteral reports whether s is a value variable holding a literal, not a channel.
+func (s *Symbol) IsLiteral() bool {
+	return s.IsValueVariable() && s.Type.Kind != types.KindChan
+}
+
+// BacksInternalChannel reports whether s is a value variable backed by a
+// program-local channel (reactive or literal) rather than an external channel.
+func (s *Symbol) BacksInternalChannel() bool {
```
emilbon99: This is never used. in production You need to go back and do an audit of code you introduced that is now dead.

arc/go/symbol/symbol.go
```diff
@@ -361,17 +394,53 @@ func (s *Symbol) Add(ctx context.Context, sym Symbol) (*Symbol, error) {
...
+	case KindInput, KindOutput, KindLoopVariable:
 		child.ID = s.addIndex()
 	}
 	s.children = append(s.children, child)
 	return child, nil
 }
 
+// nounForKind returns a human-readable noun for k, used in diagnostics so a
+// name conflict names what it collides with (variable, channel, function, ...).
+func nounForKind(k Kind) string {
```
emilbon99: Should just make this a method on `Kind` if absolutely necessary. Other thing you could consider doing is just deleting the generated stringer for `Kind` and handwriting it yourself to do this. Probably better.

docs/site/src/components/arc/external.ts
emilbon99, line 1: This folder should be called `arc` and the package should be called `Arc`

docs/site/src/components/arc-context/ArcContext.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export type ArcContext = "Flow" | "Function";
+
+export interface Info {
+  key: ArcContext;
+  name: string;
+  icon: Icon.ReactElement;
+}
+
+export const TABS: Info[] = [
```
emilbon99: Instead of defining `TABS` here, instead make a union type called `Arc.ExecutionContext = "flow" | "function" in here, then move the `TABS` definition into the tabs file. Also, this current implementation will not work with the new tabs API introduced in https://github.com/synnaxlabs/synnax/pull/2584

emilbon99 (changes requested):
Fundamental questions:

1) Why do variable nodes have multiple inputs? `f0, f1, f2, etc.`. If you just have one input last write wins falls out automatically and everything simplifies dramatically.
2) Take this example:

```
  scaled i64 := raw * 2   // scaled is DERIVED from raw. the rule is "raw * 2"
  scaled = raw * 3          // swap the rule to "raw * 3"
  scaled -> out              // whenever scaled produces a value, write it to out
```

For the vast majority of the program lifecycle, `raw * 2` will never be needed, yet it will be re-evaluated for the entire program life?

Seems to me like the design of current var nodes is made to fit a square peg into a round hole. We didn't want to make changes to the scheduler, so instead we coerced this weird set of subgraphs, multiple inputs, and selector nodes to make something sub-optimal work with the current scheduler. Am I off the mark?

arc/go/analyzer/analyzer_test.go
sy-nico: Yes, in commit [72dd59e](https://github.com/synnaxlabs/synnax/pull/2577/commits/72dd59e7c62bf4a7a3bc1df8117bfcc3f8d06113)

arc/go/analyzer/statement/statement.go
sy-nico: I agree. This was done per Patrick's preferences.

arc/go/stl/variable/variable.go
sy-nico: Resolved e3625a0

Created follow up ticket for existing offenders: [SY-4506](https://linear.app/synnax/issue/SY-4506/reuse-series-buffers-in-arc-runtime-emit-paths)

arc/go/stl/variable/variable.go
sy-nico: Resolving here, but keeping [the sibling thread](https://github.com/synnaxlabs/synnax/pull/2577/changes#discussion_r3622616513) open

arc/go/stl/channels/channels.go
sy-nico: Updated to `channelIdx`: the channel ref input's index; -1 when not alias-bound

arc/go/stl/wasm/node.go
sy-nico: A chan-typed WASM param carries a channel key for the body's `read_*`/`write_*` host calls. A literal key bakes in at Create; an alias-bound one arrives from the bind register over an edge, so the node re-reads the latest key each pass, the WASM twin of `boundKey` in stl/channels. It sits outside the data path because a binding is not a sample stream: a rebind must not trigger execution or zip with samples.

Updated inline comment.

arc/go/stratifier/stratifier.go
sy-nico: Variables hold state like channels do: reads see the held value with no same-pass ordering, and write-read feedback must not stratify as a cycle.

Updated comment.

arc/go/symbol/resolver.go
sy-nico: `Symbol.Resolve` takes variadic options (`Resolve(ctx, name, ...ResolveOption)`), so `*Symbol` doesn't satisfy the interface's exact `Resolve(ctx, name)` signature. Widening the interface instead would push compiler-only options onto every production resolver (Core channels). Updated the inline comment to say why.

arc/go/symbol/scope_test.go
sy-nico: Fixed mine (the only one this PR introduces). I was following the existing pattern in pre-existing tests across parser, stl/time, stl/selector, stl/channels, compiler/expression, and analyzer/constraints.

Follow up: [SY-4507](https://linear.app/synnax/issue/SY-4507/enforce-inline-error-assertions-in-arc-go-tests)

arc/go/stl/wasm/node.go
sy-nico: renamed

arc/go/symbol/symbol.go
sy-nico: Made it a method (`Kind.noun()`). Kept the stringer: `noun()` is many-to-one (stateful/loop → "variable", alias → "import"), so folding it into `String()` would collapse distinct kinds in debug output.

sy-nico:
Responding to [this comment](https://github.com/synnaxlabs/synnax/pull/2577/changes#pullrequestreview-4745069390),

A variable's derivations now fold into one compiler-synthesized dispatcher function. An if/else chain on the bind index calls exactly one branch per evaluation, so after the swap `raw * 2` stops executing. The deref reads a single value input; `f0, f1, f2` are gone from the value path. The bind register keeps one param per write site because the runtime binds one edge per param; last write wins across them.

No scheduler changes were needed. The old design put selection in the graph because the scheduler can't selectively run nodes. With selection in the compiled body, the dispatcher is an ordinary run-on-input-change node and an untaken branch costs nothing. A scheduler primitive for this (node enable/disable, runtime edge re-pointing) would still need every derivation compiled plus the rebind bookkeeping, just moved into the scheduler.

arc/go/runtime/node/ref.go
```diff
@@ -0,0 +1,61 @@
...
+// included in the file licenses/APL.txt.
+
+package node
+
+import "github.com/synnaxlabs/x/errors"
+
+// StringRef reads a string input that may be var-bound, falling back to its
+// configured value when it is not. Value is allocation-free until the referenced
+// variable actually changes.
+type StringRef struct {
```
emilbon99: I'm pretty sure that I already solved this problem with defaults on WASM input. Instead of this ref thing, what you do is just connect a fake input to it, write the string to a series, and then everything behaves as normal i..e you don't need any of this ref stuff. Sure its' an allocation on initialization, but that doesn't really matter because the node only gets initialized on program start. Then this entire `Ref` machinery disappears.

emilbon99:
> Responding to [this comment](https://github.com/synnaxlabs/synnax/pull/2577/changes#pullrequestreview-4745069390),
> 
> A variable's derivations now fold into one compiler-synthesized dispatcher function. An if/else chain on the bind index calls exactly one branch per evaluation, so after the swap `raw * 2` stops executing. The deref reads a single value input; `f0, f1, f2` are gone from the value path. The bind register keeps one param per write site because the runtime binds one edge per param; last write wins across them.
> 
> No scheduler changes were needed. The old design put selection in the graph because the scheduler can't selectively run nodes. With selection in the compiled body, the dispatcher is an ordinary run-on-input-change node and an untaken branch costs nothing. A scheduler primitive for this (node enable/disable, runtime edge re-pointing) would still need every derivation compiled plus the rebind bookkeeping, just moved into the scheduler.

As discussed verbally, I think this is the wrong design. This probably deserves scheduler changes instead of coercing the data flow graph around the scheduler. If you really feel like getting this PR in I will approve the design as is, but the scheduler should _absolutely_ be refactored in the future.

x/go/telem/series_factory.go
```diff
@@ -78,6 +78,18 @@ func NewSeries[T Sample](data []T) Series {
 // NewSeriesV is a variadic version of NewSeries.
 func NewSeriesV[T Sample](data ...T) Series { return NewSeries(data) }
 
+// SetSeriesV overwrites s with the given samples, reusing s's buffer capacity to
+// avoid allocation. Emit hot paths should prefer this over NewSeriesV (SY-4506).
+func SetSeriesV[T FixedSample](s *Series, values ...T) {
```
emilbon99: None of the call sites actually use the variadic form of this function, so this doesn't seem very useful. It's also not clear from the function name that this function overrides all of the original properties of the series. 

If the only place this is called is in `arc/go/stl/variable/variable.go`, this function can be completely deleted. The output time series is already guaranteed to have the proper data type and time range, and you can just use the normal value setter.

core/pkg/service/arc/status/status.go
```diff
@@ -62,8 +62,8 @@ var moduleDoc = doc.New(
 // analysis never mutates a shared symbol. Empty defaults keep inputs optional.
 func newSetSymbolType() types.Type {
 	params := types.Params{
-		{Name: "key_or_name", Type: types.String(), Value: ""},
-		{Name: "message", Type: types.String(), Value: ""},
+		{Name: "key_or_name", Type: types.String(), Value: "", AcceptsVar: true},
```
emilbon99: What is the point of `AcceptsVar`. What does this do? Why do we need to accept or deny a variable, and why can't variant `AcceptVar`?

pluto/src/arc/language.ts
```diff
@@ -116,6 +116,16 @@ const TOKEN_CONFIG = {
...
+    dark: "#61AFEF",
+    light: "#0070C1",
+    scopes: [],
+  },
+  reactiveVariable: {
```
emilbon99: If `channelAlias` and `reactiveVariable` have the same color, why do they need to exist as unique LSP language tokens? I don't think that `channelAlias` and `reactiveVariable` should be separate LSP tokens.

arc/go/stl/wasm/wasm.go
```diff
@@ -79,15 +84,21 @@ func (w *Module) Create(_ context.Context, cfg node.Config) (node.Node, error) {
...
+		chanInputs[i] = inp.Type.Kind == types.KindChan
 	}
 
 	stringOutputs := make([]bool, len(irFn.Outputs))
 	for i, out := range irFn.Outputs {
 		stringOutputs[i] = out.Type.Kind == types.KindString
 	}
 
+	selIdx := -1
+	if idx, err := cfg.State.ResolveInput("$sel"); err == nil {
```
emilbon99, line 99: I will approve with the implementation as is, but all of this feels like very hacky workaround code.

arc/go/stl/variable/variable.go
```diff
@@ -0,0 +1,139 @@
...
+	mk := func(name string) *symbol.Symbol {
+		typeVar := types.Variable("T", nil)
+		return &symbol.Symbol{
+			Name:     name,
+			Kind:     symbol.KindFunction,
+			Exec:     symbol.ExecFlow,
+			Internal: true,
+			Type: types.Function(types.FunctionProperties{
+				Outputs: types.Params{{Name: ir.DefaultOutputParam, Type: typeVar}},
+				Inputs:  types.Params{{Name: "value", Type: typeVar}},
```
emilbon99, line 39: Why isn't this using `ir.DefaultInputParam`?

arc/go/stl/variable/variable.go
```diff
@@ -0,0 +1,139 @@
...
+		if _, ok := v.ConsumeInput(v.selIdx); ok {
+			repointed = true
+		}
+	}
+	data, ok := v.ConsumeInput(0)
+	if !ok || repointed {
+		return
+	}
+	// A derivation only changes when its value does; recomputes are not events.
+	if data.DataType == v.Output(0).DataType &&
```
emilbon99: This is dangerous, and doesn't align with previous node implementations. Take a an `on` node that emits a series with the values [1,1,1,1,1] that come from a DAQ and then another series with [1,1,1,1,1] that come from the DAQ at a later time. Although these have the same value, they are distinctly different data acquisition events. Should the `on` node refire or not? 

The problem is not that you chose this model, but the fact that I'm not aware of a consistent design decision. Does a node refire when it's output value emits a new event (model I designed) vs it's output value distinctly changes i.e. equality is false (this model). Which one are we choosing and why?

emilbon99 (changes requested):

arc/go/stl/variable/variable.go
sy-nico: We could (variable nodes are branch-new, no compat issue), but we shouldn't. This mirrors the constant builtin's `value` param, which is released surface we shouldn't rename, and variable lowering emits constant nodes wired to that same param. Renaming one side leaves two spellings in the same lowering path. The param also carries the initial-value payload, not a flow-chain input.

arc/go/stl/wasm/wasm.go
sy-nico: Not sure why it feels hacky. I'm working strictly within the confines of the wasm and scheduler frameworks. Is your issue with `$sel`? No matter the architecture, there needs to be some form of demux over the compiled reassignments. `$sel` is an ordinary input resolved through the standard path; the `$` prefix just keeps it out of the user param namespace.

sy-nico:
@emilbon99 ,

> As discussed verbally, I think this is the wrong design. This probably deserves scheduler changes instead of coercing the data flow graph around the scheduler. If you really feel like getting this PR in I will approve the design as is, but the scheduler should absolutely be refactored in the future.

I disagree. The hard constraint was not to make hacky scheduler changes. I've explored multiple different alternatives and none are strictly better. If it feels like there is a better alternative, then please create the followup ticket (and link to this one) with the concrete methodology you want to see. Here's what I've explored and why I ruled each out:

### 1. Enable/disable node sets in the scheduler
Each derivation stays its own subgraph; a rebind flips the old set off and the new set on. The compiler still has to emit the bind-to-nodes table, so the demux still exists, now as mutable scheduler state instead of static code. On top of that: shared upstream nodes need refcounting so a disable doesn't kill a subgraph another consumer still uses, and every reactivated node must absorb the values that piled up while it was off, or it fires stale. That reintroduces at runtime the exact race class the current design eliminates statically. Same demux, more moving parts, new race surface, zero user-visible gain.

### 2. Runtime edge re-pointing
The register keeps one input edge and rebind mutates its source. This is only half a design: it changes where the register reads, not which nodes run. The scheduler fires any node with fresh inputs, so unselected derivations keep executing (the exact bug this PR fixes). Blocking them requires an enable or gate mechanism, which turns this into option 1 or 4. On its own it also mutates bindings that the state and gating tables assume are fixed at configure time.

### 3. Derivations as anonymous sub-scopes
Rebind becomes a scope switch, reusing stage activation machinery. Conceptually clean, but scope activation is built for transition-rate events and rebinds happen at sample rate, orders of magnitude hotter. It also tangles variable lifetime with stage entry invariants (entry one-shots, resets, latches), multiplying that surface.

### 4. Per-node execution guards
A node carries a guard ref (sel matches its tag) and the scheduler skips it on mismatch. This is the smallest real scheduler primitive and would delete the synthesized dispatcher and `$sel`. But it solves a problem we already solved, in a harder place: today the dispatcher absorbs pending values at one node; with guards, every node of every derivation subgraph accumulates fresh inputs while guarded off, and the scheduler has to run absorption as a guard-flip lifecycle event with defined same-tick ordering.

### 5. Unified clock system
Trigger-only entries, entry latches, and ref sampling are already ad-hoc clocks; a synchronous-dataflow clock system would unify them, and a bound derivation becomes a stream clocked on its bind predicate. But that means reformulating every existing gating mechanism as clock calculus, which touches the firing semantics of every node type and risks behavior changes across every existing program. This would be a ground-up redesign of the scheduler's execution semantics, not a refactor we bolt on after this PR.

In every one of these, every derivation still compiles and a bind table still exists somewhere. The demux never disappears, it just moves. The current design keeps it static and race-free with zero scheduler risk.

sy-nico:
I went ahead and removed protections so that all function params could support variables as the inputs (`aad1beebb2`). As predicted, that opened up a huge can of worms, most visibly in the timing functions (`e43dc8ac13`): `time.interval` and `time.wait` now read their span live at each evaluation, with the loop timer following live deadlines and the timing tolerance derived from the GCD of declared and literal-reassigned periods, and `stable.for` latches its duration at window start so an in-flight window completes under its starting duration. Those behaviors are deliberate decisions, each with unit, lowering, and end-to-end coverage.

So for this review, call out bugs and code nits and I'll fix them here. For behavioral changes you'd like to see, file a ticket and link this PR.

sy-nico:
@greptile, rerun

emilbon99 (approved):
LGTM. I really appreciate the detailed explanation, has me feeling much better about the current design given the options considered.

PR #2583, opened 2026-07-12
(A) SY-4385-1: Add Keyed Membership and Presence Store Primitives

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4385](https://linear.app/synnax/issue/SY-4385)

## Description

First PR in the tabs/mosaic composition stack. Introduces two general-purpose
keyed reactive contexts in pluto's `store` package, both built on the existing
`useKeyedListeners` primitive:

- **`Store.createMembership`** — a controlled *set* of keys with surgical
  per-key subscriptions (`useIsMember` / `useItem` / `useMembers` /
  `useMemberAmong`). A single mutation notifies only the keys whose membership
  actually flipped (symmetric difference).
- **`Store.createPresence`** — a single *present* key (the hovered item, the
  active item, a keyboard cursor). A change notifies only the previous and next
  holders, so at most two items re-render.

Selection is refactored onto these primitives. Because the selection context is
a select-internal concern (not a shared primitive — that role now belongs to
`Store.*`), it is lifted out of the `select/base` subpackage up into
`select/Context` as a single instance composing one membership (selection +
`onSelect`) and one presence (hover). Focus tracking is dropped.

`table`'s cell selection is migrated onto its own `Store.createMembership`
instance, so the selection context is no longer imported anywhere outside
`select`.

No user-facing behavior change: `Select.Context` / `Select.useItemState` keep
their public shape (minus `focused`), and all selection/table specs pass.

## Basic Readiness


pluto/src/store/presence.tsx
```diff
@@ -0,0 +1,124 @@
...
+> extends PropsWithChildren {
+  value?: K;
+}
+
+/**
+ * Presence is a created presence context: a provider that publishes at most one present
+ * key plus hooks bound to its own React context. Nesting is independent, so a created
+ * presence never reads or writes the presence of another.
+ */
+export interface Presence {
```
pjdotson: shouldn't the type parameter be at the top level `Presence<K extends record.Key = record.Key>` so everything has to use the same key type?

pluto/src/store/presence.tsx
```diff
@@ -0,0 +1,124 @@
...
+    useLayoutEffect(() => {
+      if (valueRef.current === value) return;
+      const changed: K[] = [];
+      if (valueRef.current != null) changed.push(valueRef.current);
+      if (value != null) changed.push(value);
+      valueRef.current = value;
+      if (changed.length > 0) notifyListeners(changed);
+    }, [value, notifyListeners]);
+    return (
+      <BaseContext value={ctx as unknown as ContextValue<record.Key>}>
```
pjdotson: shouldn't use `as unknown as`. You should be able to fix if you give a type parameter to `createPresence`

pluto/src/store/membership.tsx
```diff
@@ -0,0 +1,184 @@
...
+  member: boolean;
+  onItem: () => void;
+}
+
+/**
+ * Membership is a created set-membership context: a provider that publishes a controlled
+ * set of keys plus hooks bound to its own React context. Nesting is independent, so a
+ * created membership never reads or writes the set of another.
+ */
+export interface Membership {
```
pjdotson: shouldn't the `K` live at the top so all members use the same key type?

pluto/src/store/membership.tsx
```diff
@@ -0,0 +1,184 @@
...
+    children,
+  }: MembershipProps<K>): ReactElement => {
+    const valueRef = useRef(value);
+    const { notifyListeners, subscribe } = useKeyedListeners<K>();
+    const getValue = useCallback(() => valueRef.current, []);
+    const ctx = useMemo<ContextValue<K>>(
+      () => ({ onItem, setValue, clear, subscribe, getValue }),
+      [getValue, onItem, setValue, clear, subscribe],
+    );
+    useLayoutEffect(() => {
```
pjdotson, line 121: are you sure we need to use `useLayoutEffect` instead of `useEffect` for these?

pjdotson (changes requested):

pluto/src/store/membership.tsx
emilbon99: Yes, 100% we should be using `useLayoutEffect`.  The values here synchronously drive the way the DOM renders, so using a standard `useEffect` here means we get weird flickering with selection state. This keeps everything nice and smooth, and is not an expensive effect to run.

pluto/src/store/presence.tsx
emilbon99: See comment above.

pluto/src/store/membership.tsx
emilbon99: See comment above.

pluto/src/store/presence.tsx
emilbon99: It's not possible to do this if you want he consumers of the key type to be polymorphic. Right now the `Select` component accepts a type parameter for `Key` which means that it's `Membership` and `Presence` contexts need to be able to accept a type parameter without doing it via a factory.

emilbon99: If you disagree, take a look at the diff on the select components and tell me how they should work with a statically defined key type on the factory.

pjdotson: The point is that we should move type assertions / bad typing up, not down. I would rather have type assertions (`as unknown as`) in the `Select` component, that way when you read the select component code, you know that the select component isn't properly typed, but the presence / membership components are properly typed.

pjdotson (changes requested):

pluto/src/store/presence.tsx
emilbon99: Implemented.

pjdotson (approved):

PR #2587, opened 2026-07-12
(E) SY-4370-1: Add Getter Hooks to Flux Selectors and an Optimistic Update Lifecycle

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4370](https://linear.app/synnax/issue/SY-4370)

## Description

First slice of the console panel integration stack: the flux store primitives the panel component depends on.

* `flux.createSelector` now returns a `[useSelect, useGet]` pair. `useSelect` is the existing reactive hook; `useGet` returns a stable getter that reads the current value on demand without subscribing, so callbacks can read store state without mounting a subscription. The transform memo now also keys on args, so a selector re-derives when an arg changes even if the raw selection is referentially stable.
* `Scope.bindSelector` binds a selector pair to a scope: both the reactive hook and the getter resolve their key from the surrounding scope, overridable per call.
* Flux updates gain an optimistic-completion lifecycle: update implementations receive a required `onOptimisticComplete` callback and callers can hook `afterOptimistic`. Update methods accept an `extra` option threaded into `beforeUpdate`/`afterOptimistic`/`afterSuccess`/`afterFailure` for per-call context.
* `state.usePurePassthrough` renames `initial` to `initialValue` and both passthrough hooks now always notify `onChange`, controlled or not.
* All pluto domain query modules are migrated to the pair-return shape and thread `onOptimisticComplete` through their update implementations. `useRetrieveObservableName` is preserved in every domain the console still consumes it from.

## Basic Readiness

### Greptile Summary

This PR establishes the flux store primitives needed for the upcoming console panel integration. `flux.createSelector` now returns a `[useSelect, useGet]` pair — the reactive hook for components and a stable getter for callbacks that must read store state without mounting a subscription. Transform memoization additionally keys on args so a selector re-derives when an arg changes even if the raw selection is referentially stable.

* `Scope.bindSelector` binds both hooks to a scope, injecting the active key into the reactive hook and the getter, with per-call override support.
* `UpdateParams` gains a required `onOptimisticComplete` callback and callers can hook `afterOptimistic`; an `Extra` type parameter is threaded through all four lifecycle callbacks (`beforeUpdate`, `afterOptimistic`, `afterSuccess`, `afterFailure`).
* All 20+ domain query files are migrated to the pair-return shape and wire `onOptimisticComplete` through their update implementations.

### Confidence Score: 4/5

Safe to merge after fixing the schematic delete ordering: the schematic is still present in the store when `onOptimisticComplete` fires, and the store delete has no rollback if the server call fails.

The core flux primitives (`createSelector`, `afterOptimistic`) are well-implemented and well-tested. The domain migration is mechanically correct for all modules except `schematic/queries.ts`, where `store.schematics.delete(keys)` is pushed only after the server call completes, so the optimistic signal fires while the schematic is still in the store and any `afterOptimistic` side-effects have no rollback on server failure. Every other domain places the store mutation before the signal.

pluto/src/schematic/queries.ts — `useDelete` needs `store.schematics.delete(keys)` pushed before `await onOptimisticComplete(data)`.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| pluto/src/flux/select.ts | Core change: `createSelector` now returns a `[useSelect, useGet]` pair; transform memo keys on args identity in addition to raw value reference. New `computeSelected` helper shared by both hooks. Logic is correct and well-tested. |
| pluto/src/flux/update.ts | Adds `onOptimisticComplete` to `UpdateParams`, `afterOptimistic` lifecycle hook to `UseObservableUpdateParams`, and `Extra` type parameter throughout. `onOptimisticComplete` is correctly wired so `afterOptimistic` errors propagate to the rollback path. |
| pluto/src/scope/scope.tsx | Adds `bindSelector` that binds both the reactive hook and the getter hook to the scope; scope key defaults correctly and is overridable per call. `useCallback` dependency array `[get, scopeKey]` is correct. |
| pluto/src/schematic/queries.ts | All selectors migrated to pair-return shape correctly. However, `useDelete` calls `onOptimisticComplete` before `store.schematics.delete(keys)` (which itself runs only after the server call), so `afterOptimistic` sees the schematic still present and rollback coverage is incomplete if the server call fails. |
| pluto/src/flux/select.spec.tsx | Test coverage for the new `useGet` hook and the arg-keyed transform memoization is comprehensive: stable getter, no-subscription guarantee, transform applied, latest value on call, and deep-equal arg stability. |
| pluto/src/flux/update.spec.ts | New `afterOptimistic` tests cover: fires when `onOptimisticComplete` is called, not fired when it isn't, ordering relative to `afterSuccess`, payload shape, and rollback-on-throw. All cases are well-structured. |
| pluto/src/ranger/queries.ts | Adds `useSelectName`/`useGetName` pair and `requireRange` helper. Minor: the local variable inside `requireRange` is named `schem` (copy-paste from schematic pattern) instead of `range`. |
| pluto/src/lineplot/queries.ts | All selectors migrated to pair-return shape; `useDelete`/`useCreate`/`useRename` correctly place `onOptimisticComplete` after store mutations and before server calls. |
| pluto/src/project/queries.ts | `ListParams` is now derived from `project.RetrieveRequest` fields directly, enabling `keys` filtering. `onOptimisticComplete` placement in update implementations is correct. |
| pluto/src/flux/form.ts | Omits `onOptimisticComplete` from `FormUpdateParams`, intentionally excluding the new lifecycle from form-based updates. Correct and consistent with the form pattern. |

### Sequence Diagram

```mermaid
sequenceDiagram
    participant Caller
    participant useUpdate
    participant update (domain impl)
    participant Store
    participant afterOptimistic
    participant Server

    Caller->>useUpdate: updateAsync(data, opts)
    useUpdate->>update (domain impl): update({ data, store, rollbacks, onOptimisticComplete, ... })
    update (domain impl)->>Store: store.domain.delete/set (optimistic, push rollback)
    update (domain impl)->>useUpdate: onOptimisticComplete(output)
    useUpdate->>afterOptimistic: "afterOptimistic({ data, store, client, rollbacks, extra })"
    afterOptimistic-->>useUpdate: resolved
    update (domain impl)->>Server: client.domain.delete/create/rename
    Server-->>update (domain impl): response
    update (domain impl)-->>useUpdate: output
    useUpdate->>Caller: "afterSuccess({ data, client, extra })"
```

```mermaid
sequenceDiagram
    participant Caller
    participant useUpdate
    participant update (domain impl)
    participant Store
    participant afterOptimistic
    participant Server

    Caller->>useUpdate: updateAsync(data, opts)
    useUpdate->>update (domain impl): update({ data, store, rollbacks, onOptimisticComplete, ... })
    update (domain impl)->>Store: store.domain.delete/set (optimistic, push rollback)
    update (domain impl)->>useUpdate: onOptimisticComplete(output)
    useUpdate->>afterOptimistic: "afterOptimistic({ data, store, client, rollbacks, extra })"
    afterOptimistic-->>useUpdate: resolved
    update (domain impl)->>Server: client.domain.delete/create/rename
    Server-->>update (domain impl): response
    update (domain impl)-->>useUpdate: output
    useUpdate->>Caller: "afterSuccess({ data, client, extra })"
```

Reviews (2): Last reviewed commit: ["Merge branch < />sy-4385-mosaic< /> into sy-43..."](<https://github.com/synnaxlabs/synnax/commit/4164e4c125869e8c44d2806f078edf374f83a5bd>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=43660824>)


pluto/src/state/state.ts
```diff
@@ -70,23 +70,35 @@ export interface UsePassthroughProps<NextState extends State> {
...
+  const setAndNotify = useCallback(
+    (arg: SetArg<NextState>) => {
+      setInternal(arg);
+      onChange?.(arg);
+    },
+    [onChange],
+  );
   if (value != null && onChange != null) return [value, onChange];
-  return [internal, setInternal];
+  return [internal, setAndNotify];
```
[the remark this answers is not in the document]
emilbon99: Intentional contract change, not an accident: value decides who owns the state, onChange decides who hears about it. The new contract is documented directly above usePassthrough and the specs were updated to assert it. Consumers were exercised by the pluto and console suites, which are green with the new behavior.

pluto/src/flux/update.ts
```diff
@@ -213,8 +279,12 @@ const useObservable = <
   const store = useStore<Store>(scope);
   const addStatus = useAdder();
   const runUpdate = useCallback(
-    async (data: Input, opts: base.FetchOptions = {}): Promise<boolean> => {
-      const { signal } = opts;
+    async (
```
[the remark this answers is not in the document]
emilbon99: Acknowledged, but keeping as-is. Extra defaults to void, and making extra conditionally required for non-void Extra needs conditional tuple parameter types that complicate every call site for a hook whose only non-void consumers (e.g. useOpenTab) construct the options object at the same site that declares Extra. The optionality is documented on UpdateOptions.

emilbon99:
@greptileai review

pjdotson (changes requested):
Can you explain more why we should have APIs that look like this (from `ranger`):

`Ranger.useSelectName()`, `Ranger.useGetName()`?

I am worried that now a) we have two functions that look and sound very similar to each other, and b) by exposing a "getter" we are allowing footguns for us to write code that _won't_ use the latest value of a field:

```tsx
export const Component = () => {
  const getter = Ranger.useGetName()
  return <div>getter()</div>
}
```

and I don't see anything explain what components / use cases this is supposed to be for in the PR changes or description.

emilbon99:
> Can you explain more why we should have APIs that look like this (from `ranger`):
> 
> `Ranger.useSelectName()`, `Ranger.useGetName()`?
> 
> I am worried that now a) we have two functions that look and sound very similar to each other, and b) by exposing a "getter" we are allowing footguns for us to write code that _won't_ use the latest value of a field:
> 
> ```tsx
> export const Component = () => {
>   const getter = Ranger.useGetName()
>   return <div>getter()</div>
> }
> ```
> 
> and I don't see anything explain what components / use cases this is supposed to be for in the PR changes or description.

The above example doesn't seem that concerning to me, as its pretty rare for us to call a getter inside JSX, and I think that most react devs understand that unless the function being called is a hook i..e like `useGetter()` the value won't be reactive. 

----

As for the use case, these getter functions are used in any place where a callback wants to have access to the current value without re-rendering the component.

There are a lot of places in the console where we do:


```tsx
const MyComponent = () => {
       const store = Flux.useStore()
       const myCallback = useCallback(() => {
           // What happens if the channel doesn't exist? What happens if I need the channel status? This doesn't get it
           const ch = store.channels.get(yada)
       }, [store])
}
```

or this:

```tsx
const MyComponent = () => {
       const name = Schematic.useSelectName();
       const myCallback = useCallback(() => {
           // name gets used here
       }, [name])
      // but it never gets used in the JSX
}
```

So we just end up re-rendering tons of component uselessly

or we do even worse

```tsx
const MyComponent = () => {
       const client = Synnax.use()
       const handleError = Status.useErrorHandler();
       const myCallback = useCallback(() => {
          handleError(async () => {
              /// We uselessly retrieve the cached schematic again from the server.
              const { name } = await client.schematics.retrieve(key)
            })
       }, [client, handleError])
}
```

With these hooks:

```tsx
const MyComponent = () => {
       const getName = Schematic.useGetName()
       const myCallback = useCallback(() => {
                // Clean, curated name getter that is guaranteed to be current and leverage the cache efficiently.
                const name = getName()
       }, [getName])
}
```

This means that we get a *correct, curated API* for efficiently retrieving the current value when we don't need it reactively. 


EDIT: correctness of last example

pluto/src/scope/scope.tsx
```diff
@@ -84,5 +116,22 @@ export const create = <K extends record.Key>(name: string): Instance<K> => {
       return hook({ ...args, key } as Args);
     };
 
-  return { Provider, use, useOptional, bindHook };
+  const bindSelector = <Args extends record.Keyed<K>, Selected>([
```
pjdotson, line 119: this needs tests

pluto/src/flux/select.ts
```diff
@@ -28,27 +28,68 @@ export interface CreateSelectorParams<
...
-   * mapping does not cause re-renders on unrelated store updates.
+   * transform derives the selected value from the raw selection. It is memoized
+   * on the raw value's reference and the args: as long as select returns the same
+   * reference (e.g. a stored array that only changes when its contents change) and
+   * the args are unchanged, transform is not re-run and the previous result is
+   * returned, so a derived mapping does not cause re-renders on unrelated store
+   * updates. It does re-run when an arg it depends on changes, even if the raw
+   * selection is referentially stable.
    */
   transform?: (raw: Raw, args: Args) => Selected;
```
pjdotson, line 39: Unrelated question but does it really make sense for `transform` to be part of the Flux API instead of just a hook that providers can manually write?

e.g. `useSelectStage` for a range could just be a hook that calls `useSelectRange` and then does a transform on the returned value to get the stage.

pluto/src/ranger/queries.ts
```diff
@@ -910,7 +917,27 @@ export const { useUpdate: useRename } = Flux.createUpdate<RenameParams, FluxSubS
...
     await client.ranges.rename(key, name);
     return data;
   },
 });
+
+const requireRange = (store: FluxSubStore, key: ranger.Key): ranger.Range => {
+  const schem = store.ranges.get(key);
+  if (schem == null) throw new NotFoundError(`Range with key ${key} not found`);
+  return schem;
+};
```
pjdotson, line 930: why do we have to do this for ranges but not everything else? or does this logic already exist for everything else?

pjdotson (changes requested):
This should be done in a separate PR, but I think the names of `useSelectName` and `useGetName` are too similar and need to be changed. It should look something like this:

```ts
const name = Ranger.useName(key);
const getName = Ranger.useGetName() // or selectName = Ranger.useSelectName()
```

That way it is obvious which one returns the value and which one returns the getter function.

pluto/src/flux/select.ts
emilbon99: Imagine I have a selector like this:

```tsx
const value = useSelectFluxValue()
const transformed = useMemo(() => {
      return transform(value)
}, [value])
```

The problem with this is that because `value` changed, the _entire_ component still re-renders even though transform is memoized. 

There is probably some refactor to make the DX of the transform hook better, but this is the fundamental reason why I made transform an embedded field.

pluto/src/ranger/queries.ts
emilbon99: We have it for all of the other types as well.

emilbon99: But calling it `schem` is dumb. Will fix.

emilbon99:
> This should be done in a separate PR, but I think the names of `useSelectName` and `useGetName` are too similar and need to be changed. It should look something like this:
> 
> ```ts
> const name = Ranger.useName(key);
> const getName = Ranger.useGetName() // or selectName = Ranger.useSelectName()
> ```
> 
> That way it is obvious which one returns the value and which one returns the getter function.

https://linear.app/synnax/issue/SY-4492/refactor-flux-selector-and-getter-names

pjdotson (approved):

PR #2595, opened 2026-07-13
SY-4484: Task Autosave with Deploy-on-Start RFC

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4484](https://linear.app/synnax/issue/SY-4484)

## Description

Adds RFC 0057, which splits saving a task config from deploying it to the Driver so the Console can autosave task forms:

- The task's single `config` field becomes an autosaved draft. The Driver stops reconfiguring on `sy_task_set` (set events become a metadata-only refresh, fixing rename-restarts-task), and the `start` command absorbs deployment via a hash comparison against the live instance's config. Drift is derived from a Driver-reported hash and rack in status details, with no new commands, endpoints, or snapshot fields.
- The Core assigns the config hash on every write and persists it on the task, so no client ever hashes and the drift signal rides the metadata event without the config.
- Autosave is made safe by a two-tier schema: a lax shape schema that lets an incomplete draft persist, and a strict deploy schema validated when the user presses play.
- Task keys migrate from rack-encoded uint64s to UUIDs with `rack` as a plain field, so drafts can be created instantly without a rack and rack moves become field writes handled through the same drift and redeploy flow.
- Console task tabs move from view tabs to resource tabs with a create-then-open flow, deleting the unsaved-task sentinels, the persisted-transition args rewrite, and the dual-mode tab name logic.
- Arc converges on the same pattern through a Core deploy endpoint and a semantic program hash stamped into the task config, with no Driver changes.

The RFC is written against the shipped implementation in #2603, #2604, #2605, and #2663, and it follows the conventions recorded in `docs/tech/rfc/CLAUDE.md`.

## Basic Readiness


emilbon99:
@greptile-apps review

docs/tech/rfc/0056-260713-task-draft-deploy.md
```diff
@@ -0,0 +1,433 @@
...
+
+```
+task.config_hash != status.details.config_hash
+```
+
+The hash is content-addressed, so a no-op save never false-positives and an undone edit
+restores the original hash.
+
+Nothing else can hash the config correctly: it is a decoded map at every hop, never
+bytes (core stores `json.Marshal` output, the C++ driver gets a protobuf `Struct`,
```
pjdotson: can you fix the capitalization everywhere

docs/tech/rfc/0056-260713-task-draft-deploy.md
```diff
@@ -0,0 +1,433 @@
...
+clients get JSON or msgpack), so a client-side hash needs a byte-for-byte
+canonicalization in four languages. The TypeScript client cannot even see what core
+stored: zod parsing injects defaults and strips unknown keys. Hashing once, where the
+config is written, avoids all of it.
```
pjdotson: Note that your hashing function will need to hash equivalent JSON deterministically, e.g. make sure that `{"key1": 1, "key2": 2}`, and `{"key2": 2, "key1": 1}` both hash to the same value.

docs/tech/rfc/0056-260713-task-draft-deploy.md
```diff
@@ -0,0 +1,433 @@
...
+
+The gate lives in the integrations: each already writes its own configure statuses, so
+each decides when to stay silent. A configure driven by a pending start or an
+`auto_start` config reports; a bare boot configure logs. An auto-start task with an
+unfinished draft still errors at boot; separating drafts from deployable configs is
+deferred to section 8.
+
+## 3.5 - Drift is a running-task concept
+
+A stopped task has no live instance, hence no authoritative hash and no drift. Nothing
```
pjdotson: To understand this more, the Driver basically has two things:

1) Task bootup, e.g. parsing the config into a different shape, applying defaults.
2) Task running

And when the Driver gets a start command, it will compare the new config hash against the old, if they are the same move directly to `2`, otherwise it runs both 1 and 2?

docs/tech/rfc/0056-260713-task-draft-deploy.md
```diff
@@ -0,0 +1,433 @@
...
+pretending otherwise.
+
+## 3.2 - Start means deploy
+
+`start` becomes: fetch the task, compare its `config_hash` against the hash the running
+instance was built from, and
+
+- if no instance exists or the hashes differ: rebuild from the fresh config (the
+  existing `CONFIGURE` path), then start.
+- if the hashes match: start the existing instance, as today.
```
pjdotson: So what is the difference between starting a task and resuming a paused task?

docs/tech/rfc/0056-260713-task-draft-deploy.md
```diff
@@ -0,0 +1,433 @@
...
+This RFC splits the two:
+
+1. Drivers stop configuring on `sy_task_set`; the set event becomes a metadata-only
+   refresh, so renaming no longer restarts a task.
+2. The `start` command absorbs deployment: the driver fetches the task, compares its
+   `config_hash` against the running instance's hash, and rebuilds from the fresh config
+   when they differ or no instance exists.
+3. Core hashes `config` on every write into `config_hash`; the driver echoes the hash
+   its instance was built from in every status. Drift is
+   `task.config_hash != status.details.config_hash`; no client ever hashes.
```
pjdotson: Will the config hash be stored in Gorp? If the config hash is just computed data from the config why would we duplicate storing it?

pjdotson (changes requested):
This PR needs info about how it reacts with the auto start button as well. That is just ignored except for on Driver bootup?

docs/tech/rfc/0056-260713-task-draft-deploy.md
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Done. The RFC now follows `docs/tech/rfc/CLAUDE.md`, which landed in #2664 after this branch was cut: the Core, the Driver, the Console, Pluto, Gorp, Oracle, Arc, and Zod are capitalized throughout, and "server" is now "the Core".

The rest of the conventions came along with it: the file is renamed `0057-task-autosave-deploy-on-start.md` (no date, and 0056 was taken by #2673), the front matter is the author/date/related bullet list, headings are hierarchical and numbered from zero, and citations are `§4.2` rather than "section 4.2".

docs/tech/rfc/0056-260713-task-draft-deploy.md
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Good catch, and it holds. `hashConfig` (`core/pkg/service/task/hash.go`) marshals the decoded config map with `encoding/json`, which sorts map keys at every level, so `{"key1": 1, "key2": 2}` and `{"key2": 2, "key1": 1}` produce identical bytes and identical hashes. Arrays keep their order, which is what we want since order is meaningful in a config.

The RFC did not state this, so §4.4 is now a section on exactly that property, including the one failure mode: a config that cannot be JSON encoded, such as a NaN or infinite float arriving over msgpack, fails the write rather than hashing to something arbitrary.

Worth adding that this is only safe because the hash is computed in one place. The earlier design had all four languages hashing, which is where equivalent-JSON agreement gets genuinely hard, and it is why we moved it into the Core.

docs/tech/rfc/0056-260713-task-draft-deploy.md
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Yes, that is exactly it. Configure parses the config, applies defaults, and builds the live instance; run executes it. On a start the Driver compares the stored `config_hash` against the hash the live instance was built from: equal means go straight to (2), different or no instance at all means (1) then (2).

Your framing was clearer than the RFC's, so I took it: §3 is a new vocabulary section defining configure, run, deploy, live instance, and drift, and §4.1 states the two-phase model and the gate before §4.2 gets into the command.

One case worth naming: a snapshot task ignores start outright, since its config is a frozen record rather than a deployable one.

docs/tech/rfc/0056-260713-task-draft-deploy.md
emilbon99: *Emiliano + Claude: generated reply from our review session.*

There is no resume, and that is the point of the design rather than an omission. The wire carries `start` and `stop` only. The Console's pause icon sends `stop`, and the Driver keeps the live instance rather than destroying it, so the next `start` against an unchanged config runs that same instance. A resume is precisely the hash-matched branch of start, so it needs no second command.

This is now written out in §3, and it also resolves what was open question 2 in the old draft (whether the instance is retained after stop). It is retained, recorded in §9.1.

docs/tech/rfc/0056-260713-task-draft-deploy.md
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Yes, it is a persisted field on the task row, written by `task.Writer` on every write and ignored on writes from clients. You are right that it is derived data, so the RFC now argues the case explicitly in §4.3 instead of assuming it. Three reasons:

- `sy_task_set` strips `config` (§4.9), so the stored hash is the only drift signal the metadata event can carry. Deriving it on read would force every listener to fetch the full config on every rename.
- The value has to be comparable against a hash the Driver writes into its status. Storing a server-assigned string is what lets the Core and the Driver agree without four languages agreeing on an algorithm.
- Snapshots keep their captured hash. `Writer.Create` carries `existing.ConfigHash` forward with the config it preserves, so a snapshot's identity does not shift when the live task is edited.

The cost is one 16-character string per row.

emilbon99:
*Emiliano + Claude: generated reply from our review session.*

On auto start: your read is right, and §4.6 is a new section on it. `auto_start` is a per-task config field, not a control, so nothing about it changes shape here. What changes is how often it can fire. Configure used to run on every save, so toggling auto start took effect on the next save. Configure now runs only at boot and at start, which leaves it exactly one job:

- **At boot**: the task is configured, started, and reported on, including configuration failures. Unchanged.
- **After boot**: nothing. A start command already runs the task, so the auto-start branch and the start command arrive at the same place.

The one consequence worth knowing is that auto start is config like any other field, so flipping it autosaves and re-hashes. Flipping it on a running task drifts that task and raises the redeploy control; flipping it on a stopped task takes effect at the next boot.

I also went back over the whole document against what actually shipped in #2603, #2604, #2605, and #2663, and fixed the places where it had drifted from the implementation:

- **The two-tier form schema was missing entirely**, which is the load-bearing piece of the autosave story. The shape schema stays lax so drafts persist mid-typing; a separate strict `deployConfigZ` runs at deploy, with warning-variant issues rendering and everything else blocking the start. §6.0 covers it, along with the real deploy pipeline order.
- **Ontology placement was wrong.** The RFC said drafts parent under a Tasks group and reparent under the rack. We ship every task under one cluster-level Tasks group with no rack edges at all, which is cleaner: a rack change moves no relationships. §6.5 rewritten.
- **Arc was listed as future work** and is now built, so §6.6 covers the Core deploy endpoint, the semantic program hash, and the editor convergence.
- **Rollout was understated.** Three migrations, not two, and the tab migration is a separate pass consuming a staged legacy-to-UUID map rather than an extension of the project-layout conversion. Phases now map to the four PRs.
- **Driver behaviors the RFC never mentioned**: snapshot tasks ignore start, and a command for an unknown key only warns after confirming the task belongs to this rack.
- **Old open questions 1 through 3 are answered** from the implementation in §9 (starts are not coalesced, the instance is retained after stop, a rack move does not wait).

One gap I found rather than fixed, now §11.0: the RFC claimed a rackless start fails with a clear error. It does not. Commands are broadcast and the Core does not validate them, so a start for a task with rack 0, or for a rack whose Driver is offline, is simply unclaimed and returns no status at all. The Console dodges the first case because `onConfigure` resolves a rack before sending the start, but move a task to a disconnected rack and press play and nothing happens, silently. Options are a Core-side rejection on rack 0, a Console-side block, or a client-side timeout that reports the command as unclaimed. Happy to take your preference.

Also note the greptile P1 further up was a fair hit: boot does configure a never-started draft, so §6.0 no longer claims nothing reads it until start. §4.5 spells out that such a configure fails silently by design.

PR #2599, opened 2026-07-14
(K) SY-4490: Serve Core on Multiple Ports with Per-Listener Certificates

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4490](https://linear.app/synnax/issue/SY-4490)

## Description

Implements RFC 0044: a Core node can now serve its full API surface on several
listeners at once, each with its own address and TLS certificate. Single-listener
behavior is unchanged.

**Certificate sources.** A new `cert.Source` interface (matching the crypto/tls
`GetCertificate` callback) resolves the certificate a listener presents during the
handshake. Three implementations:
- `file` — serves a cert/key PEM pair, hot-reloading on modtime change so rotation
  needs no restart.
- `auto` — self-signs from the built-in CA, deriving SANs from the listener address.
- `tailscale` — defers to the local tailscaled daemon. It lives in its own subpackage
  (`pkg/security/cert/tailscale`) and registers through `cert.RegisterSource`, so the
  base package never imports it.

**Security provider split.** The node-wide TLS policy and outbound peer identity stay
on `Provider`; the per-listener inbound certificate is a `cert.Source`. `TLS()` is
replaced by `TLSConfigFor(src)` (per-listener inbound) plus `NodeClientConfig()` (node
outbound dialing).

**Server.** `server.Config` now takes `[]server.Listener{Address, TLS}`. `start()`
builds one cmux tree per listener; every branch (gRPC, Fiber, redirect) binds to all of
them. `SimpleHTTPBranch` was made N-listener-safe (its single-slot stop channel would
have deadlocked with multiple serve goroutines).

**Configuration.** `--listen` is now polymorphic: a scalar address (the existing form,
mapped to a file source over the node cert paths, advertised) or a list of listener
objects, each with an `address`, a `cert` block (`source`/`cert`/`key`), and optional
`advertise`/`name`. Validation enforces the RFC rules: at most one advertised listener,
unique addresses, resolvable source per listener in secure mode, file needs cert+key,
auto/tailscale must not set them. A listen list may not be combined with the global
node-cert flags (`--auto-cert`, `--node-cert`, `--node-key`); `--certs-dir` and the
`--ca-*` flags stay legal since they still locate the cluster CA and node identity.

Exactly one listener is advertised to the cluster and used as the driver address
(first listener when none is explicitly marked).

### Dependency note

The `tailscale` source pulls `tailscale.com v1.100.0` into `core`, adding ~18 indirect
modules. These come from the thin `client/local` LocalAPI socket client only — the
heavy wireguard/ebpf/wintun deps are pulled during `go mod tidy` graph resolution but
do **not** land in the final build graph. Flagging for reviewer awareness since it is a
non-trivial addition to the server module.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR lets Core serve the full API from multiple listeners. The main changes are:

- Per-listener address and certificate configuration.
- File, auto, and Tailscale certificate sources.
- Separate inbound listener TLS and outbound node TLS settings.
- Multi-listener server startup and shutdown handling.
- CLI and production docs for the new listener format.

This looks safe to merge.

| Filename | Overview |
|----------|----------|
| core/cmd/listener/listener.go | Adds listener parsing, validation, certificate source resolution, and advertised listener selection. |
| core/pkg/security/secure.go | Splits inbound listener TLS from outbound node TLS and adds advertised certificate verification. |
| core/pkg/security/cert/tailscale/tailscale.go | Adds a Tailscale-backed certificate source with configured-host fallback for empty SNI. |
| core/pkg/server/server.go | Updates server startup to bind and serve the same branches on multiple listeners. |
| core/pkg/server/simple_http.go | Updates the simple HTTP branch shutdown path for multi-listener serving. |

1. `core/pkg/security/cert/tailscale/tailscale.go`, line 1630-1633 ([link](https://github.com/synnaxlabs/synnax/blob/4bf04ee3bf9b7b506b3b11fdda1b4422981ce8f6/core/pkg/security/cert/tailscale/tailscale.go#L1630-L1633)) 

   <a href="#">P1</a> <a href="#">security</a> **Wrong SNI Persists**

   This fallback only handles clients that omit SNI. If a client connects to the listener by `localhost`, an IP address, a short hostname, or another non-Tailscale name, `hi.ServerName` is non-empty and this code passes that value through to tailscaled instead of using the configured listener host in `s.host`. Tailscaled then looks up a certificate for the wrong name, so the TLS handshake can still fail even though the listener was configured with the correct Tailscale address.
</details>

<sub>Reviews (6): Last reviewed commit: ["Assert each cert source satisfies cert.S..."](https://github.com/synnaxlabs/synnax/commit/516faecaadf5174cf4037e9a8e373f4d05b586c1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44187108)</sub>

**Context used:**


emilbon99:
@greptileai re-review

core/pkg/security/cert/tailscale/tailscale.go
```diff
@@ -0,0 +1,34 @@
...
+	"tailscale.com/client/local"
+)
+
+func init() { cert.RegisterSource(cert.SourceTypeTailscale, newSource) }
+
+// source serves certificates from the local tailscaled daemon. The daemon fetches and
+// caches them, so GetCertificate defers entirely to it.
+type source struct{ client local.Client }
+
+func newSource(cert.SourceConfig) (cert.Source, error) { return &source{}, nil }
```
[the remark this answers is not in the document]
emilbon99: The address isn't dropped. `NewSource` captures it as `s.host` (`host: cfg.Address.Host()`), and `GetCertificate` falls back to it when the ClientHello carries no SNI. The `source_internal_test` specs pin both the empty-SNI fallback and the non-empty passthrough.

The residual concern, a client dialing by IP, `localhost`, or a short hostname, is inherent to Tailscale certificates rather than something this constructor can fix. The certificate is issued by a public CA keyed to the node's tailnet FQDN, and TLS verification happens client-side against the name the client actually dialed. If a client dials a non-FQDN name, forcing the SNI to the configured host would make tailscaled serve the FQDN cert, which the client then rejects on a name mismatch. So overriding a non-empty SNI cannot rescue that client; it can only break a legitimate handshake that stated a different valid name. Those clients have to dial the `ts.net` FQDN regardless of what the server presents.

Honoring a non-empty SNI as-is, and backfilling the configured host only when SNI is absent, is therefore the correct behavior. That contract is now documented on `GetCertificate`.

emilbon99:
@greptileai re-review

core/cmd/start/listener.go
```diff
@@ -0,0 +1,189 @@
...
+		return
+	}
+	advertised := listeners[0]
+	for _, l := range listeners {
+		if l.advertise {
+			advertised = l
+			break
+		}
+	}
+	v.Ternary(
```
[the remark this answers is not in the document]
emilbon99: Fixed in 6a568b157d. Added Provider.VerifyClusterCert, which verifies the leaf against the cluster CA pool; Configs.Resolve now runs it on the advertised listener (only), so a file listener advertising a non-cluster cert fails startup with a clear error instead of breaking gossip at peer-dial time.

core/pkg/security/cert/tailscale/tailscale.go
```diff
@@ -0,0 +1,67 @@
...
+// Factory builds tailscale sources.
+type Factory struct{}
+
+var _ cert.SourceFactory = Factory{}
+
+// Type implements cert.SourceFactory.
+func (Factory) Type() string { return SourceType }
+
+// NewSource implements cert.SourceFactory.
+func (Factory) NewSource(cfg cert.SourceConfig) (cert.Source, error) {
```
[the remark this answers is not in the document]
emilbon99: Fixed in 6a568b157d. tailscale.NewSource now rejects a listener with an empty host.

emilbon99:
@greptile-apps review

core/pkg/security/secure.go
```diff
@@ -68,10 +68,37 @@ var defaultCipherSuites = []uint16{
...
+	if err != nil {
+		return err
+	}
+	leaf := c.Leaf
+	if leaf == nil {
+		if leaf, err = x509.ParseCertificate(c.Certificate[0]); err != nil {
+			return err
+		}
+	}
+	_, err = leaf.Verify(x509.VerifyOptions{Roots: p.certPool})
```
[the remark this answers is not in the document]
emilbon99: Fixed in 710b0fda26. VerifyClusterCert now takes the advertised host and passes it as DNSName to x509 Verify, so the leaf must both chain to the cluster CA and be valid for the advertised address (handles IP SANs too). Added a spec rejecting a cluster-signed cert served for the wrong host.

emilbon99:
@greptile-apps review

core/cmd/listener/listener.go
```diff
@@ -0,0 +1,254 @@
...
+// caller gates it on the insecure flag.
+func (cs Configs) ValidateAdvertiseSource() error {
+	v := validate.New("listeners")
+	if len(cs) == 0 {
+		return v.Error()
+	}
+	v.Ternary(
+		"advertise",
+		cs.advertised().Cert.Source == tailscale.SourceType,
+		"advertised listener cannot use the tailscale source; peers verify certificates against the cluster CA",
```
pjdotson: should use "Tailscale" and "Core"

core/cmd/listener/listener.go
```diff
@@ -0,0 +1,254 @@
...
+func Parse(fc cert.FactoryConfig) (Configs, error) {
+	switch v := viper.Get(FlagListen).(type) {
+	case string:
+		return Configs{scalar(v, fc)}, nil
+	case nil:
+		return Configs{scalar(viper.GetString(FlagListen), fc)}, nil
+	case []any:
+		return parseList(v)
+	default:
+		return nil, errors.Newf("[listener] - invalid listen configuration of type %T", v)
```
pjdotson: fix these bad logs

core/pkg/security/cert/source.go
```diff
@@ -0,0 +1,67 @@
...
+	FS xfs.FS
+	// Cert and Key are the PEM paths served by a file source.
+	Cert string
+	Key  string
+	// Address is the listener address. The auto source derives its SANs from it.
+	Address address.Address
+	// CertsDir and the CA paths give the auto source access to the built-in CA.
+	CertsDir   string
+	CAKeyPath  string
+	CACertPath string
```
pjdotson: might make sense to just have `KeyPath` and `CertPath` and let `cmd` figure those out from the provided `CertsDir`

core/pkg/security/cert/file/file.go
```diff
@@ -0,0 +1,116 @@
...
+
+var _ cert.SourceFactory = Factory{}
+
+// Type implements cert.SourceFactory.
+func (Factory) Type() string { return SourceType }
+
+// NewSource implements cert.SourceFactory.
+func (Factory) NewSource(cfg cert.SourceConfig) (cert.Source, error) {
+	if cfg.Cert == "" || cfg.Key == "" {
+		return nil, errors.Wrap(validate.ErrValidation, "[cert] - file source requires both a cert and a key path")
```
pjdotson: bad log

core/pkg/security/cert/file/file.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	return &c, nil
+}
+
+func (s *source) readAll(path string) ([]byte, error) {
+	f, err := s.fs.Open(path, os.O_RDONLY)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = f.Close() }()
```
pjdotson: should either attach or log this error

core/pkg/security/cert/tailscale/tailscale.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	return &local.Client{}
+}
+
+// NewSource implements cert.SourceFactory.
+func (f Factory) NewSource(cfg cert.SourceConfig) (cert.Source, error) {
+	if cfg.Cert != "" || cfg.Key != "" {
+		return nil, errors.Wrap(
+			validate.ErrValidation,
+			"tailscale certificate source must not set cert or key",
```
pjdotson: all tailscale logs should probably capitalize Tailscale

core/pkg/security/cert/factory.go
```diff
@@ -161,48 +162,60 @@ func (c *Factory) CreateNodePairIfMissing() error {
...
 	if err = c.writePEM(c.NodeKeyPath, keyP, false); err != nil {
 		return err
 	}
+	return c.writePEM(c.NodeCertPath, xpem.FromCertBytes(b) /* multi */, false)
+}
+
+// CreateNodeTLS signs an in-memory node certificate for the factory's hosts without
+// touching the filesystem. Used by the auto certificate source to give a listener a
+// self-signed identity derived from its own address.
+func (c *Factory) CreateNodeTLS() (*tls.Certificate, error) {
```
pjdotson: 1) receiver name should be `f`, 2) should this live in `cert` or `auto` package?

core/pkg/security/secure.go
```diff
@@ -68,10 +68,37 @@ var defaultCipherSuites = []uint16{
...
+	return p.baseTLSConfig(src.GetCertificate)
+}
+
+// NodeClientConfig implements TLSProvider.
+func (p *secureProvider) NodeClientConfig() *tls.Config {
+	return p.baseTLSConfig(p.getNodeCert)
+}
+
+// VerifyClusterCert implements TLSProvider.
+func (p *secureProvider) VerifyClusterCert(src cert.Source, host string) error {
```
pjdotson: should use Core instead of cluster everywhere this appears, are only verifying for one node

core/pkg/server/server.go
```diff
@@ -114,26 +128,39 @@ func Serve(cfgs ...Config) (*Server, error) {
...
-		if *s.Security.Insecure {
-			return s.serveInsecure(sCtx, mux)
+	opened := make([]net.Listener, 0, len(s.Listeners))
+	for _, l := range s.Listeners {
+		lis, err := net.Listen("tcp", l.Address.PortString())
+		if err != nil {
+			// Closing the opened listeners unblocks their serve goroutines; cancel then
+			// tears down the signal context so a partial bind leaves nothing running.
+			for _, o := range opened {
+				_ = o.Close()
```
pjdotson: should probably attach these as secondary errors

core/pkg/server/simple_http.go
```diff
@@ -55,19 +53,17 @@ func (h *SimpleHTTPBranch) Init(BranchContext) {
...
-	return <-h.stopErr
+	return nil
 }
 
 // Stop implements Branch.
 func (h *SimpleHTTPBranch) Stop() {
-	if h.server == nil {
-		return
+	if h.server != nil {
+		_ = h.server.Shutdown(context.TODO())
```
pjdotson: should probably log this error

pjdotson (changes requested):
Generally looks good. You should make sure to run `claude /security-review` locally in case it spots anything.

emilbon99:
@greptile re-review

pjdotson (approved):

PR #2603, opened 2026-07-14
G. SY-4488: Task Identity UUID Keys with Rack as Field

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

First of three stacked PRs implementing RFC 0045 (task autosave with deploy-on-start).

Task keys change from rack-scoped uint64s to UUIDs minted at creation, and the rack becomes a mutable row field instead of being packed into the key. This lets a task exist as a draft before it is bound to a rack (rack 0) and lets it move between racks without changing identity, which the deploy-on-start flow in the next PR depends on.

Changes span the Oracle task schema and hand-written migrations, the Go core and services, and the TS, Python, and C++ clients and driver, keeping the wire format consistent across all of them.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR migrates task identifiers from composite `uint64` keys (high 32 bits = rack, low 32 bits = local counter) to randomly-minted `uuid.UUID` values, with rack becoming a mutable field on the task record instead of being packed into the key. The change spans every layer — Oracle schema, Go server and migrations, C++ client and driver, TypeScript and Python clients — and is the foundational step for the deploy-on-start RFC that lets tasks exist as drafts before being bound to a rack.

- **Schema & migration**: `v56_task_uuid_key` re-keys all existing tasks atomically (statuses, ontology resources, and relationships rewritten in the same transaction), stages a legacy→UUID mapping in KV, and `v56_task_tab_uuid_keys` consumes that map to rewrite panel view-tab `taskKey` args; `Rack.TaskCounter` is removed and the mutex protecting it is dropped from the rack service.
- **Signal wire format**: the `sy_task_set` channel now emits JSON payloads (key + rack + metadata) and `sy_task_delete` emits raw UUID bytes; the C++ driver's `process_task_set`/`process_task_delete` are updated accordingly, foreign-rack filtering now uses the payload's `rack` field rather than extracting bits from the key.
- **Config hashing**: a new `ConfigHash` field (xxhash64 over canonical JSON) is computed server-side on every write and included in task status so drivers can detect config drift without re-fetching the full config.

Safe to merge with a second look at driver.go's handleTaskChange else branch and the remote-retrieve path in processCommand.

The migration chain is multi-step and complex but each step is well-tested and runs atomically. The signal wire-format changes are consistent between the Go publisher and C++ consumer. Two concerns remain: handleTaskChange's else branch calls d.delete for SET events targeting foreign-rack tasks, relying on d.delete being idempotent for unknown keys; and processCommand performs a remote task retrieve for every command to an unknown key, adding latency under bursty multi-driver scenarios.

core/pkg/service/driver/driver.go (handleTaskChange else branch and processCommand remote retrieve) and driver/task/manager.cpp (process_task_cmd command-drop behaviour).

| Filename | Overview |
|----------|----------|
| core/pkg/service/task/migrate.go | New hand-written migration that re-keys every task from legacy uint64 to UUID, rewrites statuses, ontology resources, and relationships, and stages the mapping in KV for the panel migration to consume. Logic is complex but tested. |
| core/pkg/service/driver/driver.go | handleTaskChange now calls d.delete for all events that are not (VariantSet AND own rack), including SET events for brand-new tasks on foreign racks. processCommand adds a remote retrieve to avoid false-positive warnings, adding latency for every command targeting a foreign-rack task. |
| driver/task/manager.cpp | process_task_set now parses JSON strings (key+rack) instead of raw uint64 values; process_task_delete uses series.uuids() matching the server's UUIDT wire format; process_task_cmd no longer creates spurious empty entries for unknown tasks. |
| core/pkg/service/layer.go | Task signal publisher now emits JSON for set events and raw UUID bytes for delete events; Panel service is correctly moved after the task migration runs so the staging KV map is populated before the panel table migration reads it. |
| core/pkg/service/task/writer.go | Key generation simplified to uuid.New() when Key is nil-UUID; ConfigHash is computed before MergeExisting so snapshots correctly preserve the existing hash when config is retained; rack.Writer dependency removed cleanly. |
| core/pkg/service/panel/migrate.go | convertTaskTab and MigrateTaskTabKeys correctly consume the staging KV map to rewrite legacy uint64 task keys in panel tabs to UUIDs, then drain the staging entries. |
| core/pkg/service/task/hash.go | New config hashing via xxhash64 over canonical JSON serialization; encoding/json map key sorting ensures equal configs hash identically regardless of insertion order; NaN/Inf configs correctly fail at hash time. |
| client/ts/src/task/types.gen.ts | Key type changed from string/number/bigint union to z.uuid(); statusDetailsZ task field no longer has a default (required); payloadZ key defaults to uuid.create factory called per-parse; rack and configHash fields added. |
| client/py/synnax/task/client.py | Task class gains rack and config_hash attributes; maybe_assign_def_rack now checks pld.rack==0 instead of pld.key==0; UUID auto-generation via uuid4() in __init__ and Client.create. |
| schemas/synnax/task.oracle | Key changed from uint64 to uuid alias; rack and config_hash fields added to Task and StatusDetails; task key default changed to `create` (UUID factory). |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant DB as KV / gorp DB
    participant TMig as Task Migration (v56_task_uuid_key)
    participant PMig as Panel Migration (v56_task_tab_uuid_keys)
    participant Driver as C++ Driver
    participant Server as Go Server (layer.go)

    Note over DB,Server: OpenLayer service open order

    Server->>DB: OpenService(Task) runs v56_task_uuid_key
    TMig->>DB: Read all v56.Tasks (uint64 keys)
    TMig->>DB: Write Tasks with new uuid.UUID keys + rack field
    TMig->>DB: Delete old v56.Task entries
    TMig->>DB: Rewrite statuses / ontology resources / relationships
    TMig->>DB: Stage mapping sy_task_legacy_key to UUID

    Server->>Server: PublishFromGorp(Task.Observe(), JSONT/UUIDT)

    Server->>DB: OpenService(Panel) runs v56_task_tab_uuid_keys
    PMig->>DB: Read staging map
    PMig->>DB: Rewrite panel view tab args taskKey uint64 to UUID
    PMig->>DB: Delete staging entries

    Note over Server,Driver: Runtime task lifecycle signals

    Server->>Driver: sy_task_set JSON key rack metadata
    Driver->>Driver: process_task_set parse rack skip if foreign
    Driver->>Server: retrieve task by UUID key
    Driver->>Driver: configure(task)

    Server->>Driver: sy_task_cmd JSON task UUID type key args
    Driver->>Driver: process_task_cmd lookup in entries by UUID

    Server->>Driver: sy_task_delete raw UUID bytes UUIDT
    Driver->>Driver: process_task_delete series.uuids skip if not in entries
```

1. `driver/task/manager.cpp`, line 202-222 ([link](https://github.com/synnaxlabs/synnax/blob/5fbb7c4a84eb0a2286952fa313610d77b43ef1e2/driver/task/manager.cpp#L202-L222)) 

   <a href="#">P2</a> **Commands silently dropped during task configuration window**

   The new path checks `entries.find(cmd.task)` under lock, then falls through to a remote `rack.tasks.retrieve` outside the lock. If a CONFIGURE op is processed in that gap (task gets added to `entries` and started), the command is still dropped and the warning is never logged — the driver misses the command for a task that is now live.

   The old code queued the command into an empty entry regardless, which was also broken (the entry had no task object to dispatch to). Given that both approaches share this narrow race, it may be acceptable; but the user-visible consequence is different — silently dropped command vs. queued-but-never-executed command — so worth noting in a comment.

</details>

<sub>Reviews (1): Last reviewed commit: ["Merge branch &#39;sy-4484-autosave-task-conf..."](https://github.com/synnaxlabs/synnax/commit/5fbb7c4a84eb0a2286952fa313610d77b43ef1e2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=46763262)</sub>


emilbon99:
@greptile-apps review

core/pkg/service/driver/driver.go
```diff
@@ -215,12 +218,10 @@ func (d *Driver) handleTaskChange(
...
-			} else {
-				d.delete(ch.Key)
-			}
+		if ch.Variant == change.VariantSet && ch.Value.Rack == d.rack.Key {
+			d.configure(ctx, ch.Value)
+		} else {
+			d.delete(ch.Key)
 		}
 	}
 }
```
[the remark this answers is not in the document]
emilbon99: Confirmed: `delete` guards on `d.mu.tasks[key]` and returns early for unknown keys, so a SET for a new foreign-rack task is a no-op (no Stop call, no log). The branch also goes away in #2604, where `handleTaskChange` ignores sets entirely and only handles deletes.

core/pkg/service/driver/driver.go
```diff
@@ -168,17 +168,20 @@ func (d *Driver) processCommand(ctx context.Context, frame framer.Frame) {
...
+					Entry(&tsk).
+					Exec(ctx, nil); err == nil && tsk.Rack == d.rack.Key {
+					d.cfg.L.Warn(
+						"received command for unknown task",
+						zap.Stringer("task", cmd.Task),
+					)
+				}
 				continue
 			}
 			sCtx, cancel := signal.WithTimeout(
```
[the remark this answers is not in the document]
emilbon99: This is the embedded driver running inside the core process: `Task.NewRetrieve` is a local gorp/KV read, not a network call. It also only fires for commands targeting unknown tasks, which are rare operator-initiated events. The retrieve replaced the rack-from-key decode that UUID keys removed; caching rack membership instead would add state that goes stale on rack moves for no measurable win.

core/pkg/service/layer.go
```diff
@@ -506,10 +497,29 @@ func OpenLayer(ctx context.Context, cfgs ...LayerConfig) (l *Layer, err error) {
 	if closer, err := signals.PublishFromGorp(
 		ctx,
 		l.Signals,
-		signals.GorpPublisherConfigPureNumeric(l.Task.Observe(), telem.Uint64T),
+		signals.GorpPublisherConfig[task.Key, task.Task]{
```
pjdotson: Is there a reason the task signals can't be inside the task service? I don't think channel or framer depend on it

core/pkg/service/rack/codec.gen.go
```diff
@@ -21,7 +21,6 @@ import (
 func (rv Rack) EncodeOrc(w *orc.Writer) error {
 	w.Uint32(uint32(rv.Key))
 	w.String(rv.Name)
-	w.Uint32(uint32(rv.TaskCounter))
```
pjdotson, line 43: you will need to generate new rack version if codec changes

schemas/synnax/rack.oracle
```diff
@@ -16,14 +16,15 @@ import "schemas/synnax/status"
...
 
 Key uint32 {
-    @doc value  """
-        is a composite identifier for a rack. The high 16 bits contain the node key, and
-        the low 16 bits contain the local sequential key. Racks are leased to specific
-        nodes because task configuration signals are passed through gossip operations,
-        which can take 15s+ to propagate through a large cluster. This structure
-        minimizes hops and configuration latency.
+    @doc value   """
+        is a composite identifier for a rack. The high 16 bits contain the core node
```
pjdotson: last doc was better no need to repeat `core node`

console/src/feature/arc/editor/TaskControls.tsx
```diff
@@ -23,14 +23,17 @@ const INITIAL_RACK_QUERY: rack.RetrieveParams = { integration: "arc" };
...
+  );
   const taskKeyDefined = primitive.isNonZero(taskKey);
   const [selectedRack, setSelectedRack] = useState<rack.Key | undefined>();
   const [expanded, setExpanded] = useState(false);
 
   useEffect(() => {
-    if (taskKeyDefined) setSelectedRack(task.rackKey(taskKey));
-  }, [taskKey, taskKeyDefined]);
+    if (taskKeyDefined && primitive.isNonZero(taskRack)) setSelectedRack(taskRack);
+  }, [taskKeyDefined, taskRack]);
```
pjdotson: why do we still need this tracking? Feels like this change means we should be able to remove some `useEffect`s from the Console

console/src/feature/ethercat/task/types.ts
```diff
@@ -210,8 +212,10 @@ export interface WritePayload extends task.Payload<WriteSchemas> {}
 
 export const ZERO_WRITE_PAYLOAD = {
   key: "",
+  rack: 0,
   name: "EtherCAT Write Task",
   config: ZERO_WRITE_CONFIG,
+  configHash: "",
```
pjdotson, line 218: doesn't this mean the `config` and `configHash` don't agree?

driver/opc/scan_task.cpp
```diff
@@ -68,7 +68,7 @@ bool Scanner::exec(
 }
 
 x::errors::Error Scanner::check_device_health(synnax::device::Device &dev) {
-    const auto rack_key = synnax::task::rack_key_from_task_key(this->task.key);
+    const auto rack_key = this->task.rack;
```
pjdotson, line 70: It is a little odd that all of these get the rack from the task - shouldn't we already know the rack key because everything on the driver shares the same rack?

pjdotson (changes requested):
Looked through like 1/2 of this, see comments and also fix merge conflicts and CI

console/src/feature/ethercat/task/types.ts
emilbon99: The core is the single source of hash generation, so this is treated less as a 'mismatch'  and more as a sentinel value meaning 'the task is in draft state and the config has has not been assigned'

driver/opc/scan_task.cpp
emilbon99: I feel like this is out of scope and we should just leave as is for now right?

schemas/synnax/task.oracle
```diff
@@ -57,41 +61,55 @@ Task struct<
...
+    config      Config              {
         @doc value """
             is task-specific configuration stored as JSON. Structure varies by
             task type.
         """
     }
-    internal bool = false {
+    config_hash string = "" {
+        @doc value """
+            is the server-assigned hash of config, rewritten on every write and
```
pjdotson: Core-assigned

schemas/synnax/rack.oracle
```diff
@@ -16,28 +16,27 @@ import "schemas/synnax/status"
...
-    @doc value  """
+    @go version  2
+    @doc value   """
         is a composite identifier for a rack. The high 16 bits contain the node key, and
-        the low 16 bits contain the local sequential key. Racks are leased to specific
-        nodes because task configuration signals are passed through gossip operations,
-        which can take 15s+ to propagate through a large cluster. This structure
-        minimizes hops and configuration latency.
+        the low 16 bits contain the local sequential key. A rack is leased to the node
+        named in its high bits, so writes route to the node running its driver.
```
pjdotson: use Core and capitalize Driver

schemas/synnax/task.oracle
pjdotson, line 1: Does it make sense for `config_hash` to be both something assigned in the task and shown in the task status details?

pluto/src/arc/queries.spec.tsx
```diff
@@ -588,7 +591,8 @@ describe("Arc queries", () => {
             });
             expect(childrenBefore).toHaveLength(1);
             originalTaskKey = childrenBefore[0].id.key;
-            expect(task.rackKey(originalTaskKey)).toBe(rack1.key);
+            const originalTask = await client.tasks.retrieve({ key: originalTaskKey });
```
pjdotson: should be able to call this by just passing in `originalTaskKey` instead of an object

oracle/plugin/ts/types/types.go
```diff
@@ -2501,6 +2501,9 @@ func (p *Plugin) typeRefToZodSchemaType(
 	case resolution.DistinctForm:
 		return fmt.Sprintf("typeof %s%sZ", prefix, camelCase(tsName))
 
+	case resolution.AliasForm:
```
pjdotson, line 2507: add test

oracle/plugin/cpp/types/types.go
pjdotson, line 1: add test

oracle/plugin/cpp/json/json.go
pjdotson, line 1: add test

client/ts/src/arc/client.ts
```diff
@@ -96,13 +104,23 @@ const TASK_TYPE = "arc";
 
 const taskStatusDataZ = z.null().optional();
 
-const configuringStatus = (taskKey: task.Key): task.Status<typeof taskStatusDataZ> =>
+const configuringStatus = (
+  taskKey: task.Key,
+  rack: number,
```
pjdotson: should be `rack.Key`

console/src/platform/task/useStatus.ts
```diff
@@ -21,7 +21,14 @@ const defaultStatus = <StatusData extends z.ZodType>(): task.Status<
...
   description: "",
   time: TimeStamp.now(),
-  details: { task: "", running: false, cmd: "", data: {} as any },
+  details: {
+    task: "",
+    running: false,
+    cmd: "",
+    configHash: "",
+    rack: 0,
+    data: {} as any,
```
pjdotson, line 30: bad

core/pkg/service/task/versions/v2/migrate_test.go
pjdotson, line 1: this should be inside one of the `versions` package since the logic lives there

pjdotson (changes requested):

schemas/synnax/task.oracle
emilbon99: Yes, this is how the driver communicates what config hash it is currently running for the task, and whether the task needs to be redeployed.

console/src/platform/task/useStatus.ts
emilbon99: *Emiliano + Claude: drafted in our review session.*

This gets removed in a later PR: https://github.com/synnaxlabs/synnax/pull/2605/files#diff-e2f20f2acecf685239f294260f1e95a68c5725906ae948ca23c6052e364fd7e0L15-L32

pjdotson (approved):

PR #2604, opened 2026-07-14
H. SY-4488: Deploy Task Configurations on Start Across Drivers and Clients

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Second of three stacked PRs implementing RFC 0045 (task autosave with deploy-on-start).

The separate configure-then-start lifecycle collapses into deploy-on-start: a start
command deploys the task's stored configuration, so saving a task is a plain metadata
write with no driver round trip. The C++ driver and the Go embedded driver both fetch
the stored task when a start arrives and rebuild only when its config hash differs from
the hash the live instance was built from. The Python client's configure path becomes a
plain save.

The Core assigns the config hash on every task write; clients never hash. A client
detects drift by comparing the stored hash against the hash in the task's status. A hash
stamped on a start command can force a redeploy but never confirm one, and the fetch it
triggers runs on the driver's worker pool, off the streamer loop.

The drivers now share one status model. Each live task keeps the last status it sent and
answers no-work commands from it. Starts are idempotent: a running task acks instead of
restarting. A start whose fetch fails answers the command with an error status, except
when the task was deleted. Delete teardown is silent, so a deleted task's status is not
recreated.

Integration tests cover the deploy-on-start flow end to end.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR collapses the previous configure-then-start lifecycle into a single deploy-on-start model: a start command now deploys the stored config if the running hash differs from the stored hash, and saving a task is a plain metadata write with no driver round-trip. Server-side hashing (`xxhash64` of the JSON-encoded config) is computed on every write, and both drivers compare that hash against the hash the live instance was built from to decide whether a redeploy is needed.

- **C++ external driver**: `process_start` performs an O(1) hash check; redeploys by fetching the full task row and queuing a `CONFIGURE` op only when the hash differs or no instance exists. New entry lifecycle helpers (`entry_for`, `remove`, `clear_deploy`) and a `RELEASE` op type handle rack-move teardown cleanly.
- **Go embedded driver**: `handleStart` retrieves the stored task, compares `inst.hash` against `tsk.ConfigHash`, and rebuilds via `configure()` only when needed; `AckCurrentStatus` re-stamps the live status with the command key for idempotent start acknowledgements.
- **Arc and PagerDuty factories**: `ConfigureTask` now carries `cmdKey`; boot-time config failures for non-auto-start tasks are logged silently (no command is waiting), while command-driven failures write a status that acknowledges the key.

- This PR is safe to merge. The deploy-on-start lifecycle change is well-contained, the lock ordering in the C++ driver is consistent throughout (Manager::mu → Entry::mu), the Go driver's hash-gated redeploy is race-free under the existing RWMutex, and the auto-start acknowledgment flows correctly thread the original command key through to the final status write.
- The core logic — hash comparison, deploy gating, status acknowledgment — is covered by new integration tests (C++ manager_test, Go driver_test) and the arc/pagerduty unit tests. The two findings are an inconsistent log level for expected no-handler conditions and a missing default message in a bootstrapped status; neither affects correctness or reliability in production.
- No files require special attention.

| Filename | Overview |
|----------|----------|
| driver/task/manager.cpp | Implements deploy-on-start for C++ driver: process_start checks deployed hash vs wanted hash, falls through to retrieve+CONFIGURE only when needed; locking hierarchy (Manager::mu → Entry::mu) is consistent throughout. |
| driver/task/task.h | Entry struct adds deployed_hash, op_cmd, op_config_hash fields; Factory::configure_task gains cmd_key parameter propagated throughout. NO_COMMAND sentinel and Entry::relevant() helper are clean additions. |
| core/pkg/service/driver/driver.go | Go embedded driver gains handleStart with hash-gated redeploy. instance struct wraps Task + hash for O(1) hash comparison. configure, release, delete, exec helpers correctly manage the mu lock lifecycle around stopping and starting tasks. |
| core/pkg/service/task/writer.go | hashConfig called on every Create; resolveStatus now distinguishes new vs existing tasks, giving new tasks a "has not been deployed" disabled status instead of a warning. Logic is correct for snapshot and non-snapshot paths. |
| core/pkg/service/driver/status.go | New AckCurrentStatus helper reads the live status, re-stamps config hash, rack, and cmd key, then rewrites it. Correctly handles ErrNotFound by seeding a success variant; preserves .Data from previous status. |
| core/pkg/service/arc/task/factory.go | ConfigureTask now accepts cmdKey; error statuses include cmd key for acknowledgment. Boot-time config failures for non-auto-start tasks are logged only (correct: no command is waiting). Auto-start flow delegates status ack to the subsequent Exec call. |
| core/pkg/service/pagerduty/factory.go | Mirrors arc/task/factory.go changes. Auto-start triggers pdTask.start(ctx, driver.NoCommand) from configure, then d.exec re-calls Exec with the real cmdKey which hits ackCurrent — correctly acknowledging the original command. |
| client/ts/src/task/client.ts | drifted() now compares two server-assigned hashes (no client-side canonicalization). stampConfigHash populates configHash from cached task row so the C++ driver can skip unnecessary redeploys. Command flow and cache invalidation logic are sound. |
| client/py/synnax/task/client.py | configure() is now a plain save (no driver round-trip, timeout parameter retained only for backward compat). execute_command includes config_hash from self.config_hash so the C++ driver can skip redeploys when the hash matches. |
| core/pkg/service/driver/driver_test.go | Comprehensive tests cover deploy-on-start, hash-gated rebuild, metadata-only rename (no rebuild), rack filtering, rack-move stop, and silent delete. Consistently uses atomic counters to track configure/stop invocations. |
| driver/task/manager_test.cpp | New tests cover DeployOnStart, NoConfigureOnCreate, and RackMoveStopsLiveInstanceSilently. expect_no_driver_status helper correctly filters placeholder statuses by rack key to distinguish driver-reported statuses from core-written ones. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client
    participant Core as Synnax Core
    participant Driver as Driver (C++ / Go)

    Note over Client,Driver: Task Save (no driver round-trip)
    Client->>Core: Create/Save task (config)
    Core->>Core: hashConfig(config) → config_hash
    Core-->>Client: "Task{config_hash}"

    Note over Client,Driver: Deploy-on-Start
    Client->>Core: "start command{task, config_hash}"
    Core->>Driver: task_cmd channel (start)
    Driver->>Driver: lookup deployed_hash
    alt "deployed_hash == wanted_hash"
        Driver->>Driver: exec start on live instance
        Driver->>Core: "status{cmd, running=true, config_hash}"
    else hash mismatch or no instance
        Driver->>Core: retrieve task row (full config)
        Driver->>Driver: stop old instance (if any)
        Driver->>Driver: ConfigureTask → new instance
        Driver->>Driver: exec start on new instance
        Driver->>Core: "status{cmd, running=true, config_hash}"
    end
    Core-->>Client: status (via sy_status_set)

    Note over Client,Driver: Drift Detection
    Client->>Core: retrieve task (includeStatus)
    Core-->>Client: "Task{config_hash, status.details.config_hash}"
    Client->>Client: "drifted = details.config_hash ≠ config_hash"
```

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;sy-4488-autosave-task-conf..."](https://github.com/synnaxlabs/synnax/commit/26dd0537581de6ecbca755a8e0f0c3a0397feb64) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44287105)</sub>


client/ts/src/task/drift.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Hashes a task config into the shared cross-language form: xxhash64 of the
+ * canonical JSON string, as 16 lowercase hex characters. Matches the hashes the
+ * server and drivers stamp into status details. Keys are snake-cased first,
+ * since that is how the wire codec stores the config.
+ * @param config - The task config to hash.
+ * @returns The 16-character hex hash.
+ */
+export const hashConfig = (config: unknown): string =>
+  hash.xxHash64(json.canonicalString(caseconv.camelToSnake(config)));
```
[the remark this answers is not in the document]
emilbon99: Client-side hashing was removed entirely in a later revision: core now assigns config_hash on every write, and drifted() compares two server-assigned values. No client canonicalization exists anymore, so the NodeId corruption path is gone.

driver/task/manager.cpp
```diff
@@ -219,6 +191,29 @@ void Manager::process_task_cmd(const x::telem::Series &series) {
...
+    std::lock_guard<std::mutex> lock(this->mu);
+    if (!this->entries[cmd.task]) this->entries[cmd.task] = std::make_shared<Entry>();
+    if (this->deploys_->hash(cmd.task) == hash) {
+        VLOG(1) << "queuing start for live task " << tsk;
+        this->op_queue.push_back(Op{Op::Type::COMMAND, cmd.task, {}, cmd});
+    } else {
+        VLOG(1) << "queuing deploy and start for task " << tsk;
+        this->op_queue.push_back(Op{Op::Type::CONFIGURE, tsk.key, tsk, cmd});
+    }
+    this->cv.notify_one();
```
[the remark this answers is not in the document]
emilbon99: Fixed in a later revision: the driver no longer hashes configs (the hash arrives on the stored row), and DeployState::hash returns an optional whose nullopt (never deployed) never compares equal, so a fresh task always takes the configure path.

client/ts/src/task/drift.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      "2de66015b3bdded8",
+    );
+  });
+
+  it("should hash camelCase keys as their snake_case wire form", () => {
+    expect(hashConfig({ dataSaving: true, sampleRate: 5.5 })).toEqual(
+      hashConfig({ data_saving: true, sample_rate: 5.5 }),
+    );
+  });
+});
```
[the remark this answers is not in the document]
emilbon99: Client hashing was removed in a later revision; drifted() is a plain field compare of server-assigned hashes, so there is no canonicalization left to golden-test.

emilbon99:
@greptile-apps review

emilbon99:
@greptile re-review

driver/common/status.h
```diff
@@ -190,32 +197,42 @@ struct StatusHandler {
...
+        // Ack the start so a waiting caller resolves with the failure, not a timeout.
+        status.details.cmd = cmd_key;
+        if (boot && !res.first.auto_start)
+            LOG(WARNING) << "[driver] failed to configure task " << task.name << ": "
+                         << res.second;
+        else
+            ctx->set_status(status);
+        return {std::move(res.first.task), true};
     }
     if (res.first.auto_start) {
```
sy-nico, line 240: When `auto_start` is true and a start command drove the deploy, the task starts twice. `handle_config_err` execs its own start here, then `manager.cpp:487` execs `op.cmd` on the same instance. C++ `read_task::start()` calls `stop("", false)` first, so the device is opened, closed, and reopened inside one start.

Go does not do this: `arc/task/task.go` guards `start()` on `isRunning()` and acks the command instead of restarting.

Suggest skipping the self-start when a command is driving the deploy, since the manager will exec it:

    if (res.first.auto_start && cmd_key == driver::task::NO_COMMAND) {
        ...
    }

driver/task/manager.cpp
```diff
@@ -202,21 +222,80 @@ void Manager::process_task_cmd(const x::telem::Series &series) {
...
+            const auto entry = it->second;
+            bool ours;
+            std::string deployed, wanted;
+            {
+                std::lock_guard entry_lock{entry->mu};
+                ours = entry->row.rack == this->rack.key;
+                deployed = entry->deployed_hash;
+                // A command carrying a hash states what the sender wants running
+                // without waiting for a set event to land.
+                wanted = cmd.config_hash.empty() ? entry->row.config_hash
```
sy-nico: `cmd.config_hash` is trusted to confirm a match, not just to detect a newer revision. A sender that stamps a stale hash equal to `deployed_hash` takes the fast path and starts the old config. `client.ts:692` fills it from the cached row, and `Task.executeCommand` uses `this.configHash` frozen at fetch time, so a stale hash is reachable.

Go cannot do this: `handleStart` ignores `cmd.ConfigHash` and always compares against the freshly retrieved `tsk.ConfigHash`.

Can we gate this so the sender's hash can only force a redeploy, never skip one? That keeps the documented win of reusing a live instance without waiting for a set event:

    wanted = entry->row.config_hash;
    if (!cmd.config_hash.empty() && cmd.config_hash != wanted) wanted = cmd.config_hash;

driver/task/manager.cpp
```diff
@@ -202,21 +222,80 @@ void Manager::process_task_cmd(const x::telem::Series &series) {
...
             }
         }
-        // Unknown instance: the command may target another rack's task, so only
-        // warn when the task actually belongs to this rack.
-        auto [tsk, err] = this->rack.tasks.retrieve(cmd.task);
-        if (err || tsk.rack != this->rack.key) continue;
-        LOG(WARNING) << "received command for unconfigured task " << tsk;
     }
+    // A deploy needs the config, which set events omit.
+    auto [tsk, err] = this->rack.tasks.retrieve(cmd.task);
```
sy-nico: This is a blocking gRPC call on the streamer read loop, which also dispatches `control_state` frames. A slow Core stalls control authority updates behind a task retrieve. Can this move to a worker?

core/pkg/service/driver/status.go
```diff
@@ -0,0 +1,51 @@
...
+	t task.Task,
+	cmdKey string,
+	running bool,
+) error {
+	var stat task.Status
+	err := status.NewRetrieve[task.StatusDetails](svc).
+		Where(status.MatchKeys[task.StatusDetails](t.OntologyID().String())).
+		Entry(&stat).
+		Exec(ctx, nil)
+	if errors.Is(err, query.ErrNotFound) {
```
sy-nico: There might be an issue here. On `ErrNotFound` the variant is seeded but `Message` and `Description` stay zero, so the status that answers the command has an empty message. Can you add coverage for the not-found path in `status_test.go`? Both existing specs go through a status row that already exists.

core/pkg/service/driver/driver.go
```diff
@@ -184,43 +197,132 @@ func (d *Driver) processCommand(ctx context.Context, frame framer.Frame) {
...
-			sCtx.Go(
-				func(ctx context.Context) error { return t.Exec(ctx, cmd) },
-				signal.RecoverWithErrOnPanic(),
+		}
+		return
+	}
+	d.mu.RLock()
+	inst, ok := d.mu.instances[cmd.Task]
+	d.mu.RUnlock()
+	if !ok || inst.hash != tsk.ConfigHash {
```
sy-nico, line 246: A rename does not change the config hash, so this reuses the instance built from the old row. `inst.task` still holds the old name, and every status it writes from then on (including `AckCurrentStatus`) reports it. Worth refreshing the instance's task snapshot on the match path?

driver/task/manager.cpp
```diff
@@ -349,62 +443,122 @@ void Manager::monitor_loop() {
...
             break;
         }
         case Op::Type::REMOVE: {
-            if (entry->task == nullptr) return;
-            LOG(INFO) << "deleting task " << entry->task->name();
-            entry->task->stop(false);
-            entry->task = nullptr;
+            if (entry->instance != nullptr) {
+                LOG(INFO) << "deleting task " << entry->instance->name();
+                entry->instance->stop(false);
```
sy-nico: This is a pre-existing issue (on `rc` and `main`). `stop(false)` means `propagate_state=true`, so a deleted task writes a terminal status after its row is gone. The Go driver deliberately avoids this (`driver.go`, `d.release(key, false)`, "a terminal status written afterwards would resurrect it"). Since this PR is where the two drivers' teardown paths get spelled out side by side, worth aligning or filing a new issue.

sy-nico (changes requested):
Design looks right and the test coverage is strong. But two blocking issues to resolve (or comment why it's not an issue):

- `driver/common/status.h`: `auto_start` tasks start twice on a command-driven deploy, so the device gets closed and reopened inside one start.
- `driver/task/manager.cpp`: the C++ driver trusts a sender's config hash to skip a redeploy, which Go does not. A stale hash starts the old config.

Also, the PR description is stale and says the TS client produces the same hash as the drivers. Only the Core hashes now, and `drifted()` just compares two server-assigned values. The RFC 0045 link points at multi-listener certs.

driver/common/status.h
emilbon99: I implemented a more robust fix that just makes sure a task ins't started twice i.e. has the same guard as go.

driver/task/manager.cpp
emilbon99: Implemented.

core/pkg/service/driver/status.go
emilbon99: Implemented.

core/pkg/service/driver/driver.go
emilbon99: Discussed verbally.

sy-nico (approved):

PR #2605, opened 2026-07-14
I. SY-4488: Autosave Task Configurations in the Console with Deploy-on-Start

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Third of three stacked PRs implementing RFC 0045 (task autosave with deploy-on-start).

Console task forms now edit real task rows instead of holding unsaved view state. Creating a task from the palette, tree, or a device context menu writes an untyped draft row (rack 0 unless a device or rack is given) and opens its resource tab, so a work-in-progress config survives closing the tab or the app. The play button saves the config and issues a start command, renaming a running task is a metadata-only save that no longer stops it or asks for confirmation, and a drift badge with a redeploy button appears when the saved config diverges from the running task.

Task tabs move to a single panel registry keyed by the task's ontology type, replacing the per-type view tabs. Pluto's task queries gain the autosave write path and a metadata listener. The console and pluto spec suites were converted to the draft-row semantics and run green against a live core (1977 tests).

## Basic Readiness

<h3>Greptile Summary</h3>

This PR (third in a three-PR stack for RFC 0045) converts console task forms from hold-in-view-state semantics to always-editing real task rows. Creating a task writes a draft row, opens it as a resource tab, and autosaves on every keystroke; the play button runs a validate-configure-save-start pipeline; renaming is now a metadata-only save; and a drift badge with a redeploy button surfaces when the running instance's config hash differs from the stored row's hash.

- **Backend (`writer.go`)**: `healStatus` is promoted to return the task's effective status, allowing `Create` to populate `t.Status` on the response so the console sees the current live status immediately after save.
- **Frontend (pluto/task, console/platform/task)**: `rackKey` merges into `rack`, `configHash` is added to form state, `mountListeners` narrows to metadata-only fields to prevent clobbering in-flight autosaves, and `useStatus` drops its synthetic fallback because the Flux form always suspends until the server provides a real status.
- **Panel migration (`migrate.go`)**: Legacy per-type view tabs are converted to unified `TabResource` entries under a single panel registry keyed by `task.TYPE_ONTOLOGY_ID.type`.

- This PR is safe to merge. The autosave/deploy pipeline is well-structured, the Go writer change is covered by new test cases, and the migration path from view tabs to resource tabs is fully tested.
- All findings are non-blocking style or UX concerns: discarding `deployConfigZ`'s transformed output (innocuous while schemas are pure-validation) and stale field error indicators between failed deploy attempts. The core logic — draft creation, autosave, drift detection, status healing, and panel migration — is correct and backed by 1977 passing tests against a live core.
- No files require special attention, though reviewers should verify that each task type's `getInitialValues` calls `configSchema.parse(config)` for import validation to remain strict.

| Filename | Overview |
|----------|----------|
| pluto/src/task/queries.ts | Core form-data contract change: replaces rackKey with rack, adds configHash to form state, narrows mountListeners to metadata-only updates, and conditions getCached on a non-null live status. The update path now only resets server-assigned fields (key, configHash) to avoid clobbering in-flight autosave edits — a deliberate trade-off that is correctly commented but creates a window where the form can show a server-normalised name before the listener fires. |
| console/src/platform/task/Form.tsx | wrapForm now returns a single FC<FormTabProps> instead of a Panel.Tab triple; the deploy pipeline (validate → onConfigure → saveAsync → executeCommand) replaces the old beforeSave/afterSave hooks. deployConfigZ.safeParse result is used only as a gate — raw config is forwarded to onConfigure — which is correct for pure-validation schemas but would silently drop any schema-level coercions. |
| core/pkg/service/task/writer.go | healStatus promoted to return the effective *Status (existing row or freshly-written placeholder), allowing Create to populate t.Status on every call. Logic is correct, error-return paths use zero values, and the new behaviour is covered by two new test cases in task_test.go. |
| console/src/platform/task/useStatus.ts | Drops the synthetic defaultStatus fallback; Form.useFieldValue now throws "Path status does not exist" when called before a status is available. This is intentional — the Flux form suspends until the server provides a real status — and the behaviour is documented by an explicit test. Callers that access .details.running unconditionally (Controls.tsx, Form.tsx::useIsRunning) are safe only while that Suspense guarantee holds. |
| console/src/platform/task/useCreate.ts | New helper; correctly strips key/status from getInitialValues before creating the draft row, then opens a resource tab. rack defaults to 0 when neither the caller nor getInitialValues supplies one. |
| console/src/platform/task/useDrifted.ts | Computes drift by comparing form's configHash/rack against the deployed instance's hash/rack via task.drifted. Guards on configHash == null so a form that hasn't loaded yet is never considered drifted. Well-tested in useDrifted.spec.ts including the "empty deployed hash" edge case. |
| console/src/feature/task/tab.tsx | Single unified TAB replaces per-type view tabs. restore creates a new task row from the tombstone's corpse payload with status:undefined so core seeds a fresh placeholder. If the corpse is absent from the cache (e.g., after a cold restart before the tombstone subscription re-populates), requireCorpse throws and the restore silently fails. |
| core/pkg/service/panel/versions/v0/migrate.go | MigrateTaskTabKeys updated to emit TabResource instead of TabView with args; convertNodeTaskTabs replaces rewriteNodeTaskKeys. The layoutType parameter is dropped cleanly and the change is fully covered by the updated migration_test.go assertions. |
| console/src/platform/task/createIngester.ts | Ingester now delegates all config validation to getInitialValues rather than calling configSchema.parse(data) directly. Invalid configs are caught only if each task type's getInitialValues uses strict parsing — the test fixture demonstrates the required pattern, but production implementations that don't call .parse() would silently create tasks with invalid configs. |
| client/ts/src/task/client.ts | drifted now guards against an empty deployed configHash (unknown deploy info does not imply drift). The optimistic status update on start/stop commands now carries the latest known configHash and rack forward instead of zeroing them, preventing a spurious drift indicator immediately after issuing a command. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Console
    participant FluxForm
    participant Server
    participant Driver

    User->>Console: Create task (palette / device menu)
    Console->>Server: tasks.create(draft, status:undefined)
    Server-->>Console: "task{key, configHash, status}"
    Console->>Console: "openTab({variant:"resource", resource:ontologyID(key)})"

    Note over FluxForm,Server: Autosave on every form change
    User->>FluxForm: Edit config field
    FluxForm->>Server: "tasks.create({key, rack, config, …}) [autoSave]"
    Server-->>FluxForm: "{key, configHash}"

    User->>Console: Click ▶ Play (deploy)
    Console->>FluxForm: deployConfigZ.safeParse(config)
    alt Blocked validation errors
        FluxForm-->>User: field error statuses
    else Passes (or warnings only)
        Console->>Server: onConfigure → resolve channels/rack
        Console->>FluxForm: form.set("config", newConfig, SKIP_AUTOSAVE)
        Console->>Server: "saveAsync() → tasks.create({…})"
        Server-->>Console: "{key, configHash}"
        Console->>Server: "tasks.executeCommand({task:key, type:"start"})"
        Server->>Driver: start command
        Driver-->>Server: "status{running, configHash, rack}"
        Server-->>FluxForm: onChange listener → set("status", …) + set("configHash", …)
        FluxForm-->>User: drift badge hidden (hashes match)
    end

    User->>Console: Rename task
    FluxForm->>Server: "tasks.create({key, name, …}) [metadata save, no stop]"
```

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;sy-4488-deploy-on-start&#39; i..."](https://github.com/synnaxlabs/synnax/commit/1288a46e1b90dda8e71ff4006e395d3b29102b82) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44287159)</sub>


console/src/platform/task/Form.tsx
```diff
@@ -116,65 +118,35 @@ export const wrapForm = <S extends task.Schemas = task.Schemas>({
...
+      handleError(async () => {
+        if (client == null) throw new DisconnectedError();
+        const { config, name } = form.value();
+        const [newConfig, newRack] = await onConfigure(client, config, name);
+        form.set("config", newConfig, SET_OPTIONS);
+        if (primitive.isNonZero(newRack)) form.set("rack", newRack, SET_OPTIONS);
+        if (!(await saveAsync())) return;
+        await client.tasks.executeCommand({ task: taskKey, type: "start" });
+      }, "Failed to start task");
+    }, [client, form, saveAsync, taskKey, handleError]);
```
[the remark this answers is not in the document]
emilbon99: Intentional omission: onConfigure is a wrapForm parameter captured from outside the component, and react-hooks/exhaustive-deps exempts outer-scope values, which is why lint does not flag it. Listing it would be noise.

console/src/platform/device/TaskContextMenuItems.tsx
```diff
@@ -10,14 +10,13 @@
...
 
-import { Panel } from "@/platform/panel";
 import { Task } from "@/platform/task";
 import { type Tree } from "@/platform/tree";
 
 export interface TaskContextMenuItemConfig {
   itemKey: string;
   label: string;
-  type: string;
+  useCreate: Task.UseCreate;
```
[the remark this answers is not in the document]
emilbon99: Hoisting the hook into the parent is not possible here: the parent renders a dynamic list of configs, so it would have to call a variable number of hooks, which itself violates the rules of hooks. A per-item child calling its injected hook is the correct pattern, and each item's useCreate identity is a module-level constant, so it cannot change between renders.

emilbon99:
@greptile-apps review

console/src/feature/task/external.tsx
```diff
@@ -71,20 +71,17 @@ export const FILE_INGESTERS: Import.FileIngesters = {
...
 export const SNAPSHOT_SERVICES: Range.SnapshotServices = {
   task: {
     icon: <Icon.Task />,
-    onClick: async ({ id: { key } }, { client, openTab }) =>
-      await retrieveAndOpenTab(client, key, openTab),
+    onClick: async ({ id }, { client, openTab }) => {
+      if (client == null) throw new DisconnectedError();
+      await client.tasks.retrieve({ key: id.key });
+      openTab({ variant: "resource", resource: id });
+    },
```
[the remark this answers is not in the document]
emilbon99: The discarded retrieve is the established convention for snapshot services: the schematic snapshot onClick does the identical retrieve-then-open, using it as an existence check so a deleted snapshot fails fast with an error toast instead of opening a dead resource tab. Keeping it for consistency.

pluto/src/flux/form.ts
sy-nico: Autosave doesn't have a debounce, so it saves on each char change. We talked about this the other day and one of the solutions could be on `enter` or clicking out of the text field.

console/src/platform/task/useStatus.ts
```diff
@@ -9,27 +9,7 @@
...
-    running: false,
-    cmd: "",
-    configHash: "",
-    rack: 0,
-    data: {} as any,
-  },
-});
-
 export const useStatus = <Schema extends z.ZodType>(ctx?: Form.ContextValue<Schema>) =>
-  Form.useFieldValue<task.Status>("status", { ctx, optional: true }) ?? defaultStatus();
```
sy-nico, line 35: Opening a task tab while disconnected might cause a crash? Removing the default status makes this return `null`, the type still says `task.Status` so nothing catches it, and `Controls` reads `taskStatus.details.running` straight off it.

console/src/platform/task/Form.tsx
```diff
@@ -72,29 +54,32 @@ export interface GetInitialValues<S extends task.Schemas = task.Schemas> {
...
+  Form: FC<{}>;
   type: z.infer<S["type"]>;
   onConfigure: OnConfigure<S["config"]>;
   schemas: S;
+  /**
+   * Validates the config when the user deploys. Failures render as field
+   * errors and block the start command; warning-variant issues render but
+   * don't block. Shape schemas stay lax so drafts persist through autosave.
+   */
+  deployConfigZ: z.ZodType;
```
sy-nico, line 74: Every task type now keeps a lax `configZ` and a strict `deployConfigZ` in sync by hand. If someone adds a constraint to one and not the other it silently never runs, and nothing catches that.

pluto/src/task/queries.ts
```diff
@@ -147,30 +128,39 @@ export const createForm = <S extends task.Schemas = task.Schemas>({
...
       client.tasks.onChange(key, (result) => {
         if (!query.isLive(result)) return;
-        resetFormValues(set, result.payload as task.Payload<S>);
+        // Metadata only: config changes come solely from this form's own
+        // saves, and resetting it would clobber in-flight autosave edits.
+        const payload = result.payload as task.Payload<S>;
+        set("name", payload.name, RESET_OPTIONS);
+        set("rack", payload.rack, RESET_OPTIONS);
+        set("snapshot", payload.snapshot, RESET_OPTIONS);
+        set("configHash", payload.configHash, RESET_OPTIONS);
```
sy-nico, line 163: This takes `configHash` from a remote save but not `config`, so a second console editing the same task shows the drift badge for a change it doesn't have. Clicking Redeploy there pushes the local config and silently overwrites the other save.

sy-nico (changes requested):

sy-nico:
@greptile, rerun.

console/src/platform/task/useStatus.ts
emilbon99: Two things:

1) since we removed optional this will throw, not return null.
2) We have a suspended query above the tree, which means if the task or status is not available, this component won't render and the user will be required to retry the request.

console/src/platform/task/Form.tsx
emilbon99: This is what we talked about this morning. These schemas stay for now, and I'm going to make a fix in the oracle code generator befroe we release.

pluto/src/task/queries.ts
emilbon99: Yes, this is somewhat of a larger problem with the task forms that I planned on deferring. Eventually we're going to need a more surgical way of updating the forms rather than sending the entire config every time. It will require a similar solution to the schematic, but I plan on implementing this in a later release.

Unfortunately we're just going to need to live with this behavior for th enear future.

sy-nico: Good plan. My concern with that solution is making sure that individual actions don't get missed and the engineer deploys a task that isn't configured the way they think it is. That's also an issue with the schematic editing. Don't know if/how you accounted for it in the schematic editing.

sy-nico (approved):

PR #2612, opened 2026-07-16
 SY-4493: Client-Side Cache and Dispatch Engine for Flux

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4493](https://linear.app/synnax/issue/SY-4493)

## Description

Implements phases 1 and 2 of RFC 0046 (included in this PR along with the implementation plan): the flux caching and undoable-dispatch machinery moves out of pluto and into the client package, so cached entities and dispatch history live with the client connection instead of the React provider tree.

- `client/ts/src/cache/`: a cache engine with keyed entity stores, lazy change streaming, epoch-based reconciliation, and a typed query cache registry. A detached mode (no streamer) backs offline and test use.
- `client/ts/src/dispatch/`: a dispatch controller providing replay, undo/redo, transactions, and coalescing for the six undoable document domains. Action classification (`kindOf`) lives next to each domain's `actions.ts`.
- Domain clients bind to the engine: each undoable domain exposes a `dispatcher`, entity domains expose a `store` read surface, and creates write through to the cache.
- Pluto's `flux/base` is reduced to a type-alias layer over the client's cache and dispatch modules; the provider composes per-domain store controllers through a `StoreComposer` seam assembled at the `Pluto.tsx` wiring site. Schematic's `preprocess` (edge segment augmentation) stays pluto-side as the one genuine visualization dependency.
- Behavior preserving: no wire changes. Console cluster records drop the new `cache` client-construction knob from persisted connection params.

Test infrastructure moved with it: pluto and console test wrappers now build one detached engine per wrapper with prebuilt controllers, and a shared `openContextMenu` console testutil hardens specs against text nodes being replaced by async permission resolution.

## Basic Readiness

- [ ] I have performed a self-review of my code.
- [x] I have added relevant, automated tests to cover the changes.
- [ ] I have updated documentation to reflect the changes.

<h3>Greptile Summary</h3>

This PR implements RFC 0046 phases 1 and 2: a `cache/` engine (keyed `Table` store, `Answers` query-space registry, lazy `Streamer`, epoch/reconciliation) and a `dispatch/` controller (local replay, undo/redo stacks, echo recognition, coalescing, transactions) are introduced in `client/ts`, and Pluto's `flux/base` is reduced to a thin alias layer on top of them. All 255 changed files are behavior-preserving — no wire changes.

[...]

PR #2620, opened 2026-07-24
A. SY-4493: Partial-Key API Retrieves and x Cache Primitives

emilbon99:
## Linear Issue

## Description

**PR A of 5** splitting the [SY-4493](https://linear.app/synnax/issue/SY-4493/robust-session-state-management-for-the-console) program (currently synnaxlabs/synnax#2612: client cache + unified<br>reads + session state + connection lifecycle) into reviewable, stacked PRs. This is the<br>foundation layer: additive, behavior-safe, and independent of the rest.

**Go — partial-key API retrieves.** Multi-key retrieves for arc, label, lineplot, log,<br>ontology, panel, ranger, schematic, table, and task now `Skip` `query.ErrNotFound`,<br>returning the found subset instead of erroring. This lets callers bulk existence-check<br>keys (used later by client reconciliation and the console session sweep).

**x — cache/connection primitives.** Extracts a shared `state` module into `x`, adds<br>max-interval and jitter controls to `breaker`, and adds destructor composition helpers.<br>All additive; consumed by later PRs in the stack.

Stack: A → rc, B → A, C → B, D/E → C.

## Basic Readiness

### Greptile Summary

This PR introduces two independent layers of changes: Go API services gain an `IgnoreNotFoundError` field on their `RetrieveRequest` structs so callers can bulk existence-check cached keys without treating missing entries as failures, and the TypeScript `x` package gains a `state` module, `Chain` destructor composition for rollback, and `maxInterval`/`jitter` controls on `Breaker`.

* **Go partial-key retrieves**: Ten API services add `IgnoreNotFoundError bool`; when set, `errors.Skip` removes `query.ErrNotFound` before propagating. Tests cover both the partial-survivor and all-missing cases. `label.go` deviates from this pattern (bug, see inline comment).
* `x/ts` **primitives**: New `state` module exports React-style `SetArg`/`executeSetter`/`skipUndefined`/`skipNull` helpers; `destructor.Chain` accumulates rollback callbacks to unwind optimistic writes on failure; `Breaker` gains a `maxInterval` ceiling and a `jitter` fraction. A correctness issue exists in `Chain.run()` (see inline comment).

### Confidence Score: 3/5

Two correctness defects in new code block a clean merge: `label.go` unconditionally swallows ErrNotFound regardless of the new flag, and `destructor.Chain.run()` stops at the first failing handler, leaving later optimistic writes unrolled.

Nine of ten Go services implement the IgnoreNotFoundError guard correctly; label.go is the sole outlier where the flag is dead code and prior error-propagation behavior is silently broken. The destructor Chain rollback mechanism fails to attempt all handlers when one throws, which defeats its purpose. The state module and Breaker additions are correct and well-tested.

core/pkg/api/label/label.go and x/ts/src/destructor/destructor.ts

### Important Files Changed

| Filename | Overview |
| -- | -- |
| core/pkg/api/label/label.go | Adds `IgnoreNotFoundError` to `RetrieveRequest` but the flag is never checked; `ErrNotFound` is unconditionally swallowed, breaking callers who pass `false`. |
| x/ts/src/destructor/destructor.ts | New `Chain` class for rollback composition, but `run()` stops at the first failing destructor, leaving subsequent optimistic writes unrolled. |
| core/pkg/api/arc/arc.go | Adds `IgnoreNotFoundError` to `RetrieveRequest`; guard pattern correctly skips `ErrNotFound` only when the flag is set. |
| core/pkg/api/panel/panel.go | Correctly adds `IgnoreNotFoundError` guard before the existing access-enforcement block. |
| x/ts/src/breaker/breaker.ts | Adds `maxInterval` cap and `jitter` fraction; jitter applied only to the sleep duration, not the persistent interval. |
| x/ts/src/state/state.ts | New module exporting `SetArg`, `executeSetter`, `skipUndefined`, and `skipNull` helpers; well-tested and logically consistent. |
| core/pkg/api/schematic/schematic.go | Adds `IgnoreNotFoundError` with consistent guard pattern; access enforcement correctly follows the error check. |
| core/pkg/api/ontology/ontology.go | Adds `IgnoreNotFoundError` with correct guard; also switches `var resources` to `make([]ontology.Resource, 0)` to ensure non-nil slice on empty results. |

### Sequence Diagram

```mermaid
sequenceDiagram
    participant Client
    participant APIService
    participant DB
    participant RBAC

    Client->>APIService: "Retrieve(keys=[A,B,C], IgnoreNotFoundError=true)"
    APIService->>DB: "query.Exec(keys=[A,B,C])"
    DB-->>APIService: "results=[A,C] + ErrNotFound (B missing)"
    APIService->>APIService: errors.Skip(err, ErrNotFound) to nil
    APIService->>RBAC: Enforce(subject, ActionRetrieve, [A,C])
    RBAC-->>APIService: allowed
    APIService-->>Client: "Response{[A,C]}, nil"

    Note over Client,RBAC: label.go deviation: ErrNotFound always skipped, IgnoreNotFoundError flag never checked
```

Reviews (1): Last reviewed commit: ["Add x/state module plus breaker and dest..."](<https://github.com/synnaxlabs/synnax/commit/abc676a99632f962585eed02b04d3476fc98ce4f>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=47043809>)

<h3>Greptile Summary</h3>

This PR adds two independent layers: ten Go API services gain `IgnoreNotFoundError bool` on their `RetrieveRequest` structs so callers can bulk existence-check keys and receive a partial result set instead of an error, and the TypeScript `x` package gains a `state` module, a `Chain` destructor composition helper, and `maxInterval`/`jitter` controls on `Breaker`. All changes are additive; the previously flagged issues in earlier commits (label.go ignoring the flag, destructor.ts stopping at the first failure) have been resolved in the head commit.

- **Go partial-key retrieves**: All ten services — arc, label, lineplot, log, ontology, panel, ranger, schematic, table, task — consistently guard `errors.Skip(err, query.ErrNotFound)` behind `req.IgnoreNotFoundError`. Tests cover both the partial-survivor and all-missing cases. The `ontology` service also switches `var resources` to `make([]ontology.Resource, 0)` so empty results serialize as `[]` rather than `null`.
- **`x/ts` primitives**: `destructor.Chain` accumulates rollback callbacks and runs each in its own `try/catch` on failure; `state` exports React-style `SetArg`/`executeSetter`/`skipUndefined`/`skipNull` helpers; `Breaker` gains a `maxInterval` ceiling on the persistent interval and a `jitter` fraction applied per-sleep on top of the scaled interval. All additions are covered by new tests.

Safe to merge. All changes are strictly additive: existing callers that never set IgnoreNotFoundError see identical behavior, and the new flag is correctly guarded in every service.

The IgnoreNotFoundError pattern is applied consistently across all ten Go services, including label.go which had a gap in an earlier commit. The destructor.Chain correctly isolates each rollback in its own try/catch so a failing destructor does not abort the rest. The state module and Breaker additions are self-contained with thorough test coverage. No correctness defects remain in the head commit.

No files require special attention. All key files have been reviewed and are consistent with the intended design.

| Filename | Overview |
|----------|----------|
| core/pkg/api/label/label.go | Adds IgnoreNotFoundError to RetrieveRequest; current head correctly guards errors.Skip behind the flag, consistent with all other services in this PR. |
| x/ts/src/destructor/destructor.ts | New Chain class wraps each destructor call in its own try/catch inside the for loop, ensuring all rollbacks are attempted even when one fails. |
| core/pkg/api/panel/panel.go | Correctly inserts the IgnoreNotFoundError guard between the query and the access enforcer; access enforcement runs on partial results, which is the intended behavior. |
| x/ts/src/breaker/breaker.ts | Adds maxInterval cap to the persistent interval and jitter fraction to the per-sleep duration; both changes are well-tested. |
| x/ts/src/state/state.ts | New module exporting React-style SetArg/executeSetter/skipUndefined/skipNull helpers; logically consistent and thoroughly tested. |
| core/pkg/api/ontology/ontology.go | Adds IgnoreNotFoundError with correct guard; also switches var resources to make([]ontology.Resource, 0) so an empty result serializes as [] rather than null. |
| core/pkg/api/arc/arc.go | Adds IgnoreNotFoundError with the correct conditional guard pattern; also generalizes the test helper from grantUpdateOn to grantOn(action). |
| core/pkg/api/schematic/schematic.go | Adds IgnoreNotFoundError with consistent guard before the access enforcer; new tests cover partial-survivor and all-missing cases. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client
    participant APIService
    participant DB
    participant RBAC

    Client->>APIService: "Retrieve(keys=[A,B,C], IgnoreNotFoundError=true)"
    APIService->>DB: "q.Exec(keys=[A,B,C])"
    DB-->>APIService: "results=[A,C] + ErrNotFound (B missing)"
    APIService->>APIService: errors.Skip(err, ErrNotFound) to nil
    APIService->>RBAC: Enforce(subject, ActionRetrieve, [A,C])
    RBAC-->>APIService: allowed
    APIService-->>Client: "Response{[A,C]}, nil"

    Client->>APIService: "Retrieve(keys=[A,B,C], IgnoreNotFoundError=false)"
    APIService->>DB: "q.Exec(keys=[A,B,C])"
    DB-->>APIService: "results=[A,C] + ErrNotFound (B missing)"
    Note over APIService: skip guard not applied
    APIService-->>Client: "RetrieveResponse{}, ErrNotFound"
```

<sub>Reviews (3): Last reviewed commit: ["honored IgnoreNotFoundError in label ret..."](https://github.com/synnaxlabs/synnax/commit/550e03e57cefa702a4b3cb7bd1ca69e9984bbe0b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47043809)</sub>


emilbon99:
@greptileai review

emilbon99:
@greptileai review

x/ts/src/breaker/breaker.ts
```diff
@@ -13,28 +13,37 @@ import { sleep } from "@/sleep";
...
 
   async wait(): Promise<boolean> {
-    const { maxRetries, scale, sleepFn } = this.config;
+    const { maxRetries, maxInterval, scale, jitter, sleepFn } = this.config;
     if (this.retries >= maxRetries) return false;
-    await sleepFn(this.interval);
+    let interval = this.interval;
+    if (jitter > 0)
+      interval = TimeSpan.milliseconds(
+        interval.milliseconds * (1 + Math.random() * jitter),
```
pjdotson: shouldn't the interval sometime be less than 1 as well?

x/ts/src/state/state.ts
```diff
@@ -0,0 +1,68 @@
...
+// included in the file licenses/APL.txt.
+
+import { type primitive } from "@/primitive";
+import { type record } from "@/record";
+
+/** Shape of values storable as state. */
+export type State = Exclude<primitive.Value, undefined> | record.Unknown | void;
+
+/** A function that computes the next state from the previous state. */
+export type SetFunc<S, PS = S> = (prev: PS) => S;
```
pjdotson, line 17: can you explain to me again when / where we would have previous states that are different from the actual state? I know there was some weird edge case this was needed for but I would rather we think about fixing our states so that everywhere there is a state it only has one type.

x/ts/src/state/state.ts
```diff
@@ -0,0 +1,68 @@
...
+/** Accepts a {@link SetArg} and applies it to the state it manages. */
+export type Setter<NextState extends State, PrevState = NextState> = (
+  value: SetArg<NextState, PrevState>,
+) => void;
+
+/** A {@link Setter} that accepts only literal next states, not updaters. */
+export type PureSetter<NextState extends State> = (value: NextState) => void;
+
+/** An initial state, or a function that lazily computes it. */
+export type Initial<InitialState extends State> = InitialState | (() => InitialState);
```
pjdotson: `S` would make more sense than `InitialState`, makes the types a little harder to read

pjdotson (changes requested):
what is jitter being used for?

emilbon99:
> what is jitter being used for?

It is used to prevent thundering herd issues when multiple clients that open RPCS (telemetry streams, etc.) all start retrying on the same schedule.

x/ts/src/state/state.ts
emilbon99: Yes, there are places where do this. For example in flux `retrieve.ts` we do:

```
 const query = state.executeSetter<Query, Partial<Query>>(
    querySetter,
    queryRef.current ?? {},
  );
```

The reason I did this move was purely so some of these caching mechanisms could re-use the state types. I didn't plan on changing these signatures at all, and don't think it's in scope for this PR.

My take is you have pretty strong opinions about this, and if you want to do a refactor you should take it up yourself so it gets done your (the right) way.

core/pkg/api/arc/arc_test.go
```diff
@@ -35,6 +36,27 @@ func createArc(ctx context.Context, name string) arc.Arc {
 	return a
 }
 
+var _ = Describe("Service.Retrieve", func() {
```
pjdotson: don't we normally write these as two nested `Describe` blocks, one for `"Service"` and another for `"Retrieve"`?

x/ts/src/breaker/breaker.ts
```diff
@@ -12,52 +12,86 @@ import { z } from "zod";
...
   maxRetries: z.number().optional(),
   scale: z.number().optional(),
+  jitter: z.number().min(0).max(1).optional(),
 });
 
-export interface Config extends Omit<z.infer<typeof breakerConfigZ>, "baseInterval"> {
+/** Configures a {@link Breaker}. */
+export interface Config extends Omit<
+  z.infer<typeof breakerConfigZ>,
+  "baseInterval" | "maxInterval"
```
pjdotson: why do we need to omit the `baseInterval` and `maxInterval` here instead of making the `Config` the type `z.input`?

pjdotson (changes requested):
Approved after comments addressed.

PR #2621, opened 2026-07-24
C. SY-4493: Rebind all domain services onto the query cache

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Third PR in the [SY-4493](https://linear.app/synnax/issue/SY-4493/robust-session-state-management-for-the-console) stack (base: `sy-4493-query-pkg`). Rebinds every domain service onto the query cache introduced in PR B.

**What lands here**

* `client/ts/src/actions/` — the `Controller` that owns dispatch/undo/redo and server-echo suppression.
* All 21 domain services rebound onto the cache and controller: optimistic writes with rollback, `setIfAbsent` hydration, undoable dispatch. `Synnax.cache` and per-domain `getCached` are exposed.

**Why it's safe to merge alone**<br>Additive to consumers. Pluto and console still drive the services through the pre-cutover raw-send dispatch path, preserved as a `@deprecated` overload on each domain client's `dispatch`, so no pluto/console call site changes. The flux rewrite and `flux/base` deletion that consume this substrate land in the next PR (D).

One pluto test (`Mosaic.spec` tab-scope render) is `it.skip`ped: while flux is still cache-blind, legacy raw-send dispatch mutates a tab server-side but the client keeps the tab-less `create()` copy. PR D's rebind onto the client cache restores it.

## Basic Readiness

### Greptile Summary

This PR rebinds all 21 domain service clients onto the shared `query.Cache` introduced in the previous stack PR, adding the new `Controller` class that manages local replay, undo/redo stacks, echo deduplication, and rollback for every document-table domain. The legacy raw-send `dispatch` overloads are preserved as `@deprecated` so no Pluto/console call sites break before the next PR lands.

* `actions/controller.ts` — new 594-line controller with a suite of 916-line tests covering replay, coalescing, stale-entry skipping, transactions, and echo deduplication.
* `client.ts` — wires the shared `query.Cache` through all domain constructors; exposes `cache` as a public field; adds `cache` and `onInternalError` params.
* `channel/retriever.ts` — removes the old hand-rolled `CacheRetriever` (superseded by the cache infrastructure); `Writer` no longer mutates it on create/delete/rename.

### Confidence Score: 4/5

Broadly safe additive change; two targeted fixes are needed before merging into a release branch.

The refactoring of [destructor.Chain.run](<http://destructor.Chain.run>)() from per-destructor try/catch to a single shared one means any throwing rollback function silently aborts the remaining rollbacks, leaving the cache partially rolled back. The covering test was deleted alongside this change. Separately, the Go label Retrieve handler now unconditionally skips ErrNotFound, making the IgnoreNotFoundError request field permanently inert while still advertising it in the struct.

x/ts/src/destructor/destructor.ts ([Chain.run](<http://Chain.run>) rollback completeness) and core/pkg/api/label/label.go (dead IgnoreNotFoundError field).

### Important Files Changed

| Filename | Overview |
| -- | -- |
| client/ts/src/actions/controller.ts | New Controller class: owns local replay, undo/redo stacks, echo deduplication, and remote-frame application for one document table. Well-structured with thorough tests. |
| x/ts/src/destructor/destructor.ts | [Chain.run](<http://Chain.run>)() refactored to a single try/catch, breaking the "run all destructors even when one fails" invariant; the covering test was deleted. If any rollback function throws, subsequent rollbacks are silently skipped. |
| core/pkg/api/label/label.go | Retrieve now unconditionally skips ErrNotFound, making IgnoreNotFoundError in the request struct permanently dead code; callers relying on not-found errors will silently get empty results. |
| client/ts/src/client.ts | Synnax client wires all 21 domain services onto the shared query.Cache; exposes cache as a public field; adds cache/onInternalError params with sensible defaults. |
| client/ts/src/arc/client.ts | Fully rebound onto query.Cache with optimistic writes, rollback via destructor.Chain, undo/redo via Controller, and live change-stream listeners. |
| client/ts/src/channel/retriever.ts | Removes the old hand-rolled CacheRetriever (replaced by query.Cache infrastructure); ClusterRetriever and DebouncedBatchRetriever retained unchanged. |
| client/ts/src/query/reconcile.spec.ts | New integration test: verifies cache.reconcile() tombstones documents deleted out-of-band and leaves live documents intact, across all doc-table domain services. |
| client/ts/src/ontology/store.ts | New shared store utilities for ontology resources and relationships: optimistic rename, delete, and relationship-drop helpers with rollback destructors. |
| client/ts/vite.config.ts | Adds global testTimeout: 15_000 and expect.poll.timeout; allows removing per-test timeout overrides in reader.spec.ts cleanly. |
| console/src/session/cluster/slice.ts | Omits the new cache field from the console's cluster schema so it doesn't leak into session storage; straightforward housekeeping. |

### Sequence Diagram

```mermaid
sequenceDiagram
    participant Caller as Domain Client
    participant Ctrl as Controller
    participant Table as query.Table (cache)
    participant Net as Network / Server
    participant Broadcast as Broadcast channel

    Caller->>Ctrl: dispatch(key, actions, send)
    Ctrl->>Table: replay - preprocess + reduce - set(next)
    Ctrl->>Ctrl: recordEntry(forward, inverse, targets)
    Ctrl->>Ctrl: registerOutstandingDispatch(key, dispatchKey)
    Ctrl->>Net: send(processed, dispatchKey)
    alt send succeeds
        Net-->>Broadcast: broadcast frame (dispatchKey, seq, actions)
        Broadcast->>Ctrl: applyRemote(key, seq, dispatchKey, actions)
        Ctrl->>Ctrl: own echo - skip reduce (unless disturbed)
    else send fails
        Ctrl->>Table: rollback (restore prior state)
        Ctrl->>Ctrl: stackRollback (pop undo entry)
        Ctrl-->>Caller: rethrow
    end

    Caller->>Ctrl: undo(key, send)
    Ctrl->>Ctrl: prepareUndo - find non-stale top entry
    Ctrl->>Table: replay(inverse actions)
    Ctrl->>Net: send(inverse, dispatchKey)
    Ctrl->>Ctrl: move entry undo to redo stack
```

Reviews (2): Last reviewed commit: ["Drop @deprecated tag from legacy dispatc..."](<https://github.com/synnaxlabs/synnax/commit/caac77afacff80dfc1b660dc3ce9a6999d582009>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=47055492>)

<h3>Greptile Summary</h3>

This PR rebinds all 21 domain service clients onto the shared `query.Cache`, introduces the new `Controller` class that manages local replay, undo/redo stacks, echo deduplication, and rollback, and removes the old hand-rolled `CacheRetriever` in favor of the cache infrastructure. Legacy raw-send `dispatch` overloads are preserved as `@deprecated` so no Pluto/console call sites break before the next PR.

- **`actions/controller.ts`**: New 597-line `Controller` with thorough 917-line test suite covering replay, coalescing, stale-entry skipping, echo deduplication, and transaction rollback.
- **`client.ts`**: Wires all 21 domain services onto the shared `query.Cache`; exposes `cache` as a public field; adds `cache` and `onInternalError` params.
- **`channel/retriever.ts`**: Removes the old hand-rolled `CacheRetriever` (superseded by cache infrastructure); `DebouncedBatchRetriever` retained unchanged.

Safe to merge — both previously-flagged regressions (destructor per-destructor rollback and label IgnoreNotFoundError gating) are confirmed fixed, and the additive cache-rebind is non-breaking for existing Pluto and console call sites.

The two blocking issues called out in earlier review rounds have been addressed: Chain.run() now wraps each destructor in its own try/catch so partial rollbacks can't leave the cache in a bad state, and the label Retrieve handler correctly gates the ErrNotFound skip on req.IgnoreNotFoundError again. The rest of the change is additive — deprecated overloads preserve the existing dispatch path, the controller logic is well-tested with 917 lines of unit coverage, and the optimistic-write rollback semantics are consistent across all 21 domain clients.

**Files Needing Attention:** No files require special attention — the only remaining note is a minor type-cleanliness gap in console/src/session/cluster/slice.ts (onInternalError not omitted from the session schema), which has no runtime impact.

| Filename | Overview |
|----------|----------|
| client/ts/src/actions/controller.ts | New Controller class: local replay, undo/redo stacks, echo deduplication, and remote-frame application — well-structured with thorough tests and correct rollback semantics. |
| x/ts/src/destructor/destructor.ts | Chain.run() correctly wraps each destructor in its own try/catch so all rollbacks execute even when one throws — regression from an earlier commit in the stack is fixed. |
| core/pkg/api/label/label.go | Retrieve correctly gates the ErrNotFound skip on req.IgnoreNotFoundError — the previously flagged unconditional skip is fixed and the field is meaningful again. |
| client/ts/src/client.ts | Wires all 21 domain services onto query.Cache; exposes cache as a public field; adds cache/onInternalError params with sensible defaults; removes the old CacheRetriever wrapper. |
| client/ts/src/arc/client.ts | Fully rebound onto query.Cache with optimistic writes, rollback via destructor.Chain, undo/redo via Controller, task management, and live change-stream listeners; well-structured. |
| client/ts/src/channel/retriever.ts | Removes the old hand-rolled CacheRetriever (superseded by query.Cache infrastructure); ClusterRetriever and DebouncedBatchRetriever retained unchanged; minor type assertion cleanup. |
| client/ts/src/ontology/store.ts | New shared store utilities for ontology resources and relationships: optimistic rename, delete, and relationship-drop helpers with correct rollback order. |
| console/src/session/cluster/slice.ts | Correctly omits the new `cache` field from the session schema; the new `onInternalError` function field is not omitted, leaking it into the Cluster TypeScript type even though it is never serialized. |
| client/ts/vite.config.ts | Adds global testTimeout: 15_000 and expect.poll.timeout: 5_000, allowing per-test timeout overrides to be removed cleanly. |
| client/ts/src/query/write.ts | New optimistic write helper that runs rollbacks via destructor.Chain.guard on commit failure; clean separation of concerns. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant DomainClient as Domain Client
    participant Ctrl as Controller
    participant Table as query.Table (cache)
    participant Net as Network / Server
    participant Broadcast as Broadcast channel

    DomainClient->>Ctrl: dispatch(key, actions, send)
    Ctrl->>Table: replay — preprocess + reduce → set(next)
    Ctrl->>Ctrl: recordEntry(forward, inverse, targets)
    Ctrl->>Ctrl: registerOutstandingDispatch(key, dispatchKey)
    Ctrl->>Net: send(processed, dispatchKey)
    alt send succeeds
        Net-->>Broadcast: broadcast frame (dispatchKey, seq, actions)
        Broadcast->>Ctrl: applyRemote(key, seq, dispatchKey, actions)
        Ctrl->>Ctrl: own echo — skip reduce (unless disturbed)
    else send fails
        Ctrl->>Table: rollback (restore prior state)
        Ctrl->>Ctrl: stackRollback (pop undo entry)
        Ctrl-->>DomainClient: rethrow
    end

    DomainClient->>Ctrl: undo(key, send)
    Ctrl->>Ctrl: prepareUndo — find non-stale top entry
    Ctrl->>Table: replay(inverse actions)
    Ctrl->>Net: send(inverse, dispatchKey)
    Ctrl->>Ctrl: move entry undo → redo stack
```

<sub>Reviews (4): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/sy-..."](https://github.com/synnaxlabs/synnax/commit/fb87329eab4c5a0f8842a4d98226ec06ecb66d56) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47055492)</sub>


emilbon99:
@greptileai review

x/ts/src/destructor/destructor.ts
```diff
@@ -30,13 +33,13 @@ export class Chain {
...
-      }
-    this.destructors.length = 0;
+    try {
+      this.destructors.reverse().forEach((d) => d());
+    } catch (error) {
+      console.error("failed to roll back optimistic writes", error);
+    } finally {
+      this.destructors.length = 0;
+    }
   }
```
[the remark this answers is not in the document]
emilbon99: Fixed. The rebind commit had regressed `run()` to a single try/catch around the whole loop; restored per-destructor try/catch so every rollback executes even when one throws, and restored the deleted `should run remaining destructors when one fails` test. The foundations PR already had the correct form; the regression was introduced here.

core/pkg/api/label/label.go
```diff
@@ -117,11 +117,10 @@ func (s *Service) Retrieve(
...
 		}
-		err := q.Entries(&res.Labels).Exec(ctx, nil)
-		if req.IgnoreNotFoundError && err != nil {
-			err = errors.Skip(err, query.ErrNotFound)
-		}
-		if err != nil {
+		// Missing keys are omitted from the result, not an error, so callers
+		// can existence-check cached keys in bulk.
+		if err := q.Entries(&res.Labels).
+			Exec(ctx, nil); errors.Skip(err, query.ErrNotFound) != nil {
```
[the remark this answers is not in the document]
emilbon99: Fixed. Restored the guard so the skip honors `req.IgnoreNotFoundError`, matching all eight sibling API services. The rebind commit had regressed it to an unconditional skip; the field is meaningful again.

client/ts/src/client.ts
```diff
@@ -91,6 +101,13 @@ export default class Synnax extends framer.Client {
   readonly tables: table.Client;
   readonly groups: group.Client;
   readonly imex: imex.Client;
+  /**
+   * The client's local mirror of cluster state. Live and streamed when
+   * `cache: true` (the default); detached (local-only, no streaming) when
+   * `cache: false`. Not a data access path: per-domain stores on the domain
+   * clients remain the only way to read cached records.
+   */
+  readonly cache: query.Cache;
```
pjdotson, line 110: who will call `client.cache`?

client/ts/src/client.ts
```diff
@@ -124,72 +141,133 @@ export default class Synnax extends framer.Client {
...
-    super(transport.stream, transport.unary, chRetriever);
-    this.auth = new auth.Client(transport.unary, { username, password });
-    transport.use(this.auth.middleware());
-    const chCreator = new channel.Writer(transport.unary, chRetriever);
-    this.key = uuid.create();
-    this.createdAt = TimeStamp.now();
-    this.params = parsedParams;
-    this.transport = transport;
-    this.channels = new channel.Client(this, chRetriever, transport.unary, chCreator);
+    const chRetriever = new channel.ClusterRetriever(transport.unary);
```
pjdotson, line 152: Is the channel internal retriever getting merged with the cache in another PR?

client/ts/src/client.ts
```diff
@@ -124,72 +141,133 @@ export default class Synnax extends framer.Client {
...
+      unary,
+      ontology: this.ontology,
+      cache,
+    });
+    this.imex = new imex.Client({ file: this.transport.file });
   }
 
   close(): void {
     this.connectivity.stop();
+    this.cache.close().catch(console.error);
```
pjdotson: shouldn't this use `onInternalError`?

client/ts/src/imex/client.ts
```diff
@@ -70,11 +70,15 @@ export interface ImportOptions extends Options {
  * - Tauri/Console: read the picked file into a Blob to import, and hand the export
  *   stream to the Console's downloadStream helper.
  */
+export interface ClientParams {
+  file: FileTransport;
+}
+
 export class Client {
-  private readonly fileTransport: FileTransport;
+  private readonly cfg: ClientParams;
```
pjdotson: if all of these variables are titled `cfg`, then why are the called `ClientParams`? feels like the the params and config names should match

client/ts/src/label/client.ts
```diff
@@ -40,77 +42,207 @@ const retrieveRequestZ = z.object({
...
+      const rel = labeledByRel(id, key);
+      rollback.add(
+        this.cfg.ontology.cache.relationships.set(
+          ontology.relationshipToString(rel),
+          rel,
+        ),
+      );
+    });
+    await opts.onOptimistic?.();
+    await rollback.guard(
```
pjdotson: If we follow this pattern for everything:

1. make rollback chain
2. Add rollbacks
3. Await optimistic
4. Do the server call in a rollback guard

then there should be some function we can call with parameters for 2 and 4 so that we don't have to call 1 and 3.

client/ts/src/lineplot/client.ts
```diff
@@ -40,71 +49,237 @@ export type RetrieveParams = z.input<typeof retrieveParamsZ>;
...
+  private readonly cfg: ClientParams;
+  private readonly store: query.Table<Key, LinePlot>;
+  private readonly dispatcher: actions.Controller<Key, LinePlot, Action>;
+
+  constructor(cfg: ClientParams) {
+    const { cache } = cfg;
+    // Dispatch mutates documents server-side, so fetched copies never clobber
+    // a doc holding locally replayed edits: the table hydrates if-absent.
+    const store = cache.createTable<Key, LinePlot>({
+      name: "lineplots",
```
pjdotson: should probably be `"line plots"` with a space

client/ts/src/query/retriever.ts
```diff
@@ -121,14 +124,14 @@ export abstract class Retriever<
   V extends state.State & record.Keyed<K>,
   D extends Data = V,
   SP = { key: K },
+  SN extends Params = K,
```
pjdotson: SN is confusing parameter name

pjdotson (changes requested):
See comments

client/ts/src/client.ts
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Nobody, as of D. This question pulled on the right thread: everything `cache` sold publicly was really connection lifecycle in disguise. On D and up, `client.cache` is private. Stream demand is owned by the connection machine (it re-demands the stream on every successful check), and `await client.connect()` is the public way to wait for it, so the specs use that. The epoch is a fact about the client's session continuity, not about the cache, and it was already mirrored into `connection.Status.details.epoch`, so the console session synchronizer now subscribes to `client.connection.onChange` instead of `cache.onEpoch`. It stays public on this PR only because `connect()` does not exist until D.

client/ts/src/client.ts
emilbon99: Will be done in this issue: https://linear.app/synnax/issue/SY-4533/resolve-framer-channel-lookups-through-the-query-cache

client/ts/src/client.ts
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Fixed. Close failures now report through the cache's resolved `onInternalError` sink, wrapped with context so the original error survives as `cause`. D applies the same treatment to the connection machine's `close()`, which had the same flaw.

client/ts/src/imex/client.ts
emilbon99: Which one do you prefer? Do you prefer config or params standardization?

client/ts/src/label/client.ts
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Fixed. Added `query.optimistic({ rollbacks, onOptimistic, commit })`, which owns the chain construction and the notify-then-guard sequencing, and migrated all 53 sites across the domain clients onto it. The helper takes a pre-bound `onOptimistic` thunk instead of opts plus a value, so value-bearing creates close over their optimistic value (`() => opts.onOptimistic?.(optimistic)`) with no type assertions. Sites that re-apply after success (delete/rename racing concurrent set events) keep that one trailing line, since it is semantics rather than ceremony.

client/ts/src/lineplot/client.ts
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Fixed. Also matches the singular query space in the same file, which was already named "line plot".

client/ts/src/query/retriever.ts
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Fixed. Renamed `SN` to `SingleQuery` and its pair `SP` to `SingleInput`, so the schema field now reads `schema: z.ZodType<SingleQuery, SingleInput>`, mirroring zod's `ZodType<Output, Input>` order. Generic bindings are positional, so no subclass sites changed.

emilbon99:
@greptileai review

emilbon99:
@greptileai review

client/ts/src/imex/client.ts
pjdotson: think cfg makes more sense as it matches the Core services

client/ts/src/table/actions.ts
```diff
@@ -253,3 +253,21 @@ const handlers: Handlers = {
...
 
 export const reduceAll = createReduceAll(handlers);
+
+// kindOf classifies an action batch for the undoable store's
+// per-kind coalesce window. Continuous resize gestures (a stream of
+// resize_row or resize_col actions) collapse into a single undo step;
+// successive set_cell actions on the same cell coalesce per-cell so typing
+// inside one cell collapses to one undo step but switching cells starts a
+// fresh entry; everything else is treated as a discrete user action.
+export const kindOf = (actions: Action[]): string => {
```
pjdotson, line 263: so this returns a general `"string"` instead of a string enum? latter might be a better API

pjdotson (changes requested):

client/ts/src/table/actions.ts
emilbon99: Can't be a string enum because occasionally we scope directly to an item with a specific key.

client/ts/src/view/client.ts
```diff
@@ -30,59 +38,135 @@ const retrieveRequestZ = z.object({
...
 
-const singleRetrieveParamsZ = z
-  .object({ key: keyZ })
-  .transform(({ key }) => ({ keys: [key] }));
-
-const retrieveParamsZ = z.union([singleRetrieveParamsZ, retrieveRequestZ]);
-
-export interface RetrieveSingleParams extends z.input<typeof singleRetrieveParamsZ> {}
+// A type alias, not an interface: only aliases get the implicit index
+// signature that record-constrained query types require.
```
pjdotson: bad comment

client/ts/src/query/retriever.ts
```diff
@@ -166,12 +180,42 @@ export abstract class Retriever<
       serverFields,
       watch,
     });
-    this.normalizeRequest = (p) => schema.parse(p);
+    // A bare key array the schema does not accept directly retries as
+    // `{ keys }` shorthand. A schema that strips the field would silently
+    // match every record, so that retry is rejected instead.
+    this.normalizeRequest = (p) => {
+      const res = schema.safeParse(p);
```
pjdotson: this seems really confusing to have this logic here - why can't the schemas just do the transformation for themselves of transforming a key to a `{ keys }` shorthand?

client/ts/src/query/write.ts
```diff
@@ -0,0 +1,41 @@
...
+  onOptimistic?: () => Promise<void> | void;
+  /** The network send. Its result is returned when it succeeds. */
+  commit: () => Promise<R>;
+}
+
+/**
+ * Runs an optimistic write. On commit failure the rollbacks run and the error
+ * rethrows; on success the cache keeps the optimistic state.
+ */
+export const optimistic = async <R>({
```
pjdotson, line 32: needs test

client/ts/src/schematic/symbol/client.ts
```diff
@@ -8,48 +8,57 @@
...
   keyZ,
   type New,
+  ontologyID,
   type Symbol,
   symbolZ,
 } from "@/schematic/symbol/types.gen";
-import { checkForMultipleOrNoResults } from "@/util/retrieve";
+
+export const SET_CHANNEL_NAME = "sy_schematic_symbol_set";
+export const DELETE_CHANNEL_NAME = "sy_schematic_symbol_delete";
```
pjdotson, line 27: Long term should these channel names be exported or not? If a caller could do anything they wanted from the `onSet`/`onDelete` handlers

pjdotson (changes requested):

client/ts/src/schematic/symbol/client.ts
emilbon99: They should not be exported.

pjdotson (approved):

PR #2622, opened 2026-07-24
B. SY-4493: Introduce the client query/ cache package

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Second PR in the [SY-4493](https://linear.app/synnax/issue/SY-4493/robust-session-state-management-for-the-console) stack (base: `sy-4493-foundations`). Adds the client-side caching package on its own, with nothing consuming it yet.

**What lands here**

* `client/ts/src/query/` — the read-through cache: `Cache`, `Table` (rows + tombstones + equality-silenced set), `Queries` (answer spaces), `Retriever`, and the streamer/reconcile plumbing, with unit + live-core specs.
* `framer.HardenedStreamer` gains `onReopen`/`onDrop` hooks (and a retry-forever, capped-jitter default) that the cache uses to resubscribe and invalidate across reconnects. Additive: existing callers keep the three-argument form.

**Why it's safe to merge alone**<br>Purely additive. No domain service, pluto, or console code references the package. The rebind of the 21 domain services onto it lands in the next PR (C).

## Basic Readiness

### Greptile Summary

This PR introduces the client-side `query/` caching package — `Table`, `Queries`, `Retriever`, `Cache`, and streamer plumbing — and extends `framer.HardenedStreamer` with `onReopen`/`onDrop` hooks and a retry-forever default. No existing domain service is wired to it yet; all additions are purely additive.

* `Table` is a keyed store with tombstones, equality-silenced sets, and per-row subscriptions. Its public `clear()` method bypasses `notify`, so subscribers (including all live `Queries` instances) see a stale key list after a clear, composing results against an empty table without any notification or error.
* `Queries` implements three maintenance rules (exact-key, client-checkable, server-computed) with deduped fetches, debounced wholesale refetches, and epoch-triggered invalidation on reconnect. The logic is well-guarded and the state machine is correct.
* `HardenedStreamer` changes are additive: existing callers pass fewer arguments and keep prior behavior; the new hooks are opt-in.

### Confidence Score: 4/5

Safe to merge as-is given no domain service consumes the package yet; the `Table.clear()` notification gap should be fixed before the domain wiring lands in PR C.

The only real defect is `Table.clear()` silently skipping subscriber notifications — every `Queries` instance watching that table would continue delivering stale composed data to subscribers after a clear, with no error or signal. Since no code calls `clear()` yet this isn't causing live breakage, but the method is public and exported, making it a correctness trap for the next PR that uses it.

client/ts/src/query/table.ts — `clear()` needs to notify subscribers before wiping rows and tombstones, matching the contract established by `delete()`.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| client/ts/src/query/table.ts | New keyed store with tombstones, equality-silenced sets, and per-row subscriptions — `clear()` silences all subscriber notifications, leaving active Queries instances in a stale state |
| client/ts/src/query/query.ts | Core answer-space implementation with three maintenance rules (exact-key, client-checkable, server-computed), deduped fetches, and epoch-triggered refetches; logic is correct and well-guarded |
| client/ts/src/query/cache.ts | Lazily-opened cache orchestrator wiring tables, streamer, epoch counting, and reconciliation; clean and straightforward |
| client/ts/src/query/streamer.ts | Lazy change-stream driver: channels are sorted so deletions precede other changes; frame parsing and error reporting are correct |
| client/ts/src/query/reader.ts | Typed read-surface router (`Retriever`) splitting single vs. multi params — clean abstraction with no logic issues |
| client/ts/src/framer/streamer.ts | Adds `onReopen`/`onDrop` hooks to `HardenedStreamer` and changes default breaker to retry-forever with capped jitter backoff; purely additive to existing callers |
| client/ts/src/query/types.ts | Foundational type aliases (`Params`, `Data`, `FetchOptions`, `WriteOptions`) — straightforward, no issues |

### Sequence Diagram

```mermaid
sequenceDiagram
    participant Consumer
    participant Queries
    participant Table
    participant Cache
    participant Streamer
    participant Server

    Consumer->>Queries: retrieve(query)
    Queries->>Cache: ensureStreaming()
    Cache->>Streamer: demand()
    Streamer->>Server: open change stream
    Server-->>Streamer: stream live
    Streamer-->>Cache: "onOpen → epochCount=1"
    Cache-->>Queries: onEpoch fired
    Queries->>Server: fetch(query)
    Server-->>Table: set(records)
    Table-->>Queries: notify (set event)
    Queries-->>Consumer: compose(keys) → D

    Note over Streamer,Server: Reconnect scenario
    Server->>Streamer: connection dropped
    Streamer->>Cache: onDrop
    Streamer->>Server: reconnect (HardenedStreamer retry)
    Server-->>Streamer: stream live
    Streamer-->>Cache: onReopen → bumpEpoch()
    Cache->>Server: reconcile() — refetch all table keys
    Cache-->>Queries: onEpoch → refetch maintained entries
    Server-->>Table: set/delete (reconciled values)
    Table-->>Queries: notify
    Queries-->>Consumer: updated answer
```

Reviews (2): Last reviewed commit: ["B. SY-4493: Introduce the client query/ ..."](<https://github.com/synnaxlabs/synnax/commit/9e54b207ef5847f4e551267b5612744ff2040121>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=47058167>)

<h3>Greptile Summary</h3>

This PR introduces the client-side `query/` caching package — `Table`, `Queries`, `Retriever`, `Cache`, `derived`, and streamer plumbing — and extends `framer.HardenedStreamer` with `onReopen`/`onDrop` hooks plus a retry-forever default. No domain service is wired to it yet; every addition is purely additive.

- `Table` gains a generation counter that discards stale fetches and reconcile results across `reset()`, fixing the previously reported stale-write race. `reset()` silently clears data (intentionally), while `delete()` and `set()` correctly notify subscribers.
- `Queries` implements three maintenance rules with deduped fetches, debounced bulk refetches, and epoch-triggered invalidation. The state-machine guards are correct: Rule-1 subscriptions that fire during a parallel `refetch()` beat the guard check without double-notifying, and eviction correctly handles `single`-space tombstoned keys.
- `HardenedStreamer` changes are additive: existing callers pass three arguments and keep prior behavior; `onReopen`/`onDrop` are opt-in.

Safe to merge. No domain service consumes the package yet, so the new code is exercised only by its own unit and live-core specs.

The change is purely additive — no existing consumer references the new query/ package. The previously reported stale-write race and the Table.clear() notification gap were both fixed before this review. The remaining gap (background refetch errors not forwarded to the error sink) affects diagnostics for subscription-only consumers after reconnect, but has a recovery path through an explicit retrieve() call and does not corrupt state. HardenedStreamer changes are backward-compatible.

**Files Needing Attention:** client/ts/src/query/query.ts — the refetch() error path does not call this.hooks.onError or notify subscribers, which could leave purely subscription-driven consumers with a silent stale state after a failed reconnect-triggered refetch.

| Filename | Overview |
|----------|----------|
| client/ts/src/framer/streamer.ts | Adds onReopen/onDrop hooks to HardenedStreamer and changes default breaker to retry-forever with capped jitter backoff. Purely additive; existing callers continue to work with the three-argument form. |
| client/ts/src/query/table.ts | New keyed store with tombstones, equality-silenced sets, per-row subscriptions, rollbacks, and a generation counter that discards stale fetches across reset(). Prior review issue with clear() was fixed; reset() correctly silences notifications and bumps gen. Logic is sound. |
| client/ts/src/query/query.ts | Core answer-space with three maintenance rules, deduped fetches, and epoch-triggered refetches. Logic is correct and well-guarded, but background refetch() failures from epoch callbacks are stored silently in entry state without calling the error sink or notifying subscribers. |
| client/ts/src/query/cache.ts | Lazily-opened cache orchestrating tables, answer spaces, stream epochs, and reconciliation. Previous stale-write race after reset() was fixed by generation counters in Table. Clean and correct. |
| client/ts/src/query/streamer.ts | Lazy change-stream driver with sorted channels (deletions precede other changes) and per-listener fan-out. Frame parsing and error reporting are correct. |
| client/ts/src/query/retriever.ts | Typed read-surface router splitting single vs. multi params; keys-only requests resolve through the table's fetch directly rather than the request fetch. Clean abstraction with no logic issues. |
| client/ts/src/query/derived.ts | Materialized derived-table binding: composed rows are replaced on source and watch events, with deletions correctly propagated through the source subscription. Correct and clean. |
| client/ts/src/query/types.ts | Foundational type aliases (Params, Data, FetchOptions, WriteOptions). Straightforward, no issues. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Consumer
    participant Queries
    participant Table
    participant Cache
    participant CacheStreamer as Streamer (query)
    participant HardenedStreamer as HardenedStreamer (framer)
    participant Server

    Consumer->>Queries: retrieve(query)
    Queries->>Cache: ensureStreaming()
    Cache->>CacheStreamer: demand()
    CacheStreamer->>HardenedStreamer: open(opener, config)
    HardenedStreamer->>Server: open change stream
    Server-->>HardenedStreamer: stream live
    HardenedStreamer-->>CacheStreamer: onOpen fired
    CacheStreamer-->>Cache: "onOpen → epochCount=1"
    Cache-->>Queries: onEpoch(1) → refetch maintained entries
    Queries->>Server: fetch(query)
    Server-->>Table: retrieve → set(records)
    Table-->>Queries: notify (set event, Rule-1 sub)
    Queries-->>Consumer: compose(keys) → D

    Note over HardenedStreamer,Server: Reconnect
    Server->>HardenedStreamer: connection dropped
    HardenedStreamer->>CacheStreamer: onDrop(error)
    HardenedStreamer->>Server: reconnect (retry-forever backoff)
    Server-->>HardenedStreamer: stream live
    HardenedStreamer-->>CacheStreamer: onReopen
    CacheStreamer-->>Cache: onReopen → bumpEpoch()
    Cache->>Table: reconcile() — fetch all cached keys
    Table-->>Cache: delete(vanished), set(survivors)
    Cache-->>Queries: onEpoch(N) → refetch maintained entries
    Table-->>Queries: notify (set/delete events)
    Queries-->>Consumer: updated answer
```

<sub>Reviews (4): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/rc&#39;..."](https://github.com/synnaxlabs/synnax/commit/db18283be34b593dd431fcca024c7718113087b4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47058167)</sub>


client/ts/src/query/query.ts
```diff
@@ -0,0 +1,544 @@
...
+  ensureStreaming?: () => Promise<void>;
+  /** Subscribes to connection-epoch bumps; maintained answers refetch on bump. */
+  onEpoch?: (callback: () => void) => destructor.Destructor;
+  /** Reports maintenance errors that have no caller to throw to. */
+  onError?: (error: Error) => void;
+}
+
+const DEFAULT_DEBOUNCE = TimeSpan.milliseconds(100);
+
+type EntryState<K extends record.Key, D extends Data> =
```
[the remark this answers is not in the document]
emilbon99: By design. `hashQuery` treats a `Hashable`'s `hash()` as the authoritative key for its type; uniqueness against the primitive forms is the implementor's contract, the same as for any hashed key in the codebase. A given query space is only ever fed one `Params` type, so cross-form aliasing isn't reachable in practice.

client/ts/src/query/query.ts
```diff
@@ -0,0 +1,544 @@
...
+    if (entry.refetchTimer != null) {
+      clearTimeout(entry.refetchTimer);
+      entry.refetchTimer = undefined;
+    }
+    const before = entry.state;
+    void this.params.fetch(entry.query).then(
+      (keys) => {
+        // Only apply if maintenance hasn't been torn down and no newer fetch
+        // superseded this one.
+        if (entry.state !== before) return;
```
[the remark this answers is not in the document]
emilbon99: Intentional retention. `getCached` must serve an answer after every handler has unsubscribed (per the derived-view cache design), so the entry can't be dropped on last-unsubscribe. Console query shapes are bounded per domain, so growth is bounded in practice. A generic eviction policy for high-cardinality spaces is a separate design decision, not part of this PR.

client/ts/src/query/streamer.ts
```diff
@@ -0,0 +1,169 @@
...
+          for (const { onChange, schema } of listeners) {
+            let parsed: z.output<typeof schema>[];
+            try {
+              if (!series.dataType.equals(DataType.JSON))
+                parsed = Array.from(series).map((s) => schema.parse(s));
+              else parsed = series.parseJSON(schema);
+            } catch (exc) {
+              report(exc, `failed to parse streamer change for ${name}`);
+              continue;
+            }
```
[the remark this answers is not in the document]
emilbon99: No reorder in practice. `onChange` handlers are synchronous (`(frame) => void`) and complete their `table.set`/`table.delete` before the `await` yields the microtask. Two `handleChange` calls run their synchronous bodies in arrival order, so a set-then-delete for the same key applies in order. The `await` only matters for genuinely async handlers, which this interface doesn't have.

client/ts/src/framer/streamer.ts
```diff
@@ -206,67 +206,90 @@ class BaseStreamer implements Streamer {
...
     this.opener = opener;
     this.config = streamerConfigZ.parse(config);
     const {
-      maxRetries = 5000,
+      maxRetries = Infinity,
       baseInterval = TimeSpan.seconds(1),
-      scale = 1,
+      maxInterval = TimeSpan.seconds(30),
+      scale = 2,
+      jitter = 0.25,
```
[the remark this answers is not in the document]
emilbon99: Intentional. The only production caller is the control state tracker, which should retry forever with backoff; the new defaults encode exactly that. Callers wanting bounded retries still pass their own `breakerConfig`.

emilbon99:
@greptileai review

client/ts/src/query/table.ts
```diff
@@ -0,0 +1,302 @@
...
+        this.tombstones.delete(k);
+        this.rows.set(k, value);
+        this.notify({ variant: "set", key: k, value });
+      });
+  }
+
+  clear() {
+    this.rows.clear();
+    this.tombstones.clear();
+  }
```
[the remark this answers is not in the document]
emilbon99: Fixed. `clear()` now emits a `delete` for every cleared row and tombstone key, so subscribed answer spaces drop their stale keys instead of composing against an empty table. Added a regression test asserting subscribers are notified per cleared key.

client/ts/src/query/table.ts
```diff
@@ -0,0 +1,428 @@
...
+  /** When the deletion was applied to this table. */
+  deletedAt: TimeStamp;
+}
+
+/**
+ * Presence of a key in a table: "present" when a live row exists, "tombstoned"
+ * when the row was deleted and its corpse is retained, "unknown" when the
+ * table has never seen the key or the corpse was cleared.
+ */
+export type RowStatus = "present" | "tombstoned" | "unknown";
```
pjdotson, line 30: What are examples of when the `Status` is `"unknown"`?

client/ts/src/query/table.ts
```diff
@@ -0,0 +1,428 @@
...
+
+/** A change to a single row of a {@link Table}. */
+export type TableEvent<
+  Key extends record.Key = record.Key,
+  Value extends state.State = state.State,
```
pjdotson, line 35: does it make sense that `Value` should have some shape similar to a `gorp.Entry`, e.g. it needs to have a key of type `Key`?

client/ts/src/query/table.ts
```diff
@@ -0,0 +1,428 @@
...
+> {
+  (event: TableEvent<Key, Value>): void;
+}
+
+/**
+ * How fetched records hydrate a table: "set" overwrites rows, "if-absent"
+ * preserves existing rows.
+ */
+export type HydrateMode = "set" | "if-absent";
```
pjdotson, line 51: when do we use both of these?

client/ts/src/query/table.ts
```diff
@@ -0,0 +1,428 @@
...
+
+/** Construction arguments for a {@link Table}. */
+export interface TableParams<
+  Key extends record.Key = record.Key,
+  Value extends state.State = state.State,
+> {
+  /** Receives subscriber and fetch errors that have no caller to throw to. */
+  onError: (error: Error) => void;
+  /** Overrides the deep-equality default used to silence redundant sets. */
+  equal?: (a: Value, b: Value, key: Key) => boolean;
```
pjdotson, line 68: are there performance issues for doing deep-equality checks for everything?

client/ts/src/query/table.ts
```diff
@@ -0,0 +1,428 @@
...
+    };
+  }
+
+  /**
+   * Sets the value for the given key, or applies a setter to the previous
+   * value. A setter producing null/undefined, or a value equal to the current
+   * one, is a no-op.
+   * @returns A rollback that undoes the set.
+   */
+  set(key: Key, value: state.SetArg<Value | undefined>): destructor.Destructor;
```
pjdotson, line 178: when would setters produce undefined?

client/ts/src/query/table.ts
```diff
@@ -0,0 +1,428 @@
...
+    return () => rollbacks.reverse().forEach((r) => r());
+  }
+
+  /**
+   * Writes fetched records into the table under its declared hydrate mode:
+   * "set" overwrites rows, "if-absent" leaves existing rows untouched.
+   * @returns A rollback that undoes the rows this call wrote.
+   */
+  ingest(
+    values: (Value & record.Keyed<Key>) | Array<Value & record.Keyed<Key>>,
```
pjdotson: seems like there should be a type alias for this comment type of `Value & record.Keyed<Key>`

client/ts/src/query/table.ts
```diff
@@ -0,0 +1,428 @@
...
+    const arr = array.toArray(values);
+    if (this.hydrateMode === "if-absent") return this.setIfAbsent(arr);
+    return this.set(arr);
+  }
+
+  /** Returns every row in the table. */
+  get(): Value[];
+  get(key: Key): Value | undefined;
+  get(keys: Key[]): Value[];
+  get(filter: (value: Value) => boolean): Value[];
```
pjdotson, line 223: filter option seems somewhat useless if all it is is calling `.filter` from `.get()`

client/ts/src/query/query.ts
```diff
@@ -0,0 +1,568 @@
...
+/**
+ * Declares that events on the given table affect answers in this space.
+ * `affects` maps an event to the primary keys whose membership or content it
+ * touches, "refetch" to invalidate wholesale, or null when unaffected.
+ */
+export const watch = <
+  Q extends Params,
+  K extends record.Key,
+  FK extends record.Key,
+  FV extends state.State,
```
pjdotson: bad type argument names, don't know what `FK` and `FV` mean (presumably foreign key? but just write that)

client/ts/src/query/query.ts
```diff
@@ -0,0 +1,568 @@
...
+  /** Reports maintenance errors that have no caller to throw to. */
+  onError?: (error: Error) => void;
+}
+
+const DEFAULT_DEBOUNCE = TimeSpan.milliseconds(100);
+
+type EntryState<K extends record.Key, D extends Data> =
+  | { variant: "unfetched" }
+  | { variant: "loading"; promise: Promise<D> }
+  | { variant: "ready"; keys: K[] }
```
pjdotson, line 173: why can ready be multiple keys, but loading and deleted look like they are for a single entry?

client/ts/src/query/query.ts
```diff
@@ -0,0 +1,568 @@
...
+    void this.hooks.ensureStreaming?.().catch((exc: unknown) => this.report(exc));
+  }
+
+  private report(
+    exc: unknown,
+    message = `failed to maintain ${this.params.name} answers`,
+  ): void {
+    const error = new Error(message, { cause: exc });
+    if (this.hooks.onError != null) this.hooks.onError(error);
+    else console.error(error);
```
pjdotson: why not just default `onError` to `console.error`?

client/ts/src/query/retriever.ts
```diff
@@ -0,0 +1,234 @@
...
+ * from the table and its fetch: `{ key }` params resolve one record, with
+ * deletion flipping the answer to deleted. Declare `single` only when the
+ * domain's single query is richer than a key.
+ */
+export interface RetrieverParams<
+  Z extends z.ZodType<Params>,
+  K extends record.Key,
+  V extends state.State & record.Keyed<K>,
+  D extends Data = V,
+  SP = { key: K },
```
pjdotson: bad type parameter names, what is `SP`?

client/ts/src/query/retriever.ts
```diff
@@ -0,0 +1,234 @@
...
+  private readonly isSingle: (params: unknown) => params is SP;
+  private readonly normalizeSingle: (params: SP) => Params;
+  private readonly normalizeRequest: (params: z.input<Z>) => z.output<Z>;
+
+  constructor(cache: Cache, params: RetrieverParams<Z, K, V, D, SP>) {
+    const {
+      name,
+      table,
+      request,
+      compose = (record) => record as unknown as D,
```
pjdotson: you should probably look through everywhere you have type assertions as there will be potential for weird bugs

client/ts/src/query/streamer.ts
```diff
@@ -0,0 +1,191 @@
...
+  const bHasDelete = b.includes("delete");
+  if (aHasDelete && !bHasDelete) return -1;
+  if (!aHasDelete && bHasDelete) return 1;
+  return 0;
+};
+
+/**
+ * An opened change stream: frame notifications plus close.
+ *
+ * TODO: exists only because a concrete framer.ObservableStreamer would close
```
pjdotson, line 50: when are you addressing these TODOs?

pjdotson (changes requested):
Is `query` the right name for the package if it will also be used for mutations?

client/ts/src/query/table.ts
emilbon99: *Emiliano + Claude: generated reply from our review session.*

`unknown` means the key is in neither `rows` nor `tombstones`. Three cases:

- A watch event announces a key this client never fetched. `recheckMany` sees `unknown` and backfills it through `table.retrieve` before rechecking membership.
- An exact-key subscriber attaches for a key with no row or tombstone yet. Exact-key means the query addresses a single row through `keyOf`, so maintenance tracks that one row directly. `unknown` here means no instant answer, so the normal fetch path runs.
- After `reset()` on cluster replacement, rows and tombstones are cleared, so every previously known key reverts to `unknown`.

client/ts/src/query/table.ts
emilbon99: No, there are certain things we store that can't have that like `ontology.Relationship` which has a function to produce its key. In gorp we use a method of `Key` but we can do methods here unless we want things to be classes (which we don't), so we should leave as is. 

In the future if we get rid of data structures like that it wil be something we can do, but not yet.

client/ts/src/query/table.ts
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Both get used on the next branch in the stack:

- `"set"` (default): plain record tables (label, channel, policy, ...). The server copy is authoritative, so fetches overwrite rows.
- `"if-absent"`: the dispatch-backed document tables (arc, lineplot, log, panel, schematic). Their rows are maintained by replaying local edit actions, so a fetch must never clobber a doc holding locally replayed edits. The domain's `hydrate()` decides when a fresh network doc replaces the cached one.

emilbon99: Basically it came down to problems with 'echoed` sets to data structures clobbering correct optimistic updates. Say you do create a line plot and then quickly (optimistically) dispatch an action to it. Its possible that the action updates optimistically, and then only _later_ the create signal arrives, overriding the update you made. 

I think there may be a future in which we can remove it, but it's not really a high consequence decision and its a robust solution to the problem, so I say we leave it

client/ts/src/query/table.ts
emilbon99: Maybe, but I think that doing deep equality checks is still more efficient than doing a bunch of unnecessary re-renders, and our implementation of deep equal is pretty efficient, especially for smaller data structures.

My take is: maybe and it may be worth benchmarking in the future. That said, console performance feels pretty good at the top of this stack, and I think that most of the low hanging performance gains are not from this.

client/ts/src/query/table.ts
emilbon99: Optimistic partial updates against an uncached row. Imaging you get a listener change for a key that is not in cache and you want to partially update it:

you do `cache.set(key, (v) => v == null ? undefined : {...prev, ...next})`.  If `v` is not in cache it will be undefiend and we just skip the update.

client/ts/src/query/table.ts
emilbon99: My take is that it should stay for two reasons:

1) It makes the API tighter and nicer to use, and we already have many many call sites using this. Marginal, but somewhat real.
2) In the future its likely that we'll do more optimistic storage of these data structures and keep them in things that are not arrays, so if we make that change all calls to `.get().filter()` will either be (a) broken or (b) sub-optimal as it requires conversion of the entire cache to a list first.

client/ts/src/query/streamer.ts
emilbon99: I'm going to do it in the same ticket I linked about the channel retriever above: https://linear.app/synnax/issue/SY-4533/resolve-framer-channel-lookups-through-the-query-cache

client/ts/src/query/retriever.ts
emilbon99: I've already done several passes to structurally removed type assertions, and I just did another pass. I also went through and did a deep review that all assertions won't result in weird bugs, they just help the typescript compiler see things that it normally can't.

There are less assertions now than there used to be in the production code, and I'm very confident that the query and cahche implementation is sound.

emilbon99:
> Is `query` the right name for the package if it will also be used for mutations?

I think its the best one for now. I'm open to changing it in the future but it kind of standardizes with `SQL` where a SQL `INSERT` is still a type of query ya know.

client/ts/src/query/query.ts
emilbon99: *Emiliano + Claude: generated reply from our review session.*

`ready` holds the answer's membership, so a request query that matches many records keeps all of their keys. `loading` isn't per-record either, the promise is for the whole answer, and it's what lets concurrent retrieves join the same in-flight fetch instead of firing duplicates.

`deleted` is the only one that's intentionally single. An answer can only "be deleted" when the query addresses exactly one record (exact-key queries, or spaces marked `single`). A multi-record answer never flips to deleted, it just loses members as rows get evicted. The key on `deleted` points at the tombstone so we can compose the corpse at read time instead of caching a stale copy of it.

emilbon99:
@greptileai review

client/ts/src/query/cache.ts
```diff
@@ -0,0 +1,344 @@
...
+  ): (() => Promise<void>) =>
+  async () => {
+    const keys = table.keys();
+    if (keys.length === 0) return;
+    const values = await fetchSurvivors(fetch, keys);
+    const present = new Set<Key>(values.map(({ key }) => key));
+    const vanished = keys.filter((k) => !present.has(k));
+    if (vanished.length > 0) table.delete(vanished);
+    if (values.length > 0) table.set(values);
+  };
```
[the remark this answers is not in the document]
emilbon99: *This reply was generated by Claude Code on Emiliano's behalf.*

Fixed in a1b752b309 and fb05cfaded: `Table` now bumps a private generation on `reset()` and discards any fetch or reconcile that completes across it, with regression specs for both the mid-fetch retrieve and the mid-reconcile reset. Reconciliation moved into `Table` so the generation never leaves the class.

emilbon99:
@greptileai review

pjdotson (approved):

PR #2624, opened 2026-07-24
E. SY-4493: Cut pluto flux over to the client query cache

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Fifth PR in the [SY-4493](https://linear.app/synnax/issue/SY-4493/robust-session-state-management-for-the-console) stack (base: `sy-4493-connection`). The atomic pluto flux cutover: pluto stops running its own store and reads through the client-side query cache instead.

**What lands here**

* The flux factories are rewritten: `createRetrieve` / `createList` / `createUpdate` now delegate to `client.<domain>` (`retrieve` / `subscribe` / `getCached`) rather than a pluto-side store with `mountListeners`. Every domain `queries.ts` is rebound onto the new factory.
* `flux/base` is deleted entirely. The `state` namespace moves from `@/state` to `@synnaxlabs/x`.
* The legacy raw-send `dispatch` overload is removed from the six domain clients now that pluto drives the controller form.
* Console adapts to the removed pluto surface (panel selection hooks, direct flux-store access).

This is one atomic cutover because the factory config-shape change forces every domain to conform at once.

**Verification**<br>Client 1422/1422; full pluto suite 3398/3398 (the cutover re-enables the panel Mosaic test skipped in PR C); full console suite 1998 passed / 2 skipped. All packages typecheck clean.

Two console tests are `it.skip`ped pending the session PR: the login rejected-credentials flow (routes through the session Synnax provider) and one panel-selection reconcile test (needs the reconcile synchronizer). Both re-enable in the session-state PR.

## Basic Readiness

### Greptile Summary

This is the fifth PR in the [SY-4493](https://linear.app/synnax/issue/SY-4493/robust-session-state-management-for-the-console) stack and performs the atomic cutover where pluto stops maintaining its own Redux-style store and instead reads directly through the client-side query cache. The pluto flux factories (`createRetrieve`, `createList`, `createUpdate`, `createDispatch`) are rewritten to delegate to `client.<domain>` methods (`retrieve`, `subscribe`, `getCached`, `dispatch`) rather than a pluto-side undoable store with channel listeners; `flux/Provider.tsx`, `flux/aether/provider.ts`, and the `flux/base` subtree are deleted entirely.

* Every domain's `queries.ts` (panel, schematic, arc, log, lineplot, table, ranger, channel, etc.) is migrated to the new factory shape, removing `FluxStore`/`FluxSubStore` types and the `storeKey`/`mountListeners` API.
* The `state` namespace is moved from `@/state` to `@synnaxlabs/x`, and the legacy raw-send `dispatch` overload is removed from all six domain clients.
* Console adapts: panel selection hooks drop the live-tree resolution (deferred to the upcoming session/reconcile-synchronizer PR, tracked by two skipped tests), and `useCloseResourceTabs` changes from an auto-subscribing hook to an explicit callback invoked by delete flows.

### Confidence Score: 5/5

Safe to merge. The only finding is a dead hook call in Mosaic.tsx that does nothing and has no behavioural impact.

The architecture shift is large but well-contained: every domain's queries file is consistently migrated, the removed base store is fully deleted (no partial remnants), the client query cache correctly handles fetch deduplication, live subscriptions, and the deleted-record case. The two intentionally skipped tests are accurately documented and scoped to a follow-on PR. The full test suites (client, pluto, console) pass. The single finding — a discarded return value from useCloseResourceTabs in Mosaic — is dead code with no runtime effect.

**Files Needing Attention:** console/src/feature/panel/Mosaic.tsx — the dead Panel.useCloseResourceTabs() call should be removed.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| pluto/src/flux/retrieve.ts | Core retrieve refactor: drops pluto-side store, delegates fetch/subscribe/getCached to client domain. New LocalCache per-client WeakMap deduplicates in-flight suspense fetches cleanly. |
| pluto/src/flux/list.ts | List hook replaced store-backed listeners with per-page subscriptions (subscribe/subscribeByKey). Page model handles dedup, dropped keys, and client swaps correctly. |
| pluto/src/flux/dispatch.ts | Dispatch/undo/redo hooks now delegate directly to the domain client instead of a scoped store; INERT_TX type broadened from Transaction to Transaction for correct fallback assignment. |
| pluto/src/flux/form.ts | Form factory simplified: Store generic and scope removed; rollback array removed (rollbacks were previously managed by the form itself, now the client layer handles them). |
| pluto/src/panel/queries.ts | Panel domain fully migrated: FluxStore/FluxSubStore removed, all reads/writes go through client.panels; useCloseResourceTabs signature changed to explicit callback rather than auto-subscribing synchronizer. |
| console/src/feature/panel/Mosaic.tsx | Hook call updated to useCloseResourceTabs, but the return value is discarded — the call is now dead code since the hook no longer carries side effects. |
| console/src/session/panel/selectors.ts | useSelectSelectedTabs simplified by removing live-tree resolution (Panel.useSelectSelection); intentional regression pending the reconcile-synchronizer in the session PR, tracked by a skipped test. |
| pluto/src/flux/Provider.tsx | Provider.tsx deleted entirely as part of the flux base removal; its store-scoping context is no longer needed since state now lives in the client query cache. |
| pluto/src/flux/errors.ts | New file introducing DeletedError (a typed FluxError subclass) surfaced to error boundaries when a cached answer is a deletion; carries the last known value as corpse. |
| console/src/platform/panel/useOpenTab.ts | afterOptimistic now dispatches internalSelectTab directly (panel not yet retrievable at optimistic time), correctly replacing the now-removed selectTab helper that required a cached panel. |

### Flowchart

```mermaid
flowchart TD
    subgraph Before["Before (pluto-side store)"]
        A1["React hook\n(useRetrieve/useList)"] --> B1["Flux.Provider\n(base.Client)"]
        B1 --> C1["UndoableUnaryStore\n(per-domain)"]
        C1 --> D1["Channel listeners\n(mountListeners)"]
        D1 --> E1["client.domain\nHTTP/WS"]
    end
    subgraph After["After (client query cache)"]
        A2["React hook\n(useRetrieve/useList)"] --> B2["Synnax.use()\n(direct client ref)"]
        B2 --> C2["client.domain\n.retrieve / .getCached\n.subscribe / .dispatch"]
        C2 --> D2["Client query cache\n(per-client WeakMap)"]
        D2 --> E2["HTTP/WS cluster"]
    end
    style Before fill:#ffeaea
    style After fill:#eaffea
```

Reviews (3): Last reviewed commit: ["Merge branch < />sy-4493-connection< /> into s..."](<https://github.com/synnaxlabs/synnax/commit/400cae1c7ffdeaa76fb48f90489c65c751999de8>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=47079372>)


pluto/src/flux/list.ts
```diff
@@ -259,73 +154,141 @@ export const createList =
       [listItemListeners.current],
     );
```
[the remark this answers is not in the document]
emilbon99: The parameter names already reflect the semantic roles: `key` is the data key compared against `changed`, `notify` is the callback. As you note, the logic is correct. Leaving as-is.

console/src/platform/tree/Tree.tsx
```diff
@@ -273,30 +291,57 @@ const Internal = ({ root, emptyContent }: InternalProps): ReactElement => {
...
+        listeners.delete(callback);
+        disconnect?.();
+      };
+    },
+    [client],
+  );
+
+  const getItem = useMemo(
+    () =>
+      List.createGetItem<string, ontology.Resource>(getResourceByKey, (keys) =>
```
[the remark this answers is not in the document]
emilbon99: Fixed. Removed the per-subscription `onResourceSet` listener and folded the fan-out into the single shared `useResourceSetSynchronizer`, which now dispatches to the matching key's listener set via `placeholderListenersRef`. That restores O(1)-per-event notification; `subscribe` just manages the set.

pluto/src/flux/list.ts
```diff
@@ -259,73 +154,141 @@ export const createList =
...
+            },
+            (cached: query.Cached<Data[]> | undefined) => {
+              if (cached?.variant !== "changed") return;
+              applyPageAnswer(page, cached.data);
+            },
+          );
+        pagesRef.current.push(page);
+      },
+      [client, applyPageAnswer],
+    );
```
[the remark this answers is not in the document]
emilbon99: A list answer space only settles to `ready` (recomposed keys) and emits `changed`; the `deleted` variant is reachable only for single-record spaces (`single === true`), not list queries. Mid-session invalidation recomposes to `changed` with the updated key set, and reconnect is handled by the `useEffect([client])` refetch, so there's no stale-key path to guard here.

emilbon99:
@greptileai review

emilbon99:
@greptileai review

pluto/src/access/role/queries.ts
```diff
@@ -23,69 +20,37 @@ export type RetrieveQuery = {
...
-  role.FluxSubStore
->({
+export const { useRetrieve } = Flux.createRetrieve<RetrieveQuery, access.role.Role>({
   name: RESOURCE_NAME,
-  retrieve: retrieveSingle,
-  mountListeners: ({ store, query: { key }, onChange }) => [
-    store.roles.onSet(onChange, key),
-  ],
+  retrieve: async ({ client, query: { key } }) =>
+    await client.access.roles.retrieve({ key }),
```
pjdotson: is there a reason you have to destructure `query`, shouldn't you just be able to call `await client.access.roles.retrieve(query)` for 90% of these?

pluto/src/arc/queries.ts
```diff
@@ -115,36 +40,47 @@ export interface SelectKeyParams {
...
-  return a;
+const requireArc = (client: Synnax | null, key: arc.Key): arc.Arc => {
+  const cached = client?.arcs.getCached({ key });
+  if (cached == null || query.Deleted.matches(cached))
+    throw new NotFoundError(`Arc with key ${key} not found`);
+  return cached;
+};
+
+const getArc = (client: Synnax | null, key: arc.Key): arc.Arc | undefined => {
+  const cached = client?.arcs.getCached({ key });
```
pjdotson: I don't get why things like this `requireArc` or `getArc` have the option to take in `null` for the client - shouldn't that be handled before these functions get called that way there is a consistent error message for a null client on every service?

pluto/src/arc/queries.ts
```diff
@@ -306,262 +217,67 @@ export const ZERO_FORM_VALUES: z.infer<typeof formSchema> = formSchema.parse({
...
-  retrieve: async ({ client, query, reset, store }) => {
-    if (!("key" in query) || primitive.isZero(query.key)) return;
-    reset(await retrieveSingle({ client, query: query as RetrieveQuery, store }));
+  retrieve: async ({ client, query: { key, ...rest }, reset }) => {
+    if (key == null || primitive.isZero(key)) return;
+    // Prefer the cached copy: it may hold locally replayed edits ahead of the
+    // server.
+    const cached = client.arcs.getCached({ key });
+    if (cached !== undefined && !query.Deleted.matches(cached)) return reset(cached);
+    reset(await client.arcs.retrieve({ key, ...rest }));
```
pjdotson: why isn't this handled by the Client code?

pluto/src/flux/aether/external.ts
```diff
@@ -7,8 +7,4 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-export * from "@/flux/aether/provider";
-export * from "@/flux/base/client";
-export * from "@/flux/base/store";
-export * from "@/flux/base/streamer";
-export * from "@/flux/base/types";
+export * from "@/flux/aether/verbs";
```
pjdotson: do we still need a separate `flux/aether` package?

x/ts/src/verbs/verbs.ts
```diff
@@ -0,0 +1,62 @@
+// Copyright 2026 Synnax Labs, Inc.
+//
+// Use of this software is governed by the Business Source License included in the file
+// licenses/BSL.txt.
+//
+// As of the Change Date specified in that file, in accordance with the Business Source
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+export interface Verbs {
```
pjdotson, line 11: IMO this belongs in X

pluto/src/flux/update.ts
```diff
@@ -328,13 +240,11 @@ const useObservable = <
       }
     },
     [
-      maybeClient,
-      allowDisconnected,
```
pjdotson, line 332: If we have gotten rid of the `allowDisconnected` how do we allow updating the version when no Core is connected now?

pjdotson (changes requested):

pluto/src/arc/queries.ts
emilbon99: I agree, I have a plan on doing a refactor of the public facing flux API in a PR immediately following this (things like preferring suspension, reducing the number of query variants, making things more efficient, etc). I'd like to do this in that PR: https://linear.app/synnax/issue/SY-4492/flux-simplification-and-tuning

pluto/src/flux/update.ts
emilbon99: I decided that the version updater shouldn't use flux at all, and migrated it off of it.

client/ts/src/query/query.ts
```diff
@@ -46,10 +46,25 @@ export const hash = (query: Params): string => {
  * A cached answer as delivered to consumers: the query's current data, or the
  * {@link Deleted} corpse of a record that was deleted. A corpse is never
  * delivered as bare data, so deletion cannot be mistaken for a live result.
- * Narrow with {@link Deleted.matches}.
+ * Narrow with {@link isLive}, or {@link Deleted.matches} to reach the corpse.
  */
 export type Cached<D extends Data> = D | Deleted<D>;
 
+/** True when a cached answer is live: cached at all, and not a deletion. */
+export const isLive = <D extends Data>(value: Cached<D> | undefined): value is D =>
```
pjdotson, line 54: add tests

pluto/src/access/role/queries.ts
```diff
@@ -123,82 +79,32 @@ export const changeRoleFormSchema = z.object({
...
-  return r.key;
+const retrieveUserRole = async (
+  client: Synnax,
+  userKey: user.Key,
+): Promise<access.role.Key | undefined> => {
+  const parents = await client.ontology.parents.retrieve({
+    ids: user.ontologyID(userKey),
+    types: ["role"],
+  });
+  return parents.at(0)?.id.key;
```
pjdotson, line 90: every user should have a role right - maybe we should error here instead of potentially returning `undefined`?

pluto/src/arc/queries.ts
pjdotson: sounds good

console/src/platform/arc/useCreate.ts
```diff
@@ -21,20 +20,14 @@ export interface UseCreateProps {
...
-    ),
-    afterOptimistic: useCallback(
-      ({ data: { key } }: Flux.AfterSuccessParams<arc.Arc>) =>
-        openTab({ variant: "resource", resource: arc.ontologyID(key), key: tabKey }),
-      [openTab, tabKey],
-    ),
-  });
-  return useCallback(() => update({ name: "Arc Editor", mode: "graph" }), [update]);
+  return useCallback(() => {
+    void openModal({}).then((result) => {
```
pjdotson: shouldn't you `catch` here for potential errors?

console/src/platform/arc/useCreateModal.tsx
```diff
@@ -68,78 +76,73 @@ const ArcModeSelectButton = ({
...
-                itemKey="text"
-                icon={<Icon.Text />}
-                title="Text"
-                description="Text-based editor that is best for complex automations such as control sequences"
-              />
-            </Select.Buttons>
-          </Input.Item>
+          <Form.Form<typeof Arc.formSchema> {...form}>
+            <Form.Field<string> path="name" required>
+              {(p) => (
```
pjdotson: should memoize some of these ghings

console/src/feature/log/toolbar/testutil.tsx
```diff
@@ -58,6 +58,26 @@ export interface RenderLogOptions {
   preloadedState?: (key: string) => ConsolePreloadedState;
 }
 
+// createResourceTab creates a single-leaf panel holding one resource tab that backs
+// the given log on the cluster, so the panel scope hooks a mounted tab content reads
+// (useSelectTabResource) resolve to the log's ontology ID.
+const createResourceTab = async (
```
pjdotson: this logic is duplicated a bunch, should just be a function that take in an ontology ID

console/src/feature/log/toolbar/testutil.tsx
```diff
@@ -58,6 +58,26 @@ export interface RenderLogOptions {
...
+  const doc = panel.panelZ.parse({
+    name: uniqueName("panel"),
+    root: {
+      variant: "leaf",
+      tabs: [{ variant: "resource", key: tabKey, resource: clientLog.ontologyID(key) }],
+    },
+  });
+  await client.panels.create(doc);
+  // Prime the query cache the way the mosaic's retrieve does.
+  await client.panels.retrieve(doc.key);
```
pjdotson: 2 things:

1) why do you need to call `panel.panelZ.parse`, shouldn't that be handled internally by `panels.create`
2) why do you need to prime the query cache, we don't add newly created items to the cache?

pjdotson (changes requested):

pluto/src/access/role/queries.ts
emilbon99: Should and does are two different things. In the client you can create users without roles. Do you still wan tme to error on this?

pjdotson (approved):

PR #2625, opened 2026-07-24
F. SY-4493: Session state synchronizers and tombstone UX

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Final PR in the [SY-4493](https://linear.app/synnax/issue/SY-4493/robust-session-state-management-for-the-console) stack (base: `sy-4493-flux`). Completes RFC 0046 phase 3. Panel multi-window support (tab tear-off, Open in New Window, cross-panel tab drag, RFC 0049) originally landed on this branch and has been moved up the stack to `sy-4511-panel-ux-improvements-part-1`, where it is introduced alongside its polish.

**What lands here**

* Console **session synchronizers**, one per domain (panel, project, range, schematic, status, arc, lineplot, log, table), projecting the client cache's epochs into console state. They are wired through `Session.useSynchronizers` and the `SessionSynnaxProvider`.
* **Session persistence** across L0/L1/L2.
* The panel **tombstone / restore UX**: closed resource tabs are recoverable, and the reconcile synchronizer repairs stored selections against the live tree.
* Re-enables the two tests deferred by PRs D and E (login rejected-credentials and panel-selection reconcile).

**Verification**<br>Console typecheck clean; full console suite 2026/2026 (both previously-deferred tests now pass; one unrelated schematic symbol test flaked under suite contention and passes in isolation). Client and pluto are unchanged from PR E.

## Basic Readiness

### Greptile Summary

This PR completes RFC 0046 phase 3 by adding per-domain session synchronizers, three-scope (global/cluster/project) persistence partitioning, and the panel tombstone/restore UX.

* **Session synchronizers** (`arc`, `cluster`, `lineplot`, `log`, `panel`, `project`, `range`, `schematic`, `status`, `table`) are wired through a new `Synchronizer.use` host that mounts listeners once per client, runs sequential reconcile passes at epoch boundaries, and exposes a `verified` boolean that gates the `ConnectionGuard` splash screen via the new `Settled.Provider`.
* **Partitioned persistence** replaces the single flat ring with per-scope partitions keyed by cluster and project; context switches flush the outgoing partition, load the incoming one, and apply it atomically via a `hydrate` action that the root reducer handles with `Drift.restoreWindows`.
* **Tombstone UX** renders a deleted resource tab with the corpse name plus Close/Restore buttons inside an error boundary fallback; restore re-creates the document under its original key so all existing references continue to work.

### Confidence Score: 5/5

Safe to merge — the synchronizer, persistence, and tombstone paths are well-structured, the race conditions in the partition swap are handled correctly with generation counters and swapGen, and the previously-deferred tests now pass.

The generation counter in Synchronizer.use prevents stale reconcile completions from marking the session verified, the swap logic correctly flushes the outgoing partition before loading the incoming one, and Drift.restoreWindows is careful to zero runtime-specific counters. The one minor observation about the ring-version pointer advancing on a failed KV write does not affect normal operation.

**Files Needing Attention:** console/src/session/persist/state.ts — Partition.write advances the ring-version pointer even when the underlying KV write fails, which would leave the revert chain pointing at an empty slot on a disk error.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| console/src/session/persist/state.ts | Rewrites persistence to partition state into global/cluster/project scopes with ring versioning per partition; adds Engine class, swap/hydrate middleware, and context-keyed read/write. |
| console/src/session/synchronizer/use.ts | New synchronizer host: mounts listeners once per client, runs reconcile passes at epoch boundaries, returns verified boolean; uses generation counter to prevent stale-pass completions. |
| console/src/session/panel/synchronizer.ts | New panel synchronizer with three sub-synchronizers: panel deletion pruning, selection repair vs. live project tree, window title sync, and tab-selection convergence to the live panel tree. |
| console/src/session/settled/Settled.tsx | New context provider and hook signalling workspace readiness: epoch >= 1, cluster key matches connection, no partition swap in flight, and first reconcile pass completed. |
| console/src/session/store.ts | Wires up the new partitioned persistence: defines PERSIST_SCOPES and getPersistContext, adds root reducer wrapper that handles hydrate via Drift.restoreWindows, adds Persist slice to combined reducer. |
| console/src/feature/panel/Mosaic.tsx | Adds tombstone UX: deleted-resource tabs show corpse name with Close/Restore buttons inside an error boundary fallback; non-resource deleted tabs fall through to the standard error display. |
| console/src/feature/auth/ConnectionGuard.tsx | Replaces grace-period takeover with settled-based splash; adds Splash/Trouble split so connecting shows a spinner and first-epoch errors show retry controls. |
| drift/src/middleware.ts | Broadens shouldSync to fire on any action that changes the drift slice (prevS !== nextS), enabling hydrate (a non-drift action) to trigger window restoration through the runtime. |
| drift/src/state.ts | Adds restoreWindows: merges persisted secondary windows into the live slice, keeping the main window and pre-renders from current, zeroing runtime-specific counters on restored windows. |
| console/src/platform/panel/restore.ts | New restore module mapping ontology types to restorer functions that re-create deleted documents from their corpses under their original keys. |
| console/src/platform/tree/createUseDelete.ts | Moves closeTabs from beforeUpdate to afterSuccess so tabs close only after the server confirms the delete; adds closeTabs to afterSuccess useCallback deps. |
| console/src/session/panel/slice.ts | Adds reconcileSelection action and useStartOverlaying hook; changes remove to cross-window operation and removes it from the window-key injection middleware; StoreState now extends Drift.StoreState. |

### Sequence Diagram

```mermaid
sequenceDiagram
    participant App
    participant Settled.Provider
    participant Synchronizer.use
    participant ConnectionGuard
    participant PersistMiddleware
    participant Engine

    App->>Settled.Provider: mount
    Settled.Provider->>Synchronizer.use: useSynchronizers()
    Synchronizer.use->>Synchronizer.use: listen() per synchronizer
    Synchronizer.use-->>Settled.Provider: "verified=false"
    Note over ConnectionGuard: settled=false shows Splash
    Synchronizer.use->>Synchronizer.use: "epoch>=1 triggers reconcile pass"
    Synchronizer.use-->>Settled.Provider: "verified=true"
    Note over ConnectionGuard: settled=true renders workspace
    Note over PersistMiddleware: User switches project p1 to p2
    PersistMiddleware->>PersistMiddleware: detectContextChange
    PersistMiddleware->>Engine: persist(state, oldCtx c1 p1)
    PersistMiddleware->>App: "dispatch beginSwap swapping=true"
    Note over ConnectionGuard: settled=false shows Splash again
    Engine->>Engine: write global+cluster(c1)+project(c1,p1)
    PersistMiddleware->>Engine: loadSwap(state, c1 p2, false)
    Engine->>Engine: read project(c1,p2)
    PersistMiddleware->>App: "dispatch hydrate swapping=false"
    Note over ConnectionGuard: settled=true renders p2 workspace
```

Reviews (3): Last reviewed commit: ["Fix floating act dispatches in the logou..."](<https://github.com/synnaxlabs/synnax/commit/05b0fba93daeccadddea2fb5a5e8094ec4dc2a5d>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=47080577>)


console/src/session/Context.tsx
```diff
@@ -26,3 +36,36 @@ export const Context = ({ children }: ContextProps) => {
...
+  ...Table.SYNCHRONIZERS,
+};
+
+/**
+ * Mounts every registered session synchronizer hook. Must be called below the
+ * Pluto providers, in the main window only.
+ */
+export const useSynchronizers = (): void => {
+  Object.values(SYNCHRONIZERS).forEach((useSynchronizer) => useSynchronizer());
+};
```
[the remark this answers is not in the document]
emilbon99: `SYNCHRONIZERS` is a module-level const, so the iteration order is fixed and every hook is mounted unconditionally on every render. Hook order is therefore stable, which is the invariant the Rules of Hooks protect. ESLint's `react-hooks/rules-of-hooks` doesn't flag it (verified lint-clean), and this is the established pattern for the synchronizer set. A future conditional synchronizer would be the thing to catch in review, not this call site.

console/src/feature/panel/Mosaic.tsx
```diff
@@ -29,19 +32,77 @@ import { useDispatch } from "react-redux";
...
+      await PlatformPanel.restore(resource, { client, project, corpse });
+      resetErrorBoundary();
+    }, `Failed to restore ${name}`);
+  };
+  return (
+    <Flex.Box grow align="center" justify="center" gap="small">
+      <Icon.Warning />
+      <Text.Text>{name} was deleted</Text.Text>
+      <Flex.Box x gap="small">
+        <Button.Button onClick={() => closeTabs(resource)}>Close</Button.Button>
```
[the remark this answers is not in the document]
emilbon99: `Session.Project.useSelectSelected()` returns a guaranteed non-null `project.Key`; it throws (outside a `Project.Guard`) rather than returning `undefined`. The panel mosaic always renders within a project context, so `project` is never absent when the Restore button is shown, and the failing-call scenario can't occur.

console/src/session/store.ts
```diff
@@ -135,6 +171,29 @@ interface OpenPersistReturn {
...
+    store.dispatch(Drift.closeWindow({ key: win.key }));
+  });
+  if (stored == null) return loaded;
+  Object.values(stored.windows).forEach((win) => {
+    if (win.key === Drift.MAIN_WINDOW || !win.reserved) return;
+    store.dispatch(Drift.createWindow(Drift.windowPropsZ.parse(win)));
+  });
+  return loaded;
+};
+
```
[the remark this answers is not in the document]
emilbon99: Fixed. Switched to `safeParse` per stored window and skip the unparseable entries, so a window on a stale schema can no longer abort the swap after the current windows have already closed. The target's valid windows all reopen; only genuinely corrupt persisted entries are dropped.

emilbon99:
@greptileai review

console/src/session/range/synchronizer.ts
```diff
@@ -0,0 +1,56 @@
...
+    // Unpersisted ranges are local-only; only cluster-backed ones can vanish.
+    const keys = store
+      .getState()
+      .range.ranges.filter(({ persisted }) => persisted)
+      .map(({ key }) => key);
+    if (keys.length === 0) return;
+    const found = await client.ranges.retrieve(keys);
+    const foundKeys = new Set(found.map(({ key }) => key));
+    const missing = keys.filter((key) => !foundKeys.has(key));
+    if (missing.length > 0) dispatch(remove({ keys: missing }));
```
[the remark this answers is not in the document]
emilbon99: Fixed. The epoch sweep now passes `{ keys, ignoreNotFoundError: true }`, matching every sibling synchronizer, so a deleted range key no longer throws and aborts pruning.

console/src/platform/tree/createUseDelete.ts
```diff
@@ -42,15 +42,19 @@ export const createUseDelete =
...
         async (query: Flux.BeforeUpdateParams<K | K[]>) => {
           const res = await confirm(getResource(ids));
           if (!res) return false;
-          if (beforeUpdate != null) return await beforeUpdate({ ...query, ...props });
-          return true;
+          let result: K | K[] | boolean = true;
+          if (beforeUpdate != null) result = await beforeUpdate({ ...query, ...props });
+          // The deleting console closes its own tabs; remote consoles tombstone.
+          if (result !== false) closeTabs(ids);
+          return result;
```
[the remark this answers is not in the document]
emilbon99: Fixed. Moved `closeTabs` from `beforeUpdate` into `afterSuccess`, so tabs close only after the server confirms the delete; a failed delete now leaves the tabs open instead of orphaning them. The remote-console tombstone path is unchanged.

emilbon99:
@greptileai review

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
```diff
@@ -0,0 +1,491 @@
...
+## 5.8 - Scoped persistence
+
+Persisted session state is partitioned by scope and swapped on context switch. The live
+store shape never changes; the persistence layer keys partitions by scope:
+
+| Scope     | Slices                                                               |
+| --------- | -------------------------------------------------------------------- |
+| `global`  | cluster, color, docs, theme                                          |
+| `cluster` | project, status                                                      |
+| `project` | arc, drift, haul, lineplot, log, nav, panel, range, schematic, table |
```
pjdotson: why does `range` belong in `project` but `status` belongs in `cluster`? And `cluster` should probably be scoped as `core`.

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
```diff
@@ -0,0 +1,491 @@
...
+Global exists outside any cluster: preferences plus the cluster registry and selection
+needed to pick a context. The cluster scope remembers, per cluster, which project was
+active and which statuses are favorited. The project scope is the workspace: window
+arrangement, panel and tab selection, per-document view state. The chain is deliberate:
+global names the cluster, the cluster partition names the project, the project partition
+holds the workspace, which is also why logout clears the cluster last, after the project
+selection it flushes is already cleared.
+
+- **Partitions are structured key prefixes in the one existing KV store** (`global.<v>`,
+  `cluster.<key>.<v>`, `project.<cluster>.<project>.<v>`), each with its own version
```
pjdotson: why does the project scope have its own cluster scope as well?

integration/tests/arc/lifecycle.py
```diff
@@ -244,6 +244,8 @@ def verify_sequence_execution(self) -> None:
...
         self.remove_arc(old_name)
 
-        self.log("Verifying tab removed from mosaic")
-        tab = self.console.layout.get_tab(self.new_name)
-        tab.wait_for(state="hidden", timeout=5000)
+        # A deleted resource's tab stays open and tombstones; it is not closed
+        # out from under the user. The tombstone renders inside that tab.
+        self.log("Verifying tab tombstones in place")
+        tombstone = self.console.layout.get_tombstone(self.new_name)
+        tombstone.wait_for(state="visible", timeout=5000)
```
pjdotson, line 251: should have other integration tests written for tombstoning behavior

client/ts/src/project/client.ts
```diff
@@ -30,6 +30,7 @@ const retrieveReqZ = z.object({
   searchTerm: z.string().optional(),
   offset: z.int().optional(),
   limit: z.int().optional(),
+  ignoreNotFoundError: z.boolean().optional(),
```
pjdotson, line 33: shouldn't this not be public?

console/src/feature/lineplot/link.spec.ts
```diff
@@ -37,7 +37,7 @@ describe("LinePlot.useLink", () => {
     await waitFor(async () => {
       const panelKey = Session.Panel.selectSelected(store.getState());
       expect(panelKey).toBeDefined();
-      const doc = await client.panels.retrieve(panelKey as string);
+      const doc = await client.panels.retrieve({ key: panelKey as string });
```
pjdotson: is there a fundamental reason to remove the single call to retrieve where we just pass a single key for all of the APIs? seems like we have made our API's worse by not allowing you to just pass in a string for the key.

pjdotson (changes requested):
should "close" here be red?

Image

also somewhat odd to have rename options for deleted things here:

Image

And maybe the tab name should be yellow / red in the tab selector if we have a tombstoned component.

Also, your plus button next to new panel is still broken and makes playing around with stuff harder

client/ts/src/project/client.ts
emilbon99: It has to be public right now, unfortunately, because the console needs it for state reconciliation. I think we should just refactor to structurally remove this in the futrue.

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
emilbon99: I moved status to being project scoped.

As far as the `cluster` scope, what should we do with the slice naming itself? Should everything o from cluster to core?

console/src/feature/lineplot/link.spec.ts
emilbon99: Implemented.

emilbon99:
> should "close" here be red?
> 
> Image
> also somewhat odd to have rename options for deleted things here:
> 
> Image
> And maybe the tab name should be yellow / red in the tab selector if we have a tombstoned component.
> 
> Also, your plus button next to new panel is still broken and makes playing around with stuff harder

1) Implemented
2) This refactor is going to have to wait a smidge on some tab design changes
3) The plus button is completely fixed on the UI refactor

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
```diff
@@ -0,0 +1,492 @@
+# 46 - Client Cache, Unified Reads, and Console Session State
```
pjdotson, line 1: what is the point of putting the RFC in like 10 PRs after this feature has already gotten introduce?

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
```diff
@@ -0,0 +1,492 @@
...
+   `onChange`, and `getCached`. Flux becomes a React binding layer over that surface; it
+   holds no cache of its own.
+3. Session state converges to document truth through dispatched actions, never through
+   intentional divergence. Deleted-while-open documents tombstone in place instead of
+   tearing down views. Persisted session state is partitioned into global, per-cluster,
+   and per-cluster-per-project scopes swapped on context switch.
+
+# 1 - Motivation
+
+Deletion was the sharpest edge of the split brain. A remote delete reached a suspense
```
pjdotson: Prose like "Deletion was the sharpest edge of the split brain" reads like slop, these RFCs need to be cleaned up and shortened

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
```diff
@@ -0,0 +1,492 @@
...
+onChange(params, handler): destructor.Destructor; // query-scoped subscription
+getCached(params): Cached<D> | undefined;         // synchronous, no promise
+
+type Cached<D> = D | Deleted<D>;
+```
+
+- **`retrieve`** hashes the query, dedupes against an in-flight promise, resolves
+  instantly on a maintained hit, and on a miss fetches, populates the table, and settles
+  the entry. A deleted key rejects with a plain `NotFoundError`: mount-after-delete is
+  "not found". Errors never flow through `onChange`.
```
pjdotson: we would only return `NotFoundError` iff a requested key is not found, matching the server behavior?

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
```diff
@@ -0,0 +1,492 @@
...
+only member; new uses require the same justification in review.
+
+## 5.8 - Scoped persistence
+
+Persisted session state is partitioned by scope and swapped on context switch. The live
+store shape never changes; the persistence layer keys partitions by scope:
+
+| Scope     | Slices                                                                       |
+| --------- | ---------------------------------------------------------------------------- |
+| `global`  | cluster, color, docs, theme                                                  |
```
pjdotson, line 305: what is `color`?

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
```diff
@@ -0,0 +1,492 @@
...
+Global exists outside any cluster: preferences plus the cluster registry and selection
+needed to pick a context. The cluster scope remembers, per cluster, which project was
+active. The project scope is the workspace: window arrangement, panel and tab selection,
+per-document view state, and which statuses are favorited. The chain is deliberate:
+global names the cluster, the cluster partition names the project, the project partition
+holds the workspace, which is also why logout clears the cluster last, after the project
+selection it flushes is already cleared.
+
+- **Partitions are structured key prefixes in the one existing KV store** (`global.<v>`,
+  `cluster.<key>.<v>`, `project.<cluster>.<project>.<v>`), each with its own version
```
pjdotson: Why do we need these scoping rules - the UX is if we log out of Core 1 and log into Core 2, then we want to have the same exact session state and everything else as when we were last in Core 2?

integration/console/layout.py
```diff
@@ -120,6 +120,7 @@ class LayoutClient:
         ".console-mosaic .pluto-mosaic__leaf > .pluto-tabs > .pluto-tabs__selector"
     )
     TAB_SELECTOR = f"{TAB_STRIP_SELECTOR} > .pluto-tabs__tab"
+    TOMBSTONE_SELECTOR = ".console-panel__tombstone"
```
pjdotson: ideally we wouldn't look for this CSS selector

console/src/session/synchronizer/create.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Lifecycle callbacks keeping session state consistent with the cluster.
+ * The host runs reconcile at client-ready and on every epoch bump, and mounts
+ * listen once per client, tearing it down on client swap or unmount.
+ */
+// The any defaults erase the store shape at the registry boundary: the host
+// always calls with the full session store, which satisfies every
+// synchronizer's slice-composed S.
+export interface Synchronizer<S = any, A extends Action = any> {
```
pjdotson: no `any`

console/src/session/synchronizer/create.ts
```diff
@@ -0,0 +1,97 @@
...
+  reconcile: (params: Params<S, A>) => void | Promise<void>;
+  /** Mounts steady-state subscriptions: client feeds and store watches. */
+  listen?: (params: Params<S, A>) => destructor.Destructor;
+}
+
+/**
+ * Builds a synchronizer. The hook body only captures context (flux handles,
+ * adders); execution timing belongs to the host.
+ */
+export type Use<S = any, A extends Action = any> = () => Synchronizer<S, A>;
```
pjdotson: no `any`

console/src/session/synchronizer/use.ts
```diff
@@ -0,0 +1,86 @@
...
+    const params: Params = { client, store };
+    // A pass finishing after its cold reset must not count: it verified
+    // against a cluster the client no longer mirrors.
+    let generation = 0;
+    const destructors = entriesRef.current.flatMap(
+      ([, { listen }]) => listen?.(params) ?? [],
+    );
+    const reconcile = (): void => {
+      const gen = generation;
+      void (async () => {
```
pjdotson: should `useAsyncEffect`

console/src/session/settled/Settled.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Synnax } from "@synnaxlabs/pluto";
+import {
+  createContext,
+  type PropsWithChildren,
+  type ReactElement,
+  useContext,
+} from "react";
+
+import { Cluster } from "@/session/cluster";
+import { useSynchronizers } from "@/session/Context";
```
pjdotson: weird dependency inversion to import from the top-level `session`

console/src/session/settled/Settled.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+/**
+ * Whether the workspace state is verified against the connected cluster:
+ * first contact made, the session's cluster identity agrees with the
+ * connection's, no partition swap is in flight, and a reconcile pass has
+ * completed. While false the workspace is in structural doubt and must not
+ * render; the connection guard shows a splash instead.
+ */
+export const use = (): boolean => {
```
pjdotson: needs test

console/src/session/synchronizer/use.ts
```diff
@@ -0,0 +1,86 @@
...
+ * return-to-cold: false until the first pass finishes, reset on client change
+ * and on the epoch returning to 0 (cluster replacement).
+ */
+export const use = (synchronizers: Synchronizers): boolean => {
+  const store = useStore<unknown, UnknownAction>();
+  const client = Synnax.use();
+  const [verified, setVerified] = useState(false);
+  const entries: [string, Synchronizer][] = Object.entries(synchronizers).map(
+    ([key, useSynchronizer]) => [key, useSynchronizer()],
+  );
```
pjdotson: don't quite get why we pass in the `Synchronizers` object to then transform into an array of `Synchronizer` while technically breaking the rule of hooks instead of just taking in a `Synchronizer[]`

console/src/session/store.ts
```diff
@@ -85,6 +115,19 @@ export const reducer = combineReducers({
   [Theme.SLICE_NAME]: Theme.reducer,
 }) as unknown as Reducer<State, Action>;
 
+// hydrate replaces the swapped slices wholesale on a session context switch. Drift
+// merges instead, keeping the windows this process is actually running.
+export const reducer: Reducer<State, Action> = (state, action) => {
+  if (state != null && Persist.hydrate.match(action)) {
+    const payload = action.payload as Partial<State>;
```
pjdotson: shouldn't need type assertion

console/src/session/panel/slice.ts
```diff
@@ -41,7 +42,7 @@ export const sliceStateZ = z.object({
 
 export interface SliceState extends z.output<typeof sliceStateZ> {}
 
-export interface StoreState {
+export interface StoreState extends Drift.StoreState {
```
pjdotson, line 45: why do all of these need to `extends Drift.StoreState`?

pluto/src/theming/static/theme-light.css
```diff
@@ -682,6 +684,7 @@
     --pluto-dark-gray-l10-10: #dadada1a;
     --pluto-dark-gray-l10-5: #dadada0d;
     --pluto-dark-gray-l11: #fafafa;
+    --pluto-dark-gray-l11-rgb: 250, 250, 250;
```
pjdotson, line 687: why do we need this and the `--pluto-dark-gray-l11`?

console/src/feature/panel/useResetOnRestore.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Resets the surrounding error boundary once the tab's resource exists again,
+ * so a restore from any session heals a tombstoned surface without a manual
+ * reset.
+ */
+export const useResetOnRestore = (resetErrorBoundary: () => void): void => {
+  const resource = Panel.useSelectTabResource({});
+  const client = Synnax.use();
+  useEffect(() => {
```
pjdotson: shouldn't this go through the ontology flux API instead of some custom useEffect? also needs test

console/src/feature/panel/Toolbar.tsx
```diff
@@ -7,29 +7,54 @@
...
 );
 
+// The toolbar reads the same queries as the tab's content, so a deleted
+// resource throws here too. Show a quiet placeholder; the tombstone with Close
+// and Restore lives in the mosaic.
+const DeletedResourceContent = ({
+  resetErrorBoundary,
+}: Errors.FallbackProps): ReactElement => {
+  useResetOnRestore(resetErrorBoundary);
+  return <EmptyContent message="This resource was deleted." />;
```
pjdotson: Should it be `"This resource was deleted."` or `"This Arc was deleted."`

console/src/feature/auth/ConnectionGuard.tsx
```diff
@@ -120,82 +140,61 @@ const Takeover = ({ client, status, nav }: TakeoverProps): ReactElement => {
...
+          {activeKey != null && (
+            <Button.Button
+              variant="outlined"
+              onClick={() => openConnect({ clusterKey: activeKey })}
+            >
+              Edit Connection
+            </Button.Button>
+          )}
+          <Button.Button variant="outlined" onClick={logout}>
+            Log Out
```
pjdotson: These should probably be "Retry now", "Edit connection", and "Log out".

pjdotson (changes requested):
UX feels very good

console/src/session/panel/slice.ts
emilbon99: Because they need acess to window specifci state.

pluto/src/theming/static/theme-light.css
emilbon99: Not totally sure, this was introduced by @sy-nico  and he never re-renad the static theme generator.

integration/console/layout.py
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Agreed. Dropped `TOMBSTONE_SELECTOR` entirely. The tombstone shell now carries `role="group"` with its message as the accessible name, so the client locates it with `page.get_by_role("group", name=f"{name} was deleted", exact=True)` and scopes Restore/Close inside it.

Two things this buys beyond losing the class coupling: the role query ignores hidden elements, so the portaled contents of unselected tabs no longer match (the class selector did match them, which was a latent strict-mode failure once a second tab tombstoned), and the buttons stay scoped, which matters because the tab strip's own close buttons are also named "Close".

console/src/feature/panel/useResetOnRestore.ts
emilbon99: *Emiliano + Claude: generated reply from our review session.*

Right on both counts, and it went further than the ontology swap: the hook is gone, along with `fallback.tsx`.

The reason it existed at all was structural. Reading a deleted resource threw `DeletedError` to an error boundary, which unmounted the subtree and destroyed the very subscription that would have noticed the restore. So three call sites (tab name, tab content, toolbar) each mounted an effect to rebuild that observation from outside, and the only feed that worked for every resource type was `client.ontology.onChange`. Using the ontology as the change detector for typed resources is the wrong layer regardless; every domain already has its own change handler.

The fix reads the deletion as a value before the boundary is ever involved. `Flux.createRetrieve` now also returns `useTombstone`, which reads through the same `subscribe`/`getCached` the query is already bound to and hands back the corpse, or null while the record is live. `Panel.ResourceGuard` renders the tombstone when the tab's resource is deleted and swaps back on its own when it returns, so `handleRestore` no longer resets anything. Net effect is three effects and one ontology subscription per tombstoned tab removed, replaced by a plain read.

Tests, all three mutation-verified:

- `pluto/src/flux/retrieve.spec.tsx` -> the read flips to the corpse and back with no boundary reset and no refetch.
- `pluto/src/schematic/queries.spec.tsx` -> against a live cluster: delete, then re-create from the corpse, and the real client cache flips the tombstone back to live and notifies.
- `console/src/platform/panel/tab.spec.tsx` -> the guard swaps both ways through real tab scope and a real registry entry, asserting the live children are actually gone while the tombstone shows.

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
emilbon99: Tracks the most recently used colors. `src/session/color`

pluto/src/theming/static/theme-light.css
sy-nico: I added the `-rgb` version in #2440 (SY-4316) for the `oklch` automatic theme adjustment. `theme.css` does:

```css
oklch(from rgb(var(--pluto-symbol-color, var(--pluto-gray-l11-rgb))) ...)
```

`rgb()` needs comma separated components, so the fallback can't be the hex `--pluto-gray-l11`. Hence the second variable holding the same color in the other format.

The `--pluto-dark-` copy on this line isn't referenced anywhere. `generate.ts` has emitted the whole palette a second time under a `dark-` prefix since #2040, so every var added to `css.ts` produces two lines here. The one actually in use is `--pluto-gray-l11-rgb`.

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
emilbon99: Yes, correct. You have things you want 'globally configured' settings that live everywhere, you want things that are scoped per cluster, and then you want things that are scoped per project.

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
emilbon99: This RFC got misplaced in the chain, sorry

docs/tech/rfc/0046-260716-client-cache-unified-reads-session-state.md
pjdotson: Do we still have the full client cache for the Core 2 client? couldn't that cause issues if things get deleted on Core 2 while we were gone?

emilbon99: No, we don't still have the full cache. We rebuild the client cache and, _crucially_, reconcile the sessions tate.

pjdotson (approved):

PR #2626, opened 2026-07-24
SY-4509: Driver CLI Login UX Improvements

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4509](https://linear.app/synnax/issue/SY-4509)

## Description

Improves the `synnax-driver login` flow and adds a dedicated role for driver machines.

1. Connection check before credentials: `login` now prompts for connection parameters (host, port, security) and verifies the cluster is reachable before asking for a username and password. Previously a typo in the host or port surfaced as an authentication failure after credentials were entered. To support this, the `/connectivity/check` client no longer runs auth middleware, matching the server, which serves it unauthenticated.

2. System trust store fallback: secure connections no longer require a CA certificate file. When `Secure` is enabled and no CA file is provided, the client falls back to the system trust store, matching how the Console connects to TLS clusters with publicly trusted certificates.

3. Post-login start hint: after credentials are saved, the CLI prints a hint to run `synnax-driver start -s` to start the driver and register it with the cluster.

4. New `Host` role: a built-in role for machines running the driver. It grants full access to only the resource types the driver needs: `channel`, `range`, `framer`, `range_alias`, `rack`, `device`, `task`, `arc`, and `status`. It has no access to user management, access control, or Console resources such as workspaces, schematics, and views. The role is provisioned idempotently on startup, so existing clusters receive it on upgrade with no migration.

Tests: C++ transport middleware tests verifying the connectivity check skips middleware while other clients run it, plus an unauthenticated probe against an unreachable cluster. Go enforcement tests asserting `Host` allows all actions on driver resource types and denies everything else, plus an exact match on the edit policy contents. Integration test role lists updated to include `Host`.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR improves the `synnax-driver login` UX and introduces a new `Host` built-in role for machines running the driver. Connection reachability is now verified before credentials are collected, TLS now falls back to the system trust store when no CA file is provided, and a post-login hint directs users to start the driver.

- **Login flow**: connectivity is probed unauthenticated (auth middleware is excluded from the `connectivity_check` transport client) before the username/password prompts, so network misconfiguration fails fast instead of surfacing as an authentication error.
- **Host role**: a new built-in role is provisioned idempotently on startup, granting `AllActions` on the nine driver resource types (`channel`, `range`, `framer`, `range_alias`, `rack`, `device`, `task`, `arc`, `status`) with no access to user management or console-only resources.
- **TLS fallback**: a new `secure` field on `Config` enables TLS using the system trust store when no explicit CA file is supplied, matching Console behavior; backward compatibility is preserved via `is_secure()` returning `true` when either `secure` or `ca_cert_file` is set.

Safe to merge — all changed paths are well-tested, behaviorally backward-compatible, and the key architectural decisions (unauthenticated connectivity probe, system-trust-store TLS fallback, idempotent Host role provisioning) are each directly exercised by new tests.

The connectivity-check middleware exclusion is the most load-bearing change; it is proven correct by three dedicated transport middleware tests and a full end-to-end probe test. The TLS system-trust-store path is structurally sound (gRPC uses system roots when pem_root_certs is empty). The Host role provisioning reuses the same idempotent provisionRole helper used by all other built-in roles and is verified by exact-match policy content and enforce/deny table tests. No regressions in existing APIs — is_secure() preserves the legacy CA-file-implies-secure contract.

| Filename | Overview |
|----------|----------|
| driver/cmd/login.cpp | Login flow reordered to probe connectivity before prompting credentials; adds TLS system-trust-store path and a post-login start hint. Logic is clean and well-tested. |
| client/cpp/transport.cpp | Removes connectivity_check from the middleware registration loop so the unauthenticated probe skips auth middleware; passes new `secure` flag to the Pool constructor. |
| freighter/cpp/grpc/grpc.h | Pool constructor refactored to accept an explicit `secure` flag; early-returns for insecure case; calls SslCredentials with empty opts when no CA path is supplied, correctly delegating to the system trust store. Previously flagged key-file error-variable bug is fixed. |
| client/cpp/synnax.h | Adds `secure` field to Config; updates `is_secure()` to OR it with the ca_cert_file check; passes `is_secure()` to the Transport constructor. Backward-compatible. |
| client/cpp/synnax_test.cpp | New test file covering Config round-trip, is_secure() combinations, middleware skip for connectivity_check, and unauthenticated probe against an unreachable cluster. Good coverage of the new behavior. |
| core/pkg/service/access/rbac/builtin/builtin.go | Adds Host role and hostPolicies granting AllActions on nine driver resource types; also adds ResourceTypeBuiltin to owner's retrieve policy. |
| core/pkg/service/access/rbac/builtin/provision.go | Adds HostKey to ProvisionResult and provisions the Host role idempotently between Viewer and Operator. |
| core/pkg/service/access/rbac/provision_test.go | Extends provision tests: verifies Host role exists, that all AllActions are granted on the nine driver resource types, and that non-driver resource types are denied. Policy contents are checked exactly. |
| console/src/access/role/ontology.tsx | Introduces per-role icon mapping; Host role displays a TerminalOutline icon; unknown roles fall back to the generic Role icon. |
| x/cpp/cli/cli.h | prompt() now shows [none] for empty string defaults instead of [], so optional cert file fields display their skip-ability clearly. |
| pluto/src/icon/registry.tsx | Exports IoTerminalOutline as TerminalOutline for use by the Host role icon. |
| integration/tests/console/user/register.py | Adds Host to the BUILTIN_ROLES list for integration tests. |
| integration/tests/migration/rbac_verify.py | Adds Host to BUILTIN_ROLES in the migration verification test. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant LoginCLI as login.cpp
    participant Probe as Synnax (probe)
    participant ConnCheck as connectivity_check<br/>(no auth middleware)
    participant Server

    User->>LoginCLI: synnax-driver login
    LoginCLI->>User: prompt Host, Port, Secure?
    User-->>LoginCLI: host / port / TLS options

    LoginCLI->>Probe: new Synnax(config) [no credentials yet]
    LoginCLI->>ConnCheck: "probe.connectivity->check()"
    ConnCheck->>Server: /connectivity/check (unauthenticated)
    Server-->>ConnCheck: OK or UNREACHABLE
    ConnCheck-->>LoginCLI: "ConnectionState {status, message}"

    alt "status != CONNECTED"
        LoginCLI->>User: ERROR: failed to connect (exit 1)
    else CONNECTED
        LoginCLI->>User: connection established
        LoginCLI->>User: prompt Username, Password
        User-->>LoginCLI: credentials
        LoginCLI->>Server: authenticate (auth middleware)
        Server-->>LoginCLI: token or error
        alt auth error
            LoginCLI->>User: ERROR: failed to authenticate (exit 1)
        else success
            LoginCLI->>User: credentials saved — run synnax-driver start -s
        end
    end
```

<sub>Reviews (2): Last reviewed commit: ["SY-4509: tweak error string"](https://github.com/synnaxlabs/synnax/commit/4c51747879d189f78417d10e374512e16ee403bf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47112824)</sub>


docs/site/src/pages/reference/console/users.mdx
```diff
@@ -50,6 +50,7 @@ deployment.
 | **Owner**    | Full control of deployment, including user registration and security      | All actions on all resources; user management; security settings                                                         |
 | **Engineer** | Full access to system configuration except user management                | Create/edit channels, workspaces, schematics, line plots, tasks, devices, ranges, tables, Arc code; view users and roles |
 | **Operator** | View workspaces and visualizations; control hardware and data acquisition | Control hardware tasks; edit ranges and framers; view all resources; cannot modify system configuration                  |
+| **Host**     | For machines running the Synnax driver                                    | Manage devices, tasks, and Arc automations                                                                               |
```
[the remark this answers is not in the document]
sy-nico: Non-issue. Human reviewer will provide feedback for this description. This is user documentation, so your suggestion is bad and over-explained.

freighter/cpp/grpc/grpc.h
```diff
@@ -86,9 +86,8 @@ class Pool {
             if (pem_priv_key_err)
                 LOG(ERROR) << "Failed to read client private key from " << err;
             opts.pem_private_key = pem_private_key;
-            secure = true;
         }
```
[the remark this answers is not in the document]
sy-nico: Fixed: now logs `pem_priv_key_err`. Kept structure, only the variable was wrong.

sy-nico:
@greptile, rerun

x/cpp/cli/cli.h
```diff
@@ -33,7 +33,12 @@ inline std::string prompt(
 ) {
     while (true) {
         std::string prompt_text = message;
-        if (default_value.has_value()) prompt_text += " [" + *default_value + "]";
+        if (default_value.has_value()) {
+            if (default_value->empty())
+                prompt_text += " [none]";
+            else
+                prompt_text += " [" + *default_value + "]";
+        }
```
pjdotson, line 44: add tests for this please

freighter/cpp/grpc/grpc.h
```diff
@@ -62,33 +62,32 @@ class Pool {
...
+    /// with an empty ca_path, the system trust store verifies the server. The client
+    /// certificate and key are loaded from the provided paths when non-empty.
     Pool(
         const std::string &ca_path,
         const std::string &cert_path,
-        const std::string &key_path
+        const std::string &key_path,
+        const bool secure
     ) {
+        if (!secure) return;
```
pjdotson: if secure is false, why should we still need the certificate?

client/cpp/synnax.h
```diff
@@ -110,7 +114,9 @@ struct Config {
 
     /// @brief returns true if the configuration uses TLS encryption to secure
     /// communications with the cluster.
-    [[nodiscard]] bool is_secure() const { return !this->ca_cert_file.empty(); }
+    [[nodiscard]] bool is_secure() const {
+        return this->secure || !this->ca_cert_file.empty();
```
pjdotson: Shouldn't this just return `this->secure`?

core/pkg/service/access/rbac/builtin/builtin.go
```diff
@@ -97,13 +97,41 @@ var (
...
+				{Type: ontology.ResourceTypeChannel},
+				{Type: ontology.ResourceTypeRange},
+				{Type: ontology.ResourceTypeFramer},
+				{Type: ontology.ResourceTypeRangeAlias},
+				{Type: ontology.ResourceTypeRack},
+				{Type: ontology.ResourceTypeDevice},
+				{Type: ontology.ResourceTypeTask},
+				{Type: ontology.ResourceTypeArc},
+				{Type: ontology.ResourceTypeStatus},
+			},
```
pjdotson, line 125: Why do we need these:

1) Range
2) RangeAlias

Also, which one of these actually gets edited by the Driver? I don't think the Driver ever creates channels, right? And the Driver shouldn't be able to delete status or framer as well.

pjdotson (changes requested):
see comments, huddle me if you have questions.

freighter/cpp/grpc/grpc.h
sy-nico: Reworked it so the caller picks:  `auto pool = secure ? make_shared<Pool>(ca, cert, key) : make_shared<Pool>();`

Pool drops the secure flag and the early return, so the insecure path never receives cert paths at all. No behavior change.

client/cpp/synnax.h
sy-nico: Returns `secure` now. The CA inference moved into override so pre-flag config still loads secure instead of silently dropping to plaintext.

core/pkg/service/access/rbac/builtin/builtin.go
sy-nico: I removed `RangeAlias`, but`Range` is needed for the ranges stl being release for v0.57.

`Channel` is now retrieve only. The driver never creates channels, the Console does that during task configuration, but every task resolves channel keys before opening a writer.

`Framer` is now create and retrieve only. Create opens the acquisition writer, retrieve opens the control streamer. Delete is gone.

client/cpp/synnax.h
```diff
@@ -110,7 +119,7 @@ struct Config {
 
     /// @brief returns true if the configuration uses TLS encryption to secure
     /// communications with the cluster.
-    [[nodiscard]] bool is_secure() const { return !this->ca_cert_file.empty(); }
+    [[nodiscard]] bool is_secure() const { return this->secure; }
```
pjdotson: useless now

pjdotson (changes requested):
Approved after comment addressed and you bump Console version

pjdotson (approved):

PR #2628, opened 2026-07-24
SY-4126: Fix flakey Python control test and harden integration teardown

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4126](https://linear.app/synnax/issue/SY-4126)

## Description

`tests/test_control.py::TestController::test_remains_true_for_target_percent` raced against its simulated DAQ: the DAQ fired all of its post-acknowledgement samples back-to-back, while the sequence thread only registered the `RemainsTrueFor` processor after its second `wait_until` returned. A scheduling delay of as little as 10ms (routine on CI runners, reproduced locally by injecting a sleep) meant every sample arrived before the processor was registered, `remains_true_for` observed zero samples, and the test failed. The related `test_remains_true_for_true`/`_false` tests had the same structural hazard.

This PR fixes the flake at three levels:

- **Test rewrite**: the simulated DAQ now paces its writes every 5ms and keeps publishing until the sequence signals completion, so the `remains_true_for` observation window always contains a representative sample mix no matter when it starts. The rewritten target-percent design was verified against injected scheduling delays of 10–100ms (which reliably failed the old test) and 30 consecutive full runs of the control suite.
- **Unit coverage**: the `RemainsTrueFor` percentage math is now covered by deterministic unit tests that drive the processor directly, with no streaming timing involved.
- **Client bug**: `Controller.remains_true_for` mutated the receiver's processor set directly instead of going through `add_processor`/`remove_processor`, racing the receiver thread's locked iteration of that set — a possible `RuntimeError: Set changed size during iteration` that kills the receiver thread.

It also fixes the *other* recurring flake in this suite (the failure in the Migration (Clients) run that prompted this work): channel names generated as `test_{randint(...)}` across seven test files drew from overlapping spaces as small as 10,000, producing birthday-problem collisions (`channel with name 'test_322629' already exists`). All such sites now use `random_name()` from `x.strings`, making collisions practically impossible.

### Follow-up: hardening the integration harness against the *same* cascade

While investigating a failure on this PR's arc integration run, I found a second, unrelated way this control-authority flake surfaces — this time in the integration harness itself. A flaky Playwright dispatch on the "Connected to" notification during `ConsoleCase.setup` was amplified into a run-wide failure:

- `setup` raised on the notification dispatch (15s timeout) *before* `_cleanup_pages` was initialized;
- `ConsoleCase.teardown` then hit `AttributeError: no attribute '_cleanup_pages'`;
- because `SimDaqCase.teardown` calls `super().teardown()` **before** `sim_daq.stop()`, the simulator was never stopped — leaving it holding ABSOLUTE control authority on `daq_time`;
- every subsequent SimDAQ test (and its retries) then failed with `UnauthorizedError: ... held by [Simulated DAQ]`, turning one UI flake into 3 cascading failures.

Three defensive fixes, so a single setup flake can no longer poison the rest of the run:

- **`SimDaqCase.teardown`**: run `super().teardown()` in `try`/`finally` so `sim_daq.stop()` always releases control authority.
- **`ConsoleCase`**: initialize `_cleanup_pages` at the top of `setup` (before any Playwright call that can throw) and guard it in `teardown`.
- **`NotificationsClient.close_connection`**: make the silence-button dispatch best-effort (2s timeout, no-op on timeout) instead of a 15s hard failure.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR fixes two recurring flaky test failures in the Python client suite: a data-race in `Controller.remains_true_for` that directly mutated the receiver's processor set without the lock (risking `RuntimeError: Set changed size during iteration`), and a birthday-problem collision in channel names generated from small integer spaces across seven test files.

- **Controller bug fix**: `remains_true_for` now calls `add_processor`/`remove_processor` (both lock-protected) instead of directly accessing `self._receiver.processors`; `add_processor` is moved before the `try` block, matching the existing `wait_until` pattern.
- **Test rewrites**: The three `remains_true_for` integration tests replace burst DAQ writes with continuous 5ms-paced publishing guarded by `acked`/`stop` events, ensuring the observation window always contains representative samples regardless of scheduling jitter; new `TestRemainsTrueForProcessor` unit tests cover the percentage math deterministically without streaming.
- **Collision fix**: All seven test files replace `random.randint(0, 1_000_000)` with `random_name()` (15-char alphabetic, ~52^15 space), making collisions practically impossible.

Safe to merge — the controller bug fix is minimal and correct, and the test rewrites eliminate the timing hazards through well-structured event coordination.

The lock fix in remains_true_for is straightforward: two direct set mutations are replaced with the existing lock-protected helper methods. The test rewrites are carefully designed — continuous paced publishing means the observation window is always populated regardless of scheduling delays, and the acked/stop event pattern correctly coordinates teardown. The new unit tests cover the percentage logic deterministically. The random_name() migration is a pure improvement with no behavioral side effects.

**Files Needing Attention:** No files require special attention. The core logic change is confined to the six lines in controller.py.

| Filename | Overview |
|----------|----------|
| client/py/synnax/control/controller.py | Bug fix: remains_true_for now uses add_processor/remove_processor (lock-protected) instead of mutating the processor set directly; add_processor also moved before the try block to match wait_until's pattern. |
| client/py/tests/test_control.py | Rewrites three flakey integration tests to use continuous paced DAQ publishing + acked/stop events for race-free coordination; adds new TestRemainsTrueForProcessor unit-test class with 5 deterministic tests for RemainsTrueFor percentage logic. |
| client/py/tests/conftest.py | Replaces random.randint(0, 1000000) with random_name() (15-char alphabetic string, 52^15 space) in shared fixtures to eliminate birthday-problem channel-name collisions. |
| client/py/tests/test_frame.py | Replaces 6 occurrences of randint(0, 100000) with random_name() for channel name generation; removes unused numpy import. |
| client/py/pyproject.toml | Patch version bump 0.56.0 → 0.56.1 to reflect the client bug fix. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant S as Sequence Thread
    participant R as _Receiver Thread
    participant D as DAQ Thread

    Note over S,D: remains_true_for (fixed)
    S->>R: add_processor(RemainsTrueFor) [lock protected]
    S->>S: "event.wait(timeout=duration)"
    D-->>R: frame arrives
    R->>R: _process() [acquires processor_lock]
    R->>R: processor.process(controller)
    alt "target >= 1 and cond() == False"
        R-->>S: event.set() early exit
    else normal sample
        R->>R: update actual/count
    end
    S->>R: remove_processor(RemainsTrueFor) [lock protected]
    S->>S: "return actual >= target"

    Note over S,D: Old bug (removed)
    Note over S: processors.add() without lock
    Note over R: _process() iterating processors
    Note over S: processors.remove() RuntimeError!
```

<sub>Reviews (2): Last reviewed commit: ["Bump version"](https://github.com/synnaxlabs/synnax/commit/0dfcbc7b6f202d70930745b1dd97de9c59ef9c4f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47129534)</sub>


pjdotson:
@greptile /rerun

emilbon99 (approved):

PR #2640, opened 2026-07-28
SY-4520: Fix flaky TypeScript framer test

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4520](https://linear.app/synnax/issue/SY-4520)

## Description

Fixes the flaky `client.spec.ts > should correctly write a single series of data` test and the underlying Cesium bug it exposed.

The test wrote a second `TimeStamp.now()` as its index sample instead of the writer's start timestamp. `TimeStamp.now()` is millisecond-resolution, so the two calls usually return the same value — but when the clock ticked between them, the follow-up data-only write was anchored at a timestamp that is not an exact sample in the index.

That case hit a real bug in Cesium's commit path: `idxWriter.resolveCommitEnd` returned an inexact stamp approximation whose zero-valued lower bound was then used as the commit end, producing the confusing `commit timestamp 1970-01-01T00:00:00Z must be strictly greater than the starting timestamp` error seen in CI. The commit path now rejects an inexact resolution with a clear validation error ("writer start ... cannot be resolved to an exact sample in index channel ..."), mirroring the check that already exists in the unary writer, so users hit an understandable error instead of a garbage commit.

The TypeScript test now writes its start timestamp as the index sample, making the usage correct and deterministic (verified with 10 consecutive runs against a live server; full cesium suite passes 403/403).

Also bumps Core to 0.56.13 for the cesium fix. The TS client change is test-only, so the TS Client and Console versions are unchanged.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR makes TypeScript framer test timestamps deterministic and improves Cesium’s handling of inexact index resolution.
- Rejects commits whose writer start cannot be resolved to an exact index sample.
- Adds single-sample and multi-sample regression coverage for the validation path.
- Reuses the writer start timestamp in the TypeScript framer test.
- Bumps Core from 0.56.12 to 0.56.13.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| cesium/writer_stream.go | Adds exact-index validation before committing data-only writes, preventing an invalid approximation from becoming the commit boundary. |
| cesium/writer_test.go | Adds focused regression coverage for inexact single-sample and multi-sample commit starts. |
| client/ts/src/framer/client.spec.ts | Makes the framer write test deterministic by using the writer start as its index sample. |
| core/pkg/version/VERSION | Bumps the Core patch version to publish the Cesium fix. |

<sub>Reviews (2): Last reviewed commit: ["SY-4520: Use MustBeContinuous Constant a..."](https://github.com/synnaxlabs/synnax/commit/314d29aa3c20a5a56175dc0b1bb05ebe1923312a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48003746)</sub>

**Context used:**

- Knowledge Base — [Cesium Storage Engine](https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/cesium-storage.md)


pjdotson:
@greptile /rerun

cesium/writer_stream.go
```diff
@@ -872,7 +872,22 @@ func (w *idxWriter) resolveCommitEnd(ctx context.Context) (index.TimeStampApprox
...
 	}
-	return w.idx.Stamp(ctx, w.start, w.sampleCount-1, true)
+	approx, err := w.idx.Stamp(ctx, w.start, w.sampleCount-1, index.MustBeContinuous)
+	if err != nil {
+		return approx, err
+	}
+	// An inexact approximation means w.start is not an exact sample in the index, so
+	// there is no defined timestamp for the samples written by this writer. Committing
+	// anyway would use the approximation's zero-valued lower bound as the commit end.
+	if !approx.Exact() {
```
emilbon99, line 882: There are a number of errors constructors in the `index` package that might overlap with this one. Could be worth using and/or modifying one of those so we stay consistent.

There is `index.NewDiscontinuousError`, `index.NewDiscontinuousOffsetError`, and `index.NewDiscontinuousStampError`

emilbon99 (changes requested):
LGTM after comment addressed. Please coordinate with @sy-nico for a consistent version bump.

PR #2636, opened 2026-07-27
SY-4471: Arc flow variables in the C++ runtime

sy-nico:
## Linear Issue

[SY-4471](https://linear.app/synnax/issue/SY-4471)

## Description

Brings the C++ embedded runtime to parity with the Go runtime for Arc flow variables (SY-4293). Programs using variables now behave the same whether they run on Core or on the driver.

### What changed

**1. Node state foundation** (`arc/cpp/runtime/state`)

- New accessors: `ref_input`, `ref_sourced`, `string_input`, `numeric_input`, `consume_input`, `input_fresh`, `last_changed`, `absorb_inputs`.
- New per-input `Rearm` policy replaces the single blanket re-arm on scope entry. Register reads re-arm on fresh values, deref reads on post-entry values, self-write feeders only on `Reset`.
- Inputs with no source slot are marked `NO_SOURCE` instead of silently aliasing slot 0.

**2. Variable nodes** (`arc/cpp/stl/variable`)

- `Register` holds a variable's current binding. Last write wins. Restores its initial on scope entry unless declared `$=`.
- `ExprRead` derefs a reassigned variable's dispatcher, suppressing the value that accompanies a re-point.

**3. Scheduler settle passes** (`arc/cpp/runtime/scheduler`)

- `next()` re-walks until no change lands on an already-run node, bounded by node count.
- Sequential scopes run trailing strata variable nodes every pass and defer the next step to the settle pass so it observes prior writes.

**4. Live var-bound params**

Params are read at fire time instead of baked at construction, in `channels.on`, `channels.write`, `constant`, `set_authority`, `stable.for`, `time.interval`, `time.wait`, and the driver's `ranges.create`, `ranges.end`, `status.set`.

### Why `read_write` is in this PR

`ChanDirection` gains `read_write = 3`.

The analyzer has emitted `ChanDirectionRead | ChanDirectionWrite` for channel aliases that are both read and written since SY-4293 (`arc/go/analyzer/statement/statement.go`). `ChanDirectionToPB` rejected that value with `unrecognized types.ChanDirection value: 3`, so any program with such an alias failed to serialize. `ChanDirection` is already a bitmask (`IsRead` and `IsWrite` mask rather than compare), so this only names a value the type could always hold.

Strictly a SY-4293 follow-up, kept here because it blocks this work and is 7 lines plus codegen. Schema in `schemas/arc/types.oracle`, with regenerated Go, TS, Python, C++, and proto bindings.

### Three bugs found while building this

[...]

PR #2646, opened 2026-07-28
SY-4530: Reset Function-Scoped Stateful Variables on Stage Re-entry

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4530](https://linear.app/synnax/issue/SY-4530/wasm-stateful-variables-not-resetting-on-stage-re-entry)

## Description

Stateful variables declared inside an Arc function with `$=` were never discarded when their enclosing stage was re-entered, contradicting the stage entry semantics in `arc/docs/spec.md`. A counter in a looping stage kept climbing instead of restarting, so sequences relying on a per-entry counter or elapsed-time guard ran against stale state and could trip a condition they should never have reached. Found via a customer control sequence that exited its loop early.

State was keyed by node key and only ever written, never deleted. `Reset()` cleared the node's execution flags but left the stateful maps intact. Both runtimes were affected.

This adds a per-node clear called from the WASM node's reset, which the scheduler already invokes on stage activation:

- Go: `stateful.Host.ClearNode`, called from `nodeImpl.Reset`
- C++: `stateful::Variables::clear_node`, plumbed through `state::State` and `state::Node`, called from `wasm::Node::reset`

Scope is WASM function-scoped stateful variables only. Flow and sequence-scoped variables keep their existing lifetime. There is no semantic overlap with the C++ flow variables work in #2636: that PR does not touch `arc/cpp/stl/stateful/*` or any of the Go files here, and its own stateful test covers sequence-scoped variables, which this change deliberately leaves alone. The two touch a few of the same C++ files (`runtime/state/state.{h,cpp}`, `runtime/wasm/node.h`, and two test files), but not the same lines, so any back-merge conflicts should be mechanical rather than behavioral.

Two follow-on changes:

- Six tests in `arc/go/stl/wasm/wasm_test.go` asserted the old behavior, using `Reset()` as a re-arm and expecting counters to keep accumulating across it. They now expect re-initialization.
- `arc/cpp/stl/stateful/state_test.cpp` had no `cc_test` target and had never been compiled by any workflow. Registering it surfaced `StoreStrInvalidHandleIsNoOp`, which used handle 0 as a stand-in for an invalid handle. Handle 0 is the canonical empty-string handle in both runtimes, so the test now uses an unknown handle, matching the Go equivalent.

Verified end to end against the embedded and standalone C++ drivers, and against the Go runtime.

## Basic Readiness

<h3>Greptile Summary</h3>

Adds node-scoped state cleanup on WASM node reset in both runtimes.
- Clears all function-scoped stateful variable types using the resetting node’s IR key.
- Preserves state belonging to sibling call sites and flow- or sequence-scoped variables.
- Adds Go and C++ regression coverage for stage re-entry, nested stages, repeated cycles, type coverage, and node isolation.
- Registers the previously unbuilt C++ stateful-variable test target and corrects its invalid-string-handle fixture.

The PR appears safe to merge, with consistent node-keyed reset behavior and comprehensive cross-runtime regression coverage.

Both runtimes clear the exact node key installed before WASM execution, scheduler resets align with scope activation, and the shared state stores remain isolated per runtime and synchronously accessed.

| Filename | Overview |
|----------|----------|
| arc/go/stl/wasm/node.go | Extends node reset to clear stateful slots under the same stable IR key used before execution. |
| arc/go/stl/stateful/stateful.go | Adds node-scoped deletion across every supported scalar, string, and series state map. |
| arc/cpp/runtime/wasm/node.h | Mirrors the Go reset behavior by clearing the resetting C++ WASM node’s state. |
| arc/cpp/stl/stateful/state.h | Adds per-node erasure for every C++ stateful-variable storage map. |
| arc/go/sequence_test.go | Adds end-to-end coverage for stage re-entry, repeated and nested transitions, type reset, and call-site isolation. |
| arc/cpp/runtime/wasm/node_test.cpp | Adds C++ node-level regression tests for reset reinitialization and sibling-node isolation. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Scheduler
  participant WasmNode as WASM Node
  participant Store as Stateful Store
  participant Wasm as WASM Function
  Scheduler->>WasmNode: reset() on stage activation
  WasmNode->>Store: clearNode(nodeKey)
  Scheduler->>WasmNode: next()
  WasmNode->>Store: setNodeKey(nodeKey)
  WasmNode->>Wasm: execute
  Wasm->>Store: load variable with initializer
  Store-->>Wasm: initialized value
```

<sub>Reviews (1): Last reviewed commit: ["SY-4530: Reset Function-Scoped Stateful ..."](https://github.com/synnaxlabs/synnax/commit/7c71f78e9a22f611ffbb6c188cf1e689bbf9bbab) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48136303)</sub>

**Context used:**


emilbon99 (approved):

emilbon99 (changes requested):
@sy-nico If this is going into main it needs a version bump. Otherwise LGTM

PR #2647, opened 2026-07-29
SY-4399: Upgrade Bazel dependencies

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4399](https://linear.app/synnax/issue/SY-4399/upgrade-bazel-dependencies)

## Description

Upgrades the Bazel C++ dependency stack, migrates logging off the deprecated
glog, and fixes the Linux systemd service path that the migration surfaced:

- **grpc 1.81.1 → 1.83.0, rules_cc 0.2.20 → 0.2.22.** gRPC upstreamed the
  `rules_cc` load fixes we were patching locally, so the two `load()` patches
  are deleted. Three patches remain: `grpc-disable-layering-check.patch`
  (openssl headers still trip `layering_check`) and the two upb-gen header
  patches (the strict-inclusion check they fix only runs on Linux/Windows
  toolchains).
- **glog/gflags → Abseil Logging.** glog is deprecated, so all
  `LOG`/`VLOG`/`DCHECK` sites now use `absl/log`. abseil-cpp becomes a direct
  dependency at **20260526.0**, up from the 20250814.1 that grpc/protobuf were
  already pulling in; glog and gflags are removed.
- **Custom color logging.** glog's `colorlogtostderr` has no absl equivalent, so
  `x::log::init` installs `x::log::StderrSink`, an `absl::LogSink` that colors
  WARNING lines yellow and ERROR/FATAL lines red. Color requires the existing
  `--no-color` flag to be absent *and* stderr to be a terminal (glog also
  skipped color on non-TTYs), so ANSI codes never land in journald or
  redirected log files. `init` is `std::call_once`-guarded — calls after the
  first are no-ops instead of aborting in `absl::InitializeLog`. The default
  absl stderr sink is silenced so lines aren't duplicated. `LOG(ERROR)` call
  sites that wrapped their messages in the inline `RED()`/`RESET()` helpers
  drop the wrapping — the sink already colors those lines whole. Tests in
  `x/cpp/log/log_test.cpp` register their own sinks to cover both color paths
  (line coloring, no-color output, no duplicate lines, FATAL abort, init
  behavior via death tests, string helpers).
- **Linux systemd service actually starts.** The unit's `ExecStart` has invoked
  `synnax-driver internal-start` since the template was introduced, but the
  command was never dispatched — the service printed usage and exited 1.
  `internal-start` now wraps `sub::start` in `daemon::run`, which reports
  `READY`/`WATCHDOG`/`STOPPING` over `sd_notify`, and exits with `sub::start`'s
  return code so the unit's `Restart=on-failure` policy engages when startup
  fails; the exit code also stays nonzero when the callback throws, since
  `daemon::run` swallows the exception after reporting it. On that failure
  path, shutdown sends `STOPPING=1` without a `STATUS=` line so the error
  message stays visible in `systemctl status` instead of being overwritten
  with "Stopping". `daemon::run`'s watchdog thread now blocks on a condition
  variable instead of polling an unguarded flag, fixing a data race and making
  shutdown immediate rather than lagging up to a watchdog interval. The unit
  template drops the now-dead `GLOG_*` environment lines and runs at default
  verbosity — absl reads no `GLOG_v`-style environment variables, and rather
  than hardwiring `--debug` into `ExecStart`, verbose service logs are now
  opt-in via `systemctl edit`. The unit passes `--disable-stdin-stop`: stdin
  under systemd is `/dev/null`, whose immediate EOF the stdin listener treats
  as a shutdown signal.
  `driver/daemon/daemon_test.cpp` covers the sd_notify handshake in CI by
  pointing `NOTIFY_SOCKET` at an abstract unix datagram socket and asserting
  the protocol messages, including the callback-error path.
- **`driver/cmd` and daemon API cleanup.** `cmd.h` now declares only `exec`;
  subcommands stay one per file in `driver::cmd::sub`, with their prototypes
  declared at the dispatch site in `exec.cpp` instead of a shared header.
  `daemon::run` no longer threads argc/argv through to the callback — it was
  never inspected — so `ApplicationCallback` is `std::function<void()>` and
  `internal-start` takes the arg parser like every other subcommand (all three
  daemon backends updated). The noop backend (macOS/Windows) now catches
  callback exceptions like the other two, so a throwing `start` makes
  `internal-start` exit 1 on every platform instead of hitting the crash
  handler.
- **`.clang-format` third-party regex refreshed.** The quoted-include vendor
  group still listed glog, gflags, and lua (none used anywhere) and was missing
  absl entirely, so every new `absl/` include sorted into the internal group.
  The group now covers the third-party prefixes actually included from
  first-party code (absl, benchmark, boost, gmock, google, gperftools,
  grpc/grpcpp, gtest, mbedtls, modbus, nlohmann, open62541, soem, vendor) plus
  the bare SDK headers a directory regex can't catch (`httplib.h`,
  `wasmtime.hh`, `nisyscfg*`, `LJM_*`, `LabJackM.h`). The old `/// std` /
  `/// external` / `/// module` / `/// internal` section labels are deleted
  everywhere: with `IncludeBlocks: Regroup`, comments are barriers clang-format
  won't sort across, so the labels were actively pinning includes into the
  wrong groups (they had trapped `absl/log/log.h` in the internal group in
  several OPC headers). All C++ trees are reformatted under the new rules —
  include reordering only.
- **Include hygiene.** Many files used `LOG`-family macros through transitive
  includes left over from glog; they now include `absl/log/log.h` directly, and
  `x/cpp/log/log.h` no longer re-exports it. On top of that, a
  compile-commands-driven sweep dropped every unused include clangd could find
  across the C++ trees, then load-bearing transitive includes were restored at
  their most specific location: the defining header instead of an umbrella
  (`client/cpp/channel/channel.h` rather than `client/cpp/synnax.h`, individual
  `boost/uuid/*.hpp` headers), in the `.cpp` rather than the `.h` when only the
  implementation needs it (`driver/ethercat/soem/master.cpp`), and `wasmtime.hh`
  in each file that names `wasmtime::` symbols. Platform-gated sources the local
  build can't compile (`daemon_test.cpp`, `daemon_nilinuxrt.cpp`,
  `loop_windows.cpp`) were audited by hand. A follow-up pass over the OPC
  driver dropped more unused includes, narrowed `connection.h` to
  `open62541/common.h` (the header that actually declares `UA_Client`) instead
  of `client.h`, and gave `driver/opc/read_task.h` direct includes for
  everything it names instead of leaning on transitive ones.
- **NI Linux RT daemon** drops `SetLogDestination`: the init script already
  redirects stderr to `/var/log/synnax-driver.log`, which is what
  `synnax-driver logs` reads. The glog-suffixed side files under `/var/log` are
  gone.

Verified with `bazel build //...`, `bazel test //x/cpp/...`, running
`synnax-driver version` with and without `--no-color` (no ANSI codes when
stderr is piped), and `echo STOP | synnax-driver internal-start` (runs the
driver and stops gracefully with exit 0). The systemd handshake test runs in
the Linux driver CI job; end-to-end `systemctl` behavior still needs a pass on
a systemd host.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR upgrades the Bazel C++ dependency stack and migrates driver logging and service startup behavior.
- Upgrades Abseil, gRPC, and rules_cc while removing obsolete gRPC patches and glog/gflags dependencies.
- Replaces glog with Abseil logging and adds severity-colored terminal output.
- Adds systemd daemon lifecycle notifications, exception handling, and responsive watchdog shutdown.
- Cleans up C++ dependencies, includes, command dispatch declarations, and formatting rules.

The PR appears safe to merge.

The previously reported process-wide logging configuration remains, but current production code initializes it exactly once at process startup, so no blocking runtime failure remains.

| Filename | Overview |
|----------|----------|
| MODULE.bazel | Upgrades the C++ dependency stack, removes glog/gflags, and retains the required gRPC compatibility patches. |
| x/cpp/log/log.cpp | Implements Abseil initialization and a custom stderr sink for terminal-aware severity coloring. |
| x/cpp/log/log_test.cpp | Adds coverage for colored output, duplicate suppression, fatal logging, initialization, and helper functions. |
| driver/daemon/daemon_linux.cpp | Adds systemd lifecycle notifications and condition-variable-based watchdog shutdown. |
| driver/cmd/exec.cpp | Initializes logging from command-line configuration and dispatches the internal service-start command. |

<sub>Reviews (3): Last reviewed commit: ["Drop redundant ERROR-line coloring, catc..."](https://github.com/synnaxlabs/synnax/commit/945860c2dc903a991d23233b57c5fa3d49ec3e93) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48161655)</sub>


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

emilbon99 (approved):

PR #2655, opened 2026-07-31
SY-4534: Add String Display Symbol

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4534](https://linear.app/synnax/issue/SY-4534)

## Description

Adds a `String Display` schematic symbol showing the live value of a string channel.

Neither existing symbol can do this. `value` hardcodes a numeric pipeline, so a string channel becomes `NaN` and renders blank. `stateIndicator` compares option values against numbers.

The `Value` renderer was already string native, so the symbol reuses it. Supporting changes:

- `telem/aether/remote.ts`: adds `StreamChannelStringValue`, a `StringSource` sibling of `StreamChannelValue` reading the leading sample via `Series.asString`. Variable density decodes UTF-8, `UUID` renders as a uuid string, numerics stringify. Decoding is lazy and gated on buffer length, since the client appends into the leading buffer in place.
- `vis/draw2d/canvas.ts`, `vis/value`: threads `characters` through to the atlas registry and adds a `numeric` flag gating the leading minus treatment, which mangles strings like `-IDLE-`. Both default to current behavior.
- `text/aether/atlas.ts`: sizes atlas cells from the whole charset rather than `"0"`. Digits have no descender, so tails were clipped and bled into the cell below. Baseline placement is unchanged.

Console needs no changes; its palette and properties panel are data driven off `REGISTRY`, `GROUPS` and `ELEMENT_REGISTRY`.

Two known limitations, both matching `value`: the display is blank until the channel is next written, and newlines do not render.

## Basic Readiness

<h3>Greptile Summary</h3>

Adds a DOM-rendered String Display schematic symbol for live string-channel telemetry.
- Introduces a string telemetry source that preserves variable-density values and safely invalidates in-flight subscriptions during cleanup.
- Adds Aether state synchronization, staleness handling, symbol configuration, form controls, registry wiring, and automated tests.
- Updates Console and core patch versions.

The PR appears safe to merge.

No blocking failure remains; the DOM text path resolves the Unicode-rendering concern, and generation checks plus stale-handler guards resolve the previously reported asynchronous subscription cleanup issue.

| Filename | Overview |
|----------|----------|
| pluto/src/telem/aether/remote.ts | Adds the string-channel streaming source with lazy decoding, buffer ownership, and generation-based cleanup guards; the previously reported subscription lifecycle issue is addressed. |
| pluto/src/vis/stringValue/aether/stringValue.ts | Adds worker-side string value publication and staleness tracking with source and timer cleanup. |
| pluto/src/schematic/node/general/stringDisplay/Symbol.tsx | Connects synchronized string telemetry to the new DOM-rendered schematic symbol, fully removing the previously reported fixed-atlas Unicode limitation. |
| pluto/src/schematic/node/general/stringDisplay/Primitive.tsx | Renders channel values as native React text and applies fresh or stale text coloring. |
| pluto/src/schematic/node/general/stringDisplay/Form.tsx | Adds style, telemetry, channel-selection, and staleness configuration controls. |
| pluto/src/telem/aether/remote.spec.ts | Covers string decoding, mutable leading buffers, ownership cleanup, stale handlers, and cleanup during asynchronous subscription setup. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Channel as String Channel
  participant Source as StreamChannelStringValue
  participant Worker as StringValue Aether Leaf
  participant DOM as String Display DOM
  Channel->>Source: Stream latest series
  Source->>Source: Decode leading sample as string
  Source->>Worker: Notify value change
  Worker->>Worker: Update value and staleness
  Worker->>DOM: Synchronize Aether state
  DOM->>DOM: Render native text
```

<sub>Reviews (2): Last reviewed commit: ["more tests"](https://github.com/synnaxlabs/synnax/commit/326ece517ff4e7051591ba9d3ccc5e1448ff9f1f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49162784)</sub>


pluto/src/telem/aether/remote.spec.ts
```diff
@@ -197,6 +198,156 @@ describe("remote", () => {
...
+        new Map([
+          [
+            c.channel.key,
+            new MultiSeries([new Series({ data: new Float32Array([1, 2, 3]) })]),
+          ],
+        ]),
+      );
+      await expect.poll(() => scsv.value()).toBe("3");
+    });
+
```
pjdotson, line 337: should probably have tests for all data types: int test, uint test, also test rendering of float32 that are +-0, NaN, or +-Inf, and UUID, bytes, and JSON

pluto/src/schematic/node/general/stringDisplay/external.spec.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright 2026 Synnax Labs, Inc.
```
pjdotson: a) we don't have `external.spec.ts` files anywhere, you should have one test file per prod file, e.g. `Primitive.spec.tsx`, `Symbol.spec.tsx`, or just one `stringDisplay.spec.tsx`

pluto/src/schematic/node/general/stringDisplay/external.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+  });
+
+  describe("configZ", () => {
+    it("should reject a telem spec that emits numbers", () => {
+      const config = {
+        ...StringDisplay.defaultConfig(theme),
+        telem: telem.streamChannelValue({ channel: 1 }),
+      };
+      expect(() => StringDisplay.configZ.parse(config)).toThrow();
```
pjdotson: when throw you should test what error it throws with

pluto/src/schematic/node/general/stringDisplay/Symbol.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { Theming } from "@/theming";
+import { Value as BaseValue } from "@/vis/value";
+
+const PRINTABLE_ASCII = Array.from({ length: 95 }, (_, i) =>
+  String.fromCharCode(32 + i),
+).join("");
+
+// Glyph set for this symbol's text atlas. A string channel needs full printable
+// ASCII; the default set is numeric and drops everything else.
+const CHARACTERS = `${PRINTABLE_ASCII}°µ∞ᴇ`;
```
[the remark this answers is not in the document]
pjdotson: ^this is very important issue here.

pluto/src/vis/value/aether/value.ts
```diff
@@ -44,6 +42,11 @@ const valueState = z.object({
   // host can't grow to fit the natural text width (e.g. a table cell);
   // overflow gets truncated at the cell edge instead of bleeding past.
   clip: z.boolean().default(false),
+  // Glyph set of the backing text atlas. Set when the source emits arbitrary text;
+  // the default set is numeric and drops everything else, space included.
+  characters: z.string().optional(),
```
pjdotson: why is this all flowing through the `value.ts`? Feels like `value.ts` is what we are using for numbers, and some things like `notation` or your negative number handling don't make sense for the general string case. It is also odd because the numbers use the defined set of characters to print (0-9, negative sign, E, etc.) but a string could be any valid Unicode string, so the logic for which character set to use for printing is different.

pjdotson (changes requested):
Looks good but I am concerned about mangling the code for displaying numeric values with this stuff. Also, there is no way to change stale timeout or stale color from the form.

pluto/src/vis/value/aether/value.ts
sy-nico: `value.ts`, `canvas.ts` and the atlas are back to identical with `main`. String Display now renders DOM text through its own `stringValue` aether leaf, which also drops the fixed-charset Unicode problem.

pluto/src/schematic/node/general/stringDisplay/Symbol.tsx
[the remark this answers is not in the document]
sy-nico: `CHARACTERS` and the atlas are gone; the value renders as DOM text, so any valid UTF-8 works.

pluto/src/telem/aether/remote.ts
```diff
@@ -326,6 +328,82 @@ export class StreamChannelData
...
+        const leading = data.series.at(-1);
+        if (leading != null) {
+          leading.acquire();
+          this.leadingBuffer?.release();
+          this.leadingBuffer = leading;
+          this.decodedAt = -1;
+        }
+        this.notify();
+      };
+      this.removeStreamHandler = await this.client.stream(handler, [ch.key]);
```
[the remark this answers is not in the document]
sy-nico: Addressed and filed a follow up for the pre-existing offender [SY-4544](https://linear.app/synnax/issue/SY-4544/subscription-leak-in-streamchannelvalue-and-streamchanneldata)

sy-nico:
Follow up ticket to add stale color to gauge and state indicator: [SY-4545](https://linear.app/synnax/issue/SY-4545/add-staleness-color-and-timeout-to-state-indictor-and-gauge)

pluto/src/telem/aether/remote.spec.ts
sy-nico: The selector filters on dataType.isVariable, so only string, JSON, and bytes channels can be assigned to a String Display (Numeric and UUID rendering isn't reachable here)

sy-nico:
@greptile, rerun

pjdotson (approved):
Approved and looks good. If you have time later you might want to think about how we could move the rendering for this off the main thread otherwise there could be performance issues with high-rate string data.

PR #2658, opened 2026-08-03
SY-4542: Fix flakey Cesium specs and restore relay connection buffering

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4542](https://linear.app/synnax/issue/SY-4542)

## Description

Fixes the flakey `Index-less Data Write Alignment` specs in the Cesium writer suite that panicked in CI with `index out of range [0] with length 0`, and fixes a real relay regression the investigation surfaced.

The Cesium relay delivers frames asynchronously: writers push into a shared buffered pipe that a delta-multiplier goroutine drains into per-streamer connections, and that goroutine handles frame delivery and new connections in the same `select`. Two specs seeded data before opening their streamers, so under CI load the delta goroutine could register the streamer first and then deliver the stale seed frame to it. The spec then read the seed frame in place of the write under observation and panicked indexing a series that wasn't in the frame.

Since the relay's connection channel is unbuffered, a streamer whose `Flow` has returned is guaranteed to receive every subsequent write in order. Both specs now open their streamers before any writes and drain the seed frame — asserting on the drained frame's contents so a future ordering change fails loudly at the drain rather than as an off-by-one two assertions later. The `Disjoint Domain Alignment` spec had the same race, papered over with a `runtime.Gosched()` + 20ms sleep and a raw (hang-prone) `<-outlet` read, and got the same treatment.

**Relay fix**: `openRelay` never initialized `relay.bufferSize`, so every streamer connection stream has been unbuffered since the `a9176e6db1` refactor dropped the old `relayBufferSize = 100` constant. An unbuffered connection made frames eligible for drops after the 20ms slow-consumer timeout whenever a consumer briefly stalled. Connection streams are now buffered again, with two options restoring the pre-refactor asymmetry: `WithRelayBufferSize` sizes the relay's main pipe (default 1000 frames — every write funnels through it, and writers block when it fills), and `WithStreamBufferSize` sizes each streamer's connection to that pipe (default 100 frames, the pre-refactor constant). The per-streamer buffer stays smaller because its cost is paid per streamer and a smaller buffer bounds how stale buffered frames can get before the relay drops them. `newOptions` seeds both defaults directly, so a non-positive size — including an explicit zero — fails validation instead of silently reverting to the default.

**Regression spec**: a new sleep-free `Slow Consumers` spec opens a DB with `WithRelayBufferSize(10)` and `WithStreamBufferSize(10)`, writes 2× the buffer size with a stalled consumer, then asserts every frame arrives in order. A correctly buffered relay absorbs the burst without the writer ever blocking; an unbuffered connection forces the writer to block on the full relay pipe past the slow-consumer timeout, guaranteeing drops before draining begins. Mutation-tested — reverting `connect()` to an unbuffered stream fails the spec deterministically.

**API cleanup**: audited the top-level `cesium` package for exports with no outside non-test callers. Removed the `DBStreamingConfig` type entirely — it had been unsettable since the dead-code sweep removed `WithStreamingConfig`, so its Override/Validate/merge machinery only ever produced the defaults, which are now constants. Removed the unused `ChannelCount` metric (core only reads `Metrics().DiskSize`) and inlined the writer/GC default configs at their single call sites. The streamer specs also share an `openStreamer` helper that threads the spec context through `NewStreamer` and the streamer's signal context.

The full Cesium suite passes with CI-matching flags (`--race --randomize-all`), the new spec passed 10 consecutive runs, and core builds and its `distribution/framer` suites (including the Relay suite) pass against the relay changes.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR restores distinct buffering for Cesium’s relay pipe and per-streamer connections while making writer and streamer tests deterministic.
- Adds validated relay and stream buffer options with defaults of 1000 and 100 frames.
- Initializes each relay connection with the configured stream buffer.
- Opens streamers before seed writes and explicitly drains seed frames in affected alignment tests.
- Removes unused Cesium configuration exports and the unused channel-count metric.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| cesium/options.go | Adds separate validated options and defaults for relay-wide and per-streamer buffering. |
| cesium/relay.go | Restores buffered streamer connections while preserving the existing slow-consumer timeout behavior. |
| cesium/open.go | Passes both configured buffer sizes into relay initialization. |
| cesium/streamer_test.go | Derives streamer lifetimes from the spec context and adds deterministic coverage for stalled consumers. |
| cesium/writer_test.go | Removes timing sleeps by connecting streamers before writes and explicitly draining verified seed frames. |
| cesium/db.go | Removes the unused ChannelCount metric while retaining disk-size reporting. |
| cesium/writer_open.go | Inlines the unchanged writer defaults at their sole construction site. |
| cesium/delete.go | Removes the exported GC default variable after moving its values to option initialization. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Writer[Cesium writer] -->|frames| RelayBuffer[Relay pipe<br/>default: 1000 frames]
  RelayBuffer --> Delta[Dynamic delta multiplier]
  Delta -->|20 ms slow-consumer timeout| StreamBuffer[Per-streamer buffer<br/>default: 100 frames]
  StreamBuffer --> Streamer[Streamer filtering]
  Streamer --> Consumer[Consumer outlet]
```

<sub>Reviews (2): Last reviewed commit: ["update ops"](https://github.com/synnaxlabs/synnax/commit/ad9c9d011a497ddd66d908b4164e96f436cf4132) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49743362)</sub>

**Context used:**

- Knowledge Base — [Cesium Storage Engine](https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/cesium-storage.md)


pjdotson:
@greptile /rerun

emilbon99 (approved):

PR #2660, opened 2026-08-04
SY-4529: Server-side task configuration versions and import RFC

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4529](https://linear.app/synnax/issue/SY-4529/strongly-type-tasks-on-core)

## Description

Adds RFC 0051: server-side task configuration versions and import. Today a task
holds its configuration as an opaque blob — `msgpack.EncodedJSON` in Go, a
`google.protobuf.Struct` on the wire. The Core stores the bytes and never reads
them, so it cannot migrate a stored config, cannot import one, and cannot reject
a malformed one. Every client carries its own copy of each config shape and its
own version chain.

The RFC gives the Core a typed copy of every task configuration without changing
what clients send or receive. Each task type gets an Oracle schema and its own
Gorp table. The stored configuration moves out of the task row into that table,
and the config record becomes a parent of the task. Both `type` and `config`
become resolved fields: the task row stores neither, and the task service
composes them on retrieve and decomposes them on write. The task payload keeps
its shape on every wire.

Key design points:

- The Core owns the version chain for every config type. A version bump rewrites
  the table at startup, so the Core always serves the latest shape and clients
  delete their private chains — the Console's NI `v0`/`v1` types, the Python
  client's old-shape readers, the Driver's parse tolerance.
- Each config type implements `imex.Importer`, which unblocks SY-4524. The task
  service stays the only exporter, because export routes on the ontology ID the
  caller holds and the envelope name lives on the task row. `Export` keeps its
  current flattened body shape but encodes a typed struct instead of an opaque
  map.
- Task types become a closed set. Every type has a schema, including the scanners
  and the rack status task, whose schemas are one line each over a shared empty
  base. An unknown type is a write error, and the startup migration quarantines
  any stored row it cannot convert rather than dropping it.
- The link reuses `parent_of` rather than a new relationship type, so it inherits
  the tree walk and the index. Deleting the config parent is new work: deleting a
  resource drops its own relationships and reaches no parent. The edge is
  additive: a task keeps its group parent and gains the config record as a second
  one, so the existing tree shape does not change here.
- The config record carries its own UUID, which gives it a portable identity
  across export and import and ontology standing for later per-type endpoints.
- Six phases, none of which break the wire: Oracle groundwork, schema authorship,
  tables and services, decompose/compose, import/export, client compat deletion.

The RFC keeps the task's current key, so it is independent of the SY-4488 UUID
re-key and composes with it in either order.

This PR is stacked on #2664 (SY-4548 RFC cleanup), which renames the RFC files to
dateless `NNNN-slug.md` names and adds the RFC conventions doc.

### Scope notes

An earlier draft of this RFC proposed typed task resources as root aggregates —
UUID identity, drafts, a deploy verb, and action-dispatch editing — and slimmed
the task to an execution record. That design is deferred and recorded as §6,
decision 1. The config record here is its storage half and does not foreclose it.

Device properties are out of scope in any form. The EtherCAT
`manual_<index>_<subindex>` keys and OPC NodeIds that camelCase conversion
corrupts live on `device.properties`, not on the task config, so repairing that
class is its own effort.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR adds RFC 0051, specifying server-owned, typed, versioned task configurations and server-side task import without changing the task wire payload.
- Splits task configuration into per-type Gorp records linked through ontology relationships.
- Defines composition, decomposition, migration, lifecycle, import/export, and access-control behavior.
- Organizes implementation into six independently shippable phases.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| docs/tech/rfc/0051-server-side-task-config-versions.md | Adds the server-side task-configuration RFC covering typed storage, migrations, composition, import/export, lifecycle, and rollout phases. |

<sub>Reviews (2): Last reviewed commit: ["Merge branch &#39;sy-4548-rfc-cleanup&#39; into ..."](https://github.com/synnaxlabs/synnax/commit/88b7a0a811cdc0869c8b4d13fae290452e06d652) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49899736)</sub>


pjdotson:
@greptile /rerun

emilbon99 (approved):
I fixed a few bugs and added a bunch of tests to improve missing coverage. LGTM now.

pjdotson:
> I fixed a few bugs and added a bunch of tests to improve missing coverage. LGTM now.

Did you respond to the right one? this is the task config RFC

emilbon99:
Whoops

PR #2663, opened 2026-08-04
J. SY-4384: Converge the Arc editor deploy UX onto the standard task pattern

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4384](https://linear.app/synnax/issue/SY-4384)

## Description

Stacked on #2605 (sy-4488-console-task-autosave). Converges the Arc editor's deploy flow onto the standard task pattern from RFC 0045, replacing the client-side three-call deploy with a core-owned atomic endpoint.

Previously, deploying an Arc automation was a non-transactional client-side sequence, the task config carried only the arc key so its configHash never changed, and running arcs never picked up program edits. Rack selection was lost without an explicit Deploy click.

Core changes: a dedicated `/arc/deploy` endpoint that creates or moves the automation task and its ontology relationship in one transaction, keeping the task's UUID key across rack moves. The server computes a semantic hash (xxhash64 of the program's meaning: mode, materialized text or sorted graph content, positions excluded) and serves it on every retrieve, create, and dispatch response. Deploy stamps the current hash into the task config `{arcKey, hash}`, so the existing configHash machinery, and the driver's rebuild-on-start gate, move exactly when the program's meaning changes. Zero driver changes. Undeploying (zero rack) deletes the task and is rejected while the task is running.

Console changes: picking a rack in the Arc editor now deploys immediately, so the choice is never lost, and the separate Deploy button is gone. Play deploys then starts, matching every other task form. A Redeploy button appears when the running instance has drifted from the stored program (hash or rack mismatch). The platform task controls were split into a presentational `Bar` (plain props, extras slot) and a thin form-context `Controls` wrapper; the Arc editor renders the `Bar` directly with the rack select in the extras slot, and the loose control primitives are no longer exported.

Verified against a live core: Go arc api/service suites, client (1552), full Pluto and Console suites green.

CRDT performance: the hash is computed on every retrieve, create, and dispatch, and hashing a text program requires materializing its CRDT document. The per-character document rebuild made a 1000-line program cost ~2.9ms and ~50k allocations per hash, so `x/crdt` (Go and TypeScript) now stores contiguous same-replica characters as run-length nodes, split only when a concurrent edit anchors inside a run. Wire operations, public APIs, and the shared conformance vectors are unchanged. A 1000-line hash now costs 273us and 40 allocations; client editor bootstrap drops from ~4ms to ~0.4ms. Both rewrites are pinned to the old implementations by differential tests: a frozen copy of the per-character code replays identical randomized multi-replica schedules and must match on generated operations, materialized value, per-index ids, and cross-implementation snapshot round-trips. `arc.Hash` also now trusts a pre-materialized `Raw`, and create responses materialize text so a client-supplied `Raw` can never reach the hash.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR moves Arc deployment into an atomic core endpoint, adds semantic program hashes, and adapts the Console to the standard task controls.
- Adds server-owned deploy, move, and undeploy behavior with task hash stamping.
- Exposes semantic hashes through Arc create, retrieve, and dispatch responses.
- Adds immediate rack deployment and drift-aware redeploy controls in Console.
- Regenerates Arc schema bindings across Go, TypeScript, Python, C++, and Protobuf.

The PR should not merge until Protobuf compatibility, deploy authorization, and out-of-order rack deployment are corrected.

Existing Protobuf tags are reassigned across versions, redeploys require creation permission despite updating an existing task, and concurrent rack selections can leave the task on an older selection.

**Files Needing Attention:** core/pkg/service/arc/pb/arc.proto, core/pkg/api/arc/arc.go, console/src/feature/arc/editor/TaskControls.tsx

| Filename | Overview |
|----------|----------|
| core/pkg/api/arc/arc.go | Adds hash responses and deploy authorization, but applies create permission to updates of existing deployment tasks. |
| core/pkg/service/arc/writer.go | Implements transactional deploy, move, and guarded undeploy while preserving task identity. |
| core/pkg/service/arc/hash.go | Computes a deterministic semantic hash while excluding graph layout and edge identity. |
| core/pkg/service/arc/pb/arc.proto | Adds the optional hash field but incompatibly renumbers existing Program and Status fields. |
| client/ts/src/arc/client.ts | Replaces client-side deployment orchestration with the new endpoint and updates cached hashes after dispatch. |
| console/src/feature/arc/editor/TaskControls.tsx | Deploys on rack selection and adopts task Bar controls, but concurrent rack changes can complete out of order. |
| console/src/platform/arc/useDrifted.ts | Detects running-task drift from config hash, rack, and semantic Arc hash. |
| pluto/src/arc/queries.ts | Exposes the deploy endpoint through a Flux mutation without mutation ordering. |
| schemas/synnax/arc.oracle | Adds a server-computed, non-persisted hash to the cross-language Arc schema. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Operator
  participant Console
  participant Client
  participant Core
  participant TaskStore
  Operator->>Console: Select rack or press Play
  Console->>Client: arcs.deploy(arcKey, rackKey)
  Client->>Core: POST /arc/deploy
  Core->>Core: Authorize and hash Arc
  Core->>TaskStore: Create or update task atomically
  TaskStore-->>Core: Deployment task
  Core-->>Client: "Task with {arcKey, hash}"
  Client-->>Console: Update task cache
  opt Play or Redeploy
    Console->>Client: Start task
  end
```

1. `core/pkg/service/arc/pb/arc.proto`, line 52-58 ([link](https://github.com/synnaxlabs/synnax/blob/a11a5d3a36e6857888f33ff661d1650c05794c5f/core/pkg/service/arc/pb/arc.proto#L52-L58)) 

   <a href="#">P1</a> **Protobuf fields break wire compatibility**

   When an Arc is exchanged with a version built before this PR, reassigning field 6 from `program` to `hash` and field 7 from `status` to `program` makes peers decode fields as the wrong message types, causing wire errors or loss of optional program and status data. Preserve the existing field numbers and assign `hash` a new number.

   **Knowledge Base Used:** [Oracle Codegen](https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/oracle-codegen.md)

2. `core/pkg/api/arc/arc.go`, line 1494-1502 ([link](https://github.com/synnaxlabs/synnax/blob/a11a5d3a36e6857888f33ff661d1650c05794c5f/core/pkg/api/arc/arc.go#L1494-L1502)) 

   <a href="#">P1</a> **Redeploys require creation permission**

   If a subject can update an existing Arc task but cannot create tasks, every nonzero deploy still enforces type-level task creation access before the writer reuses the existing task key, causing valid redeploys and rack moves to be denied.

</details>

<sub>Reviews (1): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/sy-..."](https://github.com/synnaxlabs/synnax/commit/a11a5d3a36e6857888f33ff661d1650c05794c5f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50147329)</sub>

<details><summary><h4>Context used (5)</h4></summary>

</details>


core/pkg/service/arc/writer.go
```diff
@@ -67,6 +75,8 @@ func (w Writer) Create(ctx context.Context, a *Arc) error {
 		if err = w.otgWriter.DefineResources(ctx, a.OntologyID()); err != nil {
 			return err
 		}
+	} else if err = w.syncTask(ctx, *a); err != nil {
```
sy-nico, line 78: `Hash` trusts any non-empty `Text.Raw`, so a create that overwrites an arc with a `Raw` that disagrees with `Doc` stamps a wrong hash into the task and the drift indicator lies until the next dispatch. This should hash the stored doc, like the create response already does instead of echoing a lying `Raw`.

pluto/src/arc/queries.ts
```diff
@@ -222,17 +223,93 @@ export const { useUpdate: useRename } = Flux.createUpdate<RenameParams>({
...
+  rack: rack.Key;
+}
+
+export const { useUpdate: useSetRack } = Flux.createUpdate<
+  SetRackParams,
+  task.Task | null
+>({
+  name: `${RESOURCE_NAME} rack`,
+  verbs: verbs.SET,
+  update: async ({ client, data }) => await client.arcs.setRack(data.key, data.rack),
```
sy-nico, line 238: The zero-rack case should be routed through `clearRack` instead of calling `setRack` directly. As written the just-deleted task keeps getting served from the local cache until the delete signal lands, and `clearRack`, which drops the cached task for exactly this case, has no callers at all.

core/pkg/service/arc/writer.go
```diff
@@ -166,17 +197,118 @@ func (w Writer) deleteChildTasks(ctx context.Context, key Key) error {
...
+		return nil, w.clearRack(ctx, existing)
+	}
+	var a Arc
+	if err = w.table.NewRetrieve().
+		Where(gorp.MatchKeys[Key, Arc](key)).
+		Entry(&a).
+		Exec(ctx, w.tx); err != nil {
+		return nil, err
+	}
+	return w.writeTask(ctx, a, rackKey, existing)
```
sy-nico: Should a rack move be rejected while the arc is running, the same way the zero-rack path is? Unbinding a running arc errors out, but moving its task to a different rack goes through with no running check, and if the old rack does not stop the instance this leaves it running detached from its task.

sy-nico (changes requested):

core/pkg/service/arc/writer.go
emilbon99: No. A rack move is ok. The 'redeploy' button will show up and then hitting it will do both the teardown to the the other rack. the reason you can't deselect is because deselecting will delete the running task. I think this UX needs to be refined but is ok for now.

sy-nico (approved):

PR #2668, opened 2026-08-06
N. SY-4299: Strongly-Typed OPC UA Task Types

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Moves the OPC UA integration onto Oracle-generated task config types, following the
NI pattern from Parts 1-3.

- **Schema**: `schemas/synnax/opc.oracle` defines `BaseChannel`, `InputChannel`
  (`channel`, `useAsIndex`), `OutputChannel` (`cmdChannel`), `ReadConfig`
  (extends `task.BaseReadConfig`, adds `device`, `arrayMode`, `arraySize`), and
  `WriteConfig` (extends `task.BaseWriteConfig`). Generated clients land in Go, TS,
  Python, and C++ (`client/*/opc`, `core/pkg/service/opc`).
- **Console**: `feature/opc/task/types.ts` parses only the generated schemas.
  Cross-field validation Oracle does not generate stays on the hand-composed deploy
  schemas: node-ID uniqueness, single index channel, array-mode size/rate coupling,
  stream-rate bounds, and command-channel uniqueness. Channel enablement follows the
  RFC 0043 polarity (`disabled`), threaded through the shared task components via the
  existing `polarity` prop.
- **Driver**: `InputChan`/`OutputChan` hold the parsed generated structs
  (`::synnax::opc::InputChannel`/`OutputChannel`) and keep only runtime state
  (resolved NodeId, fetched Synnax channel) as hand-written code.
- No legacy-key handling anywhere: configs stored with the old `enabled` polarity or
  the v0 write-channel `channel` key migrate once at Core boot (the storage cutover PR
  later in this chain).

## Basic Readiness

<h3>Greptile Summary</h3>

The PR moves OPC UA task configuration onto Oracle-generated, strongly typed schemas across clients, Console, Core, and Driver.
- Adds generated OPC UA task types for C++, Python, TypeScript, and Go.
- Updates Console forms and validation to use `disabled` polarity and generated schemas.
- Updates Driver read/write task parsing to exclude disabled channels before runtime validation.
- Extends Oracle C++ generation and tests for distinct-type defaults and inheritance ordering.

The PR appears safe to merge.

No blocking failure remains; the previously reported blank disabled-channel configuration now returns before node-ID and channel-key validation, with focused read and write regression coverage.

| Filename | Overview |
|----------|----------|
| schemas/synnax/opc.oracle | Defines the shared OPC UA channel and task configuration contracts used by all generated targets. |
| console/src/feature/opc/task/types.ts | Replaces hand-composed base shapes with generated OPC schemas while retaining cross-field deployment validation. |
| driver/opc/read_task.h | Parses generated input-channel structures and skips disabled channels before node and Synnax channel validation. |
| driver/opc/write_task.h | Parses generated output-channel structures and skips disabled channels before runtime channel construction. |
| oracle/plugin/cpp/json/json.go | Updates C++ JSON generation to construct explicit defaults for distinct string-backed types. |
| oracle/plugin/cpp/types/types.go | Adjusts generated C++ inheritance ordering consistently with resolved schema fields. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Schema["schemas/synnax/opc.oracle"] --> Oracle["Oracle code generation"]
  Oracle --> TS["TypeScript OPC types"]
  Oracle --> PY["Python OPC types"]
  Oracle --> GO["Core Go OPC types"]
  Oracle --> CPP["C++ OPC types and JSON parser"]
  TS --> Console["Console task forms and validation"]
  CPP --> Driver["OPC Driver read/write task parsing"]
  Driver --> Runtime["Enabled runtime channels"]
```

<sub>Reviews (2): Last reviewed commit: ["Skip disabled OPC channels before valida..."](https://github.com/synnaxlabs/synnax/commit/8b494e56ef1042826a66fbef0c384f95b09a2656) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50928380)</sub>

**Context used:**

- Knowledge Base — [Driver: Connecting Physical Devices to Synnax](https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/driver-devices.md)


driver/opc/read_task.h
```diff
@@ -50,12 +53,20 @@ struct InputChan {
...
+        InputChan(::synnax::opc::InputChannel::parse(parser), parser) {}
+
+    InputChan(::synnax::opc::InputChannel parsed, x::json::Parser &parser):
+        cfg(std::move(parsed)),
+        enabled(!this->cfg.disabled),
         node(types::NodeId::parse("node_id", parser)),
-        synnax_key(parser.field<synnax::channel::Key>("channel")) {}
+        synnax_key(this->cfg.channel) {
+        if (this->synnax_key == 0)
+            parser.field_err("channel", "channel must be specified");
```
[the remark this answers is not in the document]
emilbon99: *Emiliano + Claude: drafted in our review session.*

Real, and the fix needed to be wider than the suggestion. Guarding only the channel check still leaves `types::NodeId::parse` erroring on a blank disabled row, since the node id regex rejects an empty string. Both the read and write parsers now decode the generated channel and return early when it is disabled, so a disabled row is never validated at all. The per-channel `enabled` flag is gone with it, and there are regression tests in `read_task_test.cpp` and `write_task_test.cpp` pinning that an unbound disabled channel still configures.

emilbon99:
@greptile /rerun

client/py/synnax/__init__.py
sy-nico, line 1: `synnax.opc` duplicates `synnax.opcua`, which already hand-defines these same channel and config shapes.

The duplication is not cosmetic. `opcua.ReadChannel` serializes an `enabled` field, but after this PR the Driver parses only `disabled` (`InputChannel::parse`). A Python user who sets `enabled=False` sends a config with no `disabled` key, so it defaults to false and the Driver runs the channel anyway. The channel silently stays on, with no config error.

Why a new package? We wouldn't add `ni_task` beside `ni`. Can `@py output` target `client/py/synnax/opcua` so the generated types replace the hand-written ones, instead of sitting next to them?

driver/opc/write_task.h
```diff
@@ -34,19 +33,15 @@ struct OutputChan {
     /// be provided via the JSON configuration.
     synnax::channel::Channel ch;
 
-    explicit OutputChan(x::json::Parser &parser):
-        enabled(parser.field<bool>("enabled", true)),
-        node(types::NodeId::parse("node_id", parser)),
-        cmd_channel([&parser] {
-            auto ch = parser.field<synnax::channel::Key>("cmd_channel", 0);
-            if (ch == 0) ch = parser.field<synnax::channel::Key>("channel", 0);
```
sy-nico, line 42: This drops the `cmd_channel` -> `channel` fallback, and #2674's OPC legacy package says no rewrite is needed because "write channels carried cmdChannel from the first release."

`main` disagrees in two places: this fallback, and `v0OutputChannelZ` in the Console's `types.ts`. Two independent layers wouldn't defend a shape that never shipped. If they are dead code, can you confirm? Otherwise stored v0 write configs lose their command channel and fail with "channel must be specified".

driver/labjack/read_task_test.cpp
```diff
@@ -190,7 +190,7 @@ TEST(TestReadTaskConfigParse, testBasicReadTaskConfigParse) {
 
     ASSERT_EQ(cfg->sample_rate, x::telem::HERTZ * 10);
     ASSERT_EQ(cfg->stream_rate, x::telem::HERTZ * 5);
-    ASSERT_EQ(cfg->data_saving, true);
+    ASSERT_EQ(cfg->data_saving_disabled, false);
```
sy-nico, line 193: Should a labjack file should be changed in this PR?

oracle/plugin/cpp/json/json.go
```diff
@@ -1118,6 +1125,19 @@ func hasRenderableDefault(field resolution.Field, table *resolution.Table) bool
 	return jsonDefaultLiteral(field, table) != ""
 }
 
+// isDistinctType reports whether the type reference resolves to a distinct type.
+func isDistinctType(typeRef resolution.TypeRef, table *resolution.Table) bool {
```
sy-nico: `isDistinctType` here and `cppDistinctWrapper` in `cpp/types/types.go:829` are the same logic in two plugins. Can one live in a package both import, with the types plugin's version just calling `typeRefToCpp` on the result?

Second: both wrap string defaults only. Numeric defaults still go through `wrapCppTelemNumeric`, which name-matches `::telem::TimeStamp|TimeSpan|Rate|Size|Alignment`, so a distinct numeric outside that namespace emits a bare literal and fails copy-init the same way `DataType` did. Python already wraps both kinds. Should ints and floats be handled here too?

sy-nico (changes requested):
Mostly generated-code mechanics and reads clean. Two things to resolve before this merges are the `opc`/`opcua` split in the Python client, and whether the dropped v0 `channel` fallback is really unreachable.

driver/opc/write_task.h
emilbon99: Yes they are dead code.

sy-nico (approved):
The Python client never got the RFC 0043 polarity flip. `ni`, `opcua`, `ethercat`, `modbus`, `labjack`, and `http` all still write `enabled` and `data_saving`. The Driver reads `disabled` and `data_saving_disabled`, so those keys go nowhere and the channel runs anyway, with no error.

Already live on the parent: `client/cpp/ni/json.gen.h:313` parses `disabled`, `client/py/synnax/ni/types.py:160` writes `enabled: bool = True`. Not this PR's fault, shouldn't block it, but it needs an issue before the chain ships.

PR #2669, opened 2026-08-06
O. SY-4299: Oracle schemas for the remaining task integrations

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Authors Oracle schemas and generated clients (Go, TS, Python, C++) for every
remaining task integration: LabJack, Modbus, EtherCAT, HTTP, Arc, and PagerDuty.
This is schema and codegen only. The console and Driver rewires for these
integrations follow after the storage cutover, so they land with no legacy compat
code.

- Wire shapes match what the consoles write today, with the RFC 0043 polarity flip
  (`disabled`, `dataSavingDisabled`) applied. Old-shape stored configs migrate once
  at Core boot in the storage cutover PR.
- EtherCAT standardizes on `sub_index`, which the Driver already parses; the Console
  currently writes `subindex` and gets fixed at its rewire.
- HTTP models the endpoint/field structure the console writes (headers and query
  params as entry arrays, tagged static/generated write fields, enum value maps).
- Arc gains a generated `TaskConfig` (module key plus runtime loop tuning) extending
  the shared task base. PagerDuty gains `Alert` and `TaskConfig`.
- Fixes a Go generator gap: inline union variants with enum or scalar defaults
  produced no `ApplyDefaults`/`Validate` while containing structs still called them.
  NI's generated types pick up the resulting enum fills and membership checks.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR adds Oracle schemas and generated Go, TypeScript, Python, and C++ task-integration types, while correcting inline-union defaulting and validation generation.
- Adds schemas and generated clients for LabJack, Modbus, EtherCAT, HTTP, Arc, and PagerDuty
- Changes HTTP static write values from object-only `record?` to unrestricted optional `any?`
- Adds TypeScript coverage for scalar, object, array, and absent HTTP static values
- Generates defaults and enum validation for inline Go union variants

The PR appears safe to merge because the previously reported HTTP static-value validation defect is resolved and no blocking failure remains.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| schemas/synnax/http.oracle | Defines HTTP task contracts and now permits scalar, object, array, or absent static write values. |
| client/ts/src/http/types.spec.ts | Covers generated validation for string, number, boolean, object, array, and absent static values. |
| oracle/plugin/go/types/union.go | Extends Go union generation to emit variant defaulting and validation behavior. |
| oracle/plugin/go/types/types.go | Coordinates generated Go default and validation methods for the updated union behavior. |
| core/pkg/service/http/types.gen.go | Regenerates Go HTTP task types with an optional unrestricted static value. |
| client/cpp/http/types.gen.h | Regenerates C++ HTTP task types using optional generic JSON for static values. |
| client/py/synnax/http/types_gen.py | Regenerates Python HTTP task types using an optional unrestricted value. |
| client/ts/src/http/types.gen.ts | Regenerates TypeScript HTTP validation using an optional unknown static value. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  S[Oracle task schemas] --> G[Oracle code generation]
  G --> GO[Go task types]
  G --> TS[TypeScript client types]
  G --> PY[Python client types]
  G --> CPP[C++ client types]
  H[HTTP static value as optional any] --> G
  U[Inline union defaults and validation] --> G
```

<sub>Reviews (2): Last reviewed commit: ["Type the HTTP static write value as any"](https://github.com/synnaxlabs/synnax/commit/26f40d6597e7a4ed9cc0d13389795e4480f0d33f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50935030)</sub>

**Context used:**


schemas/synnax/http.oracle
```diff
@@ -0,0 +1,305 @@
...
+
+    @doc value "carries the fields every write body field shares."
+}
+
+WriteField union on type extends BaseWriteField {
+    static {
+        json_type JSONType = number {
+            @doc value "is the JSON type the value is serialized as."
+        }
+        value     record?  {
```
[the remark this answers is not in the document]
emilbon99: *Emiliano + Claude: drafted in our review session.*

Correct, `record` was the wrong declaration. `json_type` only ranges over string, number, and boolean, and the driver already reads the field as a bare `x::json::json`, so the object-only decoders were wrong in C++ and TypeScript both. The field is now `any?`, which generates `z.unknown().optional()`, `std::optional<x::json::json>`, `Any | None`, and a Go `*any`. Pinned with a new spec in `client/ts/src/http/types.spec.ts` covering scalars, objects, arrays, and the absent case.

emilbon99:
@greptile /rerun

oracle/plugin/go/types/union.go
```diff
@@ -168,8 +185,10 @@ func processUnion(entry resolution.Type, data *templateData) unionData {
 			validateHasOwn,
 			validateSkip,
 		)
-		vd.NeedsApplyDefaults = len(vd.DefaultRecurse) > 0
-		vd.NeedsValidate = len(vd.ValidateRecurse) > 0
+		vd.NeedsApplyDefaults = len(vd.DefaultRecurse) > 0 ||
+			len(vd.DefaultFills) > 0
+		vd.NeedsValidate = len(vd.ValidateRecurse) > 0 ||
```
sy-nico, line 190: This makes inline union variants enforce enum membership, and the regenerated `core/pkg/service/ni/types.gen.go` picks up 80 new `IsValid` checks. Nothing calls `Validate()` on those types yet, so this is inert here, but it stops being inert at the storage cutover. When migrated legacy NI configs get validated, what happens to one carrying an empty or legacy enum string? Does the migration normalize those, or does the task fail to configure on boot?

schemas/synnax/http.oracle
```diff
@@ -0,0 +1,305 @@
...
+        @doc value "is the command channel whose writes trigger this endpoint."
+    }
+    fields       WriteField[] {
+        @doc value "contains additional static or generated body fields."
+    }
+
+    @doc value "is a single HTTP endpoint written to by a write task."
+}
+
+WriteConfig struct {
```
sy-nico: `WriteConfig` hand-rolls `device` and `auto_start` instead of extending `task.BaseWriteConfig`, and `pagerduty.oracle:TaskConfig` does the same with `auto_start`. Both look like they are avoiding the inherited `data_saving_disabled`, but `arc.oracle:TaskConfig` extends `task.BaseConfig` and takes the unused field anyway. Can we pick one? Either split a base without `data_saving_disabled` or have all three extend.

schemas/synnax/arc.oracle
```diff
@@ -219,3 +220,39 @@ Arc struct {
...
+    @doc value "selects how the Arc runtime loop schedules execution."
+    @go output "core/pkg/service/arc/task"
+    @pb omit
+}
+
+TaskConfig struct extends task.BaseConfig {
+    arc_key        Key           {
+        @doc value "is the key of the Arc module the task executes."
+    }
+    execution_mode ExecutionMode = auto {
```
sy-nico, line 233: `core/pkg/service/arc/task/factory.go` reads only `ArcKey` and `AutoStart`. Nothing consumes `execution_mode`, `rt_priority`, `cpu_affinity`, or `lock_memory` yet. Is the runtime wiring landing in a later PR in this stack, or should these wait until there is something to read them?

schemas/synnax/ethercat.oracle
```diff
@@ -0,0 +1,153 @@
...
+@ts output  "client/ts/src/ethercat"
+@py output  "client/py/synnax/ethercat"
+@cpp output "client/cpp/ethercat"
+@go output  "core/pkg/service/ethercat"
+
+PDOAddress struct {
+    index      uint16 = 0 {
+        @doc value "is the CoE object dictionary index of the PDO entry."
+    }
+    sub_index  uint8 = 0 {
```
sy-nico, line 24: The description says the Console gets fixed at its rewire, but the mismatch is not new here. On `main`, `rc`, and this branch the Console writes `subindex` while `driver/ethercat/channel/channel.h:ManualInput` parses `sub_index` with no default. That reads like manually-addressed EtherCAT channels are already broken today, independent of this PR. Is there an issue tracking it?

sy-nico (changes requested):
Generated bulk reads clean, and the schemas match what the consoles write today.

One question on the Go union generator fix. It adds enum membership checks to NI, which is inert today since nothing calls `Validate()` yet, but it becomes load-bearing at the storage cutover. What does a migrated config with a stale enum value do there?

The other comments are about consistency. Arc's four new runtime fields have no reader yet, and three configs that all want to skip `data_saving_disabled` handle it two different ways. Arc inherits it unused, while HTTP write and PagerDuty copy the base fields by hand to avoid it.

schemas/synnax/arc.oracle
emilbon99: This all gets used in C++ already. The go runtime ignores these fields because it doesn't support RT.

oracle/plugin/go/types/union.go
emilbon99: We'll call `ApplyDefaults()` before validate, which will populate the enum.

schemas/synnax/ethercat.oracle
emilbon99: Good catch. I fixed this on #2674 which is the right place for it.

sy-nico (approved):

PR #2673, opened 2026-08-06
SY-4552: Move the telemetry client out of Pluto and into the client

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4552](https://linear.app/synnax/issue/SY-4552)

## Description

Implements phases 1 through 3 of RFC 0056. Pluto's telemetry client moves into the
`framer` package of `client/ts` as the `framer.Feed`, opened with `client.openFeed()`
in the `openWriter` family and owned by its caller.

Architectural changes:

- `pluto/src/telem/client/` (cache, reader, streamer, and their suites) is deleted. The
  cache internals live in `client/ts/src/framer/cache/` and stay unexported; the public
  addition is the `Feed` facade plus its handler, subscription, and transform types.
- A subscription is durable intent. Registering demand is synchronous and cannot fail
  for per-key reasons: the multiplexer owns retry, reconnect replay, socket key repair,
  and restart of a dead stream loop, and reports per-key data-plane status instead of
  throwing.
- Failure is per key rather than collective. A reader batch cycle rejects only the
  callers whose channels were in the failed fetch, and `close()` rejects pending reads
  with a typed error instead of stranding them.
- `channel.Retriever` and `ClusterRetriever` are deleted. Framer declares a plain
  function seam for channel resolution and `Synnax` binds it to the cached
  `channel.Client`, which severs the channel, framer, and query import cycle. The query
  cache drops its duplicated structural stream interfaces and takes a plain
  `framer.StreamOpener`.
- `channel.DebouncedBatchRetriever` is deleted. Its one non-duplicated capability,
  cross-call coalescing, moves to the query table fetch seam, so every consumer of
  `client.channels.retrieve` benefits rather than telemetry alone.
- A new `debounce.Batcher` in `x` owns the shared batching skeleton: requests queued
  with settle handles, a window timer that later enqueues do not reset, and close-time
  rejection. The query table and the telemetry reader both build on it.
- Disconnected clusters no longer need a noop client. Factories hold `null` and sources
  emit a single "cluster disconnected" status, then recover when the provider swaps in a
  connected core.
- The Pluto telem provider opens its own Feed with the WebGL series transform, so the
  transform lives with the thread that renders. It hands the Feed its instrumentation
  and reports close failures to the status aggregator.

Core-side changes:

- The Core stamps a streamed series with the time range its index established for the
  write group, so a client buffer starts at a real timestamp instead of the wall clock
  at arrival. A series without a stamp still falls back to the wall clock.
- The streamer's open acknowledgment fires after the relay applies the demand, so a
  write issued after a caller sees the ack can no longer be missed.

Fixes shipped along the way: sources carry generation guards so cleanup during an
in-flight read cannot leak stream subscriptions or acquired series, the dynamic buffer
drops a re-sent sample instead of duplicating it and resets across a single-sample gap,
constructor defaults survive explicit `undefined` props, and a live-core spec pins the
end-to-end Feed surface (reads, transform threading, stream delivery, close).

Phases 4 and 5 stay out of scope. The flux metadata bindings wait on the worker-safe
`flux.Retrieve` binding from the RFC 0055 read unification, so telem sources keep their
hand-rolled `valid` flags and the channel lookup ahead of `stream()` still has no retry.
The `Subscription` status surface ships unconsumed until the "no data" symbol treatment
exists to receive it.

## Basic Readiness


client/ts/src/framer/cache/cache.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+      const cache = new Cache();
+      cache.get(1);
+      cache.close();
+      expect(() => cache.get(1)).toThrow(UnexpectedError);
+    });
+    it("should close its entries", () => {
+      const cache = new Cache();
+      const entry = cache.get(1);
+      cache.close();
+      expect(() => entry.read(new MultiSeries([]).timeRange)).toThrow(UnexpectedError);
```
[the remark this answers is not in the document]
emilbon99: *Emiliano + Claude: drafted in our review session.*

This line is 87 characters, inside the 88-character limit (the limit is 88 inclusive). Prettier passes the file unchanged.

emilbon99:
@greptile re-review

emilbon99:
@greptileai review

client/ts/src/framer/cache/streamer.ts
```diff
@@ -0,0 +1,367 @@
...
+            this.setStatus(k, STREAMING);
+          } catch (e) {
+            // One channel's write failure must not stop the stream for the rest.
+            this.setStatus(k, status.fromException(e, "failed to buffer channel"));
+          }
+        if (changed.size !== 0)
+          this.entries.forEach((e) => {
+            if (e.removed) return;
+            try {
+              e.handler(changed);
```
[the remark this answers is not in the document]
emilbon99: *Emiliano + Claude: drafted in our review session.*

Intentional. The shared stream fans the full changed-key map out to every handler, and the StreamHandler doc in this file states that handlers may receive keys they did not subscribe to. Every consumer looks up only its own keys via res.get(key), so no data leaks into consumer state. Per-subscription filtering would rebuild a map per handler per frame at up to 60 Hz for no correctness gain.

client/ts/src/client.ts
```diff
@@ -151,30 +152,21 @@ export default class Synnax extends framer.Client {
...
-              },
-              (error) => this.conn.notify({ type: "stream.drop", error }),
-            );
-            // Reads start when the ObservableStreamer is constructed below,
-            // so onOpen fires strictly before any frame or reconnect.
-            onOpen?.();
-            this.conn.notify({ type: "stream.live" });
-            return new framer.ObservableStreamer(hardened);
-          }
+        ? async (config) => await this.openStreamer(config)
```
sy-nico, line 165: This resolves channels via `retrieveChannels` → `this.channels.retrieve` → the query cache, so the cache's own change stream resolves through the cache. §4.4 says that bootstrap "stays on the raw stream opener to avoid circularity," but §4.5 deletes `ClusterRetriever` and §7 rejects a raw resolution path outright. I don't think both can hold, so §4.4's sentence probably just goes.

The code ask is smaller: `query/streamer.ts:187` uses `opened ??= Promise.resolve().then(open)`, and that microtask hop is the only thing standing between here and infinite recursion. Nothing in `streamer.spec.ts` covers the re-entrant demand. Worth a regression test so nobody "simplifies" it back.

pluto/src/telem/aether/remote.ts
```diff
@@ -95,11 +115,18 @@ export class StreamChannelValue
...
+    if (client == null) {
+      this.onStatusChange?.(DISCONNECTED_STATUS);
+      return;
+    }
     try {
-      this.valid = true;
       this.removeStreamHandler?.();
-      const ch = await this.client.retrieveChannel(this.props.channel);
-      const handler: client.StreamHandler = (res) => {
+      const ch = await client.channels.retrieve(this.props.channel);
```
sy-nico, line 127: §5 puts the flux metadata bindings in phase 2, but the sources still hand-roll `valid` + `generation`. So this line is still the §1 bug: a failed lookup sets `valid = false` with no retry, and the Feed's retry only starts after `feed.stream()`.

Possible reason it's missing: §4.6 cites `pluto/src/flux/aether/retrieve.ts` as precedent, and that lives on the SY-4493/SY-4494 stack, not on `rc`. Is phase 2's binding waiting on that to land?

pluto/src/telem/aether/remote.ts
```diff
@@ -111,7 +138,8 @@ export class StreamChannelValue
         // Just because we didn't get a new buffer doesn't mean one wasn't allocated.
         this.notify();
       };
-      this.removeStreamHandler = await this.client.stream(handler, [ch.key]);
+      if (generation !== this.generation) return;
+      this.removeStreamHandler = client.feed.stream(handler, [ch.key]).close;
```
sy-nico, line 142: `Subscription` exposes `status(key)` and `onStatusChange()`, both implemented and spec'd, but every consumer destructures `.close` and drops the rest (also :341 and `log/aether/telem/sources.ts:210`). Nothing downstream can observe a reconnect or a per-key buffer error, so §4.7's "statuses are plumbed through every source" isn't true yet.

docs/tech/rfc/0056-client-telemetry-layer.md
```diff
@@ -0,0 +1,379 @@
...
+its own `openStreamer`.
+
+### 4.6 The pluto side: bindings replace the wrapper client
+
+`pluto/src/telem/client/` is deleted: `Core`, `NoopClient`, the telem aether provider's
+wrapper, and the frame cache all go. The telem aether provider opens a Feed from the
+current client with the GL series transform and closes it when the client identity
+changes, so the transform lives with the thread that renders. The fire-and-forget close
+race, the `prevCore` deadlock, and the NoopClient that throws `NotFoundError` into every
+symbol mounting during login all disappear structurally.
```
sy-nico: This says the fire-and-forget close race "disappears structurally." `prevCore` and `NoopClient` are genuinely gone, but `provider.ts:59` is `this.feed?.close().catch(console.error)`, which is the same shape as before.

docs/tech/rfc/0056-client-telemetry-layer.md
```diff
@@ -0,0 +1,379 @@
...
+  migration. No coexistence window.
+- **Phase 3: the framer merge and the cycle severing.** The module moves into `framer`
+  as the `Feed` with `openFeed` on `framer.Client`; `Synnax` drops its auto-built
+  instance and the construction-site transform option; the pluto telem provider opens
+  its own Feed with the GL transform; `channel.Retriever`, `ClusterRetriever`, and the
+  query cache's structural stream interfaces are deleted per §4.5.
+- **Phase 4: symbol status treatment.** The "no data" visual state and status plumbing
+  into schematic symbols. Split from phase 3 so rendering regressions bisect separately.
+
+**Compatibility.** No wire or persisted formats change. Behavioral deltas are
```
sy-nico: Two gaps here. "No wire or persisted formats change" holds for schemas but not payload semantics: `cesium/writer_stream.go` now stamps streamed series and `Dynamic.allocStart` reads that stamp.

The export list also names 2 of 5 removals. Also gone: `ClusterRetriever`, `createDebouncedBatchRetriever`, and `query.ObservableStream`/`StreamOpener`/`StreamOpenerHooks`.

client/ts/src/framer/cache/streamer.ts
```diff
@@ -0,0 +1,362 @@
...
+          await this.releaseStream();
+          this.breaker.reset();
+          ins.L.info("streamer closed, no keys to stream");
+          return;
+        }
+        if (
+          desired.size === this.sentKeys.size &&
+          Array.from(desired).every((k) => this.sentKeys.has(k))
+        )
+          return;
```
sy-nico, line 283: This early-out has no `this.streamer != null`, and `run()`'s `finally` (:354) mutates `streamer`/`sentKeys` without holding `connMu`. If I'm reading the ordering right:

1. `reconcile` holds the mutex, awaiting `streamer.update()`.
2. Run loop dies, so `finally` nulls `streamer`, clears `sentKeys`, and queues a reconcile (blocked on the mutex).
3. `update()` resolves, so `reconcile` writes `sentKeys = desired` over the cleared set.
4. Queued reconcile: `desired === sentKeys` hits the early-out, `streamer` still null.

No stream, no retry, status reads `success`. Narrow window, but it's the permanent demand/reality divergence §4.1 exists to kill. `this.streamer != null &&` closes it. Should the `finally` also take `connMu`?

client/ts/src/query/table.ts
```diff
@@ -131,18 +141,40 @@ export class Table<
...
     this.hydrateMode = hydrate;
+    this.fetchBatcher =
+      fetch == null
+        ? null
+        : new debounce.Batcher({
+            interval: fetchDebounce,
+            exec: async (entries) => {
+              const keys = new Set<Key>();
+              entries.forEach(({ req }) => req.forEach((key) => keys.add(key)));
+              const rows = await fetch(Array.from(keys));
```
sy-nico, line 169: If this throws, `Batcher` rejects every entry in the window, including callers whose keys weren't involved. That's the defect §1.2 cites as the reason to delete `DebouncedBatchRetriever`.

Per-key resolution currently holds only for "key not returned" (the `mine` filter at :174), not for throws. Per-entry settle in scope here, or follow-on?

client/ts/src/framer/cache/streamer.ts
```diff
@@ -0,0 +1,362 @@
...
+    ins.L.debug("adding stream handler", { keys });
+    // Deliver existing buffers before registration so a throwing handler leaves no
+    // demand behind.
+    const initial: Map<channel.Key, MultiSeries> = new Map(
+      keys.map((key) => [
+        key,
+        new MultiSeries(array.toArray<Series>(cache.get(key).leadingBuffer)),
+      ]),
+    );
+    handler(initial);
```
sy-nico, line 169: Live delivery wraps `handler()` in try/catch, but this one propagates out of `stream()`. I get the intent, but `feed.ts:83` documents the public surface as "failures surface as per-key statuses on the subscription, never as rejections." Worth carving the caller-bug case out there.

client/ts/src/framer/cache/streamer.ts
```diff
@@ -0,0 +1,362 @@
...
+  private async releaseStream(): Promise<void> {
+    const prev = this.streamer;
+    const prevLoop = this.runLoop;
+    this.streamer = null;
+    this.runLoop = null;
+    prev?.close();
+    if (prevLoop != null) await prevLoop;
+  }
+
+  private remove(entry: Entry): void {
```
sy-nico, line 216: `remove()` always arms a 5s timer, and `close()` only clears `reconcileTimer`. The callback itself is harmless since `scheduleReconcile` no-ops once closed, but the handles stay pending after the streamer is gone.

cesium/writer_stream.go
```diff
@@ -642,6 +644,15 @@ func (w *idxWriter) write(
 			incrementedSampleCount = true
 			w.hasUncommittedData = true
 			series.Alignment = alignment
+			// The index series carries the timestamps for the whole group, so data
+			// series can be stamped with a real time range without consulting the
+			// index on disk.
+			idxTimeRange = telem.TimeRange{
+				Start: telem.ValueAt[telem.TimeStamp](series, 0),
+				End:   w.idx.highWaterMark + 1,
+			}
```
sy-nico, line 653: What's the `+ 1` on `End`? Exclusive-end convention, or a guard?

Separately, this stamp is what `Dynamic.allocStart` now reads to set buffer start, falling back to wall clock when the range is zero. So a new Console on an old Core degrades rather than breaks, which seems intentional but isn't written down anywhere.

client/ts/src/framer/cache/streamer.ts
```diff
@@ -0,0 +1,362 @@
...
+const LOADING: status.Crude = { variant: "loading", message: "subscribing" };
+const STREAMING: status.Crude = { variant: "success", message: "streaming" };
+const RECONNECTING: status.Crude = { variant: "warning", message: "reconnecting" };
+
+/**
+ * Multiplexes durable subscriptions onto one shared frame stream. Demand
+ * registration is synchronous and cannot fail per key; all convergence work
+ * surfaces as per-key statuses instead of thrown errors.
+ */
+export class Streamer {
```
sy-nico: This class collides with the transport `Streamer` interface, which is why line 26 has to import that one as `Base`. `StreamOpener` is also declared twice in the package with different signatures, here at :38 and at `streamer.ts:96`.

RFC §2 already names this thing "Multiplexer" and nothing in the code uses it. Worth taking that name for the class, and distinguishing the two openers.

sy-nico (changes requested):
PR description is stale. It still describes the module as `telem.Client`, a sibling of the query and channel clients, which RFC 0056 §7 in this same PR rejects. The code ships `framer.Feed` with `openFeed()`. It also says "phases 1 and 2" while the diff covers phase 3: the framer merge, `openFeed`, and the `channel.Retriever` deletion.

client/ts/src/framer/cache/dynamic.ts
```diff
@@ -0,0 +1,268 @@
...
+      }
+    }
+    let curr = this.curr;
+    if (curr == null)
+      curr = this.allocCurr(res, series.alignment, this.allocStart(series), series, 0);
+    else {
+      // overlap > 0: the incoming series steps back into samples the current buffer
+      // already holds. overlap < 0: there is a gap between the buffer and the series.
+      const overlap = Number(curr.alignment + BigInt(curr.length) - series.alignment);
+      if (overlap > 1) {
```
sy-nico: Carried over unchanged from the old pluto cache, so not introduced here. But the comment two lines up says `overlap > 0` steps back into held samples and `overlap < 0` is a gap, while the code branches on `> 1` and `< -1`. That leaves `overlap === 1` writing a duplicated sample and `overlap === -1` writing across a one-sample gap as if contiguous.

Since §4.2 now writes the alignment contract down, is the ±1 deliberate tolerance or drift between the comment and the code?

sy-nico (changes requested):

cesium/writer_stream.go
emilbon99: (a) It's an end-exclusive convention.

(b) It wasn't designed for backwards compatibility, and version 0.56 consoles will break on 57 anyway. It's designed for handling certain cases where we can't assign timestamps.

docs/tech/rfc/0056-client-telemetry-layer.md
emilbon99: Corrected to clarify that there is a wire semantics change and we are reshaping APIs.

pluto/src/telem/aether/remote.ts
emilbon99: Fixed this in the RFC. Plumbing the statuses is somethign that will come later.

client/ts/src/client.ts
emilbon99: Pinned with a regression test, and made the code path more explicit.

pluto/src/telem/aether/remote.ts
emilbon99: Corrected this in the RFC to state that we haven't implemented yet, and will implemented later.

client/ts/src/query/table.ts
emilbon99: Two things:

1) Corrected the RFC. Our new implementation also throws on collective failure, so that was bad reasoning.
2) per-key resolution holds only for key not returned, which is the behavior we want. Anything else is a wire failure that affects every request by default.

emilbon99:
> PR description is stale. It still describes the module as `telem.Client`, a sibling of the query and channel clients, which RFC 0056 §7 in this same PR rejects. The code ships `framer.Feed` with `openFeed()`. It also says "phases 1 and 2" while the diff covers phase 3: the framer merge, `openFeed`, and the `channel.Retriever` deletion.

Fixed the RFC description.

client/ts/src/framer/cache/streamer.ts
```diff
@@ -269,8 +276,9 @@ export class Streamer {
           return;
         }
         if (
+          this.streamer != null &&
```
sy-nico, line 279: Add a test for this

client/ts/src/framer/cache/streamer.ts
```diff
@@ -189,6 +192,8 @@ export class Streamer {
     const { instrumentation: ins } = this.props;
     if (this.reconcileTimer != null) clearTimeout(this.reconcileTimer);
     this.reconcileTimer = null;
+    this.removalTimers.forEach(clearTimeout);
```
sy-nico, line 195: Needs tests

pluto/src/telem/aether/provider.ts
```diff
@@ -56,8 +72,15 @@ export const createProvider = (
     }
 
     private closeFeed(): void {
-      this.feed?.close().catch(console.error);
+      const { feed } = this;
+      if (feed == null) return;
+      // Cleared before the close is dispatched so a re-entrant swap cannot close the
+      // same feed twice.
```
sy-nico, line 78: Add test that forces a re-entrant swap

pluto/src/telem/aether/provider.spec.ts
```diff
@@ -111,9 +119,38 @@ describe("telem.Provider", () => {
...
+    update();
     expect(spy).toHaveBeenCalledTimes(1);
   });
+
+  it("hands the feed instrumentation so cache anomalies are not silenced", () => {
+    const core = stubCore();
+    mockUse.mockReturnValue(core);
+    mountProvider((client) => telem.createFactory(client));
+    const [props] = core.openFeed.mock.calls[0] as [{ instrumentation?: unknown }];
+    expect(props.instrumentation).toBeDefined();
```
sy-nico: This test only checks that instrumentation is defined, so a NOOP would still pass. Assert something actually logs through it.

docs/tech/rfc/0056-client-telemetry-layer.md
```diff
@@ -247,22 +249,29 @@ interfaces (both carried TODOs naming this cycle), and takes a plain
 `framer.StreamOpener`. The `Synnax` wiring for the query cache shrinks to handing over
 its own `openStreamer`.
 
+That opener resolves its own channel names through the cache, so the first open
+re-enters the demand that triggered it. The streamer memoizes the open before the opener
+body runs (`query/streamer.ts:182`), so the re-entrant demand joins the open in flight
```
sy-nico: Should reference the function name, not line number (in case it changes)

docs/tech/rfc/0056-client-telemetry-layer.md
sy-nico: Should say `**MultiplexedStreamer**`

client/ts/src/channel/client.ts
```diff
@@ -763,7 +763,6 @@ export class Client extends query.Retriever<
     });
   }
 
-  /** Fetches channel payloads. Missing channels are omitted, not thrown. */
```
sy-nico: Why was this dropped?

x/ts/src/telem/series.bench.ts
```diff
@@ -33,12 +33,6 @@ describe("data access", () => {
...
-  let sink = 0;
-  bench("partial buffer", () => {
-    sink += partial.data.length;
-  });
-  bench("full buffer", () => {
-    sink += full.data.length;
-  });
-  void sink;
+  bench("partial buffer", () => void partial.data.length);
+  bench("full buffer", () => void full.data.length);
```
sy-nico: What is this measuring? The `sink` accumulator was what kept the read alive. `void` discards the result, so the JIT is free to drop the subarray allocation this is timing. If the sink went to satisfy lint, can you find another way to "consume" the value?

sy-nico (changes requested):
Previous round is addressed. Asking for tests on the fixes that landed without them, plus some RFC and naming nits.

client/ts/src/channel/client.ts
emilbon99: Felt it was unnecessary, but I added it back.

sy-nico: Fair, you can keep removed. I'll mark this resolved either way.

sy-nico (approved):
Approved pending non-integration tests are passing. If integration tests are failing, ensure it is due to pre-existing issues on `rc` and not a failure introduced on this branch.

PR #2674, opened 2026-08-06
Q. SY-4299: Cut task storage over to per-type config records

emilbon99:
# Issue Pull Request

## Linear Issue

## Description

Stacked on #2671. This is the RFC 0051 phase 4 cutover: task configs stop living
as opaque JSON on the task row and become typed records in the per-integration
config stores.

- The task writer decomposes on create. The config decodes and validates through
  the type's registered store, persists under a stable record UUID that is
  `parent_of` the task in the ontology, and the config hash is computed over the
  canonical stored form (record key excluded, so equal configs hash equally
  across tasks). Snapshots keep their frozen record. A task with an unregistered
  type fails validation, and the config registry is now a required task service
  dependency.
- The task retrieve is hand-written now (the `@retrieve` directive is gone from
  the schema). It composes configs back by resolving each task's config record
  parent through the new index-backed `ontology.RetrieveParents` helper.
- Every stored record type gains a `CustomTypeName` table prefix
  (`opc_read_config`, `arc_task_config`, ...). Gorp derives table prefixes from
  bare type names, so the identically named `ReadConfig`/`TaskConfig` types
  across integrations would have shared one keyspace, and `TaskConfig` rows were
  swallowed by `Task` table scans.
- Boot migration `v57_task_config_records` walks stored tasks, normalizes every
  legacy JSON shape from the shim inventory (`data_saving`/`enabled` polarity
  flips, camelCase keys, NI device copy-down and the `ai_freq_voltage` alias,
  OPC/LabJack command channel renames, EtherCAT sub-index spellings, HTTP
  record-to-list conversions, PagerDuty alert polarity), renames the task types
  `arc` -> `arc_task` and `Rack Status` -> `rack_status`, recomputes hashes, and
  quarantines unconvertible rows under a KV prefix with a loud log.
- Internal tasks keep an ontology resource without a group parent, replacing the
  boot-time resource cleanup.
- The Arc task config schema gains the `hash` field Deploy stamps for driver
  rebuild detection; Go, C++, Python, and Console task type constants move to
  `arc_task` and `rack_status`.

The branch also carries the client and driver cutover that makes the chain
deployable:

- Scan tasks share a generated `task.BaseScanConfig` (`rate`, `disabled`), the
  v3 migration renames `scan_rate` and flips the enabled polarity for all six
  scan types, and the HTTP static write value becomes an `any` so primitive
  values decode.
- The built-in RBAC policies splice in the config record types
  (`taskcommon.ConfigResourceTypes`, RFC 0051 §4.9). Without this, ontology
  retrieves that surface a config record parent were denied, which blanked the
  Console's parent range button.
- The C++ driver parsers for LabJack, Modbus, EtherCAT, HTTP, NI, and OPC
  extend the generated types and parse through the generated parsers; the C++
  JSON generator renders the uuid `create` sentinel so locally parsed configs
  mint record keys.
- The Console storage schemas come verbatim from the generated client with
  deploy validation layered on top, and the legacy polarity machinery is gone.
- The hand-written Python config models are deleted; opcua folds into opc.
- Specs across TS, Python, and the Console create tasks with closed-set types
  and assert against the canonical server configs.

The whole chain still ships as one release.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR moves task configurations from opaque task-row JSON into typed, per-integration records and updates retrieval, migration, clients, drivers, schemas, and Console behavior for the new storage model.
- Persists canonical typed configuration records linked to tasks through ontology relationships.
- Migrates legacy task configurations and standardizes task types and generated cross-language models.
- Updates drivers, SDKs, RBAC policies, and Console forms to consume the composed configuration shape.

The PR appears safe to merge because no blocking failure remains from the previously reported writer issues.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| core/pkg/service/task/writer.go | Decomposes task configs into typed records, preserves snapshots, and performs type-change replacement within the request transaction. |
| core/pkg/service/task/retrieve.go | Composes task configurations by resolving and reading their ontology-linked typed records. |
| core/pkg/service/task/versions/v3/migrate.go | Migrates legacy task JSON into canonical per-type configuration records. |
| core/pkg/service/task/common/config.go | Defines the registry and typed configuration-store operations used by task persistence. |
| core/pkg/service/ontology/ontology.go | Adds parent retrieval support used to resolve task configuration records. |
| schemas/synnax/task.oracle | Updates the task schema and shared scan configuration model for generated clients. |
| core/pkg/service/access/rbac/builtin/builtin.go | Extends built-in access policies to cover task configuration resource types. |
| console/src/platform/task/types.ts | Aligns Console task configuration schemas with the generated canonical client models. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Client
  participant API
  participant Writer as Task Writer
  participant Store as Typed Config Store
  participant TaskDB as Task Table
  participant Ontology

  Client->>API: Create or update task
  API->>Writer: Create(task) in transaction
  Writer->>Store: Decode, validate, and persist config
  Store-->>Writer: Canonical config
  Writer->>Writer: Compute canonical config hash
  Writer->>TaskDB: Persist task row without config JSON
  Writer->>Ontology: Link config record as task parent
  Writer-->>API: Composed task
  API-->>Client: Task with canonical config
```

<sub>Reviews (3): Last reviewed commit: ["Skip disabled Modbus channels before val..."](https://github.com/synnaxlabs/synnax/commit/b80fa51f79b301322d6b6f1785a8b96d0bf5722c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50961452)</sub>

**Context used:**


core/pkg/service/task/writer.go
```diff
@@ -80,31 +84,115 @@ func (w Writer) healStatus(ctx context.Context, stat *Status) (*Status, error) {
...
+		}
+	} else {
+		if recordExists && existing.Type != t.Type {
+			oldStore, ok := w.configs.Store(ontology.ResourceType(existing.Type))
+			if ok {
+				if err := oldStore.Delete(ctx, w.tx, recordKey); err != nil {
+					return err
+				}
+			}
+			recordExists = false
```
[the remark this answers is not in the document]
emilbon99: Fixed in 656c88a4f8: the writer now stores the replacement record first and deletes the old type's record only after that write succeeds, with a regression spec covering the rejected-config path. (Reply written by Claude.)

emilbon99:
@greptile /rerun

core/pkg/service/task/writer.go
```diff
@@ -80,31 +84,118 @@ func (w Writer) healStatus(ctx context.Context, stat *Status) (*Status, error) {
...
+		}
+		// Delete the old type's record only after the new config is stored, so a
+		// rejected config cannot destroy the prior one.
+		if typeChanged {
+			oldStore, ok := w.configs.Store(ontology.ResourceType(existing.Type))
+			if ok {
+				if err := oldStore.Delete(ctx, w.tx, oldRecordKey); err != nil {
+					return err
+				}
+			}
```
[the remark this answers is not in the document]
emilbon99: *Emiliano + Claude: drafted in our review session.*

The premise does not hold here: there is no task write path that runs without a transaction. Every task route binds through `fgorp.CreateWriteUnaryHandler(db, ...)` in `core/pkg/api/layer.go:542`, which runs the handler inside `db.WithTx` and aborts the whole transaction on any error, so the config-store write, the old-record delete, the task row, the status, and the ontology writes all commit together or not at all. Imports take the same path: `imex` threads its `tx` into `s.NewWriter(tx).Create`. Nothing in the Core constructs a task writer with a nil transaction, and a type change can only arrive through the API.

emilbon99:
@greptile /rerun

schemas/synnax/ni.oracle
```diff
@@ -1475,14 +1475,14 @@ DigitalWriteConfig struct extends WriteConfig {
...
 
     @doc value  "configures an NI digital write task."
-    @go version 0
+    @go version 2
     @go marshal
     @go migrate
 }
 
-ScanConfig struct extends task.KeyedConfig {
+ScanConfig struct extends task.BaseScanConfig {
```
sy-nico, line 1483: Are we dropping the ignored models list? `ScanConfig` doesn't have an `ignored_models field`, but `driver/ni/scan_task.h` still parses it.

core/pkg/service/task/versions/v3/migrate.go
```diff
@@ -0,0 +1,262 @@
...
+		// Stored blobs are all console-era, which every store's legacy rewrite
+		// handles as version 0.
+		converted, err := store.Normalize(0, t.Config)
+		if err != nil {
+			if err := quarantine(ctx, tx, ins, otgW, t, err.Error()); err != nil {
+				return err
+			}
+			continue
+		}
+		if err := store.Write(ctx, tx, t.Key, converted); err != nil {
```
sy-nico, line 141: Doesn't this delete a task that fails to decode or validate during the upgrade? `stageAndRemove` drops the row, the status, and the ontology resource, and nothing reads `QuarantineKVPrefix` back, so it's just gone with no way to get it back.

schemas/synnax/opc.oracle
```diff
@@ -90,14 +90,14 @@ WriteConfig struct extends task.BaseWriteConfig {
...
     @doc value  "configures an OPC UA write task."
-    @go version 0
+    @go version 1
     @go marshal
     @go migrate
 }
 
-ScanConfig struct extends task.KeyedConfig {
+ScanConfig struct extends task.BaseScanConfig {
     @doc value  "configures an OPC UA scan task, which carries no settings."
```
sy-nico: This says the scan task carries no settings, but it extends `BaseScanConfig` now, so it has `rate` and `disabled`. Same line is stale in `ni.oracle`, `modbus.oracle`, `ethercat.oracle`, and `http.oracle`; `labjack.oracle` is the only one that got updated.

core/pkg/service/ontology/ontology.go
```diff
@@ -166,6 +166,30 @@ func (o *Ontology) RelationshipExists(
 		Exists(ctx, o.cfg.DB.OverrideTx(tx))
 }
 
+// RetrieveParents returns the parents of each of the given IDs, keyed by child ID.
+// Reads are executed against tx, falling back to the underlying database when tx is
+// nil. IDs with no parents are absent from the result.
+func (o *Ontology) RetrieveParents(tx gorp.Tx, ids ...ID) (map[ID][]ID, error) {
```
sy-nico: Is `RetrieveParents` meant to be public API, or just a test helper? The only callers I can find are `ontology_test.go` and `task_test.go`.

core/pkg/service/task/writer.go
```diff
@@ -147,8 +198,27 @@ func (w Writer) CreateMany(ctx context.Context, tasks *[]Task) error {
...
+		Where(gorp.MatchKeys[Key, Task](key)).
+		Entry(&t).
+		Exec(ctx, w.tx); err != nil {
+		if !errors.Is(err, query.ErrNotFound) {
+			return err
+		}
+		found = false
+	}
+	if found {
+		if store, ok := w.configs.Store(t.Type); ok {
```
sy-nico, line 218: This is the only thing that purges a config record, and `rack.Writer.DeleteGuard` deletes a rack without touching its tasks, so those tasks now orphan a config record on top of the row itself. The orphaned row is pre-existing, but the record is new, and nothing else will ever collect it. We should resolve this here or somewhere in the chain.

sy-nico: Also, what happens if an integration is dropped from the registry in a later build? This `ok` guard skips the delete, and `Create` rejects unknown types, so every record that integration owned would just sit there with nothing able to reach it.

sy-nico (changes requested):

core/pkg/service/ontology/ontology.go
emilbon99: Removed

schemas/synnax/opc.oracle
emilbon99: Fixed all doc comments

schemas/synnax/ni.oracle
emilbon99: Added back

core/pkg/service/task/versions/v3/migrate.go
emilbon99: What would you prefer we do? The tasks aren't lost we'd just have to release a patch to bring it back ya know

core/pkg/service/task/writer.go
sy-nico: Reopening the second half, it got closed along with the rack fix. But we still need to handle an integration being dropped from the registry in a later build: this `ok` guard skips the delete, and `Create` rejects unknown types, so every record that integration owned sits there with nothing able to reach it.

Or address here or with an inline comment.

core/pkg/service/task/versions/v3/migrate.go
sy-nico: We'd have to release a patch to fix, sure, but my concern is with the initial catch by the operator that there's an issue and the task just isn't there. (Panic will ensue)

Could the decode/validate branch keep the row and skip the config write instead? `composeConfigs` already treats a missing record as a nil config, so the task would still list and open, and it'd fail at deploy rather than vanish. Retired and unknown types deleting is fine, that's deliberate.

sy-nico (changes requested):
Design looks right, per-type stores are a better home for this than the inline config. I just need more discussion/assurance on the open threads.

core/pkg/service/task/writer.go
emilbon99: To be honest with you I don't think it's that important. It's unlikely for us to drop an integration, and the number of tasks that would be stranded would likely be small (under a few dozen at _most_). The user wouldn't see any visible implications from dropping it, just a slightly larger storage size.

That said, we would and should write migrations to remove tasks if we delete integrations, I just think that should be an explicitly written migration rather than a system that could accidentally delete data.

core/pkg/service/task/versions/v3/migrate.go
emilbon99: Discussed verbally.

sy-nico (approved):

PR #2676, opened 2026-08-07
SY-4286: Show a stale color on schematic state symbols

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4286](https://linear.app/synnax/issue/SY-4286)

## Description

Schematic symbols that read a state channel now change color when that channel<br>stops sending. Each symbol carries its own stale color and stale timeout, both<br>editable from the symbol form. An unset color resolves to the warning shade of<br>the theme in use.

image

image

### What staleness means

Staleness measures arrival, not sample time. It tells you if the source still<br>sends. A source that delivers old data keeps reading live, and a source that<br>sends more slowly than its timeout reads stale while it is healthy. Give a<br>channel that writes on change a timeout longer than the longest gap you expect<br>between changes. A symbol reads as stale only after data arrives and then stops.

### Design

A new `staleness.Provider` runs one periodic sweep for every source below it.<br>Sources register with the Provider and report each arriving sample. One shared<br>timer replaces the per-symbol timers, and it stops when the last source<br>unregisters. This replaces a `setTimeout` per sample in `stringValue`, and a<br>draw-time measurement in `value` that was correct but never asked for the repaint<br>that would show it.

The Provider keeps no copy of the staleness it reports. Each source holds its own<br>last reported value, so a transition is a change against the source itself.

`Value` and `Gauge` draw on the worker, so they keep staleness in internal state<br>and a transition costs a repaint and nothing else. The DOM-rendered symbols push<br>it through aether state. Two helpers hold the registration shape each side needs.

### Behavior changes

* `WithinBounds` and `RollingAverage` reported a sample only when their output<br>changed, so a steady channel turned its symbol stale while data still arrived.<br>Both now report every sample. `RollingAverage` also becomes a true sliding<br>window, and a NaN sample holds the average at NaN until it slides out.
* `Setpoint` loses its telem source. It opens at zero instead of the current<br>channel value, and the `source` spec a saved config carries is dropped the next<br>time that schematic is saved.
* `Gauge` loses `backgroundTelem`. Nothing supplied it.
* The switch symbol takes the color from its config. The primitive read the input<br>theme and ignored the color the form already offered.
* The channel field on every symbol form now reads `Channel`.
* The staleness fields carry their defaults into the form, so opening the<br>properties of a symbol saved before this branch marks the schematic changed.
* `stalenessTimeout` must be positive. Both the symbol schema and the aether<br>schema carry the rule.
* The schematic reference page explains what staleness measures and lists the<br>symbols that now carry a telemetry tab.

A follow-up issue covers the NaN sentinel the telem `Source` contract uses for an<br>absent value:<br>[SY-4564](https://linear.app/synnax/issue/SY-4564/give-telem-sources-an-explicit-absent-value-instead-of-nan).

## Basic Readiness

### Greptile Summary

The PR adds shared arrival-based staleness tracking and stale-state colors to telemetry-backed schematic symbols. The latest implementation also resets registration state when a mounted symbol changes telemetry sources.

* Adds a shared staleness provider and registration lifecycle.
* Integrates stale state, timeout, and color configuration across supported symbols.
* Updates telemetry transformations, tests, forms, and schematic documentation.

### Confidence Score: 5/5

The PR appears safe to merge.

No blocking failure remains.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| pluto/src/vis/staleness/aether/staleness.ts | Implements shared staleness registration and now replaces source-specific arrival state when the telemetry source changes. |
| pluto/src/vis/staleness/Provider.tsx | Mounts the staleness provider into the React and Aether provider hierarchy. |
| pluto/src/telem/aether/transformers.ts | Updates telemetry transformations so repeated arrivals continue to propagate and rolling averages use sliding-window behavior. |
| pluto/src/vis/staleness/aether/staleness.spec.ts | Covers registration lifecycle, source replacement, timeout transitions, and first-sample semantics. |
| pluto/src/pluto/Pluto.tsx | Adds the shared staleness provider around telemetry-backed Pluto content. |

### Sequence Diagram

```mermaid
sequenceDiagram
  participant Channel as Telemetry channel
  participant Source as Aether source
  participant Registration as Staleness registration
  participant Provider as Staleness provider
  participant Symbol as Schematic symbol
  Channel->>Source: sample arrives
  Source->>Registration: received()
  Registration->>Symbol: report live
  Provider->>Registration: periodic sweep
  Registration->>Symbol: report stale after timeout
  Note over Source,Registration: Changing source identity creates a fresh registration
```

Reviews (3): Last reviewed commit: ["SY-4286: Tie a staleness registration to..."](<https://github.com/synnaxlabs/synnax/commit/a87270f66776031dc99a5c0232ba8dad6e97b12b>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=51285389>)

**Context used:**

* Knowledge Base — [Pluto Component and Visualization Library](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/pluto-components.md>)


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

docs/site/src/pages/reference/console/schematics.mdx
```diff
@@ -240,11 +241,22 @@ The telemetry tab can have the following properties:
...
     </tr>
   </tbody>
 </table>
 
+Staleness measures arrival. It tells you if the channel still sends data. It does not
+tell you if the newest sample is recent. A channel that writes on change can pass its
+stale timeout while it is healthy. Give such a channel a timeout longer than the longest
+gap you expect between changes. A symbol registers as stale only after data arrives and
+then stops.
+
```
sy-nico, line 260: Do we really need this paragraph? You're already adding an entry in the Property table. The only useful part (imo) of this section is `Give such a channel a timeout longer than the longest gap you expect between changes. A symbol registers as stale only after data arrives and then stops.`

Even then, I think it could be reworded and maybe just wrapped in a note component.

docs/site/src/pages/reference/console/schematics.mdx
```diff
@@ -231,7 +232,7 @@ The telemetry tab can have the following properties:
       </td>
       {
         <td>
-          The amount of seconds without input channel telemetry before the element state
+          The number of seconds without data from the channel before the symbol
```
sy-nico: Preference for `The time in seconds` instead.

pluto/src/vis/staleness/Provider.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { type CrudeTimeSpan, TimeSpan } from "@synnaxlabs/x";
+import { type PropsWithChildren, type ReactElement } from "react";
+
+import { Aether } from "@/aether";
+import { staleness } from "@/vis/staleness/aether";
+
+// A quarter second keeps the reporting error small against timeouts measured in
+// seconds.
+const DEFAULT_SWEEP_INTERVAL = TimeSpan.milliseconds(250);
```
sy-nico: I think this could be bumped up to 500ms. 250ms is fine either way, so I won't block on this.

pluto/src/vis/staleness/aether/staleness.ts
```diff
@@ -0,0 +1,273 @@
...
+    source,
+  );
+};
+
+/**
+ * Resolves the color that stale content renders in.
+ * @param c - The configured staleness color. An unset (ZERO) color resolves to the
+ * theme's warning shade.
+ */
+export const resolveColor = (
```
sy-nico, line 267: Can you extend the adaptive `oklch()` coloring?

Image

Image

This is a pre-existing issue, but it would be nice if you could address it here.

sy-nico (changes requested):
Minor changes requested.

The screenshots in the PR description showing the different symbols in their various states is great. Can you verify everything looks good in the dark theme? No need to include screenshots.

pluto/src/vis/staleness/aether/staleness.ts
pjdotson: Going to leave this out of the current PR and into this issue: https://linear.app/synnax/issue/SY-4582/figure-out-schematic-coloring-oklch-story.

sy-nico (approved):

PR #2677, opened 2026-08-07
SY-4429: Return error pairs and pack every JSON type in the Any conversions

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4429](https://linear.app/synnax/issue/SY-4429)

## Description

`x::json::to_any` gave back an empty `google::protobuf::Any` on every failure, so a `Status` with unconvertible details went on the wire with no details and no error. The `record` conversions swallowed the same way with `.first`, which hit a task config, a task command's arguments, a device's properties, and a view's query.

`to_any` now returns `std::pair<google::protobuf::Any, x::errors::Error>` and accepts every JSON type. It packs a `google.protobuf.Value` instead of a `google.protobuf.Struct`, reports a `VALIDATION` error when the value does not convert (for example, an object nested deeper than the protobuf JSON parser accepts), and an `INTERNAL` error when `PackFrom` fails. The generated encoders leave the proto field unset when the value is null, so null and absent share one wire encoding. `from_any` converts an unset `Any` to null, unpacks both `Value` and `Struct`, and names the type URL when it can unpack neither.

The Go side mirrors this. `x/go/pb.AnyToPBAny` packs every JSON type in a `Value`, and `AnyFromPBAny` plus the generated `FromPBAny` translators accept `Value`, `Struct`, and typed protos.

Only generated code calls `to_any`, so the Oracle C++ `pb` plugin propagates the error. The generic template braces each `if constexpr` arm, so the JSON-field and type-param conversions emit bare statements inside those arms; the `record` conversions scope their own braced blocks. `oracle sync` regenerated the status, task, device, and view proto headers.

## Behavior change

`to_proto()` now fails where it made an empty object before. `status::Client::set` builds the full request before it sends, so one status that does not convert sends none of the batch, and the error names the offending status key. This is the wanted trade: a silently dropped payload is worse than a reported error. All callers in the repository use object-shaped details, so no known path changes.

On the wire, new senders pack `Value`, and new receivers still accept `Struct` from peers on releases before value packing. The reverse skew does not hold: a peer built from a release before value packing cannot unpack `Value`-packed details and reports a validation error on retrieve. Statuses are transient, so the error clears when both peers run the new release.

## Tests

`any_test.cpp` covers the pair API, a round-trip for every JSON type, the unknown-type error, and the propagated conversion error. Protobuf rejects an object nested more than 100 levels deep, which gives the `record` conversions a reachable failure; a shared `x::json::deeply_nested_object()` helper in `x/cpp/json/testutil` builds it. A task config, a command's arguments, a task status's data, and a status batch each prove that the error reaches the caller. Go specs round-trip every JSON type through `AnyToPBAny` and `AnyFromPBAny`. The generator specs assert the new emitted shapes and that the dropped dead branches stay out.

## Also in this PR

`CLAUDE.md` gains a prose-length rule, unrelated to [SY-4429](https://linear.app/synnax/issue/SY-4429/rewrite-to-any-to-return-an-error-pair-instead-of-swallowing-errors). Chat replies now count as prose, so they follow ASD-STE100 with the rest. A new subsection caps the length of chat replies, commit messages, and PR descriptions.

## Basic Readiness

### Greptile Summary

The PR makes C++ JSON-to-Protobuf conversions return and propagate typed errors instead of silently producing empty payloads.

* Updates `x::json::to_any` to validate inputs and return an error pair.
* Updates Oracle generation and regenerated C++ protobuf conversions.
* Adds conversion and propagation tests for status, task, command, and record data.
* Adds concise prose guidance to `CLAUDE.md`.

### Confidence Score: 5/5

The PR appears safe to merge.

No blocking failure remains.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| x/cpp/json/any.h | Changes `to_any` to validate object-shaped input and propagate Struct and packing failures. |
| oracle/plugin/cpp/pb/pb.go | Updates generated JSON, record, and generic conversion paths to return conversion errors. |
| client/cpp/status/status.h | Adds the failing status key to batch conversion errors before any request is sent. |
| client/cpp/status/proto.gen.h | Regenerates status conversion code for the new error-pair API. |
| client/cpp/task/proto.gen.h | Propagates task config, command argument, and status data conversion errors. |
| x/cpp/json/any_test.cpp | Covers successful conversion, invalid shapes, type-specific messages, and nested conversion failures. |

### Flowchart

```mermaid
flowchart LR
  J[JSON value] --> C[x::json conversion]
  C -->|success| P[Protobuf Struct or Any]
  C -->|failure| E[Typed conversion error]
  P --> G[Generated to_proto]
  E --> G
  G -->|success| R[RPC request]
  G -->|failure| A[Return error without sending]
```

Reviews (2): Last reviewed commit: ["rf"](<https://github.com/synnaxlabs/synnax/commit/59a0d31d31905c2e9dd3b04e7b0987b9d29ebc4c>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=51289741>)

<details><summary>Context used (3)</summary>

* Knowledge Base — [Client SDKs](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/client-sdks.md>)
* Knowledge Base — [Oracle Codegen](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/oracle-codegen.md>)
* Knowledge Base — [x: Shared Cross-Language Utilities](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/x-shared-utils.md>)

</details>


pjdotson:
@greptile /rerun

x/cpp/json/any.h
```diff
@@ -18,14 +19,26 @@
 #include "x/cpp/json/struct.h"
 
 namespace x::json {
-inline google::protobuf::Any to_any(const json &j) {
+/// @brief Converts json to a google::protobuf::Any that holds a Struct.
+/// @param j The JSON to convert. It must be an object or null. A null value converts to
+/// an empty object. Any other value returns a VALIDATION error.
+/// @returns A pair containing the Any and an error if one occurred.
+inline std::pair<google::protobuf::Any, errors::Error> to_any(const json &j) {
+    if (!j.is_null() && !j.is_object())
```
emilbon99: Why does this have to be an object? Why can't it be a primitive? If it has to be an object, maybe the function should be called `object_to_any`

x/cpp/json/any.h
```diff
@@ -18,14 +19,26 @@
...
-    auto [s, err] = to_struct(obj);
-    if (err) return any;
-    if (!any.PackFrom(s)) return google::protobuf::Any{};
-    return any;
+    if (!any.PackFrom(s))
+        return {{}, errors::Error(errors::INTERNAL, "failed to pack Struct into Any")};
+    return {any, errors::NIL};
 }
 
 inline std::pair<nlohmann::json, errors::Error>
```
emilbon99, line 38: Same here, should consider renaming this `object_from_any` if we do the same withe other function.

emilbon99 (changes requested):
Approved after my comment is addressed.

x/cpp/json/any.h
pjdotson: allowed them to be primitives and kept the same names

PR #2678, opened 2026-08-07
SY-4522: Server-side export and delete for schematic symbol groups

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4522](https://linear.app/synnax/issue/SY-4522)

## Description

The Core now exports and deletes a schematic symbol group. This is phase 4 of RFC 0052,<br>less import, which lands with [SY-4525](https://linear.app/synnax/issue/SY-4525/add-server-side-import-for-schematic-symbol-groups).

* **Export moves to the Core.** `symbol.Service.ExportGroup` serializes every symbol in<br>the group beside a `manifest.json`. It rejects a child that is not a symbol, two<br>symbols that take one file name, and a symbol that takes the manifest's name. File<br>names compare case-folded and Unicode-normalized, because a bundle extracts onto<br>case-insensitive filesystems.
* **The codec names its own extension.** `x/go/http` gains `FileCodec`: a codec that<br>also reports the extension its output takes. `ExportGroup` takes an encoder instead of<br>hardcoding JSON, so one point governs both the serialization and the extension every<br>file in the bundle carries.
* **File names survive every platform.** A new `x/go/os` helper replaces the path<br>separators, the characters Windows reserves, and the control characters Windows<br>forbids, drops the trailing dots and spaces Windows drops, pushes a name off a Windows<br>device name (`NUL` becomes `_NUL`), and shortens a name until it and its extension<br>together fit the 255 bytes ext4, APFS, and NTFS take. A name that sanitizes to nothing<br>comes back as a single underscore. An extension that fills a file name by itself is a<br>validation error, because no name can rescue one. Shortening can push two names onto<br>one file, which the exporter rejects. The Console's TypeScript sanitizer holds to the<br>same rules, character class and length limit included, for the name it gives the<br>downloaded archive.
* **The manifest goes to version 2**, carrying `{version, type, name}` only. Membership<br>is inferred from the files beside it, per RFC 0052 section 6.6.
* **Zip on the wire.** `POST /api/v1/schematic/symbol/group/export` responds with a zip<br>archive. A new `x/go/encoding/zip` encoder archives a file map, sorted by name, so<br>equal maps encode to equal bytes. It refuses an entry that is not a leaf: an empty<br>name, a name holding a path separator, and a dot entry. The Core builds the archive in<br>memory and writes it in one response; the Console reads it as a stream. The endpoint<br>is HTTP-only.
* **Group delete moves to the Core.** `DeleteGroup` removes the group and its symbols in<br>one transaction. The api reads the members through the same transaction it deletes<br>under, so the symbols it enforces delete access on and the symbols it removes come<br>from one snapshot, and it hands those same members to the service, so a delete walks<br>the ontology once. Export and delete share one membership check, so both name the<br>child that is not a symbol and both fail before any write lands.
* **Access is enforced on the group first.** Export and delete both check the group<br>before they read a member. A caller the group is closed to cannot read a member's name<br>out of a validation error, and cannot make the Core serialize a group it may not see.<br>Members are enforced on after the read, as before.
* **The symbol and schematic services require a group service.** It was optional, which<br>left `DeleteGroup` checking for one at request time. `ServiceConfig.Validate` now<br>rejects a nil group at construction.
* **Freighter fix.** The unary server encoded errors with the response encoder, which a<br>zip encoder cannot represent. The error path now negotiates against its own encoder<br>set, falls back to JSON, and sets the matching content type. Endpoints that emit JSON<br>or MessagePack keep the behavior their specs pinned.
* **API split.** The symbol endpoints move from `api/schematic` to<br>`api/schematic/symbol` and drop the `Symbol` suffix every name carried. Routes are<br>unchanged, so the wire and every client stay as they were.
* **Client.** `deleteGroup` drops the deleted symbols, the group's record, and the<br>group's relationships from the local caches, so a deleted group leaves a list the<br>moment the call returns instead of waiting for the change stream. This matches<br>`delete` and `groups.delete`.
* **Console.** Export streams the archive into the file the user picks. The group<br>importer reads both manifest versions, so the round trip stays green until server-side<br>import lands. Symbols and symbol groups rename in place instead of through a modal,<br>and the symbol modals now read `Schematic > Symbols > Create` and<br>`Schematic > Symbols > Group > Delete`.
* **Quiet test clients.** The Core refuses the change stream to a client built for a<br>restricted policy. Such a client no longer logs the refusal it is meant to get. The<br>log printed after the spec file finished, which vitest counted as an unhandled error.

## Behavior changes

* Exporting an empty group used to warn and write nothing. It now downloads an archive<br>carrying the manifest alone.
* The "a directory already exists" replace prompt is gone. The save dialog owns the<br>overwrite.
* Export writes a version 2 manifest only. A Console built from `main` reads version 1,<br>so it cannot import an archive this build writes. Import still reads both versions.

## Basic Readiness

### Greptile Summary

The PR moves schematic symbol-group export and deletion into the Core while adding ZIP transport, cross-platform filename sanitization, client cache cleanup, and Console integration.

* Exports deterministic ZIP bundles containing a version 2 manifest and serialized symbols.
* Deletes groups and their symbol members transactionally with access enforcement.
* Adds matching Go and TypeScript filename sanitization, including Windows device-name handling.
* Updates Freighter error negotiation, client APIs, Console import/export flows, and related tests.

### Confidence Score: 5/5

The PR appears safe to merge.

No blocking failure remains; both previously reported Windows reserved-name failures are corrected by prefixing reserved basenames before composing the exported filenames.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| x/go/os/filename.go | Adds byte-limited, cross-platform filename sanitization and now prefixes Windows-reserved device names. |
| x/ts/src/strings/sanitizeFileName.ts | Mirrors the Go sanitizer and correctly produces safe archive names such as `_CON.zip`. |
| core/pkg/service/schematic/symbol/bundle.go | Serializes group members into collision-checked filenames alongside a version 2 manifest. |
| core/pkg/service/schematic/symbol/group.go | Implements transactional symbol-group deletion using validated membership. |
| console/src/feature/schematic/symbol/export.ts | Replaces directory-based export with streaming download of the server-generated archive. |
| client/ts/src/schematic/symbol/client.ts | Adds group export and deletion endpoints with post-success cache eviction. |
| freighter/go/http/unary_server.go | Separates error-response encoding negotiation from endpoint response encoding. |

### Sequence Diagram

```mermaid
sequenceDiagram
    participant Console
    participant TSClient as TypeScript Client
    participant CoreAPI as Core API
    participant SymbolService as Symbol Service
    participant Storage

    Console->>TSClient: exportGroup(groupKey)
    TSClient->>CoreAPI: POST group/export
    CoreAPI->>SymbolService: authorize and export members
    SymbolService->>Storage: read group and symbols
    Storage-->>SymbolService: group snapshot
    SymbolService-->>CoreAPI: sanitized deterministic ZIP
    CoreAPI-->>TSClient: ZIP stream
    TSClient-->>Console: stream to sanitized archive filename

    Console->>TSClient: deleteGroup(groupKey)
    TSClient->>CoreAPI: POST group/delete
    CoreAPI->>SymbolService: authorize and delete transactionally
    SymbolService->>Storage: delete symbols and group
    Storage-->>SymbolService: commit
    CoreAPI-->>TSClient: success
    TSClient->>TSClient: evict group, symbols, relationships
```

Reviews (4): Last reviewed commit: ["Merge branch < />rc< /> into sy-4522-move-sche..."](<https://github.com/synnaxlabs/synnax/commit/603dfec8be09b5d5124e6763370d66df7f979f42>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=51290761>)


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

core/pkg/service/schematic/symbol/bundle.go
```diff
@@ -0,0 +1,113 @@
...
+	// manifestBaseName is the manifest's file name without an extension. The codec
+	// supplies the extension, so every serialization has one recognition point.
+	manifestBaseName = "manifest"
+)
+
+// GroupManifest is the body of manifest.json in a symbol group bundle. Membership is
+// inferred from the files beside the manifest, so it names no members.
+type GroupManifest struct {
+	// Version governs the manifest schema and the bundle's layout rules.
+	Version uint64 `json:"version"`
```
emilbon99: Nit: Why do we need a uint64 for this? Are we going to have over three billion versions of this? Is there standardization of what types you use for a counter version? I have a `version.Counter` type? Should we use that or should that become deprecated?

core/pkg/service/schematic/symbol/group.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// retrieveGroup returns the group identified by key together with its children. It
+// returns query.ErrNotFound if no group has key.
+func (s *Service) retrieveGroup(
+	ctx context.Context,
+	tx gorp.Tx,
+	key group.Key,
+) (root ontology.Resource, children []ontology.Resource, err error) {
+	err = s.cfg.Ontology.NewRetrieve().
```
emilbon99: Should we have an early return on error here?

core/pkg/service/schematic/symbol/group.go
```diff
@@ -0,0 +1,97 @@
...
+	return s.cfg.Group.NewWriter(tx).Delete(ctx, key)
+}
+
+// symbolIDs returns the ontology ID of every child, and a validation error naming the
+// first child that is not a schematic symbol.
+func symbolIDs(children []ontology.Resource) ([]ontology.ID, error) {
+	ids := make([]ontology.ID, 0, len(children))
+	for _, child := range children {
+		if child.ID.Type != ontology.ResourceTypeSchematicSymbol {
+			return nil, errors.Wrapf(
```
emilbon99: What did you end up deciding on the partial success semantics? I'm imagining by some odd reason something that is not a symbol ends up in this group. A customer comes to us and says 'I can't export symbols' now we have to either remove this code or do a painful migration to fix. Is there actually any danger in ignoring or warning on non-schematic symbol child types?

Are we embracing a partial success model or not?

core/pkg/service/schematic/symbol/group_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/synnaxlabs/synnax/pkg/service/group"
+	"github.com/synnaxlabs/synnax/pkg/service/ontology"
+	"github.com/synnaxlabs/synnax/pkg/service/schematic/symbol"
+	"github.com/synnaxlabs/x/errors"
+	"github.com/synnaxlabs/x/query"
+	. "github.com/synnaxlabs/x/testutil"
+	"github.com/synnaxlabs/x/validate"
+)
+
+// Group specs run on the per-spec tx so every row rolls back, but the fixtures they
```
emilbon99: Nit: this comment doesn't seem that useful to me.

x/ts/src/filename/sanitize.ts
emilbon99, line 1: Nit: the utilities should probably be in the same package in every language, that way its easy to know what to reach for regardless of what environment you are in.

x/go/encoding/json/json.go
```diff
@@ -19,37 +19,33 @@ import (
...
 	if err := json.NewDecoder(r).Decode(value); err != nil {
 		data, _ := io.ReadAll(r)
 		return encoding.SugarDecodingErr(data, value, err)
 	}
 	return nil
 }
 
-// EncodeStream implements the encoding.Encoder interface.
+func (*codec) Encode(_ context.Context, value any) ([]byte, error) {
+	b, err := json.Marshal(value)
```
emilbon99, line 46: Should there be an early return on this in the case of an error?

client/ts/src/schematic/symbol/client.ts
```diff
@@ -196,6 +200,54 @@ export class Client extends query.Retriever<typeof retrieveMultiParamsZ, Key, Sy
...
+   * @param key - the key of the group to delete.
+   * @throws {ValidationError} if the group holds a resource that is not a symbol.
+   */
+  async deleteGroup(key: group.Key): Promise<void> {
+    const groupID = group.ontologyID(key);
+    const rels = this.cfg.ontology.cache.relationships;
+    // Read the members before the delete drops the relationships naming them.
+    const memberKeys = rels.get((r) => matchChildRel(r, groupID)).map((r) => r.to.key);
+    await this.cfg.unary.send(
+      "/schematic/symbol/group/delete",
```
emilbon99, line 235: Have you put together a consistent theory on whether operations like this should be handled by the symbol service or by the group service? I feel like we're introducing a new pattern here without a shared agreement on how it should be done. To me this entire function could have been much simpler if it just passed a call through to the groups client.

console/src/feature/schematic/symbol/export.ts
```diff
@@ -9,109 +9,50 @@
...
+
 interface ExportGroupParams {
   client: Client | null;
   group: group.Group;
-  handleError: Status.ErrorHandler;
   addStatus: Status.Adder;
-  confirm: Modals.PromptConfirm;
 }
 
+// The Core owns membership, symbol serialization, file naming, and the manifest, and
```
emilbon99: This comment either needs to be removed or use proper JSDoc

console/src/feature/schematic/symbol/export.spec.ts
```diff
@@ -70,82 +67,21 @@ describe("exporting a symbol", () => {
 });
 
 describe("Schematic.Symbol.useExportGroup", () => {
```
emilbon99, line 69: Does this suite still have the level of quality it should? Was reduced to only one spec.

console/src/feature/schematic/symbol/types.ts
```diff
@@ -20,14 +20,34 @@ const manifestZ = z.object({
...
   name: z.string(),
   symbols: manifestZ.array(),
 });
 
-export interface GroupManifest extends z.infer<typeof groupManifestZ> {}
+/**
+ * The Core-written manifest. Membership is every JSON file beside it, so the manifest
+ * declares no members.
+ */
+const coreGroupManifestZ = z.object({
```
emilbon99, line 37: Why does this belong here? Shouldn't it be in the client? Is this something that will be purged in the future? Worried about desync across languages.

client/ts/src/testutil/access.ts
```diff
@@ -12,8 +12,18 @@ import { id } from "@synnaxlabs/x";
...
 import type Synnax from "@/client";
+import { AccessDeniedError } from "@/errors";
 import { createTestClient } from "@/testutil/client";
 
+/**
+ * Reports whether the error, or any error it wraps, is an access denial.
+ * {@link AccessDeniedError.matches} reads only the error it is given, so a wrapped
+ * denial hides its type behind the wrapper.
+ */
+const isDenial = (err: unknown): boolean =>
```
emilbon99, line 23: I think this is a workaround, and error matching should be improved. File a linear issue for this if you haven't already.

emilbon99 (changes requested):

console/src/feature/schematic/symbol/types.ts
pjdotson: yes, only because importing is still client-side. Update comment next to it.

client/ts/src/testutil/access.ts
pjdotson: https://linear.app/synnax/issue/SY-4568/improve-typescript-error-matching

console/src/feature/schematic/symbol/export.spec.ts
pjdotson: Yes, have Core and client tests now and changed the behavior of warning on empty groups.

core/pkg/service/schematic/symbol/bundle.go
pjdotson: It should be a `uint` because we start counting from 0. It isn't persisted as a
fixed-size unsigned integer here or in any other format I could imagine us using
(JSON,YAML,TOML all just would have generic number types). Presumably the other
`versions.Counter` should get deprecated because it is a signed type but I believe Aspen
stores it on disk so it would require a migration. Moved this to a `uint8`.

client/ts/src/schematic/symbol/client.ts
pjdotson: Right now the group client errors if you delete a group with children (as it would leave
those children orphaned). But if you wanted to delete a symbol group in a single
transaction, you need to make a single API call. I figured that the options were
something like:

1) add a `DeleteOptions` to the `GroupDeleteRequest` which allows for deleting a group and
   its children in one call. But every existing delete request is just an array of keys
   (except channels which can be deleted by name), so now we have an options format for
   deleting a group
2) Make schematic symbol group deletion explicitly managed by the symbol service, which
   is more code but now aligns towards the idea of "services have more control over the
   particular shape of the ontology graph".
3) Don't delete a group and its children at the same time.
4) Add a new ontology type for schematic symbol groups.

I don't particular know which one is the right solution, but both of the first two are
introducing a new pattern. I _think_ (but am not completely sure) that we want to move
towards something like the direction of 2 where the ontology graph is managed more via
server-side services. But these questions aren't really answered right now.

core/pkg/service/schematic/symbol/group.go
pjdotson: Decide to make this a warning but should see other comment, eventually we need to figure
out the ontology graph stuff more.

client/ts/src/schematic/symbol/client.ts
emilbon99: Copy, I'm ok with this since it's a two way decision, but we should probably prioritize defining a standard relationship management architecture for 58 or 59.

x/go/encoding/msgpack/msgpack.go
```diff
@@ -23,42 +22,50 @@ import (
...
 
-func (c *codec) ContentType() string { return "application/msgpack" }
+func (*codec) ContentType() string { return "application/msgpack" }
 
-// Encode implements the encoding.Encoder interface.
-func (c *codec) Encode(_ context.Context, value any) ([]byte, error) {
-	b, err := msgpack.Marshal(value)
-	return b, encoding.SugarEncodingErr(value, err)
+func (*codec) Decode(_ context.Context, data []byte, value any) error {
+	err := msgpack.Unmarshal(data, value)
```
emilbon99, line 33: Seems like `DecodeStream` is using a pool optimized fast path? Why doesn't this one do the same?

x/go/encoding/msgpack/msgpack.go
```diff
@@ -23,42 +22,50 @@ import (
...
+	}
+	return nil
 }
 
-// DecodeStream implements the encoding.Decoder interface.
-func (c *codec) DecodeStream(_ context.Context, r io.Reader, value any) error {
-	if err := msgpack.NewDecoder(r).Decode(value); err != nil {
-		data, _ := io.ReadAll(r)
-		return encoding.SugarDecodingErr(data, value, err)
+func (*codec) Encode(_ context.Context, value any) ([]byte, error) {
```
emilbon99, line 52: Same here: why not optimized?

x/go/encoding/orc/codec.go
```diff
@@ -139,3 +86,29 @@ func (c *codec) DecodeStream(ctx context.Context, rd io.Reader, value any) error
 	}
 	return c.Decode(ctx, data, value)
 }
+
+func (*codec) Encode(_ context.Context, value any) ([]byte, error) {
```
emilbon99, line 97: Fine with this, but please double check that the changes to this code don't introduce unnecessary heap allocations or performance regressions. This is very hot path.

x/go/gorp/migrate_test.go
```diff
@@ -243,7 +243,7 @@ var _ = Describe("Migrate", func() {
 
 	Describe("Error handling", func() {
 		It("Should not commit when a migration fails", func(ctx SpecContext) {
-			testDB := gorp.Wrap(memkv.New())
+			testDB := gorp.Wrap(memkv.New(), gorp.WithCodec(msgpack.Codec))
```
emilbon99: If you are going to pass in these options everywhere, why don't you consolidate into a helper so we don't have to keep redoing this on every new spec/suite?

emilbon99 (changes requested):
Approved after comments addressed. Nice work

x/go/encoding/msgpack/msgpack.go
pjdotson: `msgpack.Unmarshal` function already pulls from a pool internally.

x/go/encoding/msgpack/msgpack.go
pjdotson: same answer as above

PR #2683, opened 2026-08-10
SY-3482: Arc boolean support

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-3482](https://linear.app/synnax/issue/SY-3482)

## Description

Arc now has a first-class `bool` type. Previously, boolean logic was expressed through `u8` values, which made comparisons, channel writes, and flow routing lossy and inconsistent across runtimes.

### Language

- `bool` is a real type: `KindBool` in the type system, a `bool` type keyword, and `true`/`false` literals.
- Comparisons (`==`, `!=`, `<`, `>`, `<=`, `>=`) return `bool` instead of `u8`.
- Logical `and`/`or`/`not` require `bool` operands.
- Truthiness in `if`, `for`, and `=>` conditions: `bool` values, non-zero numerics, and non-empty strings are truthy.
- `str()` converts bool values, and format strings accept `bool` placeholders. `bool()` casts are rejected with an error that tells the user to write the comparison explicitly (for example `x != 0`).

### Flow and runtime

- Bool channels, bool stateful variables, and bool series (with element-wise logical operators and scalar broadcast).
- `select` routes on `true`/`false` keys, and `stable_for` preserves its input type instead of coercing to `u8`.
- Node params accept bool literals.
- The C++ embedded runtime mirrors all of the above, including bool support in `x/cpp/telem` series ops and the generated `x/go/telem/op` operators.

### Tooling and diagnostics

- LSP completion, hover, and semantic tokens gain coverage for `bool`, the literals, and the new operators. Remaining highlighting and formatting gaps are tracked for follow-up.
- Operand-type diagnostics report the offending type for logical operators (for example, `cannot use i32 in and operation`).

### Schemas and bindings

- `bool` added to `schemas/arc/types.oracle` with regenerated Go, TS, Python, C++, and proto bindings.

### Tests and docs

- New integration suite `integration/tests/arc/boolean.py`, expanded Go and C++ unit tests, and updated C++ E2E driver tests for the bool operators.
- Benchmarks for the series logical operators in both runtimes.
- The Arc language spec and the full docs-site Arc reference (operators, types, syntax, statements, concepts, and how-to guides) reflect the new semantics.

Remaining polish items are tracked separately and will land in follow-up PRs.

## Basic Readiness

<h3>Greptile Summary</h3>

Arc now has a first-class `bool` type across the language, compiler, runtimes, schemas, tooling, and documentation.
- Adds boolean literals, typing, comparisons, logical operations, truthiness, and formatting.
- Adds boolean channels, state, series operations, routing, and cross-runtime WASM support.
- Regenerates cross-language bindings and adds unit and integration coverage.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| arc/go/analyzer/expression/expression.go | Adds boolean expression analysis and enforces boolean operands for logical operators. |
| arc/go/compiler/expression/logical.go | Compiles scalar and series boolean logical operations to matching WASM operations. |
| arc/cpp/stl/series/series.cpp | Adds C++ host functions for boolean series allocation, indexing, and logical operations. |
| arc/go/stl/series/series.go | Adds the corresponding Go runtime host functions for boolean series. |
| schemas/arc/types.oracle | Adds `KindBool` as the schema source for regenerated cross-language Arc types. |
| arc/go/parser/ArcLexer.g4 | Adds boolean language tokens while retaining `^` only as exponentiation. |
| x/cpp/telem/series.h | Adds boolean telemetry series operations used by the embedded Arc runtime. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Source[Arc source with bool values] --> Parser[Parser and analyzer]
  Parser --> Compiler[Go WASM compiler]
  Compiler --> Program[Compiled Arc program]
  Program --> GoRuntime[Go runtime]
  Program --> CppRuntime[C++ embedded runtime]
  GoRuntime --> Telemetry[Boolean channels and series]
  CppRuntime --> Telemetry
```

<sub>Reviews (3): Last reviewed commit: ["header"](https://github.com/synnaxlabs/synnax/commit/3dee78164a470f2a800f9fd5747ed5125f48c6c8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51681868)</sub>

- Knowledge Base — [x: Shared Cross-Language Utilities](https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/x-shared-utils.md)
</details>


x/go/telem/op/gen.go
```diff
@@ -76,12 +76,18 @@ var operations = []Operation{
 // Modulo operations - uses % for integers, math.Mod for floats
 var moduloIntOp = Operation{Name: "Modulo", Op: "%"}
 
-// Logical operations only for uint8 (boolean) types
+// Bitwise logical operations for uint8.
```
pjdotson: shouldn't this be all integer types?

pjdotson (changes requested):
Some comments:

1) really need to fix color choice for LSP here. Why are boolean literals different colors than numeric and string literals? why is the boolean literal the same color as return? It is also very hard to tell the difference between the blue and the green in light mode. You should really just copy one-for-one the color scheme of another programming languages for which semantic tokens belong in which set of colors
2) why are we adding support for `&&`/`||`/`!`? we already have `and`/`or`/`not`. Now we have two different ways to express the exact same thing in this language.
3) Why did we add a keyword alias for `xor`? Have never seen that in a programming language before, only `^`
4) Bitwise operators don't and shouldn't be in the same PR.
5) Is there a need for getting bitwise operators out this release? we may have a better transition of `^` to `**` if we have a release or two where we deprecate the exponentiation meaning of `^` with an LSP warning.
6) not sure if this is preexising:

```
x := 34
y := 42.0 < x //shouldn't this be disallowed because we are comparing a float to an integer?
```
7) Why should this syntax be allowed:

```
if 34 {
    ...
}
```

Why don't we force the values to `if` statements to be booleans? Backwards compatibility?
7) What are the floating point number conversion rules for the `bool` operator? I am assuming that `+0`,`-0`, and `NaN` go to `false` and everything else goes to `true`?
8) Shouldn't the type outputs of `select` move from `u8` to `bool`? is that true for any other builtin functions (e.g. `stable_for`)?
9) If you are adding bitwise operators you should add left shift and right shift operators
10) If you are adding bitwise operators they should work on series.
11) Do the series operators also return `bool` for relational operators?

x/go/telem/op/gen.go
sy-nico: Removed the leftover from the u8-as-boolean era. The bitwise op set now generates these kernels for all integer types, so this u8-only set was a duplicate with no remaining callers.

sy-nico:
1. Fixed. It's supposed to match the same green as the string/numeric constants, but I broke it at some point.

2. Most languages only use keyword OR symbols. Python uses keywords for logical and symbols for bitwise. C++ has symbols and keywords for boolean, but the tokens are rarely used in practice. I've removed the boolean symbols.

3. Dropped the `xor` keyword. (C++, PowerShell, PHP, SQL, and a few others use `xor`)

4. This PR doesn't add a second feature. It splits one conflated operator family into two typed ones. Making and/or/not bool-only removes the old truthiness behavior on integers, and the bitwise operators are where that capability lands. Shipping only half the split leaves integers  with no logical or bitwise operators.

5. Deferring `^` would mean shipping the bitwise family with a hole in it, so I'd rather do the flip in one release. The transition risk is narrower than it looks: `^` on float operands fails loudly with a type error (exponent usage in sequences is almost all float math), so the silent-change case is integer-only and rare. For that, I added an LSP warning on every `^` this release ("`^` is now bitwise xor; use `**` for 
exponent"). Given how few Arc users exist today, we can also audit their sequences for `^` directly.

6. Pre-existing, but only in the `function` context. This correctly throws a type mismatch error in the `flow` context. [SY-4616](https://linear.app/synnax/issue/SY-4616/arc-function-variables-type-mismatch-handling)
image

7. Yes, backwards compatibility. Truthy `if`/`=>` on numerics is released, documented behavior (non-zero is true, non-empty string is true), and bool operands still work unchanged. This PR only tightens the logical operators, which were never documented as truthy.


7b. The logic remains the same, anything that is `!=0` is `true` (including `NaN`). Anything `=0` is `false`

8. Initially tried this, but it breaks the following automations:
```
select{} -> { false: next }
select{} -> { false: valve_cmd }
```
Users would have to change it to
```
select{} -> { false: true => next }
select{} -> { false: true -> valve_cmd }
```
 
`stable_for` returns the input value and maintains the type. It’s a debounce function, not a boolean “has been stable for Xs? True/false” question.


9. Added

10. Added

11. Yes. Relational operators on series return a bool series, in both the type system and the runtime. They were u8 series before this PR.

pjdotson (changes requested):
discussed in person

sy-nico:
Follow ups:
- [SY-4619: `select{}` refactor](https://linear.app/synnax/issue/SY-4619/arc-select-refactor)
- [SY-4620: LSP rework](https://linear.app/synnax/issue/SY-4620/arc-lsp-highlighting-rework)

arc/go/parser/ArcParser.g4
```diff
@@ -179,7 +179,15 @@ routingTable
     ;
 
 routingEntry
-    : IDENTIFIER COLON flowNode (flowOperator flowNode)* (COLON IDENTIFIER)?
+    : routingKey COLON flowNode (flowOperator flowNode)* (COLON IDENTIFIER)?
+    ;
+
+// TRUE and FALSE are lexer keywords but also valid routing-table keys
+// (select's outputs). Without these the IDENTIFIER-only rule rejects them.
+routingKey
```
pjdotson, line 187: link to linear issue since this is hacky

arc/go/parser/ArcLexer.g4
```diff
@@ -89,6 +94,14 @@ AND         : 'and';
 OR          : 'or';
 NOT         : 'not';
 
+// Bitwise
+AMP         : '&';
+PIPE        : '|';
+TILDE       : '~';
+CARET       : '^';
+LSHIFT      : '<<';
+RSHIFT      : '>>';
```
pjdotson: We should have compound assignment support for all of these operators:
- Six arithmetic: `+=`, `-=`, `*=`, `/=`, `%=`, `**=`
- Five binary bitwise:  `&=`, `|=`, `^=`, `<<=`, `>>=`

arc/go/parser/parser_test.go
```diff
@@ -1020,17 +1077,47 @@ any{ox_pt_1, ox_pt_2} -> average{} -> ox_pt_avg`)
...
+					parser.ParseExpression("a > 5 @ b < 10"),
+				).Error().
+					To(MatchError(ContainSubstring("token recognition error at: '@'")))
+			})
+
+			It("Should report error for repeated bitwise operators", func() {
+				// '&&' lexes as two AMP tokens and fails at the parser rather than
+				// the lexer.
+				Expect(
+					parser.ParseExpression("a && b"),
```
pjdotson: would we also error if we did `34 and 56`?

arc/go/analyzer/expression/expression.go
```diff
@@ -29,6 +29,21 @@ import (
 func isBool(t basetypes.Type) bool    { return t.IsBool() }
 func isNumeric(t basetypes.Type) bool { return t.IsNumeric() }
 
+// isInteger accepts untyped integer constants: they resolve to an integer type
+// through unification or the i64 default.
+func isInteger(t basetypes.Type) bool {
```
pjdotson: why would this be different than `t.IsInteger()`?

arc/go/analyzer/expression/expression.go
```diff
@@ -114,9 +129,70 @@ func Analyze(ctx context.Context[parser.IExpressionContext]) {
...
+		(wants == "integer" && unwrapped.IsBool())
+	if !fits {
+		return ""
+	}
+	return " Did you mean " + counterpart + "?"
+}
+
+// operatorHint states the operand types op takes and appends operatorSuggestion.
+// It is empty for operators outside the table.
+func operatorHint(op string, t basetypes.Type) string {
```
pjdotson: imo these hints aren't very helpful. I think it is unlikely people will be doing
bitwise arithmetic and if they are they wont need these hints.

arc/go/analyzer/types/infer_test.go
```diff
@@ -1017,3 +1063,74 @@ var _ = Describe("Type Inference", func() {
...
+		),
+		Entry(
+			"incompatible ^ operand keeps the series element type",
+			"s ^ x",
+			types.Series(types.I64()),
+		),
+		Entry(
+			"incompatible & operand keeps the series element type",
+			"s & x",
+			types.Series(types.I64()),
```
pjdotson: is this the same behavior as already exists for arithmetic operators? Should we require
casting long-term or no?

arc/cpp/types/types.cpp
```diff
@@ -273,6 +278,12 @@ to_sample_value(const x::json::json &value, const Type &type) {
         case Kind::F64:
             return value.is_number() ? std::optional(value.get<double>())
                                      : std::nullopt;
+        case Kind::Bool:
+            if (value.is_boolean())
+                return static_cast<uint8_t>(value.get<bool>() ? 1 : 0);
+            if (value.is_number())
+                return static_cast<uint8_t>(value.get<double>() != 0 ? 1 : 0);
+            return std::nullopt;
```
pjdotson: Why are we casting to `uint8_t` instead of `bool`?

arc/go/compiler/wasm/types.go
```diff
@@ -145,6 +145,42 @@ func binaryOpcode(op string, t types.Type) (Opcode, error) {
...
+		return OpI32Or, nil
+
+	case "<<":
+		if is64bit {
+			return OpI64Shl, nil
+		}
+		return OpI32Shl, nil
+
+	case ">>":
+		if strings.HasPrefix(t.String(), "u") {
```
pjdotson: shouldn't this be `if t.Unsigned()` or something like that?

arc/go/runtime/node/state_test.go
```diff
@@ -1594,6 +1594,31 @@ var _ = Describe("ProgramState", func() {
...
+							Outputs: types.Params{
+								{Name: outputParam, Type: types.Bool()},
+							},
+						}},
+					}
+					prog, diagnostics := graph.Analyze(ctx, g, nil)
+					Expect(diagnostics.Ok()).To(BeTrue())
+					s := node.New(prog)
+					n := s.Node("test")
+					*n.Output(0) = telem.NewSeriesV[bool](!last, last)
```
pjdotson: nit: unneeded type assertion

arc/go/stl/op/registry.go
```diff
@@ -15,9 +15,12 @@ import (
 )
 
 var (
-	logicalOps = map[string]op.Binary{orSymbolName: op.OrU8, andSymbolName: op.AndU8}
-	unaryOps   = map[string]op.Unary{notSymbolName: op.NotU8}
-	typedOps   = map[string]map[telem.DataType]op.Binary{
+	logicalOps = map[string]op.Binary{
+		orSymbolName:  op.OrBool,
+		andSymbolName: op.AndBool,
```
pjdotson: these probably could just be called `op.Or` and `op.And` if they only act on booleans

arc/go/stl/stable/stable.go
```diff
@@ -177,12 +178,13 @@ func (s *forNode) Next(ctx node.Context) {
...
-			*s.Output(0) = telem.NewSeriesV[uint8](currentValue)
+		if s.lastSent == nil || !bytes.Equal(s.lastSent, s.value) {
+			out := s.Output(0)
+			out.Resize(1)
+			copy(out.Data, s.value)
 			*s.OutputTime(0) = telem.NewSeriesV[telem.TimeStamp](s.now())
-			s.lastSent = &currentValue
+			s.lastSent = bytes.Clone(s.value)
 			ctx.MarkChanged(0)
 		}
```
pjdotson, line 189: why didn't code like the previous style work? we are calling `copy` and `bytes.Clone`
here, are there allocations happening?

arc/docs/spec.md
```diff
@@ -100,14 +105,22 @@ FloatLiteral ::= Digit+ '.' Digit* | '.' Digit+  // defaults to f64
...
 
-All types have default zero values: integers/floats `0`, string `""`, channels return
-zero on first read before write.
+All types have default zero values: integers/floats `0`, `bool` `false`, string `""`,
+channels return zero on first read before write.
 
 ### Type Casting
 
-Explicit casting between numeric types:
+Explicit casting between primitive types:
```
pjdotson, line 123: should we allow casting boolean values to integers?

x/cpp/telem/telem.h
```diff
@@ -1110,6 +1110,8 @@ class DataType {
             return this->cast(*static_cast<const uint32_t *>(value));
         if (value_type == details::UINT64_T)
             return this->cast(*static_cast<const uint64_t *>(value));
+        if (value_type == details::BOOLEAN_T)
+            return this->cast(*static_cast<const uint8_t *>(value));
```
pjdotson, line 1114: shouldn't the be a `bool`

arc/cpp/types/types.cpp
```diff
@@ -273,6 +278,12 @@ to_sample_value(const x::json::json &value, const Type &type) {
         case Kind::F64:
             return value.is_number() ? std::optional(value.get<double>())
                                      : std::nullopt;
+        case Kind::Bool:
+            if (value.is_boolean())
+                return static_cast<uint8_t>(value.get<bool>() ? 1 : 0);
```
pjdotson, line 283: shouldn't the be returning a `bool` instead of a `uint8_t`

x/cpp/telem/series.h
```diff
@@ -736,6 +791,10 @@ class Series {
             *reinterpret_cast<uint16_t *>(base_ptr) = static_cast<uint16_t>(value);
         } else if (dt == UINT8_T) {
             *reinterpret_cast<uint8_t *>(base_ptr) = static_cast<uint8_t>(value);
+        } else if (dt == BOOLEAN_T) {
+            *reinterpret_cast<uint8_t *>(base_ptr) = static_cast<uint8_t>(
```
pjdotson, line 785: this also probably should be `bool`

x/cpp/telem/series.h
```diff
@@ -1069,6 +1128,7 @@ class Series {
         if (dt == UINT32_T) return this->at<uint32_t>(index);
         if (dt == UINT16_T) return this->at<uint16_t>(index);
         if (dt == UINT8_T) return this->at<uint8_t>(index);
+        if (dt == BOOLEAN_T) return this->at<uint8_t>(index);
```
pjdotson, line 1121: should be `bool` too

x/cpp/telem/series.h
```diff
@@ -1099,6 +1159,8 @@ class Series {
             output_partial_vector(os, s.values<uint64_t>());
         else if (dt == UINT8_T)
             output_partial_vector_byte(os, s.values<uint8_t>());
+        else if (dt == BOOLEAN_T)
+            output_partial_vector_byte(os, s.values<uint8_t>());
```
pjdotson, line 1153: here

x/cpp/telem/series.h
```diff
@@ -1361,80 +1423,241 @@ class Series {
         }
     }
 
-    /// @brief Series > Series comparison. Returns UINT8_T Series with 0/1 values.
+    /// @brief Series-Series bitwise AND. Returns a new Series.
+    /// @throws std::runtime_error on floating-point types or length mismatch.
+    Series operator&(const Series &other) const {
+        this->check_bitwise("bitwise AND");
```
pjdotson: why do we throw for bitwise if we don't throw for others?

arc/go/stl/series/series.go
```diff
@@ -120,7 +144,7 @@ func NewSymbols() []*symbol.Symbol {
 			resultOut,
 		),
 		symbol.InternalHostFunc(
-			"not_u8",
+			"not_bool",
```
pjdotson: can't these functions just be called `"and"`, `"not"`, and `"or"`

arc/go/stl/series/series.go
```diff
@@ -256,17 +369,200 @@ type seriesOps[T any] struct {
...
+				if s1.Len() != s2.Len() {
+					panic("arc panic: series length mismatch in bitwise operation")
+				}
+				result := telem.Series{DataType: s1.DataType}
+				fn(s1, s2, &result)
+				return s.Store(result)
+			}).Export(entry.name + suffix)
+	}
+	return builder
+}
```
pjdotson: a lof of duplicated logic in the `WithFunc` blocks

docs/site/src/pages/reference/control/arc/reference/operators.mdx
```diff
@@ -48,31 +44,31 @@ result := 10.0 / 3.0 // 3.333...
...
-| `<`      | Less than             | `3 < 5`  | `1`    |
-| `>`      | Greater than          | `5 > 3`  | `1`    |
-| `<=`     | Less than or equal    | `3 <= 3` | `1`    |
-| `>=`     | Greater than or equal | `5 >= 5` | `1`    |
+| `==`     | Equal                 | `5 == 5` | `true` |
+| `!=`     | Not equal             | `5 != 3` | `true` |
+| `<`      | Less than             | `3 < 5`  | `true` |
+| `>`      | Greater than          | `5 > 3`  | `true` |
+| `<=`     | Less than or equal    | `3 <= 3` | `true` |
+| `>=`     | Greater than or equal | `5 >= 5` | `true` |
```
pjdotson: table would make more sense if left was always 3 and right was always 5

docs/site/src/pages/reference/control/arc/reference/types.mdx
```diff
@@ -226,8 +213,14 @@ b i64 := i64(a) // truncates to 3
...
 s str := str(n) // numeric to string: "42"
+
+active := bool(n) // numeric to bool: n != 0
+level u8 := u8(active) // bool to numeric: 1 or 0
 ```
 
+<Note.Note variant="info">
+  Casting a string to a numeric or boolean is not supported.
+</Note.Note>
 ### Casting Rules
```
pjdotson, line 227: add space

pjdotson (changes requested):

arc/go/parser/ArcLexer.g4
sy-nico: [SY-4622 Followup](https://linear.app/synnax/issue/SY-4622/compound-assignment-for-bitwise-operators)

arc/go/parser/parser_test.go
sy-nico: Yes, error message is 
```
cannot use integer in and operation: and takes bool operands. Did you mean & (bitwise and)?
```

arc/go/analyzer/expression/expression.go
sy-nico: `t.IsInteger()` matches only the concrete sized integers (u8 through i64). The analyzer runs pre-unification, so an untyped literal like the 1 in `1 << 3` is still an integer constant at this point, and `IsInteger()` rejects it. This helper widens the check for that case. `IsNumeric()` already includes the constant kinds, which is why `isNumeric` above is a plain forward and this one is not. Widening `IsInteger()` itself would change a versioned types method that backs concrete-kind switches like `IntegerMaxValue`, so the check stays local to the analyzer.

arc/go/analyzer/types/infer_test.go
sy-nico: Same behavior. The bitwise infer functions share `inferBinaryType` with the arithmetic ones, and these entries mirror the existing arithmetic tests (`data_series + float_var` infers `series i64`). The analyzer still rejects `s & x` with a type mismatch. This only pins the recovery type so downstream expressions don't cascade.

arc/go/compiler/wasm/types.go
sy-nico: Yes, I was copying the pre-existing pattern used for `%`. 
Fixed to `t.IsUnsignedInteger()`, which the rest of `binaryOpcode` already uses. Also updated the `%` case to match.

arc/go/analyzer/expression/expression.go
pjdotson: I am a little confused here - the method name implies that a type of `KindIntegerConstant` is numeric but isn't an integer. why couldn't the `IntegerMaxValue()`/`IntegerMinValue()` methods also work for an integer constant - surely we don't allow integer constants greater than `1<<64-1`

arc/go/stl/op/registry.go
sy-nico: Renamed to `op.And`/`op.Or`/`op.Not`.

arc/go/stl/stable/stable.go
sy-nico: The previous style was the bug that read every input with `ValueAt[uint8]` and re-emitted a `uint8` series, so a bool or f32 passing through came out retyped. Preserving the input type means handling raw element bytes, and `inputData.At(i)` returns a view into the input series buffer, so storing it without `bytes.Clone` would alias memory the runtime reuses on later ticks. Allocations only happen on a value change or emission, not per tick, and the old path also allocated there (`NewSeriesV` per emission); `Resize(1)` + `copy` reuses the output buffer in steady state.

arc/docs/spec.md
sy-nico: Yes, that's intentional. Explicit casts like `u8(flag)` give 0/1, which is useful for counting and math on conditions.

x/cpp/telem/telem.h
sy-nico: `SampleValue` has no `bool` alternative, so booleans are carried as `uint8_t` 0/1 throughout the C++ layer. Adding `bool` to the variant means updating every visitor,. Follow-up: [SY-4623](https://linear.app/synnax/issue/SY-4623/add-a-bool-alternative-to-the-c-samplevalue-variant).

x/cpp/telem/series.h
sy-nico: Same reply as this [comment](https://github.com/synnaxlabs/synnax/pull/2683#discussion_r3806129070)

x/cpp/telem/series.h
sy-nico: Same reply as this [comment](https://github.com/synnaxlabs/synnax/pull/2683#discussion_r3806129070)

x/cpp/telem/series.h
sy-nico: Same reply as this [comment](https://github.com/synnaxlabs/synnax/pull/2683#discussion_r3806129070)

arc/cpp/types/types.cpp
sy-nico: Same reply as this [comment](https://github.com/synnaxlabs/synnax/pull/2683#discussion_r3806129070)

arc/go/stl/series/series.go
sy-nico: These follow the per-element-type suffix convention every series import uses (`compare_eq_f32`, `element_add_u8`, ...). Logical ops only instantiate for `bool`, so `and_bool`/`or_bool`/`not_bool` are the only ones. The Arc-level operators are already bare `and`/`or`/`not`.

x/cpp/telem/series.h
sy-nico: The others do throw: `apply_binary_op` throws on length and type mismatch for every operator. Bitwise has one extra invalid case (floats, where `&`/`|`/`^` don't exist), and without the check the `if constexpr` fallback would silently return the lhs unchanged.

arc/go/stl/series/series.go
sy-nico: Deduped into a shared `bindSeriesBinary` helper. This touches the pre-existing `bindSeriesOps`/`bindCompareOps` closures too, and `and_bool`/`or_bool` now length-check like the rest.

arc/go/stl/series/series.go
pjdotson: right, my point is the `_bool` suffix doesn't really make sense unless you would have `and_u8`.

sy-nico:
@greptile, rerun

arc/go/stl/stable/stable.go
pjdotson: Ah okay. Isn't there a `Samples()` iterator you could use?

x/cpp/telem/telem.h
pjdotson: Make sure SY-4623 gets in before we release then.

arc/docs/spec.md
pjdotson: Do we use this internally? do we know if people will use it?

Not every language has this, so I would rather be conservative about it because it is much easier to add it in later than take it out once it already exists.

schemas/arc/versions/types/v1.oracle
```diff
@@ -7,7 +7,37 @@
...
+    variable                     = 14
+    numeric_constant             = 15
+    integer_constant             = 16
+    float_constant               = 17
+    exact_integer_float_constant = 18
+    function                     = 19
+    sequence                     = 20
+    stage                        = 21
+    var_ref                      = 22
+    bool                         = 23
```
pjdotson: should just add `bool` to `v0.Kind`, this is a superset so no need for a new type

docs/site/src/pages/reference/control/arc/reference/operators.mdx
```diff
@@ -82,42 +78,49 @@ in_range := pressure >= 100 and pressure <= 500
...
 
-| Operator | Name        | Example | Result |
-| -------- | ----------- | ------- | ------ |
-| `-`      | Negation    | `-5`    | `-5`   |
-| `not`    | Logical NOT | `not 1` | `0`    |
+| Operator | Name        | Example    | Result        |
+| -------- | ----------- | ---------- | ------------- |
+| `-`      | Negation    | `-5`       | `-5`          |
+| `~`      | Bitwise NOT | `~12`      | `-13` (`i64`) |
+| `not`    | Logical NOT | `not true` | `false`       |
```
pjdotson: somewhat odd to repeat the notes for `~` and `not`. Maybe should just move the unary negation operation into the arithmetic operators section and delete this.

arc/cpp/stl/stable/stable.h
```diff
@@ -87,10 +89,15 @@ class StableFor : public runtime::node::Node {
             const auto &input_data = this->state.input(this->input_idx);
             const auto &input_time = this->state.input_time(this->input_idx);
             if (input_data->size() > 0) {
+                const auto density = input_data->data_type().density();
+                const auto *data = reinterpret_cast<const uint8_t *>(
+                    input_data->data()
+                );
                 for (size_t i = 0; i < input_data->size(); i++) {
```
pjdotson, line 96: does this (and the Go implementation) work for variable-length channels?

pjdotson (changes requested):
Have you benchmarked for performance regressions in either runtime?

Mostly there generally looks good.

arc/go/analyzer/expression/expression.go
sy-nico: Widened `IsInteger()` to accept integer constants and reduced the analyzer helper to a plain forward; `IntegerMaxValue()`/`IntegerMinValue()` already return the i64 bounds for constants via their default branch.

arc/go/stl/stable/stable.go
sy-nico: There is, but it yields sample bytes without an index, and this loop needs the index to read the parallel timestamp series. Tracking a counter by hand alongside the iterator is no cleaner, and the yielded samples alias the buffer too, so the clone stays either way.

arc/docs/spec.md
sy-nico: Talked offline. We are in a transitional state. Some functions now output `bool`, but DI/DO hardware channels are still numeric until they get bool typing. Users need a way to command those channels in the meantime, e.g. `u8(true) -> some_do_channel`, so we keep bool to integer casts. `bool(x)` stays rejected.

arc/cpp/stl/stable/stable.h
sy-nico: Did not work and would compile. The error would not appear until after the program was stopped. I instead added an explicit check to block variable length channels as inputs.

sy-nico: The check lives in `ResolveNodeTypes` with the stable node names hard-coded. I tried moving it into `stable.go` via a symbol hook, but that required a new field on `symbol.Symbol`, a `ResolveNodeTypes` signature change, and resolver plumbing in both the text and graph paths. Too much machinery for one node's rule, so I kept the analyzer check.

sy-nico: Implementing a fix for it to work on variable-length channels would require plumbing that is out of scope for this PR. I just made the minimal edit towards stability while we discuss how this *should* behave.

sy-nico:
Stripping out bitwise support

sy-nico:
@greptile, rerun. Spoke offline and decided to revert support for bitwise ops

docs/site/src/pages/reference/control/arc/reference.mdx
```diff
@@ -17,7 +17,7 @@ imperative constructs, and standard library.
 | [Syntax](/reference/control/arc/reference/syntax)                     | Comments, identifiers, literals, imports, blocks            |
 | [Types](/reference/control/arc/reference/types)                       | Primitives, series, channels, timestamps                    |
 | [Variables](/reference/control/arc/reference/variables)               | Literal, stateful, and channel variables, assignment, scope |
-| [Operators](/reference/control/arc/reference/operators)               | Arithmetic, comparison, logical, and flow operators         |
+| [Operators](/reference/control/arc/reference/operators)               | Arithmetic, comparison, logical, flow operators             |
```
pjdotson: fix typo

integration/tests/arc/boolean.py
pjdotson, line 1: one potentially nice test you could add - testing `and` and `or` short-circuit evaluation. So have a function that takes in a boolean input and returns a boolean output but has a side-effect of writing to a channel, and then put it on the right side of `false and` and `true or` and test that it never gets written to.

arc/go/text/text_test.go
```diff
@@ -3522,6 +3522,34 @@ sensor -> stable.for{duration=1s} -> output`
 				},
 			)
 
+			It(
+				"Should reject a variable-length stable.for input",
```
pjdotson, line 3526: Can you make an issue to allow `stable.for` w/ variable-length inputs in the future.

arc/go/analyzer/types/infer_test.go
```diff
@@ -1017,3 +1037,26 @@ var _ = Describe("Type Inference", func() {
...
+
+var _ = Describe("Series operand inference", func() {
+	seriesResolver := []symbol.Symbol{
+		{Name: "s", Kind: symbol.KindVariable, Type: types.Series(types.I64())},
+		{Name: "x", Kind: symbol.KindVariable, Type: types.String()},
+	}
+
+	DescribeTable("should propagate series operands through operator inference",
+		func(bCtx SpecContext, expr string, expected types.Type) {
+			Expect(inferExprType(bCtx, seriesResolver, expr)).To(Equal(expected))
```
pjdotson, line 1049: in the future (not this PR) you might want to consider refactoring these so that series of different types have different kinds. Not completely sure if that makes sense or not and does not block this PR.

arc/go/analyzer/statement/statement_test.go
```diff
@@ -626,6 +626,36 @@ var _ = Describe("Statement", func() {
 				(*ctx.Diagnostics)[0].Message,
 			).To(ContainSubstring("undefined symbol: x"))
 		})
+
+		DescribeTable(
+			"numeric conditions are accepted without warnings",
```
pjdotson: I thought we were accepting these with warnings?

arc/go/analyzer/expression/expression.go
```diff
@@ -685,11 +698,13 @@ func isValidCast(source, target basetypes.Type) bool {
...
-	if target.Kind == basetypes.KindString && source.IsNumeric() {
+	if target.Kind == basetypes.KindString &&
+		(source.IsNumeric() || source.Kind == basetypes.KindBool) {
 		return true
 	}
 	if source.Kind == basetypes.KindString || target.Kind == basetypes.KindString {
 		return false
 	}
-	return source.IsNumeric() && target.IsNumeric()
+	return (source.IsNumeric() || source.Kind == basetypes.KindBool) &&
```
pjdotson, line 716: should we add a warning for `target.IsNumeric()` and `source.Kind == basetypes.KindBool`? Feel like this is also something we do for backwards compatibility but wouldn't if we were building the language from ground-up

arc/go/analyzer/resolve.go
```diff
@@ -110,5 +110,29 @@ func ResolveNodeTypes(
 			missingRequiredInput = true
 		}
 	}
-	return !missingRequiredInput
+	return checkStableInputs(nodes, diag) && !missingRequiredInput
+}
+
+// checkStableInputs rejects stable.for nodes whose resolved input is not a
+// fixed-size type: the runtimes compare and copy raw fixed-density samples.
+func checkStableInputs(nodes ir.Nodes, diag *diagnostics.Diagnostics) bool {
```
pjdotson, line 118: why is this the job of the analyzer? shouldn't it live in `arc/go/stl/stable`?

arc/go/compiler/compiler_test.go
```diff
@@ -1748,6 +1756,45 @@ var _ = Describe("Compiler", func() {
...
+				Expect(power).ToNot(BeNil())
+
+				results := MustSucceed(power.Call(ctx))
+				Expect(results).To(ConsistOf(expected))
+			},
+			Entry("negative base", "i64", "-2 ^ 2", uint64(4)),
+			Entry(
+				"negative base odd exponent",
+				"i64",
+				"-2 ^ 3",
```
pjdotson: this is a bad test, (-2)^3 == -(2^3) == -8

arc/go/compiler/compiler_test.go
```diff
@@ -1748,6 +1756,45 @@ var _ = Describe("Compiler", func() {
 			},
 		)
 
+		DescribeTable(
+			"unary minus binds to the base: -2 ^ 2 is (-2) ^ 2",
```
pjdotson: note that in Python, JavaScript, PHP (which are the three most common languages w/ exponentiation), `-2 ** 2` would be -4 as exponentiation is higher precedance than unary not.

arc/go/compiler/expression/cast_test.go
```diff
@@ -375,6 +391,45 @@ var _ = Describe("Type Cast Compilation", func() {
...
+		Entry("integer literal", `bool(1)`),
+		Entry("float literal", `bool(1.5)`),
+		Entry("bool literal", `bool(true)`),
+		Entry("str literal", `bool("hello")`),
+		Entry("typed numeric", `bool(i32(5))`),
+	)
+
+	DescribeTable("should compile bool to numeric casts",
+		expectExpression,
+		Entry("bool to u8", `u8(true)`, types.U8(), OpI32Const, int32(1)),
```
pjdotson, line 414: you also might want to make it so that the only bool to numeric conversion supported is `bool` -> `u8`

pjdotson (changes requested):

arc/go/text/text_test.go
sy-nico: [SY-4650](https://linear.app/synnax/issue/SY-4650/enable-variable-lenth-inputs-for-stablefor)

arc/go/analyzer/expression/expression.go
sy-nico: I don't think we should. At least, not until we implement all of the work to say for certain that it isn't needed.

arc/go/analyzer/resolve.go
sy-nico: I wished. Unfortunately, because it can take any non-variable data type, the check needs to live here unless we implement a bunch of new plumbing to support it (which opens a door that is out of scope for this release).

pjdotson: Copy, so it should stay like this for this release and when we add `stable.for` support for variable-length data types it gets deleted?

arc/go/compiler/expression/cast_test.go
sy-nico: I'd rather not implement this here. Restricting to `u8` makes bool the only type in Arc with one legal cast target, and it fights the literal defaults. A bare integer resolves to `i64`, so `count += u8(flag)` needs an extra widening cast.

integration/tests/arc/boolean.py
```diff
@@ -0,0 +1,440 @@
...
+    return v
+}
+
+func mark_called(v bool) bool {
+    bool_sc_called = true
+    return v
+}
+
+func do_short_circuit(a bool) {
+    bool_sc_and_out = false and mark_skipped(a)
```
pjdotson: should have `false or mark_called` as well

arc/docs/spec.md
```diff
@@ -28,17 +28,18 @@ IntegerType ::= 'i8' | 'i16' | 'i32' | 'i64' | 'u8' | 'u16' | 'u32' | 'u64'
...
 
-**Type defaults**: Integer literals default to `i64`, float literals to `f64`.
+**NumericType defaults**: Integer literals default to `i64`, float literals to `f64`.
 
 ### Boolean Semantics
 
-Type `u8` serves as boolean: `0` is false, non-zero is true. Logical operators normalize
-to `0` or `1` with short-circuit evaluation.
+Comparisons and logical operators return `bool`. Logical operators (`and`/`or`/`not`)
+take `bool` only. No implicit `bool`/numeric conversion; `if` and `=>` also accept
```
pjdotson, line 36: do we have warning text for both `if` and `=>` if we have to do an implicit `bool` cast?

pjdotson (changes requested):

arc/go/compiler/expression/cast_test.go
pjdotson: add warning for boolean-> integer type case

arc/docs/spec.md
pjdotson: make sure warning for both of them

arc/go/analyzer/types/infer_test.go
sy-nico: https://linear.app/synnax/issue/SY-4652/refactor-series-behavior

arc/go/analyzer/resolve.go
sy-nico: Yes. I included this in the ticket to make sure it gets removed.

PR #2688, opened 2026-08-10
E. SY-4381: Redesign the panel and tab chrome

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4381](https://linear.app/synnax/issue/SY-4381/redesign-pluto-and-console-ui)

## Description

Stacked on the menus PR and final PR of the stack. After this merges, the tree matches the original #2649 branch exactly, plus the two `client.spec.ts` awaits from the first PR.

- The `--pluto-tabs-*` contract and panel mosaic land: mosaic tabs shrink-wrap their content, the selected tab floors at its content width, strips scroll with a vertical mouse wheel, and tab contents start-align.
- The panel selector tab machinery (`platform/selector/tab.ts`) arrives, wiring `createEmptyTab`, `useOpenTab`, and the overlay trigger into the app triggers and mosaic.
- Schematic symbol groups move onto tabs and the schematic toolbar adopts content tab sizing.
- Panel name editing gets a single canonical rename target, and session panel triggers land.
- The last bridges from the design-system PR are removed: the old `platform/tree/useConfirmDelete` is deleted, the `tree/external` re-export is dropped, and the three skipped `nav/bar/Left` tests are restored.

## Basic Readiness

- [ ] I have performed a self-review of my code.
- [ ] I have added relevant, automated tests to cover the changes.
- [ ] I have updated documentation to reflect the changes.

<h3>Greptile Summary</h3>

The PR redesigns Console panel and tab chrome while adding focused-tab overlay behavior, selector tabs, keyboard-trigger scoping, and project panel seeding.

- Extends Pluto tabs with configurable sizing, alignment, overflow, wheel scrolling, and revised styling.
- Adds panel overlay rendering, context-menu composition, empty states, retry handling, and trigger scoping.
- Seeds new projects with a panel containing the component selector and migrates delete confirmations to the modal API.

<h3>Confidence Score: 4/5</h3>

The project-seeding failure path should be fixed before merging so a successfully created project cannot be left without its required initial panel.

Panel creation is launched through a fire-and-forget mutation after the project has already been saved, so a realistic request failure leaves the new project only partially initialized; the listener cleanup and writing-guide issues are non-blocking.

**Files Needing Attention:** console/src/platform/project/useCreateModal.tsx, pluto/src/tabs/Selector.tsx, pluto/src/panel/Mosaic.tsx

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
[...]

PR #2670, opened 2026-08-06
SY-4494: Unify the Flux read surface on suspense

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4494](https://linear.app/synnax/issue/SY-4494)

## Description

Flux had four ways to read the same record: a `Result`-shaped `useRetrieve`, stateful
and observable variants, ad-hoc selectors, and an effect-driven `createForm`. Each one
rendered at least one frame of the wrong answer before the real one arrived, and each
carried its own error and loading plumbing that callers had to re-implement.

This collapses them onto two idioms. Render-path reads suspend and return data
directly, so there is no placeholder frame and failures land at the nearest error
boundary. Event and callback reads go to the domain client, which already owns the
cache. Selectors are minted from the same query definition through `createSelector`
instead of being hand-written per domain, and query definitions are split into
worker-safe modules so Aether can bind the same definitions React does.

`createForm` was the last effect-based read. It now suspends on `retrieve` and takes a
`getCached` projection that resolves warm reads synchronously. `query` became nullable,
which collapsed seven per-domain `FormQuery` types and let three write-only forms drop a
dead `retrieve`.

RFC 0055 records the read-unification model this implements.

## Basic Readiness

- [ ] I have performed a self-review of my code.
- [ ] I have added relevant, automated tests to cover the changes.
- [ ] I have updated documentation to reflect the changes.

<h3>Greptile Summary</h3>

This PR consolidates Flux reads around suspense-backed domain queries and adds equivalent cache-backed reads for Aether.

- Replaces result, observable, and ad hoc selector APIs with direct suspense reads and generated selectors.
- Changes forms to suspend during initial retrieval and support nullable queries for create mode.
- Adds referentially stable maintained query snapshots for React external-store consumers.
- Moves worker-safe query definitions into Aether-compatible modules and migrates Console consumers.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge because no concrete blocking or non-blocking defect remained after checking the new retrieval, cache, form, and Aether lifecycles.

Maintained query snapshots are stable, query and client changes clean up prior subscriptions, stale asynchronous reads are discarded, forms reset from the correct query values, and cluster invalidations retain a refetch path.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| pluto/src/flux/retrieve.ts | Introduces the unified suspense, result, selector, cache-subscription, and retrieval lifecycle without an accepted defect. |
[...]

PR #2689, opened 2026-08-10
F. SY-4494: Unify the Flux read surface on suspense

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4494](https://linear.app/synnax/issue/SY-4494)

## Description

Replaces #2670, re-pointed onto the SY-4381 five-PR stack (#2688) after the giant redesign PR was carved up; GitHub does not allow base changes on stacked PRs. Review comments live on #2670.

Flux had four ways to read the same record: a `Result`-shaped `useRetrieve`, stateful
and observable variants, ad-hoc selectors, and an effect-driven `createForm`. Each one
rendered at least one frame of the wrong answer before the real one arrived, and each
carried its own error and loading plumbing that callers had to re-implement.

This collapses them onto two idioms. Render-path reads suspend and return data
directly, so there is no placeholder frame and failures land at the nearest error
boundary. Event and callback reads go to the domain client, which already owns the
cache. Selectors are minted from the same query definition through `createSelector`
instead of being hand-written per domain, and query definitions are split into
worker-safe modules so Aether can bind the same definitions React does.

`createForm` was the last effect-based read. It now suspends on `retrieve` and takes a
`getCached` projection that resolves warm reads synchronously. `query` became nullable,
which collapsed seven per-domain `FormQuery` types and let three write-only forms drop a
dead `retrieve`.

RFC 0055 records the read-unification model this implements.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR consolidates Flux reads around suspense-backed domain queries and adds equivalent cache-backed reads for Aether.

- Replaces result, observable, and ad hoc selector APIs with direct suspense reads and generated selectors.
- Changes forms to suspend during initial retrieval and support nullable queries for create mode.
- Adds referentially stable maintained query snapshots for React external-store consumers.
- Moves worker-safe query definitions into Aether-compatible modules and migrates Console consumers.

The PR appears safe to merge because no concrete blocking or non-blocking defect remained after checking the new retrieval, cache, form, and Aether lifecycles.

Maintained query snapshots are stable, query and client changes clean up prior subscriptions, stale asynchronous reads are discarded, forms reset from the correct query values, and cluster invalidations retain a refetch path.

| Filename | Overview |
|----------|----------|
| pluto/src/flux/retrieve.ts | Introduces the unified suspense, result, selector, cache-subscription, and retrieval lifecycle without an accepted defect. |
| pluto/src/flux/form.ts | Changes form initialization to cache-first suspense reads and correctly resets values when nullable queries change. |
| pluto/src/flux/aether/retrieve.ts | Adds a cache-seeded Aether observer with query deduplication, stale-fetch suppression, and lifecycle cleanup. |
| client/ts/src/query/query.ts | Interns composed answers while maintained so React external-store snapshots remain referentially stable. |
| client/ts/src/channel/client.ts | Prevents range-scoped alias requests from being approximated from records that do not contain alias data. |
| pluto/src/schematic/queries.ts | Migrates schematic reads and selectors onto the unified Flux query definition. |
| pluto/src/lineplot/range/aether/provider.ts | Migrates Aether range retrieval to the shared observer and closes it during component deletion. |
| console/src/platform/device/wrapTaskForm.tsx | Replaces the removed device provider pattern with a wrapper that reads the selected device through the unified query surface. |
| docs/tech/rfc/0055-260805-flux-read-unification.md | Documents the cache-owned, suspense-first read model introduced by this change. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  C[Console or Pluto component] --> H[Flux suspense hook]
  A[Aether component] --> O[Flux Aether observer]
  H --> Q[Domain query definition]
  O --> Q
  Q --> D[TypeScript domain client]
  D --> K[Query and record cache]
  K --> H
  K --> O
  D --> R[Core retrieval API]
```

<sub>Reviews (1): Last reviewed commit: ["Merge branch &#39;sy-4381-redesign-pluto-and..."](https://github.com/synnaxlabs/synnax/commit/0fad8b3ecbbaf19e5c13b425fccf4931544be7e3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50964172)</sub>

</details>


client/ts/src/query/query.ts
```diff
@@ -8,38 +8,52 @@
...
   TimeSpan,
+  TimeStamp,
 } from "@synnaxlabs/x";
 
 import { AccessDeniedError, NotFoundError } from "@/errors";
 import { Deleted } from "@/query/deleted";
 import { type Table, type TableEvent } from "@/query/table";
 import { type Data, type FetchOptions, type Params } from "@/query/types";
 
+const hashes = new WeakMap<object, string>();
```
emilbon99, line 26: We don't use module level mutable globals. fix this.

emilbon99: Also audit for other module level globals.

client/ts/src/query/query.ts
```diff
@@ -225,19 +266,24 @@ export class Queries<
   K extends record.Key = record.Key,
   V extends state.State = state.State,
 > {
-  private readonly entries = new Map<string, Entry<Q, K, D>>();
+  private readonly entries = new Map<string, Entry<Q, K, D, V>>();
   /** Composed corpses interned per tombstone and query, so deleted answers
    *  stay referentially stable. Identity composes reuse the tombstone. */
   private readonly corpses = new WeakMap<Deleted<V>, Map<string, Deleted<D>>>();
+  /** Composed entry-less answers interned per row and query. */
+  private readonly rowAnswers = new WeakMap<object, Map<string, D>>();
```
emilbon99: Fix this name

client/ts/src/query/query.ts
```diff
@@ -601,49 +745,71 @@ export class Queries<
...
-    const { table, matches } = this.params;
-    const keys = entry.state.keys;
-    const member = keys.includes(key);
-    const rec = table.get(key);
-    if (rec == null) {
-      if (!member) return;
-      return this.evict(entry, keys, key);
+    if (entry.state.variant !== "ready") return false;
+    const { table, matches, single } = this.params;
+    const working = [...entry.state.keys];
```
emilbon99: See if we can simplify the control flow in this function. It's hard to follow.

client/ts/src/query/query.ts
```diff
@@ -601,49 +745,71 @@ export class Queries<
...
-  private recheckMany(entry: Entry<Q, K, D>, keys: K[]): void {
+  private recheckMany(entry: Entry<Q, K, D, V>, keys: K[]): void {
     const { table } = this.params;
-    keys.forEach((key) => this.recheck(entry, key));
+    this.recheckKeys(entry, keys);
     const missing = keys.filter((key) => table.status(key) === "unknown");
     if (missing.length === 0) return;
     // Backfill through the table's fetch so membership can be rechecked;
     // fetch-less tables serve cached rows only and the recheck is a no-op.
     table
```
emilbon99: We should almost ALWAYS be preferring async/await. Let's see if we can reduce the number of `.then().catch()` calls.

client/ts/src/query/retriever.ts
```diff
@@ -260,13 +285,10 @@ export abstract class Retriever<
...
     handler: ChangeHandler<D> | ChangeHandler<D[]>,
   ): destructor.Destructor {
-    const single = this.trySingle(params);
-    if (single != null)
-      return this.singleSpace.onChange(single.query, handler as ChangeHandler<D>);
-    return this.requestSpace.onChange(
-      this.normalizeRequest(params),
-      handler as ChangeHandler<D[]>,
-    );
+    const route = this.route(params);
```
emilbon99: See if we can make this a bit simpler. instead of introducing new types and having the same switch see if there is better structural elimination

client/ts/src/query/table.ts
```diff
@@ -340,19 +356,23 @@ export class Table<
...
+    );
 
     return () =>
-      toDelete.forEach(({ key: k, value }) => {
-        if (value == null) return;
-        this.tombstones.delete(k);
-        this.rows.set(k, value);
-        this.notify({ variant: "set", key: k, value });
-      });
+      this.batch(() =>
```
emilbon99, line 371: How does this work? Does `notify` still do independent notifications that reduce efficiency?

console/src/feature/arc/toolbar/Toolbar.tsx
```diff
@@ -80,31 +80,33 @@ const Content = () => {
...
+            <List.Items<arc.Key, arc.Arc>
+              full="y"
+              emptyContent={<EmptyContent onCreate={create} />}
+              onContextMenu={menuProps.open}
+            >
+              {({ key, ...p }) => (
+                <ArcListItem
+                  key={key}
+                  {...p}
+                  onRename={(name) => handleRename({ key, name })}
```
emilbon99: We should probably shift the compose signatures for these callbacks down into the list item, that we can just pass handleEdit and handleRename directly so we aren't messing with memoization.

console/src/feature/ethercat/device/queries.spec.ts
emilbon99, line 1: Test quality in this file is still good even though so much was removed?

console/src/feature/labjack/task/Read.tsx
```diff
@@ -12,6 +12,7 @@ import { Flex, Form as PForm, Icon } from "@synnaxlabs/pluto";
 import { deep, errors, id, primitive } from "@synnaxlabs/x";
 import { type FC, useCallback } from "react";
 
+import { use } from "@/feature/labjack/device/queries";
 import { Select } from "@/feature/labjack/device/Select";
 import { SelectPort } from "@/feature/labjack/device/SelectPort";
 import * as Device from "@/feature/labjack/device/types";
```
emilbon99, line 18: Should we really be using`import *`? Kinda violates our policies.

console/src/feature/schematic/symbol/MissingForm.tsx
```diff
@@ -28,7 +28,7 @@ import { CSS } from "@/platform/css";
 export const MissingForm = (): ReactElement => {
   const form = Form.useContext();
   const openEditModal = Edit.useModal();
-  const symbolGroup = Schematic.Symbol.useRetrieveGroup({ query: {} });
+  const { data: symbolGroup } = Schematic.Symbol.useResultGroup({});
```
emilbon99: This one should be a suspended query I think.

console/src/feature/schematic/toolbar/Symbols.tsx
```diff
@@ -574,14 +574,14 @@ export const Symbols = (): ReactElement => {
   const isRemoteGroup = group.keyZ.safeParse(groupKey).success;
 
   const [searchTerm, setSearchTerm] = useState("");
-  const symbolGroup = Schematic.Symbol.useRetrieveGroup({ query: {} });
+  const { data: symbolGroup } = Schematic.Symbol.useResultGroup({});
```
emilbon99, line 573: This should also suspend I think.

console/src/feature/task/Toolbar.tsx
```diff
@@ -197,32 +197,34 @@ const Content = () => {
...
+            <List.Items<task.Key, task.Task>
+              full="y"
+              emptyContent={<EmptyContent />}
+              onContextMenu={menuProps.open}
+            >
+              {({ key, ...p }) => (
+                <TaskListItem
+                  key={key}
+                  {...p}
+                  onStopStart={(command) => handleListItemStopStart(command, key)}
```
emilbon99: Should see if we should push the signature refactors for these callbacks down so we can pass props directly instead of as new callbacks on every time.

console/src/feature/user/Toolbar.tsx
```diff
@@ -18,7 +18,7 @@ import { Tree } from "@/platform/tree";
 import { User as PlatformUser } from "@/platform/user";
 
 const Content = (): ReactElement => {
-  const { data: groupID } = User.useRetrieveGroupID({});
+  const { data: groupID } = User.useResultGroupID({});
```
emilbon99: This should suspend right?

console/src/platform/arc/useTask.ts
```diff
@@ -22,31 +22,25 @@ const notDeployedYet = (name: string) =>
   status.create({ name, variant: "disabled", message: "Not deployed yet" });
 
 export const useTask = (key: arc.Key, name: string): UseTaskReturn => {
-  const tsk = Arc.useRetrieveTask({ arcKey: key });
+  const { data: tsk } = Arc.useResultTask({ arcKey: key });
```
emilbon99, line 25: Should this suspend?

console/src/platform/panel/tab.tsx
emilbon99: Should destructure service.

console/src/platform/task/ChannelName.tsx
```diff
@@ -88,35 +61,86 @@ export const ChannelName = ({
...
-  formName: string | null,
-  defaultName: string,
-) => {
-  if (data != null) {
-    if (primitive.isNonZero(data.alias)) return data.alias;
-    return data.name;
-  }
-  if (primitive.isNonZero(formName)) return formName;
-  return defaultName;
+export const ChannelName = (props: ChannelNameProps): ReactElement => {
```
emilbon99, line 120: Double check that this is as clean as possible.

console/src/platform/tree/Tree.spec.tsx
```diff
@@ -392,4 +445,126 @@ describe("Tree.Tree", () => {
...
+            <PTree.Item {...rest} onDoubleClick={onDoubleClick}>
+              {resource.name}
+            </PTree.Item>
+          );
+        },
+      }),
+    });
+
+    // Lets the tree's own synchronizers process everything the fence observed.
+    const settle = async () =>
```
emilbon99: This seems like a hack.

console/src/platform/user/Badge.tsx
```diff
@@ -16,12 +16,12 @@ import { CSS } from "@/platform/css";
 import { Session } from "@/session";
 
 export const Badge = (): ReactElement | null => {
-  const { data: u } = User.useRetrieve({}, { addStatusOnFailure: false });
+  const { data: remoteUsername } = User.useResultUsername({});
```
emilbon99, line 19: Shouldn't we be suspending?

console/src/platform/view/Frame.tsx
```diff
@@ -77,7 +77,7 @@ export const Frame = ({ resourceType, icon, children }: FrameProps): ReactElemen
   const staticViewKeys = useMemo(() => staticViews.map((v) => v.key), [staticViews]);
   const [selected, setSelected] = useState(staticViews[0].key);
   const hasUpdatePermission = Access.useUpdateGranted(view.ontologyID(selected));
-  const [editable, setEditable] = useState(hasUpdatePermission);
+  const [editable, setEditable] = useState(false);
```
emilbon99, line 80: Double check on why we made this change.

pluto/src/access/queries.ts
```diff
@@ -70,62 +66,34 @@ export const isGranted = ({
...
-    return client.access.policies.onChange({ for: sub }, (result) => {
-      if (!isLive(result)) return;
-      const next = evaluate(result);
-      if (next === prev) return;
-      prev = next;
-      handler(next);
-    });
+    return client.access.granted.onChange(sub, { objects, action }, handler);
+  },
+  // The full query passes through: its stable identity is what keys the
```
emilbon99: Remove this useless comment.

pluto/src/code/Provider.tsx
```diff
@@ -105,8 +105,15 @@ interface ContextValue {
   getLanguage: (name: string) => Language | undefined;
 }
 
+// One rejected promise, not one per call: use() re-renders on an uncached promise, so a
+// fresh rejection each time livelocks the boundary instead of settling it.
+const NOT_MOUNTED: Promise<Monaco> = Promise.reject(
+  new Error("Code.Provider is not mounted"),
+);
+NOT_MOUNTED.catch(() => {});
```
emilbon99, line 115: Weird. Why are we doing this? Is this hacky?

pluto/src/flux/aether/retrieve.ts
```diff
@@ -0,0 +1,136 @@
...
+  onError?: (error: Error) => void;
+}
+
+/**
+ * Holds one query's answer live for an aether component: seeds synchronously from the
+ * domain cache, fetches on a miss, subscribes for changes, and pushes updates through
+ * `onChange`. Call {@link update} from `afterUpdate` and {@link close} from
+ * `afterDelete`. Repeated updates with an equal query and client are no-ops.
+ */
+export class Retrieve<Query extends query.Params, Data extends query.Data> {
```
emilbon99, line 48: Verify that we actually like this API shape.

pluto/src/lineplot/range/aether/provider.ts
```diff
@@ -58,61 +59,50 @@ interface ProviderProps {
   timeRange: TimeRange;
 }
 
+// Widens fetch boundaries to whole minutes so panning does not requery on every
+// viewport nudge.
+const QUANTUM = TimeSpan.minutes(1);
```
emilbon99: We need to figure out what is going on here.

pluto/src/lineplot/queries.ts
```diff
@@ -440,23 +336,20 @@ export const useSelectRule = Scope.bindHook((params: SelectRuleParams): DerivedR
...
-      requireLinePlot(client, key)
-        .rules.filter((r) => r.axis === axisKey)
-        .map((r) => r.key),
-    equal: (a, b) => a.length === b.length && a.every((v, i) => v === b[i]),
-  }),
+export const useAxisRuleKeys = Scope.bindHook(
+  createSelector<string[], AxisRulesParams>(
+    ({ rules }, { axisKey }) =>
+      rules.filter((r) => r.axis === axisKey).map((r) => r.key),
+    (a, b) => a.length === b.length && a.every((v, i) => v === b[i]),
```
emilbon99: Use standardized comparison primitives.

docs/tech/rfc/0055-260805-flux-read-unification.md
sy-nico, line 1: Three mismatches between the code and RFC:

1. §5.1 specifies `useCached`, an undefined-returning decoration read. It was never built. `useResult` and `createResultSelector` shipped in its place.

2. §5.6 puts "the `Result` read plumbing in `retrieve.ts`" on the kill list, and §5.0 states "two idioms, no middle". The Result read path survived with ~28 call sites plus five channel selectors (`useResultAlias`, `useResultName`, `useResultStatus`, `useResultDataType`, `useResultAliasAndName`). RD3 ("undefined-returning selectors, rejected") reads as false against the code.

3. Every hook name differs. The RFC says `useRetrieve` and `useEnsureRetrieved`; the code exports `use` and `useEnsure`.

I think the code is correct, so these just need updating.

pluto/src/flux/form.ts
```diff
@@ -155,34 +204,26 @@ export const createForm =
...
+    // Form state is built once, so a query pointing at a different record has
+    // to replace it.
+    const readQuery = useRef(memoQuery);
+    const valuesRef = useRef(values);
+    valuesRef.current = values;
+    useEffect(() => {
+      if (readQuery.current === memoQuery) return;
+      readQuery.current = memoQuery;
+      form.reset(valuesRef.current);
+    }, [memoQuery, form]);
```
sy-nico: Change it to `useLayoutEffect`.

`useEffect` runs after paint, so a query change paints one frame of the previous record's values before this reset lands.

client/ts/src/channel/client.ts
```diff
@@ -610,36 +610,6 @@ export class Client extends query.Retriever<
     return isSingle ? res[0] : res;
   }
 
-  /**
-   * Returns the cached answer to the given query without touching the
-   * network, or undefined when nothing is cached. Unfetched filter queries
-   * are approximated from the record store when possible.
-   */
```
sy-nico, line 617: Was removing the whole approximation intended, or just the alias-scoped case? RFC 0055 §5.1 only asks that a request the cache cannot fully serve return undefined. As deleted, every unfetched channel filter query now costs a round trip even when all matching records are already in the table.

pluto/src/flux/retrieve.ts
```diff
@@ -503,208 +312,475 @@ const awaitCreation = <Query extends query.Params, Data extends query.Data>(
...
+      successResult(`retrieved ${name}`, derived),
+    );
+  const settled = local.settled.get(query.hash(memoQuery));
+  if (settled != null)
+    return hold(["settled", settled], () =>
+      "data" in settled
+        ? successResult(`retrieved ${name}`, settled.data)
+        : errorResult(`retrieve ${name}`, settled.error),
+    );
+  ensureFetch(
```
sy-nico: This starts a fetch during render. `inFlight` dedupes it today, but a definition whose `getCached` serves at settle time writes no `settled` entry, so once the space tears down the next render fetches again with nothing to stop it. Move it into an effect.

Same thing for line 737

pluto/src/flux/aether/retrieve.ts
```diff
@@ -0,0 +1,136 @@
...
+        if (result === undefined && this.result !== undefined) return;
+        this.set(result);
+      });
+    const cached = getCached?.(params);
+    if (cached !== undefined) return this.set(cached);
+    void this.fetch(params);
+  }
+
+  /** Refetches the current query, bypassing the deduplication guard. */
+  refetch(): void {
```
sy-nico, line 100: Bump `generation` here.
`refetch` reuses the current generation, so two overlapping refetches both pass the staleness check and the one that resolves last wins regardless of which started last.

client/ts/src/query/query.ts
```diff
@@ -8,38 +8,50 @@
...
   TimeSpan,
+  TimeStamp,
 } from "@synnaxlabs/x";
 
 import { AccessDeniedError, NotFoundError } from "@/errors";
 import { Deleted } from "@/query/deleted";
 import { type Table, type TableEvent } from "@/query/table";
 import { type Data, type FetchOptions, type Params } from "@/query/types";
 
+const hashes = new WeakMap<object, string>();
```
sy-nico, line 26: This is a package-level mutable singleton, which the architecture rules ban. Correctness holds because `Params` is readonly, but every cache in the process shares one memo table with no owner. Move it onto the cache instance.

client/ts/src/query/query.ts
```diff
@@ -408,244 +517,295 @@ export class Queries<
...
-    if (entry.handlers.size === 0) return;
-    const result = this.cachedOf(entry);
-    entry.handlers.forEach((handler) => {
+  /** Notifies subscribers when any of the keys is a member of the ready
+   *  answer; otherwise just drops the memo so the next read recomposes. */
+  private touchMembers(query: Query<P, K, D, V>, keys: K[]): void {
+    query.composed = undefined;
+    if (query.state.variant !== "ready") return;
+    const { keys: members } = query.state;
+    if (keys.some((key) => members.includes(key))) this.touch(query);
```
sy-nico: `includes` inside `some` makes this O(events x members) on every batched write to a rule-3 space. Can you build a `Set` from one side before the scan?

pluto/src/channel/queries.ts
```diff
@@ -76,39 +88,64 @@ export const ZERO_FORM_VALUES: z.infer<
...
-export const { useRetrieve, useRetrieveStateful, useRetrieveObservable } =
+export const { use, useResult, useEnsure, createSelector, createResultSelector } =
   Flux.createRetrieve<RetrieveQuery, channel.Channel>({
-    name: RESOURCE_NAME,
-    retrieve: async ({ client, query: { key, rangeKey } }) =>
-      await client.channels.retrieve(key, { rangeKey }),
-    subscribe: ({ client, query }, handler) => client.channels.onChange(query, handler),
-    getCached: ({ client, query }) => client.channels.getCached(query),
+    ...retrieveDefinition,
+    onChange: retrieveDefinition.onChange,
```
sy-nico: `onChange` is already in the spread above, so this line is a no-op and can be dropped. Same for line 134.

pluto/src/ranger/queries.ts
```diff
@@ -182,31 +173,21 @@ export const useForm = Flux.createForm<FormQuery, typeof formSchema>({
...
+  Label.useResultLabelsOf(key == null ? null : { id: ranger.ontologyID(key) }).data;
 
-export type ListQuery = Omit<ranger.RetrieveRequest, "names">;
+export { type ListQuery } from "@/ranger/aether/queries";
 
 export const useList = Flux.createList<ListQuery, ranger.Key, ranger.Range>({
-  name: PLURAL_RESOURCE_NAME,
-  retrieve: async ({ client, query }) => await client.ranges.retrieve(query),
+  ...listDefinition,
+  onChange: listDefinition.onChange,
```
sy-nico: Same thing as the channels: `onChange` is already in the spread above and can be dropped.

client/ts/src/query/table.ts
```diff
@@ -479,15 +573,15 @@ export const createDeleteListener = <
       channel,
       schema,
       onChange: (changed) => {
-        table.delete(key(changed));
+        table.delete(changed.map(key));
```
sy-nico: `map` passes the index and array to `key` as extra arguments. Every current `key` takes one param so nothing breaks today, but `createSetListener` wraps its call for this reason. Match it: `changed.map((c) => key(c))`.

sy-nico (changes requested):

client/ts/src/query/query.ts
emilbon99: I already looked at this. I dislike it too, but unfortunately I think it's going to have to stay in this case. It's hard to move it into the `client` because it has a bunch of external callers that use it for things that are not directly cache related. The reason i'm ok with it is because it's comparable to a `sync.Pool` usage in go, where its very common to declare something like

```go 
var bufferPool = &sync.Pool{}

func shortLifetimeBufferUser() string {
    buf := bufferPool.Get()
} 
```

`hashes` serves the exact same purpose as this example that we are totally ok with in go: pool commonly used objects for a performance gain. 

The other thing that makes me ok with this is the fact that it is not public.

So, not perfect but I think it's the best solution.

client/ts/src/channel/client.ts
emilbon99: Yeah removing the whole approximation was intended. It was causing problems. The cache still has a quality fast path for the most common key and name based lookups.

client/ts/src/query/query.ts
sy-nico: Copy. Just leave a comment so that someone in the future (me) does not copy this pattern without understanding why it's fine in this case.

sy-nico (approved):

PR #2690, opened 2026-08-10
SY-4562: Remove density and normalize the TypeScript channel constructor

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4562](https://linear.app/synnax/issue/SY-4562)

## Description

The TypeScript channel constructor accepted a `density` argument that it never
read. No caller in the client, Pluto, or the Console supplied it. This PR
removes it, then cleans up the rest of the constructor argument type.

That type was `New` intersected with four overrides. All four are now gone:

- `internal` duplicated a field that `New` already carries.
- `density` was unused.
- `status` and `operations` existed because of a mismatch. `New` is the crude
  input side of the channel schema, but the constructor stored these two fields
  raw into class fields declared in the parsed form. The overrides hid the
  mismatch and made callers supply values that were already parsed.

The constructor now normalizes `operations` and `status`, which is what it
already did for `key`, `index`, and `dataType`. A crude operation gets its
`resetChannel` and `duration` defaults, a crude status time becomes a
`TimeStamp`, and a null label array becomes `undefined`. Only `frameClient`
remains as an override, because it is not part of the schema.

The audit also found a dropped value. The constructor stores `concurrency`,
but the `payload` getter did not include it, so `payloadZ.parse` applied the
`exclusive` default on each round trip. The client caches a channel as
`sugar(stripComposed(ch.payload))`, so a virtual channel created with shared
concurrency reported `exclusive` after it entered the cache. The cache table
also compares payloads for equality, so a concurrency change did not
propagate. The getter now carries the field.

## Behavior change

The constructor now throws on a malformed status, the same way it already
threw on a malformed key. Every status that reaches it today is already
parsed, so no current path is affected. `statusZ.parse` also drops unknown
keys that the previous spread kept.

## Verification

`pnpm check-types` passes across all 21 workspace packages, `pnpm lint:client`
is clean, the full client suite passes (1578 tests) against a live core, and
Pluto passes (3521 tests).

## Basic Readiness

<h3>Greptile Summary</h3>

The PR normalizes crude channel operation and status inputs in the TypeScript constructor and preserves concurrency through payload serialization.

- Removes the unused density constructor argument and redundant type overrides.
- Parses operations and statuses into their normalized forms.
- Includes concurrency in channel payloads and adds focused round-trip and normalization tests.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| client/ts/src/channel/client.ts | Normalizes constructor inputs and includes concurrency in serialized channel payloads. |
| client/ts/src/channel/channel.spec.ts | Adds focused tests for concurrency round trips, operation defaults, and status validation. |

<sub>Reviews (2): Last reviewed commit: ["SY-4562: Normalize operations and status..."](https://github.com/synnaxlabs/synnax/commit/24d580fb2a92bc73e129457d5385bc228b8773a1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51910870)</sub>

**Context used:**

### DIFF
--- a/client/ts/src/channel/channel.spec.ts
+++ b/client/ts/src/channel/channel.spec.ts
@@ -7,7 +7,7 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
-import { DataType, id, TimeRange, TimeStamp } from "@synnaxlabs/x";
+import { control, DataType, id, TimeRange, TimeStamp } from "@synnaxlabs/x";
 import { beforeAll, describe, expect, it, test, vi } from "vitest";
 
 import { Channel } from "@/channel/client";
@@ -365,6 +365,19 @@ describe("Channel", () => {
 
       const retrieved = await client.channels.retrieve(channel.key);
       expect(retrieved.name).toEqual(updated.name);
+    });
+  });
+
+  describe("payload", () => {
+    it("keeps shared concurrency through a round trip", () => {
+      const ch = new Channel({
+        name: "shared",
+        dataType: DataType.FLOAT32,
+        virtual: true,
+        concurrency: control.Concurrency.shared,
+      });
+      expect(ch.payload.concurrency).toEqual(control.Concurrency.shared);
+      expect(new Channel(ch.payload).concurrency).toEqual(control.Concurrency.shared);
     });
   });
 });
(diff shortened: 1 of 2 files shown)


pjdotson:
@greptile /rerun

emilbon99 (approved):

PR #2691, opened 2026-08-10
SY-4396: Make the dot the only Arc completion trigger

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4396](https://linear.app/synnax/issue/SY-4396)

## Description

Press enter after a comma in the Arc text editor and the keystroke did not make a new
line. It accepted whatever the suggestion menu had selected. Typing `func f(a f32,` and
then enter produced `func f(a f32,and{})`.

The Arc language server listed the comma as a completion trigger character, so the menu
opened the moment the comma landed with its first item selected. Monaco binds enter to
"accept the selected suggestion", so the next keystroke went to the menu.

**The dot is now the only completion trigger.** The comma, the colon, the brace, the
angle brackets and the minus all leave the list. `parser.LiteralEQ` leaves it too,
because it is `==`, and a trigger character must be one character, so the entry could
never fire.

Monaco selects the first item of every menu it raises automatically, and the editor
option that stops this (`editor.suggest.selectionMode`) applies to every trigger
character at once. The choice is all or nothing. A menu after a dot must keep its
selection, because tab and enter must still accept a member. Every other separator can
end a line, where the menu takes the enter key from the user, so those characters lose
the menu instead.

Completions stay available everywhere. Type a prefix and the quick suggestions open,
filtered as before. Ctrl+Space raises the full list at any position, for example
directly after `myTask{`. Monaco always selects an item for an explicit Ctrl+Space, so
tab and enter still accept there.

The result agrees with the language servers for other languages. The Go server gopls and
python-lsp-server also advertise the dot alone. No major server — clangd,
rust-analyzer, Pyright, typescript-language-server, Eclipse JDT LS, Roslyn, Intelephense
— makes the comma or the brace a completion trigger. Each of them makes the comma a
signature help trigger instead. Arc has no signature help provider yet. That is the
feature which belongs after a comma, and it is the follow-up to this change.

The server itself never reads the trigger character. Completion is driven by the cursor
position and the token stream, so the smaller set changes when the client asks, not what
comes back.

Three specs cover the change: one pins the advertised trigger set, one proves the server
still completes the remaining parameters on a wrapped line after a comma when the user
types a prefix, and the existing prefix specs prove the parameter and channel lists stay
reachable. The parameter, authority and stage specs also move their server construction
into a `BeforeEach`, which removes fourteen copies of the same block.

Verified with `ginkgo ./lsp/` and `golangci-lint fmt --diff ./lsp/...`. The editor
behavior after the new trigger set is not yet checked by hand in the Console.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR makes the dot the only automatic Arc completion trigger while retaining prefix-based and manually requested completions.
- Narrows the language server’s advertised completion trigger characters to `"."`.
- Adds coverage for the advertised trigger set and wrapped parameter completion.
- Consolidates repeated language-server setup in completion specifications.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| arc/go/lsp/server.go | Narrows automatic completion activation to member access through the dot trigger. |
| arc/go/lsp/completion_test.go | Verifies the new trigger set and preserves completion coverage after commas and across existing contexts. |

<sub>Reviews (2): Last reviewed commit: ["SY-4396: Fold the completion server setu..."](https://github.com/synnaxlabs/synnax/commit/35dd144cbdd6354a5aa57c7895bc105cb3dc38fa) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51922564)</sub>

**Context used:**


pjdotson:
@greptile /rerun

emilbon99:
Shifting this to Nico. @sy-nico Can you verify that other autocomplete scenarios stay nice please?  and review the PR

sy-nico (approved):
Tested manually. Much happiness.

PR #2693, opened 2026-08-11
SY-4531: Explicit versioned schema files replace Oracle snapshots

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4531](https://linear.app/synnax/issue/SY-4531)

## Description

Deletes `schemas/snapshots/` and the mint-time freezer. Schema version history now
lives in hand-owned version files at `schemas/<domain>/versions/<resource>/vN.oracle`:
full declarations for shapes that changed at N, alias lines to the defining version for
the rest. The versions directory is the sole version authority — `@go version` tags are
gone, and membership in the current version file marks a type persisted.

Each live schema is now a generated projection of its version chain, merged with the
live file's own annotations (outputs, language bindings, validation, wire-only types).
Shape edits happen in version files; a hand edit to version-owned live content fails
the `versions` gate, which names the version files as authority. RFC 0053 documents the
model.

Versioning model:

- Imports split on the persistence boundary: stored references pin a version file,
  resolved references import the dependency's live schema. The gate enforces placement
  and fails any stored pin that lags its dependency chain.
- A resource that stops being persisted ends its chain with a tombstone: an empty
  v(N+1).oracle. Earlier versions stay frozen, the live file returns to hand ownership,
  and nothing current may store a reference into an ended chain. Arc `Program` ends at
  v1 this way; its v0 keeps the frozen byte layout with a generated codec and a hand
  msgpack decoder for pre-Orc rows.
- Version files declare only the persisted surface. Runtime-only shapes (arc ir's
  `Scope`, `IR`, `Node`, ...) live in the live schema alone and evolve freely.
- Docs are version-owned: a redeclaration without `@doc` inherits the chain-resolved
  doc, in frozen packages and the live projection alike.

Generator changes:

- Codecs are explicit: a struct or union gets one iff its declaration carries
  `@go marshal`; `@go marshal hand` marks hand-written codec methods. Generation fails
  when a codec's persisted graph reaches an untagged struct or union.
- `oracle migrate <resource>` is a scaffolder that writes an all-alias v(N+1) file;
  `oracle snapshot` and the legacy migrate flow are deleted.
- Generated migrations leave non-bridgeable fields (union mismatches, cross-resource
  moves, non-castable primitive changes such as uint64 -> uuid) to the hand-written
  migration.
- Generated Go files alias-suffix a pinned import only when it collides with a live
  import of the same dependency.
- The schema formatter keeps one blank line between the license header and a following
  comment, so tombstone files stop fighting the copyright script.

Module cleanup, output-neutral (`oracle sync` is a no-op after it): dead helpers,
write-only fields, and the unused multi-language resolver formatters are gone;
duplicated helpers consolidated into `plugin/internal/arrays`,
`plugin/cpp/internal/includes`, and `plugin/cpp/naming`; the LSP completions, hover
docs, and VS Code grammar now describe the real `@`-sigil syntax; the `persistence`
gate gets its own exit code (13).

Current generated output is byte-identical at cutover; all `oracle check` gates pass
and `oracle sync` is at fixpoint.

Schema-content changes riding the same branch:

- The rack status task config moves to its own `rack_task` schema; task config bases
  stay in `task` without a `Base` prefix; snake_case schema namespaces camelize in
  generated TS. Config record types declare at their chains' current versions.
- `pagerduty.Alert.status` is typed `status.Key` instead of a bare string, and the
  Python status package now exports `Key`. Panel tab actions type their `key` as
  `TabKey`.
- A doc pass over every schema: corrected inaccurate docs (channel aggregation reset
  semantics, NI counter units and strain-gauge fields, symbol variants, snapshot
  fields, the LabJack thermocouple port) and normalized casing for acronyms, eponymous
  units, and proper nouns (Hertz, Ohms, ID, Unix, the Driver).
- Deleted the orphaned `x/go/telem/telem.pb.go`, left behind when its proto moved to
  `x/go/telem/pb/`.

## Basic Readiness


arc/cpp/types/types.gen.h
```diff
@@ -48,17 +48,17 @@ enum class Kind : std::uint8_t {
...
-    Variable = 16,
-    NumericConstant = 17,
-    IntegerConstant = 18,
-    FloatConstant = 19,
-    ExactIntegerFloatConstant = 20,
-    Function = 21,
-    Sequence = 22,
-    Stage = 23,
-    VarRef = 24,
+    Chan = 12,
```
emilbon99, line 51: You should verify that this won't cause problems with stored data.

emilbon99 (changes requested):
I'm not going to read this entire PR. I'm ok with the architecture.

Please be 1000000000000000000000000% sure that you are not breaking migrations, and then you are good to merge.

arc/cpp/types/types.gen.h
pjdotson: doesn't, the stored Go data generated numbers from `iota`

PR #2695, opened 2026-08-11
SY-4523: Add server-side export for projects

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4523](https://linear.app/synnax/issue/SY-4523/add-server-side-export-for-projects)

## Description

Moves project export from the Console to the Core. The Core walks the project's<br>ontology descendants and builds a bundle: one envelope per member document, one file<br>per panel with root-relative resource references, group children as directories, and a<br>`manifest.json` naming the project at the root. The bundle travels as a zip archive,<br>and the Console streams the response straight to the file the user picks.

The walk skips children with no registered exporter along with the panel tabs that<br>reference them, drops empty groups, and reserves `manifest.json` and `LAYOUT.json` at<br>the root. Two members of one directory that resolve to one file name keep distinct<br>names through a numeric suffix, assigned in a sorted walk so repeated exports produce<br>identical bundles. Panels sync `parent_of` edges to the tasks their tabs reference,<br>and those tasks export into the panel's directory. `x/go/encoding/zip` now accepts<br>nested entry paths and validates each name as a relative forward-slash path.

The Console gains an interim importer for the bundle format, so exports round-trip<br>until server-side import replaces it before release. Console downloads consolidate on<br>a streamed `useDownload` hook with typed extensions that derive the save filters, and<br>the CSV download moves to `platform/framer`. RFC 0052 is condensed and updated to<br>match the implementation.

## Basic Readiness

### Greptile Summary

The PR moves project export to Core and adds a streamed ZIP download and interim bundle importer in Console.

* Adds deterministic server-side project bundle generation over ontology descendants.
* Streams project archives through the TypeScript client and Console download flow.
* Adds bundle path handling, filename deduplication, panel reference resolution, and round-trip coverage.

### Confidence Score: 5/5

The PR appears safe to merge because no blocking failure remains within the scope of this follow-up review.

No blocking failure remains.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| core/pkg/service/project/bundle.go | Builds the deterministic project bundle by walking descendants, exporting supported resources, and organizing nested entries. |
| core/pkg/api/project/project.go | Exposes project export through the Core API as a streamed ZIP response. |
| client/ts/src/project/client.ts | Adds the TypeScript client method that downloads a project archive through the file transport. |
| console/src/feature/project/commands.tsx | Replaces client-side directory export with the streamed server-side project download. |
| console/src/feature/project/import.ts | Adds the explicitly interim importer that reconstructs groups, resources, panel references, and project selection from bundle members. |
| x/go/encoding/zip/zip.go | Extends ZIP entry handling to support validated nested relative paths. |

### Sequence Diagram

```mermaid
sequenceDiagram
  participant Console
  participant TSClient as TypeScript Client
  participant Core
  participant Ontology
  participant Exporters
  Console->>TSClient: export project
  TSClient->>Core: GET /project/export
  Core->>Ontology: walk project descendants
  loop Exportable member
    Core->>Exporters: build member envelope
    Exporters-->>Core: JSON document
  end
  Core-->>TSClient: streamed ZIP archive
  TSClient-->>Console: ReadableStream
  Console->>Console: save project.zip
```

Reviews (3): Last reviewed commit: ["SY-4523: Mark the client-side project im..."](<https://github.com/synnaxlabs/synnax/commit/53253a5016a7f74a7a3c79dc3aaac88ff0af2108>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=52414687>)


pjdotson:
@greptile /review

pjdotson:
@greptileai review

core/pkg/service/panel/bundle_test.go
```diff
@@ -0,0 +1,200 @@
...
+			),
+		}
+		root := rootOf(MustSucceed(panel.EncodeBundle(p, refs)))
+		Expect(root).To(HaveKeyWithValue("variant", "leaf"))
+		Expect(root["tabs"]).To(ConsistOf(
+			HaveKeyWithValue("resource", "kept.json"),
+		))
+	})
+
+	It("Should collapse a split whose both sides were emptied", func() {
```
emilbon99, line 118: Why does this logic belong inside of a bundle? Is it because we're intentionally purging certain views from the bundle? Like a docs view or something?

console/src/platform/import/external.ts
```diff
@@ -10,4 +10,5 @@
 export * from "@/platform/import/import";
 export * from "@/platform/import/ingestBatch";
 export * from "@/platform/import/ingester";
+export * from "@/platform/import/parsableFile";
```
emilbon99: parseable is the more common and in my opinion correct spelling of this word

console/src/platform/import/parsableFile.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+// License, use of this software will be governed by the Apache License, Version 2.0,
+// included in the file licenses/APL.txt.
+
+import { describe, expect, it } from "vitest";
+
+import { Import } from "@/platform/import";
+
+describe("isParsableFile", () => {
+  it("should accept plain .json files at any depth", () => {
+    expect(Import.isParsableFile("Main.json")).toBe(true);
```
emilbon99: I feel like it would be better to name this function `Import.canParseFile` instead of this weird `parsable` thingy

console/src/platform/export/use.ts
```diff
@@ -14,65 +14,56 @@ import {
...
+      }, `Failed to export ${name}`);
+    },
+    [client, handleError, download],
+  );
 };
 
+export interface ResourceParams {
+  id: ontology.ID;
+  /** The resource name. It names the exported file. */
+  name: string;
```
emilbon99: Why doesn't the name come from the internal export logic itself? Why does the caller need to supply this?

emilbon99 (changes requested):

console/src/platform/export/use.ts
pjdotson: updated

core/pkg/service/panel/bundle_test.go
pjdotson: yes, we are intentionally purging range views

emilbon99: Why are we intentionally purging range views?

emilbon99 (changes requested):
Approved once my reply to an open thread is addressed

core/pkg/service/panel/bundle_test.go
pjdotson: should a range be part of a project? If we start exporting ranges do we need to export labels and child ranges for that range as well?

Decided that the best first step was to just export w/out ranges and then come back later in with a better understanding of what that looks like.

PR #2696, opened 2026-08-12
SY-4570: Acquire aether worker resources on commit, not during render

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4570](https://linear.app/synnax/issue/SY-4570/aether-leaks-worker-components-from-discarded-react-renders)

## Description

Aether acquired real resources during React's render phase. `Aether.Provider` spawned its `Worker` from the `Store` constructor, and `useLifecycle` posted a create message to the worker mid-render.

React discards renders, and nothing reclaimed what a discarded one built. `useUniqueKey` mints a fresh key per render, so an orphan registers at a different path than the survivor and no same-path collision cleans it up. It stays in the worker tree, alive, receiving context propagation.

Measured on a 7-component tree, counting worker-side constructor and `afterDelete` calls:

| case | created | deleted |
| --- | --- | --- |
| plain mount | 7 | 0 |
| StrictMode | 14 | 7 (correct, 7 net alive) |
| under Suspense | 21 | 0 (14 permanently orphaned) |

A suspending tree renders three times: the render that suspends, the retry, and the render after the promise resolves. All three registered. For `synnax.Provider` that meant three live clients and three websockets where one was wanted, which is the source of the intermittent `retrieve.spec.tsx` failure in `Test - Pluto`: an orphan's unbounded reconnect loop logs `failed to open streamer` into whatever test later installs a `console.error` spy.

### What changed

`Store.register` becomes `Store.stage`. Staging validates and parses, then returns a handle the caller owns. The store keeps no reference and the worker is told nothing. A component staged by a discarded render is reclaimed with the caller's ref, so there is nothing to sweep.

`Handle.delete` becomes `attach` / `detach`, driven from `useLifecycle`'s layout effect. The handle outlives the effect so StrictMode's cleanup-then-setup remount, which does not re-render in between, re-attaches the same component.

Attaches buffer and flush on a microtask, stable-sorted by path depth. React runs layout effects children before parents, but the worker rejects a child whose parent it has never seen; an ancestor's path is always shorter, so depth order restores the invariant without tracking render order.

`Store.connect()` replaces the constructor's eager `ensureAttached`, and the Provider calls it from a layout effect. A tree with no aether components still connects, so a worker that fails to load still reports it.

Handle state now lives on the handle rather than in a `snapshots` map. That map existed only to bridge the unregister-then-register window, which no longer exists, so it and `Store.getSnapshot` are gone.

`test/provider.tsx` had the same defect: it built the mock pair and worker tree inside `useMemo`.

### Batching

Buffering attaches until the flush made batching free to add, so both wires now carry an array instead of one message. A shared `Batcher` buffers messages and posts them as one array on the next microtask. There is no `batch` variant: an array at the top level makes nesting unrepresentable and removes the receiver's is-this-a-batch branch.

Measured on a 211-component tree (1 root, 10 composites, 200 leaves), real `postMessage` over a `MessageChannel`:

| | one call per message | one call per commit |
| --- | --- | --- |
| main-thread post | 0.556 ms | 0.352 ms |

The `postMessage` saving is the fixed per-call cost, not payload size. The larger win is on the receiving side: every delivery is one task, so a plot pushing state back from twenty lines used to cost twenty React renders and now costs one.

Each side catches each message in a batch on its own, so one failure reports and the rest still apply.

### Per-update allocations

`Node` rebuilt its `toString()`, three span labels, and its whole `Context` object (five closures) on every `afterUpdate`. All are constant for a node's lifetime, and noop instrumentation discards the labels. Caching them, measured three runs each on the same tree:

| | before | after |
| --- | --- | --- |
| propagate one provider change (210 `afterUpdate`s) | 0.062 ms | 0.057 ms |
| 200 leaf state updates | 0.093 ms | 0.078 ms |

For context on where the rest goes: with a realistically sized schema, `zod.parse` is 71% of the worker's per-update cost, and the main thread parses the same states again before sending them. Collapsing that double parse is the biggest number left on the table and is not attempted here.

### One fix found along the way

`flush` drained pending creates and notified their subscribers. A subscriber that attached another component landed in the queue after the flush had already committed to its batch, and nothing scheduled the next one, so that create sat until an unrelated send. The flush now re-schedules when the queue is non-empty.

### Verification

A new regression test mounts three leaves behind a Suspense boundary and asserts three worker components. With render-phase acquisition restored it reports nine, so the test pins the actual defect. Two existing pins that described the old behavior are rewritten against the new invariants rather than dropped.

Two more tests pin the batch: one commit of five components arrives as a single message, and every parent precedes its children inside it.

Full pluto suite (3485 tests) and full console suite (2016 tests) pass against a live core. `retrieve.spec.tsx` ran clean 8 times.

Nothing outside `pluto/src/aether/` changes, so this does not conflict with the SY-4381 / SY-4299 stack.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR defers Aether worker and component acquisition until React commits, preventing resources from leaking out of discarded renders.
- Replaces render-phase registration with staged, attachable handles.
- Buffers and depth-orders component creation, state transfers, and method invocations.
- Moves worker startup into the Provider layout lifecycle.
- Updates the mock Provider and adds Suspense, StrictMode, and deferred-connection coverage.

The PR should not merge until detached handles discard queued method calls so obsolete operations cannot execute against a newly attached worker component.

The new reusable-handle lifecycle aborts pending asynchronous callers during detach but retains their outbound messages, allowing stale side effects to run after StrictMode re-attachment; the test helper also loses isolation when one returned provider is mounted concurrently.

**Files Needing Attention:** pluto/src/aether/store.ts, pluto/src/aether/test/provider.tsx

| Filename | Overview |
|----------|----------|
| pluto/src/aether/store.ts | Introduces staged handles and commit-time flushing, but detached handles retain queued method calls that execute after re-attachment. |
| pluto/src/aether/main.tsx | Moves worker connection and component attachment into layout effects while deliberately retaining handles across StrictMode remounts. |
| pluto/src/aether/test/provider.tsx | Avoids render-time mock worker creation, but makes all mounts of one returned provider share the same communications pair. |
| pluto/src/aether/main.spec.tsx | Adds focused regression coverage for Suspense-discarded renders, deferred worker startup, and handle state across detach and re-attach. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant React
  participant Handle
  participant Store
  participant Worker
  React->>Handle: stage during render
  React->>Handle: attach during layout effect
  Handle->>Store: queue component
  Store-->>Store: microtask depth-sort
  Store->>Worker: create component
  React->>Handle: detach during cleanup
  Handle->>Store: abort lifecycle requests
  Note over Store,Handle: pending invokes must also be discarded
  React->>Handle: re-attach same handle
  Store->>Worker: create fresh component
```

<sub>Reviews (1): Last reviewed commit: ["SY-4570: Acquire aether worker resources..."](https://github.com/synnaxlabs/synnax/commit/27c9ea5fba6626949d8f34b9568ba7b726d99250) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52430841)</sub>

**Context used:**

### DIFF
--- a/pluto/src/aether/aether/aether.ts
+++ b/pluto/src/aether/aether/aether.ts
@@ -861,11 +861,7 @@ export class Root extends Composite<typeof aetherRootState> {
     component._invokeMethod(params);
   }
 
-  /** Instantiates the registered component `type` at `path` under `parent`, wired to this
-   * tree's sender and instrumentation. Throws {@link UnexpectedError} if `type` is not in
-   * the registry. Does not mount the result; pass it as the `create` callback of
-   * {@link Composite._updateState}, which does. */
-  create({
+  private create({
     path,
     type,
     parent,
--- a/pluto/src/aether/aether/aether.ts
+++ b/pluto/src/aether/aether/aether.ts
@@ -771,13 +771,15 @@ export class Root extends Composite<typeof aetherRootState> {
       create: shouldNotCallCreate,
     });
     // Messages are handled sequentially: concurrent component updates lead to races, so
-    // the tree is implicitly serialized via the message queue.
-    root.comms.handle((msg) => {
-      try {
-        root.handle(msg);
-      } catch (e) {
-        root.comms.send({ variant: "error", error: errors.encode(e) });
-      }
+    // the tree is implicitly serialized via the message queue. Each message is caught
+    // on its own so one failure reports and the rest of the batch still applies.
+    root.comms.handle((messages) => {
+      for (const msg of messages)
+        try {
+          root.handle(msg);
+        } catch (e) {
+          root.comms.send({ variant: "error", error: errors.encode(e) });
+        }
     });
     return root;
   }
--- a/pluto/src/aether/aether/aether.ts
+++ b/pluto/src/aether/aether/aether.ts
@@ -13,15 +13,18 @@ import { deep, errors, type record, shallow, state, zod } from "@synnaxlabs/x";
 import { z } from "zod";
 
 import {
+  Batcher,
   type MainInvokeRequest,
   type MainMessage,
   type MainUpdateRequest,
   type Sender,
   type WorkerComms,
+  type WorkerMessage,
   wrapWorkerScope,
 } from "@/aether/aether/message";
 
 export {
+  Batcher,
   createMockPair,
   type MainComms,
   type MainMessage,
@@ -196,6 +199,16 @@ export abstract class Node implements Component {
    * their resolved provider (or `null` if unresolved). Non-`null` only while
    * {@link afterUpdate} is running, so reads outside that window are not tracked. */
   private ctxReads: Map<string, Node | null> | null = null;
+  /** {@link toString} and the span labels derived from it, built once. `type` and `path`
+   * never change, and every update would otherwise rebuild these strings only for noop
+   * instrumentation to discard them. */
+  private readonly label: string;
+  private readonly afterUpdateSpan: string;
+  protected readonly updateStateSpan: string;
+  private readonly deleteSpan: string;
+  /** Built on first use and reused. The object closes over `this` alone, so a fresh one
+   * per {@link afterUpdate} would be six identical allocations per update. */
+  private cachedCtx: Context | null = null;
 
   constructor({
     path,
@@ -206,8 +219,12 @@ export abstract class Node implements Component {
   }: ComponentConstructorProps) {
     this.type = type;
     this.path = path;
+    this.label = `${type}(${path[path.length - 1]})`;
+    this.afterUpdateSpan = `${this.label}:afterUpdate`;
+    this.updateStateSpan = `${this.label}:updateState`;
+    this.deleteSpan = `${this.label}:delete`;
     this.sender = sender;
-    this.instrumentation = instrumentation.child(this.toString());
+    this.instrumentation = instrumentation.child(this.label);
     this._parent = parent;
     this._depth = this._parent == null ? 0 : this._parent._depth + 1;
     this.childCtxValues = new Map();
@@ -226,11 +243,11 @@ export abstract class Node implements Component {
   }
 
   toString(): string {
-    return `${this.type}(${this.key})`;
+    return this.label;
   }
 
   private get ctx(): Context {
-    return {
+    return (this.cachedCtx ??= {
       get: <P>(key: string): P => {
         const provider = this.readCtx(key);
         if (provider == null)
@@ -249,7 +266,7 @@ export abstract class Node implements Component {
         this.childCtxValues.set(key, value);
         if (trigger) this.childCtxChangedKeys.add(key);
       },
-    };
+    });
   }
 
   /** Resolves `key` to the nearest ancestor publishing it (self excluded) and records
@@ -269,7 +286,7 @@ export abstract class Node implements Component {
    * resulting subscriptions. Used for both state-driven updates and context-driven
    * re-runs. */
   protected runAfterUpdate(): void {
-    const endSpan = this.instrumentation.T.debug(`${this.toString()}:afterUpdate`);
+    const endSpan = this.instrumentation.T.debug(this.afterUpdateSpan);
     const prevReads = this.ctxReads;
     this.ctxReads = new Map();
     this.childCtxChangedKeys.clear();
@@ -348,7 +365,7 @@ export abstract class Node implements Component {
    * must not call this. */
   _delete(path: readonly string[]): void {
     try {
-      const endSpan = this.instrumentation.T.debug(`${this.toString()}:delete`);
+      const endSpan = this.instrumentation.T.debug(this.deleteSpan);
       this.validatePath(path);
       this._deleted = true;
       this.afterDelete(this.ctx);
@@ -517,7 +534,7 @@ export abstract class Leaf<
     if (this.deleted) return;
     try {
       this.initializeMethods();
-      const endSpan = this.instrumentation.T.debug(`${this.toString()}:updateState`);
+      const endSpan = this.instrumentation.T.debug(this.updateStateSpan);
       this.validatePath(path);
       const state_ = zod.parse(this._schema, state, { label: this.toString() });
       if (this._state != null)
@@ -748,10 +765,15 @@ export class Root extends Composite<typeof aetherRootState> {
     instrumentation = alamos.Instrumentation.NOOP,
     registry,
   }: RootProps) {
+    // Every component in the tree shares this one sender, so a message batch spans the
+    // whole tree rather than a single component.
+    const sender = new Batcher<WorkerMessage>((values, transfer) =>
+      worker.send(values, transfer),
+    );
     super({
       path: [Root.KEY],
       type: Root.TYPE,
-      sender: worker,
+      sender,
       instrumentation,
       parent: null,
     });
@@ -778,7 +800,7 @@ export class Root extends Composite<typeof aetherRootState> {
         try {
           root.handle(msg);
         } catch (e) {
-          root.comms.send({ variant: "error", error: errors.encode(e) });
+          root.sender.send({ variant: "error", error: errors.encode(e) });
         }
     });
     return root;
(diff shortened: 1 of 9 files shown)


pluto/src/aether/store.ts
```diff
@@ -350,49 +341,88 @@ export class Store {
...
+    else {
+      const idx = this.queued.indexOf(entry);
+      if (idx !== -1) this.queued.splice(idx, 1);
+    }
+    entry.phase = "staged";
+    entry.controller.abort(
+      new Error(displaced ? "Component re-registered" : "Component deleted"),
+    );
+    // A fresh controller so a StrictMode remount can re-attach this same handle.
+    entry.controller = new AbortController();
```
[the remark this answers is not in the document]
emilbon99: Correct, and it is reachable: StrictMode detaches and re-attaches inside a single commit, and the flush is a microtask, so an invoke issued during the mounting render is always still queued when the detach lands. The abort rejected the caller while the message went out anyway on the next flush.

Fixed in 762d22f. `transfer` is deliberately left alone in the same spot: those Transferables belong to state the next flush still owes the worker, whereas an invoke is an event whose caller has already been told it failed.

Pinned by `should drop invokes queued before a detach rather than replay them on re-attach`, which fails without the one-line clear.

(Reply written by Claude at Emiliano's direction.)

pluto/src/aether/test/provider.tsx
```diff
@@ -7,23 +7,22 @@
...
+  // Built once per provider rather than per render: constructing the worker tree is
+  // resource acquisition, and a render React discards would otherwise strand one.
+  const [workerSide, mainSide] = aether.createMockPair();
+  aether.render({ worker: workerSide, registry });
 
-    return <Provider worker={worker}>{children}</Provider>;
-  };
+  const TestProvider: FC<PropsWithChildren> = ({ children }) => (
+    <Provider worker={mainSide}>{children}</Provider>
+  );
```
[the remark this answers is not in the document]
emilbon99: Right. Moving the pair out of `useMemo` fixed the discarded-render leak but changed the scope from per-mount to per-`createProvider` call, which shares the worker tree across concurrent mounts.

Fixed in 762d22f by making the comms lazy instead. The pair and worker tree are built on first use, and the only thing that touches them is the Store's `connect()`, which now runs on commit. So `useMemo` allocates a closure per mount and nothing else, and a discarded render strands nothing.

(Reply written by Claude at Emiliano's direction.)

emilbon99:
@greptile re-review

pluto/src/aether/main.tsx
```diff
@@ -146,53 +149,41 @@ export const useLifecycle = <
...
-  const getSnapshot = useCallback(
-    () => ctx.store.getSnapshot<StateSchema>(path),
-    [ctx.store, path],
-  );
+  // The handle outlives the effect so StrictMode's cleanup-then-setup remount, which
+  // does not re-render, re-attaches this same component.
+  useLayoutEffect(() => {
+    handle.attach();
+    return () => handle.detach();
+  }, [handle]);
```
sy-nico, line 175: The handle is staged once, so this effect only ever tracks itself. If `ctx.path` or `ctx.store` changes mid-life (component moves under a different `Composite`, or a different `Provider`), it keeps the original path and store silently. The docs call this out for `aetherKey` but not for context. Is this intended?

pluto/src/aether/main.tsx
sy-nico, line 78: Can `dispose()` move into the `connect()` layout effect? Teardown living in a `useSyncExternalStore` subscribe cleanup means connect is layout and dispose is passive. It works because React runs layout destroys before passive destroys, but that ordering is doing a lot of load-bearing work for a pair that could just sit in one effect.

pluto/src/aether/store.ts
```diff
@@ -239,12 +297,13 @@ export class Store {
    * `Worker`, and aborts in-flight invokes. The store remains usable: a subsequent
    * send lazily re-attaches via a fresh `Worker`. Idempotent. */
   dispose(): void {
```
sy-nico, line 299: Should `dispose()` also reset `entries`? It clears the worker tree and kills the worker, but every entry stays in the map marked `live`. If the store reconnects, those entries still believe they exist on a worker that has never heard of them, so the next `setState` sends an update for an unknown path. Safe today only because children detach before the Provider disposes.

pluto/src/aether/aether/aether.ts
```diff
@@ -837,7 +861,11 @@ export class Root extends Composite<typeof aetherRootState> {
     component._invokeMethod(params);
   }
 
-  private create({
+  /** Instantiates the registered component `type` at `path` under `parent`, wired to this
+   * tree's sender and instrumentation. Throws {@link UnexpectedError} if `type` is not in
+   * the registry. Does not mount the result; pass it as the `create` callback of
+   * {@link Composite._updateState}, which does. */
+  create({
```
sy-nico: nit: this went public only so `test/driver.ts` could call it. Production has no caller. Can the driver reach it another way, or does it stay public with a note that it exists for the test driver?

pluto/src/aether/test/provider.tsx
```diff
@@ -12,16 +12,31 @@ import { type FC, type PropsWithChildren, useMemo } from "react";
...
 ): FC<PropsWithChildren> => {
   const TestProvider: FC<PropsWithChildren> = ({ children }) => {
-    const worker = useMemo(() => {
-      const [workerSide, mainSide] = aether.createMockPair();
-      aether.render({ worker: workerSide, registry });
-      return mainSide;
-    }, []);
-
+    // Per mount, so two concurrent mounts never share a worker tree.
+    const worker = useMemo(() => lazyComms(registry), []);
```
sy-nico: React can throw away a `useMemo` result and recompute it, so it does not guarantee one comms per mount. `Provider` uses a ref for exactly this reason. Should you use a ref here too?

pluto/src/aether/aether/message.ts
sy-nico: Comment is stale now. Discarded renders no longer leak components, so `clear` is not cleaning up after them anymore.

pluto/src/aether/aether/message.ts
```diff
@@ -96,37 +96,95 @@ export const NOOP_MAIN_COMMS: MainComms = { send: () => {}, handle: () => {} };
...
+    onmessage = (e: MessageEvent<MainMessage[]>) => handler(e.data);
   },
 });
 
+/** Buffers messages and hands them to `flushTo` as one array on the next microtask.
+ *
+ * Both directions batch. `postMessage` has a fixed per-call cost that dominates the
+ * payload at these sizes, and each delivery is one task on the receiving thread, so a
+ * batch is also one React render instead of one per message. */
+export class Batcher<M> {
```
sy-nico, line 120: Needs tests for `clear()` leaving `scheduled` true, and for `beforeFlush` sending into the batch it precedes. Only covered indirectly through the store right now.

sy-nico (changes requested):

pluto/src/aether/main.tsx
emilbon99: This has to stay here because of the way react orders hooks across parents and children. If we run this cleanup in an effect, children would cleanup against a disposed store, causing aunch of weird behaviors and errors.

pluto/src/aether/main.tsx
emilbon99: *Emiliano + Claude: drafted in our review session.*

Intended. `ctx.store` cannot change mid-life at all: `Provider` holds the store in a ref, so a new store means the Provider itself remounted and every child unmounted first. `ctx.path` can only change if someone hands `Composite` a changing `path` directly, since every normal path comes from a parent `useLifecycle`, which is frozen the same way.

Worth saying the old code was worse here, not better. The `[ctx.store, path]` cleanup deleted the handle on a path change, but the re-register was `handleRef.current ??=` in render, so a path change deleted the worker component and did not recreate it until some unrelated later render.

Added a line to the `useLifecycle` doc stating identity is fixed at mount.

pluto/src/aether/aether/message.ts
emilbon99: Updated comment.

sy-nico (approved):

PR #2700, opened 2026-08-13
SY-4578: Add a centered mode control to the table

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4578](https://linear.app/synnax/issue/SY-4578/add-a-centered-mode-control-to-the-table)

## Description

A table always sat in the top left corner of its tab, which looks unbalanced when the
table is much smaller than the tab. A new per-table toggle centers it instead, reachable
from the hover controls at the top right of the tab and from the right-click menu. The
setting lives in session state next to the indicator visibility setting, so it is saved
with the workspace.

Centering cannot be pure CSS. Value cells draw on the canvas from box coordinates that
are relative to the table origin, so a DOM-only offset would leave the values behind. A
single computed offset feeds both the frame's transform and the cell boxes, which keeps
the two render paths in agreement. The offset clamps to zero per axis, so an overflowing
table stays put on that axis and can still center on the other.

The offset holds for the duration of a pointer press, and releases on pointer up or
cancel. Row and column resizes dispatch on every frame of the drag, so a live offset
would move the table by half of each dragged pixel and pull the grabbed edge out from
under the cursor.

Unlike the indicators item, the context menu entry appears while editing too, since
centering applies in both modes.

`Row` now takes `x` and `y` instead of `position`, since the parent owns the cursor
origin, and the two hardcoded `4.5 * 6` indicator sizes became `INDICATOR_SIZE`.

## Basic Readiness

`pluto/src/table/Table.spec.tsx` mounts the table and pins the centering itself: both
axes when it fits, clamping to zero on an axis it overflows, recentering when the table
changes size, holding the offset while the pointer is down, and releasing it on pointer
cancel. The last two were each verified to fail when the behavior they cover is removed.
The spec stubs `ResizeObserver`, because the `@juggle` polyfill reads computed styles
that jsdom leaves empty and so never reports a size.

`pluto/src/table/ContextMenu.spec.tsx` covers the menu entry, and the session slice and
selectors are covered console-side.

Not covered: the console toolbar toggle. Mounting the tab content in a console spec
fails with `[Root.create] - Table not found in registry`, because `pluto/src/ether.ts`
exports `lineplot`, `value`, `diagram` and `eraser` but not `table`, so the console
cannot hand its test aether provider the table components. That gap is worth its own
issue, since it blocks console specs for the whole table renderer rather than just this
change.

<h3>Greptile Summary</h3>

The PR adds a persisted per-table centering preference and keeps DOM and canvas geometry aligned through a shared offset.
- Adds toolbar and context-menu controls for switching between centered and top-left alignment.
- Persists the centering preference in the table session slice.
- Holds the current offset during pointer interactions and releases it on pointer up or cancellation.
- Adds focused coverage for state handling, menu behavior, centering, overflow, resizing, and pointer cancellation.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| pluto/src/table/Table.tsx | Implements shared, clamped DOM/canvas centering and correctly releases held offsets on pointer up or cancellation. |
| pluto/src/table/Table.spec.tsx | Covers centering, per-axis overflow clamping, resize recentering, offset holding, and pointer-cancellation release. |
| console/src/session/table/slice.ts | Adds the persisted centered preference with default, explicit-set, and toggle behavior. |
| console/src/feature/table/Table.tsx | Connects the session preference to the Pluto table and exposes the toolbar toggle. |
| pluto/src/table/ContextMenu.tsx | Adds an optional centering menu action available in both editing and viewing modes. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Control[Toolbar or context-menu toggle] --> Session[Persisted table session state]
  Session --> Table[Pluto Table centered prop]
  Table --> Offset[Compute clamped shared offset]
  Offset --> Frame[Translate DOM table frame]
  Offset --> Cells[Offset canvas cell boxes]
  Pointer[Pointer press] --> Hold[Hold current offset]
  Hold --> Frame
  Hold --> Cells
  Release[Pointer up or cancel] --> Offset
```

<sub>Reviews (2): Last reviewed commit: ["Throw explicitly when the table frame is..."](https://github.com/synnaxlabs/synnax/commit/02185752196823d84adc97ee40d6b89c6a36dd5b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52833737)</sub>


pluto/src/table/Table.tsx
```diff
@@ -462,15 +477,47 @@ export const Table = ({
...
+      x: half(box.width(region), totalCol + indicatorSize),
+      y: half(box.height(region), totalRow + indicatorSize),
+    };
+  }, [centered, region, totalCol, totalRow, indicatorSize]);
+  // A live offset would slide a dragged edge out from under the cursor.
+  const [held, setHeld] = useState<xy.XY | null>(null);
+  const handlePointerDown = useCallback(() => {
+    if (!centered) return;
+    setHeld(liveOffset);
+    window.addEventListener("pointerup", () => setHeld(null), { once: true });
```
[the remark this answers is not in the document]
emilbon99: Fixed in cb6aec7. The listener pair now removes itself on either `pointerup` or `pointercancel`, matching how `Cursor.useDrag` tears down its own window listeners. Pinned by "releases the held offset on pointer cancellation" in `pluto/src/table/Table.spec.tsx`, which I verified fails when the `pointercancel` listener is removed.

(Reply written by Claude on my behalf.)

pluto/src/table/Table.tsx
```diff
@@ -462,15 +477,47 @@ export const Table = ({
...
+  }, [centered, liveOffset]);
+  const offset = centered ? (held ?? liveOffset) : xy.ZERO;
```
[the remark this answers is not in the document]
emilbon99: Fair, and now addressed in cb6aec7. `pluto/src/table/Table.spec.tsx` mounts the table and covers the offset on both axes, clamping to zero on an axis it overflows, recentering when the table changes size, holding the offset during a press, and releasing it on cancel. The hold and cancel cases were each verified to fail when the behavior is removed.

Two limits worth stating. The spec stubs `ResizeObserver`, because the `@juggle` polyfill reads computed styles that jsdom leaves empty and so never fires. And the assertions are on the frame transform, not on the canvas cell boxes, since the text cells used here draw nothing; DOM and canvas agreement rests on both reading the same `offset` value.

The console toolbar toggle is still uncovered: mounting the tab content fails with `[Root.create] - Table not found in registry`, because `pluto/src/ether.ts` does not export `table`.

(Reply written by Claude on my behalf.)

emilbon99:
@greptile re-review

console/src/feature/table/Table.tsx
```diff
@@ -104,13 +113,25 @@ const TableControls = (): ReactElement | null => {
...
   return (
     <Vis.Controls x>
+      <Button.Toggle
+        value={centered}
+        onChange={handleToggleCentered}
+        size="small"
+        tooltipLocation={location.BOTTOM_LEFT}
+        tooltip={centered ? "Align table to top left" : "Center table"}
+      >
+        {centered ? <Icon.Align.Left /> : <Icon.Align.XCenter />}
```
sy-nico: You should swap these. When left-aligned, the button shows centered and vice versa. I still think the button should be outlined when not centered and filled when centered/active.

Otherwise it looks like the button mode contradicts the centered mode.

Image

pluto/src/table/Table.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+    vi.stubGlobal("ResizeObserver", ImmediateResizeObserver);
+    vi.spyOn(HTMLElement.prototype, "getBoundingClientRect").mockReturnValue(RECT);
+    // Canvas.useRegion measures against the lower2d canvas and bails without it.
+    const canvas = document.createElement("div");
+    canvas.className = "pluto-canvas--lower2d";
+    document.body.appendChild(canvas);
+    wrapper = await createAsyncSynnaxWrapper({
+      client,
+      additionalRegistry: aetherTable.REGISTRY,
+      renderContext: canvasTest.record(),
```
sy-nico: Add tests for the canvas boxes, not just the frame transform. `canvasTest.record()` is wired up here but never asserted, so dropping the offset from `cellXOrigin` or `rowYCursor` would strand the values behind and every test would still pass.

pluto/src/table/Table.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { canvasTest } from "@/vis/render/test";
+
+const client = createTestClient();
+
+const SURFACE_WIDTH = 1000;
+const SURFACE_HEIGHT = 600;
+const COL_SIZE = 72;
+const ROW_SIZE = 36;
+// Indicator strips occupy one row and one column outside the stored sizes.
+const INDICATOR_SIZE = 4.5 * 6;
```
sy-nico: Should import from `pluto/src/table/Indicator.tsx:26` instead of redeclaring.

sy-nico (changes requested):

sy-nico (approved):

PR #2708, opened 2026-08-14
SY-4560: Cluster-wide control state service

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4560](https://linear.app/synnax/issue/SY-4560/fix-control-state-provider)

## Description

Control state was published per node on `sy_node_<host>_control`, and the current<br>snapshot was injected into the first frame of every streamer. Two problems followed. A<br>client subscribed to one node's channel, so on a multi-node cluster it saw only the<br>transfers that node arbitrated and silently missed the rest. And the streamer carried<br>control-state assembly it had no other reason to know about, so every consumer of live<br>telemetry paid for a feature only the control views use.

Control state is now a read path of its own, described in RFC 0058.

Cesium stops writing its own digest channel and exposes an observable instead. The<br>digest writer wrote control updates back into Cesium through a virtual channel, which<br>needed a dedicated inlet, a shutdown path, and three guards elsewhere in the writer to<br>keep that writer from cleaning itself up. All of that is gone; `ControlStates()` stays<br>as the local snapshot.

A new distribution `control` service makes the state cluster-wide. `Retrieve` partitions<br>keys by leaseholder and answers the gateway bucket locally, and `OnChange` merges the<br>host's observable with a subscription to each peer. A service-layer `control` service<br>sits above it, answers reads, and is the only writer of one free virtual channel,<br>`sy_control`, carrying every transfer in the cluster. A third API service exposes the<br>retrieve.

Clients read through the query cache. The TypeScript client's `StateTracker` is replaced<br>by a `query.Retriever` whose table is keyed by the channel under control, so absence in<br>the table is the uncontrolled state and a release deletes the entry rather than storing a<br>sentinel. Pluto's `Control.StateProvider` shrinks to `Control.Colors`, which now does one<br>thing: assign each subject a color. Its two consumers, the legend and the authority<br>source, read control state from the client through `flux.Retrieve`. The Driver subscribes<br>to `sy_control` and seeds its mirror from the retrieve. Python gets the retrieve only;<br>nothing in the tree streams control state from Python.

`sy_node_<host>_control` is deleted outright, with no aliasing release. A Driver older<br>than the Core fails its channel lookup and starts no tasks, so a standalone Driver must<br>be upgraded with the Core.

## Basic Readiness

New specs cover each layer: `core/pkg/distribution/control` for leaseholder routing and<br>peer subscriptions, `core/pkg/service/control` for the channel's shape and for publishing<br>an acquire and a release, and `core/pkg/api/control` for the retrieve and its access<br>check. `client/ts/src/control/state.spec.ts` drives a real writer against a live Core and<br>asserts the client's table mirrors the acquire and the release.

The Cesium specs that used to observe the digest channel were rewritten onto<br>`OnControlUpdate` rather than dropped, so the same transfers are still asserted through<br>the replacement. Specs whose only subject was deleted machinery went away with it: the<br>`ConfigureControlUpdateChannel` validation pair, the digest-channel delete spec already<br>covered by the unclosed-writer spec above it, and the streamer's digest forwarding.

Not covered: the node-departure path in the distribution service, which needs a cluster<br>membership change mid-subscription and has no fixture today.

The C++ has not been compiled or run. That includes the new<br>`driver/task/manager_test.cpp` spec for control-state seeding, which is written but<br>unverified.

### Greptile Summary

The PR replaces node-local control digests with a cluster-wide control read and update service.

* Adds distributed control retrieval and peer subscriptions.
* Publishes transfers on `sy_control` and exposes retrieval in all client SDKs.
* Seeds the Driver control mirror from retrieval and updates Pluto control consumers.

### Confidence Score: 4/5

The keyed control retrieval defect must be fixed before merge because controlled free channels are reported as uncontrolled.

`Retrieve` partitions free keys separately but never reads that partition, so valid keyed requests silently omit current control state.

**Files Needing Attention:** core/pkg/distribution/control/service.go

### Important Files Changed

| Filename | Overview |
| -- | -- |
| core/pkg/distribution/control/service.go | Adds leaseholder-routed retrieval and peer state subscriptions, but keyed retrieval drops controlled free channels. |
| core/pkg/service/control/service.go | Adds the service-layer retrieval facade and publishes distributed transfers on `sy_control`. |
| driver/task/manager.cpp | Subscribes to `sy_control` and seeds the Driver authority mirror before processing streamed updates. |
| client/ts/src/control/state.ts | Defines keyed control-state and transfer schemas for query-cache storage. |
| cesium/control.go | Replaces the internal digest writer with an observable control-update interface. |

### Sequence Diagram

```mermaid
sequenceDiagram
  participant Client
  participant API as Control API
  participant Dist as Distribution control
  participant Host as Host Cesium
  participant Peer as Peer control service
  participant Signal as sy_control
  Client->>API: Retrieve(channel keys)
  API->>Dist: Retrieve(keys)
  Dist->>Host: ControlStates()
  Dist->>Peer: Retrieve(peer-owned keys)
  Host-->>Dist: Local states
  Peer-->>Dist: Peer states
  Dist-->>API: Cluster states
  API-->>Client: Controlled channels
  Host-->>Dist: Control transfer
  Peer-->>Dist: State snapshot
  Dist->>Signal: Publish transfer
```

Reviews (1): Last reviewed commit: ["SY-4560: Cluster-wide control state serv..."](<https://github.com/synnaxlabs/synnax/commit/0b8fac35945e8f305bc51890fb895e0bb348beb1>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=53380038>)

<details><summary>Context used (5)</summary>

* Knowledge Base — [Core Server](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/core-server.md>)
* Knowledge Base — [Cesium Storage Engine](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/cesium-storage.md>)
* Knowledge Base — [Client SDKs](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/client-sdks.md>)
* Knowledge Base — [Driver: Connecting Physical Devices to Synnax](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/driver-devices.md>)
* Knowledge Base — [Pluto Component and Visualization Library](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/pluto-components.md>)

</details>


core/pkg/distribution/control/service.go
```diff
@@ -0,0 +1,458 @@
...
+func (s *Service) Retrieve(
+	ctx context.Context,
+	keys ...channel.Key,
+) ([]State, error) {
+	if len(keys) == 0 {
+		return s.retrieveAll(ctx)
+	}
+	batch := s.router.Batch(keys)
+	states := statesFromStorage(s.cfg.TS.ControlStates())
+	out := filterByKeys(states, batch.Gateway)
```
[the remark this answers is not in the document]
emilbon99: Investigated with Claude Code: this is a false positive. A free channel has no storage registration, so its key can never appear in `TS.ControlStates()`, and adding `batch.Free` to the filtered key set changes nothing.

- `core/pkg/distribution/channel/create.go:112` states the rule: free channels have no storage registration, and their writes flow through the distribution relay instead of the storage engine.
- `core/pkg/distribution/framer/writer/free.go:128` returns `Authorized: true` for every request without consulting an arbiter, so control over a free channel is never arbitrated and no transfer is ever produced.
- `core/pkg/distribution/framer/iterator/service.go:61` rejects free keys on read.

No subject controls a free channel, so its absence from the result is the documented contract of `Retrieve`.

client/cpp/control/control.h
```diff
@@ -0,0 +1,77 @@
...
+                    .subject = subject,
+                    .resource = pb.resource(),
+                    .authority = static_cast<x::control::Authority>(pb.authority()),
+                }
+            );
+        }
+        return {states, x::errors::NIL};
+    }
+
+private:
```
emilbon99, line 73: Move to top of class and remove `private`

docs/tech/rfc/0058-control-state-service.md
```diff
@@ -0,0 +1,330 @@
...
+than the noop most endpoints get, because the Driver reads control state over gRPC.
+
+Per §3.1 this service never grows a `/control/stream` endpoint. `sy_control` is the
+stream.
+
+### 4.4 TypeScript client: control as a query domain
+
+`control.Client` becomes a `query.Retriever` over a table keyed by channel key, in the
+shape of `label.Client` (`client/ts/src/label/client.ts:61`): `fetch` calls
+`/control/retrieve`, and one listener parses `Update` from `sy_control` and applies each
```
pjdotson, line 197: Does `sy_control` carry the whole state or just updates (was a little unclear in the above section)? If it is just updates, isn't there a worry w/ clients missing some earlier entries from between `sy_control` and `/control/retrieve`?

core/pkg/distribution/control/control.go
```diff
@@ -0,0 +1,107 @@
...
+// Package control makes the control state of a Synnax cluster readable and observable
+// from any node. Control is arbitrated by the storage engine on each channel's
+// leaseholder, so the Service fans retrieves out to leaseholders and merges peer
+// subscriptions into a single cluster-wide stream of transfers.
+package control
+
+import (
+	"github.com/synnaxlabs/synnax/pkg/distribution/channel"
+	"github.com/synnaxlabs/synnax/pkg/storage/ts"
+	xcontrol "github.com/synnaxlabs/x/control"
```
pjdotson: Nit: no need to alias

core/pkg/distribution/control/control.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return &State{
+		Subject:   s.Subject,
+		Resource:  channel.Key(s.Resource),
+		Authority: s.Authority,
+	}
+}
+
+// updateFromStorage converts a storage-layer control update into a distribution-layer
+// update.
```
pjdotson: bad comment here and in a lot of places in this PR

core/pkg/distribution/control/control.go
```diff
@@ -0,0 +1,107 @@
...
+func updateFromStorage(u ts.ControlUpdate) Update {
+	transfers := make([]Transfer, 0, len(u.Transfers))
+	for _, t := range u.Transfers {
```
pjdotson: use `lo.Map`

core/pkg/distribution/control/service.go
```diff
@@ -0,0 +1,440 @@
...
+		return nil, err
+	}
+	s := &Service{
+		cfg:     cfg,
+		router:  proxy.BatchFactory[channel.Key](cfg.Cluster.HostKey()),
+		updates: observe.New[Update](),
+	}
+	s.mu.peers = make(map[node.Key]*peer)
+	sCtx, cancel := signal.Isolated(signal.WithInstrumentation(cfg.Instrumentation))
+	s.sCtx = sCtx
```
pjdotson: why are we using isolated contexts and storing them on a struct here?

core/pkg/distribution/control/service.go
```diff
@@ -0,0 +1,440 @@
...
+			s.addPeer(ctx, n.Key)
+		}
+	}
+}
+
+// addPeer opens a subscription to target and starts reading it. The stream is opened
+// here rather than inside the reader so that the subscription is live before addPeer
+// returns, leaving no window in which a joining peer's transfers go unseen.
+func (s *Service) addPeer(ctx context.Context, target node.Key) {
+	peerCtx, stop := context.WithCancel(context.WithoutCancel(ctx))
```
pjdotson, line 315: what is the point of doing this `WithCancel(WithoutCancel)`?

core/pkg/distribution/control/service.go
```diff
@@ -0,0 +1,440 @@
...
+	if err != nil {
+		return err
+	}
+	for {
+		if openErr == nil {
+			b.Reset()
+			openErr = s.readPeerStream(ctx, target, stream)
+		}
+		if ctx.Err() != nil {
+			return nil
```
pjdotson: shouldn't we return `ctx.Err` here and below?

core/pkg/distribution/control/control_test.go
pjdotson, line 1: not really enough tests here for failure cases (e.g. unable to connect to peers, losing connection from peer, etc.)

core/pkg/distribution/transport/grpc/control/transport.go
pjdotson, line 1: needs test

core/pkg/distribution/transport/grpc/transport.go
pjdotson, line 1: needs test

core/pkg/distribution/transport/mock/control/network.go
pjdotson, line 1: needs test

core/pkg/distribution/transport/mock/network.go
pjdotson, line 1: needs test

core/pkg/distribution/control/pb/translator.go
pjdotson, line 1: needs test

driver/control/state.h
```diff
@@ -47,8 +47,16 @@ class States {
...
+    void set(const std::vector<x::control::State<synnax::channel::Key>> &next) {
+        std::unique_lock lock(this->mu);
+        this->states.clear();
+        for (const auto &state: next)
+            this->states[state.resource] = state;
+    }
+
     void apply(const x::telem::Series &series) {
-        if (series.data_type() != x::telem::STRING_T) return;
+        if (series.data_type() != x::telem::JSON_T) return;
```
pjdotson: shouldn't we actually error or something here? when would we ever reach this condition?

client/ts/src/control/client.ts
```diff
@@ -7,27 +7,97 @@
...
+import { z } from "zod";
 
-export const CONTROL_STATE_CHANNEL_NAME = "sy_node_1_control";
+import { type channel } from "@/channel";
+import { keyZ } from "@/channel/types.gen";
+import { type KeyedState, keyedStateZ, updateZ } from "@/control/state";
+import { query } from "@/query";
+
+/** The free virtual channel carrying control updates for the entire cluster. */
+export const CHANNEL_NAME = "sy_control";
```
pjdotson: I thought channel names like this weren't exported at all anymore

driver/task/manager.cpp
```diff
@@ -198,6 +198,18 @@ x::errors::Error Manager::run(std::function<void()> on_started) {
     return c_err;
 }
 
+void Manager::seed_control_states() {
+    auto [states, err] = this->ctx->client->control.retrieve();
+    if (err) {
+        // The mirror treats an absent entry as uncontrolled and the Core arbitrates
+        // every write, so an unseeded driver starts optimistic and converges as
```
pjdotson: capitalize Driver

core/pkg/service/control/control.go
```diff
@@ -0,0 +1,31 @@
...
+	// Transfer is a transition of control over a single channel. A nil From is an
+	// acquire; a nil To is a release.
+	Transfer = control.Transfer
+	// Update is a batch of transfers that occurred atomically.
+	Update = control.Update
+)
+
+// ChannelName is the name of the free virtual channel carrying control updates for the
+// entire cluster.
+const ChannelName = "sy_control"
```
pjdotson: why does this need to be public?

core/pkg/transport/grpc/control/handler.go
pjdotson, line 1: need tests

core/pkg/service/control/service.go
```diff
@@ -0,0 +1,129 @@
...
+	ctx context.Context,
+	keys ...channel.Key,
+) ([]State, error) {
+	return s.cfg.Control.Retrieve(ctx, keys...)
+}
+
+// Close stops publishing updates and releases the service's resources.
+func (s *Service) Close() error { return s.shutdown.Close() }
+
+func (s *Service) encode(
```
pjdotson: feel like this should be something that is exposed via the `signals` package which gives an option that is just JSON marshalling the type T or something.

pjdotson (changes requested):
looking mostly good

docs/tech/rfc/0058-control-state-service.md
emilbon99: Two things:

1) It is only going to transfer updates. I have a plan in mind to completely refactor the control transfer and related data structures to make them more useful, but I want to defer to 0.58.
2) Yes there is a narrow race where we can keep stale data. I have filed an issue for it: https://linear.app/synnax/issue/SY-4597/close-clientquery-synchronization-races

core/pkg/distribution/control/service.go
emilbon99: `ctx` is a short lived context, so we can't use it for managing long lived process. That means we can either use `context.Background()` or we can use `WithCancel` to get all of the context values and other info without taking it's lifetime. 

This was released in go 1.21 and is recommended by the go team for exactly this use case.

core/pkg/distribution/control/service.go
emilbon99: Got rid of the context on a struct and refactored to use `signal.WithCancel`. See the answer to the comment below to understand why using `context.WithoutCancel` is warranted.

pjdotson (approved):

PR #2754, opened 2026-08-20
SY-4653: Pin pre-v0.54 key normalization to frozen shapes

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4653](https://linear.app/synnax/issue/SY-4653/pre-v054-task-and-workspace-rows-break-the-v057-migration-chain)

## Description

Gorp prepended a key-normalization migration to every table, decoding pre-v0.54 rows<br>with the head entry type. That binds frozen bytes to a type that keeps moving, and it<br>has rotted three times:

* The task table's head key changed from uint64 to UUID, so a Core last opened by v0.53<br>or earlier fails to decode at `task.OpenService` and never boots.
* The workspace table was renamed to project, so old-prefix workspace rows are never<br>found: the project chain sees zero workspaces, records itself applied, and every<br>workspace silently vanishes on a direct v0.53 to v0.57 upgrade.
* `CustomTypeName` renamed `Pair` to `KVPair` in v0.54.0. Every release since has looked<br>for legacy rows under `msgpack("KVPair")`, so pre-v0.54 range key-value pairs have<br>never been migrated by any release. This one was found by auditing the other tables<br>and is fixed here too.

`OpenTable` no longer prepends anything, and `TableConfig.NormalizeKeys` is gone.<br>`NormalizeKeysMigration` now takes the legacy type name as an argument instead of<br>deriving it from the head type, so the name is a literal that cannot drift. Every table<br>first stored by v0.53 or earlier pins its own normalization in its `versions/v0`<br>package, beside the frozen shape it decodes, and runs it first in its chain. `panel` and<br>`task/config` were created after v0.54, hold no old-prefix rows, and pin nothing.

Tests: one spec per `versions/v0` package seeds a raw pre-v0.54 row and carries it<br>through the pinned migration; `core/pkg/service/normalize_keys_test.go` asserts all 24<br>chains open with normalization, so dropping a pin fails a test instead of orphaning<br>rows; `SetPreV54Row` moved to `x/go/gorp/testutil`.

One related hazard is left open: renaming a head entry type still silently changes both<br>the key prefix and the migration namespace. Pinning normalization does not address it.

## Basic Readiness

### Greptile Summary

The PR moves pre-v0.54 key normalization from implicit table-open behavior into explicit, frozen migration chains.

* Adds pinned normalization migrations for all tables that can contain pre-v0.54 rows.
* Uses frozen entry shapes and literal legacy type names for task, workspace, and range key-value migrations.
* Adds migration-chain and raw legacy-row coverage.

### Confidence Score: 5/5

The PR appears safe to merge.

No blocking failure remains.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| x/go/gorp/table.go | Removes automatic normalization from `OpenTable` and exposes normalization as an explicit migration with a frozen legacy type name. |
| core/pkg/service/task/versions/migrations.go | Runs frozen task-key normalization before the task migration chain. |
| core/pkg/service/project/versions/migrations.go | Runs workspace-prefix normalization before workspace-to-project migrations. |
| core/pkg/service/ranger/kv/versions/v0/migrate.go | Pins the legacy `Pair` type name so old range key-value rows are discovered. |
| core/pkg/service/normalize_keys_test.go | Verifies that every applicable service migration chain starts with key normalization. |

### Flowchart

```mermaid
flowchart LR
  A[Pre-v0.54 row] --> B[Frozen v0 normalization]
  B --> C[Current key prefix and encoding]
  C --> D[Later schema migrations]
  D --> E[Current table]
```

Reviews (2): Last reviewed commit: ["SY-4653: Use qualified ginkgo imports in..."](<https://github.com/synnaxlabs/synnax/commit/1a6c3709fd20b42416b9b3ed6a94d17ff553ffe6>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=54992139>)

**Context used:**

* Knowledge Base — [Core Server](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/core-server.md>)
* Knowledge Base — [x: Shared Cross-Language Utilities](<https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/x-shared-utils.md>)


emilbon99:
@greptile re-review

x/go/gorp/testutil/legacy.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/synnaxlabs/x/encoding/msgpack"
+	"github.com/synnaxlabs/x/kv"
+	"github.com/synnaxlabs/x/testutil"
+)
+
+// SetPreV54Row writes a raw row in the key format releases before v0.54 used:
+// msgpack(typeName) + msgpack(key), with an msgpack-encoded value. typeName is the type
+// name the writing release reported, which a later rename can move away from. Returns
+// the full key so callers can assert the row is gone after normalization.
+func SetPreV54Row(
```
pjdotson, line 26: add test

pjdotson (changes requested):
Looks good but should we change Gorp to not determine the prefix based off of getting the type name and instead have it be set explicitly?

emilbon99:
> Looks good but should we change Gorp to not determine the prefix based off of getting the type name and instead have it be set explicitly?

I think there may be a way in the future to do this by passing the type name via the table, but one of the things to be careful about is the fact that the type names are carried across different data structure versions. I want to defer for now.

PR #2757, opened 2026-08-20
SY-4656: Sweep status favorites with a keyed existence check

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4656](https://linear.app/synnax/issue/SY-4656/evict-status-favorites-only-on-proof-of-deletion)

## Description

The status synchronizer reconciled favorites by listing every status on the Core and dropping any favorite absent from that list. That listing can be older than the favorite it is compared against — `Retriever.retrieve` joins an in-flight fetch, so it can answer as of before the caller's own write — and the user then loses a favorite whose status still exists.

Measured at the same instant inside `reconcile`, the cached read against a cache-bypassing one:

```
nCached: 1341   inCached: []              <- cached answer, favorite absent
nFresh:  1342   inFresh: ["gnPim5GnXfL"]  <- direct read, favorite present
```

Absence from an unbounded listing is not proof of deletion. Absence from a keyed answer is: the Core matched those exact keys and reported nothing.

Statuses were the one domain whose retrieve had no `ignoreNotFoundError`, which is why the synchronizer was hand-rolled rather than built on `Synchronizer.createRemover` like every other session slice — a keyed sweep would have thrown on the first deleted favorite. This adds the flag through the stack and puts statuses on the shared remover:

- **Core** — `RetrieveRequest.IgnoreNotFoundError` skips `query.ErrNotFound`, matching the label service.
- **Transport** — `ignore_not_found_error = 8` on the gRPC `RetrieveRequest`, carried both ways through the translator.
- **Client** — one field on `retrieveRequestZ`. Deliberately not in `serverFields`: the flag changes error behavior, not which records match, so `{keys, ignoreNotFoundError}` stays maintained by the client-checkable rule.
- **Console** — the synchronizer drops from 57 lines to 34. The `onSet` notification forwarding becomes its own registry entry; it has no state to repair, so `Callbacks.reconcile` is now optional, mirroring `listen`.

The sweep also drops the reconcile from one listing of every status on the Core (~1300 in a loaded test cluster) to a single keyed request.

One tradeoff: status retrieve enforces access over the records it returns, so one inaccessible favorite now fails the sweep for all of them, where per-key lookups would isolate it. The sweep fails closed and keeps the favorites, so this is a missed repair rather than a wrong eviction.

The synchronizer spec picked up the sweep coverage Greptile flagged as missing, including a regression test for this bug — a favorite created moments before the sweep, waiting on the `verified` flag so it cannot pass before a reconcile pass completes. Flipping `ignoreNotFoundError` to `false` fails that sweep spec, so the path from the TS schema through the Core is genuinely exercised.

Fixes the intermittent `status toolbar > should drop favorites whose status no longer exists` failure.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR replaces full status listing reconciliation with a keyed existence sweep and carries optional not-found suppression through the TypeScript client, transport, and Core API.
- Uses the shared session remover to evict only favorites whose requested keys are absent.
- Makes reconciliation optional for notification-only synchronizers.
- Adds Core and Console coverage for partial keyed retrieval, deleted favorites, surviving favorites, and newly created favorites.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| console/src/session/status/synchronizer.ts | Replaces hand-written full-list reconciliation with the shared keyed remover and separates notification forwarding. |
| console/src/session/status/synchronizer.spec.ts | Adds integration coverage for connected deletion, keyed sweeping, and retention of live and newly created favorites. |
| console/src/session/synchronizer/create.ts | Allows state-free synchronizers to omit reconciliation. |
| console/src/session/synchronizer/use.ts | Safely invokes reconciliation only when the callback is present. |
| client/ts/src/status/client.ts | Adds the optional not-found suppression flag to the validated status retrieval request. |
| core/pkg/api/status/status.go | Implements partial keyed retrieval by suppressing only not-found errors when requested. |
| core/pkg/api/status/status_test.go | Covers partial results, all-missing results, and default not-found behavior. |
| core/pkg/transport/grpc/status/handler.go | Translates the new retrieval flag in both transport directions. |
| core/pkg/transport/grpc/status/status.proto | Adds the backward-compatible boolean field to the status retrieval request. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Console
  participant TS as TypeScript status client
  participant Core as Core status API
  participant Store as Session store
  Console->>TS: "retrieve({keys: favorites, ignoreNotFoundError: true})"
  TS->>Core: keyed RetrieveRequest
  Core-->>TS: statuses that still exist
  TS-->>Console: existing statuses
  Console->>Store: removeFavorites(requested keys not returned)
```

<sub>Reviews (2): Last reviewed commit: ["SY-4656: Sweep status favorites with a k..."](https://github.com/synnaxlabs/synnax/commit/b645eebef3e3da701a48d5f9a64b4140392aeb7e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55016442)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Context used - CLAUDE.md ([source](https://github.com/synnaxlabs/synnax/blob/main/CLAUDE.md))
</details>


pjdotson:
@greptile /rerun

emilbon99 (approved):

PR #2759, opened 2026-08-20
SY-4643: Add trigger hints and fix the shortcuts they advertise

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4643](https://linear.app/synnax/issue/SY-4643/look-for-places-where-we-can-add-trigger-hints-to-context)

## Description

An audit of every keyboard trigger in Pluto and the Console, matching each against the button or menu item that performs the same action, then closing the gaps. Most of the value is not the hints themselves: the audit found five shortcuts that were broken, and the pattern that let them break.

**Shortcuts that did nothing.** The OPC UA, HTTP, and Modbus connect modals and the schematic symbol editor each rendered a `SaveHelpText` for Ctrl+Enter while no `trigger` was bound to their submit button. `Input.DateTime` did the same in its footer. The keys were advertised and inert.

**Table undo and redo were cursor-gated.** They registered with `region: tableElRef`, which requires the mouse pointer to sit inside the table's box. Moving the pointer to the toolbar silently disabled them. They now use the shared `Triggers.useUndoRedo`, so they are tab-wide like every other visualization; erase keeps its region so Delete elsewhere cannot clear cells.

**The palette tooltip never opened.** It wrapped `Dialog.Frame`, whose props omit `ref` and route `onPointerEnter` into dialog context, so neither the tooltip's ref nor its hover handlers reached a DOM node. It now sits on the button, which is `Tooltip.wrap`ped.

**Focus mode split.** Ctrl+L only enters focus mode; Escape is the only way out. The hints follow: the tab menu shows Ctrl+L as "Focus" and switches to `esc` once overlaid.

**Read-only diagrams no longer rewrite themselves.** `Diagram.useTriggers` gated undo, redo, and paste behind a new `editable` flag, matching the context menu that already withheld them. Copy and the selection shortcuts stay live so a read-only diagram is still navigable. Latent in the Console, where both props carry the same value, but real for any other consumer of the component.

**Keycaps print what the key prints** — `esc`, `delete`, `tab` on macOS and the capitalized forms on Windows, matching the existing Command/Option split. `KEYS` gained `Equal` and `Minus`, which let the symbol editor's zoom shortcuts drop a hand-rolled `window` keydown listener for `Triggers.use`; those keys were previously unrepresentable, which is why it existed.

**Hints added** where a trigger and a control already both existed: the palette's "Open new window" command (this needed a `trigger` field on `Command.create`, so every future command gets it free), the modal close and Confirm's Cancel buttons, the mosaic create button and empty state, the diagram fit-view button, and the table's erase item.

**Consolidation.** A new `Menu.UndoRedoItems` replaces the duplicated entries in the schematic and Arc editors and adds them to the table, the log, and the line plot, none of which offered undo from a menu. `CREATE_TAB_TRIGGER` and `ERASE_TRIGGER` became exported constants so a hint cannot drift from its handler.

### Testing notes

Hint assertions tie back to the exported constant rather than restating a literal, so drift fails the build. The revived shortcuts assert real outcomes — a symbol written to the cluster, channel data deleted, the picker closed. The six highest-value tests were verified to fail against the original code.

Two paths are untested and deliberately so. The Console line plot's menu wiring cannot be reached: the plot renders empty under jsdom and its menu content is internal, so covering it would mean exporting a symbol purely for tests. The mosaic create button's tooltip needs an editor, and the button does not render for the test client; the constant it hints is covered by the Ctrl+T test in `app/triggers/use.spec.tsx`.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR audits keyboard shortcuts across Console and Pluto, repairs inert or incorrectly scoped bindings, and adds matching trigger hints to their controls.
- Adds save, create, erase, focus, palette, fit-view, zoom, undo, and redo trigger wiring and indicators.
- Moves symbol-editor zoom to bare-key triggers so preview zoom no longer runs alongside native browser zoom.
- Gates mutation triggers and menu items by editability while retaining read-only navigation actions.
- Consolidates shared undo/redo menu entries and exports reusable trigger constants.
- Adds unit and integration coverage for shortcut outcomes and displayed hints.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| console/src/feature/schematic/symbol/edit/Preview.tsx | Replaces window-level Ctrl/Cmd zoom listeners with strict bare-key Pluto triggers, adds repeat cleanup, and displays matching zoom hints. |
| console/src/feature/schematic/symbol/edit/triggers.ts | Defines reusable bare-key zoom modes that avoid coupling preview zoom to browser-reserved control chords. |
| pluto/src/triggers/Provider.tsx | Extends shared key handling and presentation behavior used by the newly wired shortcut controls. |
| console/src/feature/panel/ContextMenu.tsx | Makes focus hints state-aware and withholds focused-tab shortcuts from menus opened on background tabs. |
| pluto/src/vis/diagram/useTriggers.ts | Gates mutation shortcuts by editability while preserving copy and selection behavior. |
| pluto/src/table/Table.tsx | Moves undo and redo to shared tab-wide trigger handling while retaining region-scoped erase behavior. |

<sub>Reviews (2): Last reviewed commit: ["Formatting"](https://github.com/synnaxlabs/synnax/commit/f428fb408b83ed16e74d962b43d059ec92770265) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55016624)</sub>

**Context used:**


console/src/app/palette/Palette.spec.tsx
```diff
@@ -74,6 +74,22 @@ describe("Palette", () => {
     expect(input.value).toBe("");
   });
 
+  it("should offer both palette shortcuts when the button is hovered", async () => {
+    await renderAppPalette();
+    const btn = document.querySelector<HTMLElement>(".console-palette__btn");
+    if (btn == null) throw new Error("palette open button not found");
```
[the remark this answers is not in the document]
pjdotson: *Patrick + Claude: drafted in our review session.*

Half right. The history framing was a real miss, so I cut "the footer has always advertised X, but nothing bound it" from all seven and kept the part that says why the assertion proves anything.

The rest stays. Those comments carry contracts the code cannot show: why the chord helper holds Control down through the whole sequence, why the menu entry advertises Escape instead of Control+L, why the keycaps need aria-hidden, and which assertion is a positive control. Our rule bans restating the next line, not explaining a non-obvious one.

console/src/feature/schematic/symbol/edit/Preview.tsx
```diff
@@ -25,6 +34,18 @@ import { FileDrop } from "@/feature/schematic/symbol/edit/FileDrop";
 import { HandleOverlay } from "@/feature/schematic/symbol/edit/Handles";
 import { CSS } from "@/platform/css";
 
+export const ZOOM_TRIGGERS: Triggers.ModeConfig<"in" | "out" | "reset" | "default"> = {
+  defaultMode: "default",
+  modes: {
+    in: [["Control", "Equal"]],
```
emilbon99: A bit weird that this trigger presents itself as "Equal" at the triggers text in the tooltip.

emilbon99 (changes requested):
Why did you decide to make the keycap border colors invisible on text buttons?

I also fixed an issue where hovering would cause the keycap text to shift. See my commit. It does need to be fixed although you can reply my fix with yours if you want.

console/src/feature/schematic/symbol/edit/Preview.tsx
pjdotson: image

pjdotson:
@greptile /rerun

pjdotson:
> Why did you decide to make the keycap border colors invisible on text buttons?
> 
> I also fixed an issue where hovering would cause the keycap text to shift. See my commit. It does need to be fixed although you can reply my fix with yours if you want.

I changed the border colors to show again.

emilbon99 (approved):

PR #2760, opened 2026-08-20
SY-4642: Standardize renaming

pjdotson:
## Linear Issue

[SY-4642](https://linear.app/synnax/issue/SY-4642/standardize-renaming)

## Description

Renaming a resource now always edits its name in place. The rename modal survives only where it is not a rename: the two creation flows and the device identifier, which edits a hidden property.

Range and Arc rows in the ontology tree registered no context menu, so right-clicking gave only "Reload Console". Both now carry their full action sets, rename included. Renaming a channel shown under a range alias swaps to the real name, edits that, and restores the alias on commit; internal channels lose the item entirely, where it used to retry for a second and reject.

Access checks gate the editable field itself, not just the menu item, so double-click and `Ctrl+E` can no longer bypass a check the menu enforces. This covers mosaic tabs, panel pills, the project selector, schematic symbols and groups, workspace views, and the Modbus and OPC UA write-task menus.

Four `Text.Editable` defects are fixed: an unchanged commit dispatched no event and left every tree rename pending forever, the `asyncEdit` listeners leaked, an untouched injected commit reported a rename instead of a cancel, and leaving an edit repainted from `value` and flashed the old name for a frame. All are mutation-verified.

A local range could not be renamed at all — the rename aborted before writing the slice instead of skipping only the server call. `useRename` moves to the session layer beside that slice, with a spec for the persisted, local, rollback, and unfavorited paths.

Smaller fixes: "Add to active plot" now gates on a focused line plot rather than any focused tab. The Core list accepts a rename to the same name. The project selector drops double-click rename and shows a menu on "New project". The line plot title commits on blur instead of renaming per keystroke. User, role, project, and symbol renames became optimistic. Arc deletes drop session state in `beforeUpdate`, matching the four trees that already do; a range delete does the same but rolls back, so a rejected delete keeps the favorite at its place in the list along with the selection it carried.

Layout: the view pill drew its rename outline outside itself; a long range name in the toolbar now ellipsizes and stays inside its box while editing; the Arc explorer list no longer sits pinned to the bottom of its pane.

Task and Arc renames do not stop or reconfigure a running task — deployment is start-triggered and hash-gated since SY-4488, and the name is in neither config hash. The stale Arc confirm dialog is deleted.

## Note for reviewers

Split from one branch alongside #2761 and #2762, both merged. `rc` is merged in, so the `platform/view/Frame.tsx` and `toolbar/Symbols.tsx` overlaps are resolved here.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR standardizes resource renaming around inline editing and adds permission-aware rename controls. It also repairs the two previously reported lifecycle issues:
- Failed range deletions now restore the removed favorites, their ordering, and their prior selection.
- Injected editable values now compare unchanged commits against the injected value and resolve them as cancellations.
- Rename mutation hooks and context menus are consolidated across resource types.
- Tests cover rename permissions, optimistic updates, rollback behavior, aliases, and editable-field lifecycle behavior.

The PR appears safe to merge because both previously reported failures are resolved and no blocking failure remains.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| console/src/session/range/useDelete.ts | Snapshots session ranges and selection before optimistic deletion and restores them through the mutation rollback. |
| console/src/session/range/slice.ts | Adds index-aware restoration that reinstates removed ranges and their selected state. |
| console/src/feature/range/tree.tsx | Adds the full range context menu and routes deletion through the rollback-aware session hook. |
| pluto/src/text/Editable.tsx | Corrects injected-value unchanged detection, event-listener cleanup, cancellation signaling, and committed-text repaint behavior. |
| console/src/platform/tree/createUseRename.ts | Supplies the underlying resource name when inline editing a derived display value such as a channel alias. |

<sub>Reviews (3): Last reviewed commit: ["SY-4642: Roll the range slice back when ..."](https://github.com/synnaxlabs/synnax/commit/9f6b0dd57d9824613b4235aed861088759fdc3cf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55019420)</sub>


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

console/src/session/range/useDelete.ts
```diff
@@ -0,0 +1,52 @@
...
+import { useGetSliceState } from "@/session/range/selectors";
+import { remove, restore } from "@/session/range/slice";
+
+/**
+ * Deletes a range, dropping the copy in the session slice and restoring it, along with
+ * the selection it carried, when the server delete fails. A caller's own beforeUpdate
+ * runs first and can still cancel the delete.
+ */
+export const useDelete: Flux.UseUpdate<Ranger.DeleteParams> = (params) => {
+  const { beforeUpdate } = params ?? {};
```
emilbon99: Nit: should default this in the argument instead of here.

console/src/feature/range/Toolbar.tsx
```diff
@@ -133,12 +112,18 @@ const listItem = Component.renderProp((props: BaseList.ItemProps<string>) => {
...
+        align="center"
+        gap="small"
+        className={CSS.BE("range-list-item", "name")}
+      >
         <Ranger.StageIcon timeRange={timeRange} />
         <Text.MaybeEditable
           id={`text-${key}`}
           level="p"
           value={name}
+          overflow="ellipsis"
```
emilbon99: I've started defaulting to using overflow="fade" on items like these as I think it looks nicer. Should consider changing.

emilbon99 (changes requested):
LGTM after comments addressed.

PR #2761, opened 2026-08-20
SY-4658: Gate mutating Console actions on access checks

pjdotson:
## Linear Issue

[SY-4658](https://linear.app/synnax/issue/SY-4658/gate-mutating-console-actions-on-access-checks)

## Description

An audit of every mutating surface in the Console — context menus, create buttons, task commands, tree drag-moves, and palette commands — found actions offered without checking the matching grant. Each one only failed once the Core rejected the write.

Fixed:

- **Workspace views**: the delete item rendered whenever a non-static view was selected, with no delete check.
- **Schematic symbols and symbol groups**: both context menus deleted server resources with no delete check.
- **Child ranges**: the add button in the range overview had no create check, while the same action in the range tree and explorer menus is gated.
- **Line plot "Create range from selection"**: creates a real range, ungated.
- **Labels**: the per-label delete button and both "New label" buttons — the modal's and the one in every label picker footer — had no checks at all. The palette command that opens the modal gates on update, so a user with update but neither create nor delete saw buttons that could not work. The picker withholds the whole `footer` prop rather than rendering an empty one, because the select moves its border onto a wrapper as soon as it is handed a footer at all.

Checked and left alone: line plot annotations sit behind the plot's own update gate; the Core list is local connection state with no server resource; device connect modals are reachable only through already-gated menu items; tree drag-moves check update on both the row and the drop predicate; and "group from selection" correctly checks update on the moved resources plus create on the group.

## Testing

Every gate gets a granted and a denied spec against a real Core. The denied spec
first waits on a control the same subject *is* granted, so the absence is the gate
and not a check still in flight:

- **Labels** (`platform/label/useEditModal.spec.ts`): the create and delete gates
  control each other. A create-only subject sees "New label" but no per-label
  delete; a delete-only subject sees the reverse.
- **Label picker** (`platform/label/Select.spec.tsx`): the footer create button.
- **Views** (`platform/view/view.spec.tsx`): the context menu delete item.
- **Symbols** (`feature/schematic/toolbar/Symbols.spec.ts`): the symbol and
  symbol-group delete items.
- **Child ranges** (`feature/range/overview/Overview.spec.tsx`): the add button.

Each gate was verified by mutation: deleting it, or swapping the action it checks,
turns the matching spec red.

Two supporting changes: the label spec's synchronous `getAddButton()` became an
awaited `findAddButton()`, since a gated button renders only after its query
resolves; and `renderOverview` now primes the panel and renders suspended, which a
[...]

PR #2762, opened 2026-08-20
SY-4659: Apply optimistic writes to remaining client mutations

pjdotson:
## Linear Issue

[SY-4659](https://linear.app/synnax/issue/SY-4659/apply-optimistic-writes-to-remaining-client-mutations)

## Description

Several client methods updated their cache only after the Core answered, so the UI lagged a round trip and had no rollback when the write failed. An audit of every `createUpdate` across all `queries.ts` files, plus every client method that touches a store, found the remaining gaps.

Made optimistic in `client/ts`:

<details><summary></summary>

* `ranges.delete` — range rows lingered in the tree until the Core answered. It now also clears ontology relationships, which every comparable delete already did.
* `ranges.setAlias` / `deleteAlias` — the most visible of the set, since aliases are the name shown on channel rows. The channel store reads from the alias store, so the optimistic write genuinely reaches the rows.
* `ranges.kv.set` / `kv.delete` — range metadata.
* `schematics.symbols.deleteGroup` — dropped the group, its member symbols, and both relationship directions only after the round trip.
* `statuses.set` — the key was already generated client-side during request encoding, so normalizing through `statusZ` up front gives every status an identity to write against. That is the same schema `setReqZ` already applies, so validation behavior is unchanged.
* `devices.create` — device keys come from the caller, so a create is optimistic like any other write. `WriteOptions` is a new trailing parameter on the schema-bearing overloads, so no existing call site changes. Normalizing through `deviceZ` applies the schema `createReqZ` already carries.

</details>

`panel.rename` was already optimistic but accepted no options, so `Panel.useRename` and `Schematic.useRename` fired `onOptimisticComplete` before the client touched the cache and an `afterOptimistic` handler read the old name. It now takes `WriteOptions` like every other rename, and both hooks call through it.

Query wiring: `Group.useDelete`, `Label.useDelete`, and `Status.useDelete` never passed `onOptimistic` even though their client methods already accepted `WriteOptions`. `Schematic.Symbol.useDeleteGroup` is new — the Console previously called the client directly from a bespoke hook, so it received no optimism, rollback, or Flux status handling. `EtherCAT.Device.useToggleEnabled` returned its request parameters rather than the updated devices, and is now optimistic end to end.

Correctly left alone: creates and copies that assign identity on the Core; `arcs.setRack` returns a Core-built task and the arc payload carries no rack; `tasks.useCommand` sends a command rather than writing state; and Flux forms exclude `onOptimisticComplete` by design, since form state already gives immediate feedback.

## Cache fix

Writing the new optimistic paths surfaced a defect in the cache that predates them. A derived table kept a corpse for an entry whose source tombstoned nothing, so a create whose commit failed read back as a deleted record carrying a value that never existed, and a subscribed `retrieve` rejected with "was deleted" instead of refetching. `Table.evict` now removes an entry without keeping a corpse, `bindDerived` mirrors the source's tombstone, and an exact-key answer returns to unfetched when the table kept none — the same presence test rule 2 already applied.

## Tests

Every changed path is covered. Client specs assert the cache state observed from inside `onOptimistic` and the rollback that follows a failed write, for range deletes, aliases, metadata pairs, symbol-group deletes, status sets, device creates, and panel renames. Pluto specs assert that `afterOptimistic` runs before each mutation commits, and that the rename hooks see the renamed record, and cover the new `Symbol.useDeleteGroup` along with `Ranger.useDelete`, `useUpdateKV`, and `useDeleteKV`, which had no specs at all. The cache fix carries its own cases in `table.spec.ts`, `derived.spec.ts`, and `query.spec.ts`.

## Note for reviewers

Split from one branch alongside synnaxlabs/synnax#2760 and synnaxlabs/synnax#2761. synnaxlabs/synnax#2761 has landed on `rc` and this branch carries the merge. synnaxlabs/synnax#2760 is still open and overlaps this one on `pluto/src/schematic/symbol/queries.ts` in a separate region, so a rebase should be clean.

## Basic Readiness

### Greptile Summary

The PR adds optimistic cache updates, rollback behavior, and Flux callback wiring for the remaining TypeScript client mutations.

* Adds optimistic writes for devices, ranges, aliases, range metadata, symbol groups, statuses, and panel renames.
* Corrects cache eviction and derived-table tombstone handling.
* Adds client and Pluto tests for optimistic state, callback order, and rollback behavior.

### Confidence Score: 5/5

The PR appears safe to merge.

No blocking failure remains.

### Important Files Changed

| Filename | Overview |
| -- | -- |
| client/ts/src/status/client.ts | Adds normalized optimistic status writes and rollback support, with focused unit coverage. |
| client/ts/src/query/table.ts | Adds eviction without tombstones so failed optimistic creates return cache entries to an unknown state. |
| client/ts/src/query/query.ts | Returns exact-key queries to the unfetched state when their underlying entry is evicted. |
| client/ts/src/query/derived.ts | Makes derived tables mirror whether source deletions retain tombstones. |
| client/ts/src/ranger/client.ts | Adds optimistic range deletion and alias updates with rollback behavior. |
| client/ts/src/schematic/symbol/client.ts | Adds optimistic symbol-group deletion across records and ontology relationships. |
| console/src/feature/ethercat/device/queries.ts | Connects device toggles to optimistic client writes and returns updated devices. |
| pluto/src/status/queries.ts | Wires Flux optimistic callbacks into status set and delete operations. |

### Sequence Diagram

```mermaid
sequenceDiagram
  participant UI as Flux hook
  participant Client as TypeScript client
  participant Cache as Local cache
  participant Core as Core
  UI->>Client: Start mutation
  Client->>Cache: Apply optimistic update
  Client-->>UI: Run optimistic callback
  Client->>Core: Commit mutation
  alt Commit succeeds
    Core-->>Client: Return committed value
    Client->>Cache: Apply committed value
    Client-->>UI: Report success
  else Commit fails
    Core-->>Client: Return error
    Client->>Cache: Roll back update
    Client-->>UI: Report failure
  end
```

Reviews (2): Last reviewed commit: ["rf"](<https://github.com/synnaxlabs/synnax/commit/014a5b9cfe80222de0c8d803a89c12a096a32418>) | [Re-trigger Greptile](<https://app.greptile.com/api/retrigger?id=55019516>)


pjdotson:
@greptile /rerun

pluto/src/status/queries.ts
```diff
@@ -46,9 +48,16 @@ export interface SetParams {
...
-    if (Array.isArray(statuses)) await client.statuses.set(statuses, { parent });
-    else await client.statuses.set(statuses, { parent });
+  update: async ({
+    client,
+    data,
+    data: { statuses, parent },
+    onOptimisticComplete,
+  }) => {
+    const onOptimistic = async () => await onOptimisticComplete(data);
+    if (Array.isArray(statuses))
```
emilbon99: A little bit odd that we need this conditional no?

emilbon99 (changes requested):

pluto/src/status/queries.ts
emilbon99: Should we modify the `set` overloads?

PR #2763, opened 2026-08-20
SY-4619: Require routing entries to be full flow statements

sy-nico:
## Linear Issue

[SY-4619](https://linear.app/synnax/issue/SY-4619)

## Description

Removes an ambiguity in output routing tables. `select{} -> { false: next }` read like the routed value flowed into the entry, but both select outputs are u8 pulses, so what actually ran was `1 -> next`. What fed the entry depended on the routing function's output types, invisibly at the call site.

The fix: everything to the right of a routing key must be a full flow statement. A routing key only gates its entry; no value flows from the router into it, and the entry defines its own source (`true: true => next`, `high: "high" -> log`). Bare targets are now compile errors. Inline `stage {}`/`sequence {}` bodies are the only bare exception, and param-mapped entries (`high: chan: a`) are unchanged.

Output types stay unconstrained on purpose rather than being forced to `bool`:

- Enforcing `bool` outputs would be a lot of extra plumbing for no behavior change.
- The root issue was ambiguity about what runs, and the strictness rule resolves it.
- It keeps the door open for the originally intended feature: a future, more thought-out syntax can still pass the routed value along.
- Truthiness is already a documented concept, so any-typed outputs behave intuitively.

Behavior change: the old passthrough form forwarded per-sample values; an entry's own constant fires once per trigger batch. Migration is loud since the old form no longer compiles.

Also fixes a gap this exposed: the WASM named-output store was missing the `bool` case, so a `bool` named output failed to compile. New runtime specs read every output type back and cover the remaining store widths.

Docs (`flow.mdx`, `functions.mdx`, `spec.md`, hover docs) are reframed around flow routing and decision making instead of output types and values.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR removes output-routing ambiguity by requiring each routing entry to define a complete flow statement, while retaining inline stage and sequence bodies and parameter-mapped entries as supported forms. It also completes boolean named-output storage and updates the Arc documentation and tests to describe the new gating semantics.
- Output routing keys now gate independently sourced entry flows instead of forwarding routed values.
- Bare routing targets are rejected during flow analysis.
- Boolean named outputs use byte-width WASM stores consistent with their memory representation.
- Runtime, analyzer, integration, hover, specification, and reference documentation coverage is updated.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| arc/go/analyzer/flow/flow.go | Enforces full routing statements, analyzes entry heads without routed input, and preserves type checking for parameter-mapped entries. |
| arc/go/compiler/statement/variable.go | Adds boolean named outputs to the one-byte WASM store path, matching boolean allocation density. |
| arc/go/routing_table_test.go | Exercises explicit branch values, mixed named-output widths, falsy routed outputs, inline bodies, transitions, and batch-level routing behavior. |
| arc/go/text/text_test.go | Updates expected IR topology and stratification so routing outputs gate entry-owned source nodes. |
| docs/site/src/pages/reference/control/arc/reference/flow.mdx | Reframes output routing around branch gating and complete entry flows rather than routed-value forwarding. |
| arc/docs/spec.md | Documents full-statement routing entries, inline-body exceptions, and the absence of routed-value passthrough. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
    A[Upstream value] --> B[Function with named outputs]
    B -->|Output fires| C{Routing key}
    C -->|Gates| D[Entry-owned source]
    D --> E[Entry flow target]
    C -. No routed value forwarded .-> D
```

<sub>Reviews (2): Last reviewed commit: ["SY-4619: Updated integration test"](https://github.com/synnaxlabs/synnax/commit/a3218760ecc5d83a587a7c24a70648cc15ba88b5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55033091)</sub>

### DIFF
--- a/arc/cpp/runtime/wasm/node_test.cpp
+++ b/arc/cpp/runtime/wasm/node_test.cpp
@@ -1866,8 +1866,8 @@ func labeler(x i64) (label str, value i64) {
     value = x * 2
 }
 )arc" + input_name +
-        " -> labeler{} -> {label: " + label_out_name + ", value: " + value_out_name +
-        "}";
+        " -> labeler{} -> {label: \"ok\" -> " + label_out_name + ", value: 0 -> " +
+        value_out_name + "}";
 
     auto str_st = std::make_shared<stl::strings::State>();
     auto series_st = std::make_shared<stl::series::State>();
--- a/arc/cpp/stl/selector/selector.h
+++ b/arc/cpp/stl/selector/selector.h
@@ -32,12 +32,11 @@ inline const std::string false_param = "false";
 inline constexpr size_t TRUE_OUTPUT_IDX = 0;
 inline constexpr size_t FALSE_OUTPUT_IDX = 1;
 
-/// @brief Select routes a boolean input to its "true" or "false" outputs:
-/// true values produce samples on the "true" output; false values produce
-/// samples on the "false" output. Both outputs emit a value of 1 for each
-/// matched sample — the output value signals "this route fired" rather than
-/// echoing the input — so the output can drive a truthy-gated transition
-/// whether the matched route is true or false.
+/// @brief Select gates the "true" or "false" branch of a routing table: true
+/// input samples fire the true output, false samples fire the false output.
+/// Each fired output emits 1 so truthy-gated transitions fire on either
+/// branch. The value never reaches an entry; entries define their own
+/// sources.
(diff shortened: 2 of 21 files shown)


sy-nico:
@greptile, rerun. Updated documentation.

PR #2774, opened 2026-08-20
SY-4680: Minor UI tweaks

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4680](https://linear.app/synnax/issue/SY-4680/minor-ui-tweaks)

## Description

A group of small interaction and appearance fixes across the documentation site, Pluto,
and the Console.

**Documentation site navigation.** The selected page lost its marker. The RFC 0050
redesign moved the Pluto list item selected state onto background tokens, and the page
navigation sets `background: none` on tree items, so nothing was left to see. The blue
accent bar is restored in the site's own stylesheet.

**List rows that hold controls.** The line plot lines tab and the log channels list are
rows of form controls with nothing to select, yet they carried hover and press states.
They now pass `preventClick`, which already gates those states in `Button.css`, and the
cursor follows it. The log toolbar's hand-rolled version of the same effect is removed.

**Button focus.** `preventClick` sets `tabIndex` to -1, and the button then cancelled the
press default for every press that reached it, not only presses on the chassis. Focus
never landed on a nested input, so the alias field in a log channel row could not be
clicked. The cancel now applies only when the chassis itself is what the browser would
focus. This also repairs the range metadata rows, which had the same latent break.

**Log viewport.** It is out of the tab order and no longer paints a focus ring, matching
the diagram. Only the diagram was ever patched for this.

**Visualization controls.** A double click on the control cluster no longer reaches the
visualization behind it, where it opens the toolbar. The line plot cluster now sits
where the log's and the schematic's do: its wrapper padding already supplies the inset
the others get from the controls' own offset, so the offset is zeroed and the range
annotation special case disappears. The cluster stays visible for keyboard focus only,
because a click leaves focus on the button and the line plot, unlike the other two, has
no focusable root to take it back.

Tests cover the two behavioral changes: the Button press-default guard and the double
click containment. The rest is CSS, and jsdom applies no stylesheets, so an assertion
there would pass with the rules deleted.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR applies minor interaction and appearance refinements across the documentation site, Pluto, and Console.
- Restores selected-page navigation styling and adjusts visualization control placement and focus behavior.
- Makes control-bearing list rows inert without blocking their nested inputs.
- Restores the post-open controller guard while preserving user-requested acquisition during disabled edit state.

The PR appears safe to merge.

The previously reported disabled-acquisition race is resolved by rechecking the disabled state after the writer opens, while the explicit-acquisition exception is intentional and covered by the overlap test; no blocking failure remains.

| Filename | Overview |
|----------|----------|
| pluto/src/telem/control/aether/controller.ts | Restores the post-open disabled-state check and distinguishes explicit acquisition so automatic writes stand down without discarding user-requested control. |
| pluto/src/telem/control/aether/controller.spec.ts | Covers delayed automatic acquisition, explicit acquisition while disabled, and the shared-promise overlap ordering. |
| pluto/src/button/Button.tsx | Limits press-default cancellation to the button chassis so nested focusable controls can receive focus. |
| console/src/platform/vis/Controls.tsx | Contains double-clicks within visualization control clusters while allowing callers to override the handler. |

<sub>Reviews (2): Last reviewed commit: ["SY-4680: Resize the scale preview and cl..."](https://github.com/synnaxlabs/synnax/commit/0037c91d676bbcbcf0f8152a5695b8da03911e8b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55350219)</sub>

### DIFF
--- a/console/src/feature/http/task/ContextMenu.tsx
+++ b/console/src/feature/http/task/ContextMenu.tsx
@@ -14,7 +14,7 @@ import { Task } from "@/platform/task";
 
 export interface ContextMenuProps {
   keys: string[];
-  onDelete: (keys: string[]) => void;
+  onRemove: (keys: string[]) => void;
   onDuplicate?: (keys: string[]) => void;
   onRename?: (key: string) => void;
 }
@@ -22,7 +22,7 @@ export interface ContextMenuProps {
 export const ContextMenu = ({
   keys,
   onDuplicate,
-  onDelete,
+  onRemove,
   onRename,
 }: ContextMenuProps) => {
   const isPreview = Task.useIsPreview();
@@ -41,7 +41,7 @@ export const ContextMenu = ({
             </Menu.Item>
           )}
           <Menu.Divider />
-          <Base.DeleteItem onClick={() => onDelete(keys)} />
+          <Base.RemoveItem onClick={() => onRemove(keys)} />
           <Menu.Divider />
         </>
       )}
--- a/console/src/feature/http/task/Form.css
+++ b/console/src/feature/http/task/Form.css
@@ -11,8 +11,15 @@
 
 .console-task-configure--http_read,
 .console-task-configure--http_write {
+    /* Both halves are load-bearing: the list must refuse to shrink, and the detail
+       column must be allowed to, or the list absorbs everything it will not give up. */
     .console-endpoint-list {
         width: 250px;
+        flex-shrink: 0;
+    }
+
+    .console-endpoint-details-pane {
+        min-width: 0;
     }
 
     .console-endpoint-details {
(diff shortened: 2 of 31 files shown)


emilbon99:
@greptile /rerun

*Emiliano + Claude: drafted in our review session.*

The P1 on `controller.ts` is fixed in 6b2639a. The post-open check is back, but it now
lives in `doAcquireImpl` where the writer becomes non-null, so it covers the retry
branch as well as the entry path.

It reads a latch set by `acquire()` rather than `disabled` alone. `doAcquire` shares one
promise across callers, so the flag on its own cannot say who asked, and a bare re-check
tears down a writer the user explicitly asked for.

Two tests: one for the stand-down, one pinning that an explicit acquire while disabled
keeps its writer.

PR #2799, opened 2026-08-22
SY-4635: Rework Console session state

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4635](https://linear.app/synnax/issue/SY-4635)

## Description

An audit of `console/src/session/` against what 0.56 shipped turned up a set of
problems in how the Console stores its session state. This fixes them.

**Cores are keyed by `host:port`.** The key used to be generated, then replaced with the
Core's own key on first connect. Every partition written under the old key was stranded
by that rename, and the swap it triggered dropped the user back to project selection.
One dev machine's store held 40 orphaned Core partitions and 80 orphaned project
partitions. An address never changes underneath a running session, so the rename, its
synchronizer, and the `LOCAL`/`DEMO` literal keys are all gone. Two entries at one
address now collapse into one.

**Stored slices are parsed on read.** Only `status` ever declared a migrator, so
everything else was cast straight out of the KV into the reducers. A live store had
`"port": "9090"` — a string where the schema says `z.number()` — sitting on disk
unnoticed. Persistence scopes now carry each slice's schema, a slice that fails it falls
back to its initial state, and a slice in no scope fails at `Persist.open` rather than
silently not persisting. The per-slice migrator hook is deleted; schema evolution is
Zod's job. Every persisted slice, `drift` included, carries a `version`.

**A window is a viewport.** Panels live on the Core and any number of windows may show
one at once, but arc, line plot, log, schematic, and table view state was shared across
them, so panning in one window moved the other. Each now nests its documents under a
window key, as `nav` and `panels` already did. Window keys are minted fresh per open and
nothing ever dropped them, so a close now emits `window/removed` and each window-keyed
slice drops its entry.

**The store moved.** The browser backend was `localStorage`, whose few-megabyte cap a
heavy workspace outgrows with twelve state slots in play, and whose quota errors surface
only as a failed write. Desktop now writes `session.json`; the browser uses IndexedDB.

**0.56 state is carried forward.** Nothing read the old store, so upgrading lost the
Core registry — hosts and credentials — along with theme and recent colors. A one-shot
seed carries those plus the workspace selection out of `persisted-state.json` on first
open. That file is only ever read: nothing writes or clears it, so a rollback to 0.56
still finds its state. Layouts and per-document view state are deliberately left behind,
since the Core migrates workspaces into projects and panels.

Also renames cluster to Core throughout the Console, deletes the docs slice and its
in-app tab in favour of opening the browser, drops two empty `PERSIST_EXCLUDE` arrays,
and rewrites the stale persistence and layout sections of `console/CLAUDE.md`.

### Worth a reviewer's attention

- Deep links are now `synnax://core/<host:port>`. Existing links were already dead once
  the key format changed. The tradeoff: two people reaching one Core by different
  addresses can no longer resolve each other's links, where a shared cluster key used to
  make that work.
- The desktop data directory did not move. `appLocalDataDir` is the conventional home
  for this, but moving it would orphan the legacy file the seed has to read, so it wants
  its own migration later.
- Two entries resolving to the same address merge silently, last write winning.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR reworks Console session persistence, Core identity, and window-scoped view state.
- Keys Core records by address and updates deep links.
- Validates persisted slices with Zod and migrates selected 0.56 state.
- Moves browser persistence to IndexedDB and desktop persistence to `session.json`.
- Separates document view state by window and removes it only after Drift removes the window.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| console/src/session/window/removed.ts | Detects actual Drift window removal after close or process completion, so deferred closes retain live session state. |
| console/src/session/window/removed.spec.ts | Covers immediate close, deferred close, completed deferred close, main-window retention, and sibling-window isolation. |
| drift/src/state.ts | Defines the window lifecycle transitions used by the removal middleware. |
| console/src/session/persist/state.ts | Reworks partitioned persistence and schema-based restoration across session scopes. |
| console/src/session/store.ts | Integrates the new persistence scopes, legacy migration, and window-removal middleware. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
    Close[Request window close] --> Check{Active process?}
    Check -->|Yes| Deferred[Keep window and session state]
    Check -->|No| Removed[Drift removes window]
    Deferred --> Complete[Process completes]
    Complete --> Removed
    Removed --> Cleanup[Dispatch window/removed]
    Cleanup --> Slices[Remove window-keyed slice state]
```

<sub>Reviews (4): Last reviewed commit: ["refactored slice architecture"](https://github.com/synnaxlabs/synnax/commit/8de9a8a89e3414b5cf64f31197f6cd060bd4c22e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55994670)</sub>


emilbon99:
@greptile-apps review

console/src/session/store.ts
```diff
@@ -203,8 +207,8 @@ export const createStore = async (opts: CreateStoreOptions = {}): Promise<Store>
       initial: ZERO_STATE,
       scopes: PERSIST_SCOPES,
       getContext: getPersistContext,
-      migrators: PERSIST_MIGRATORS,
       exclude: PERSIST_EXCLUDE,
+      seed: Legacy.migrate,
```
emilbon99: I hate the word seed. Claude overuses it. Isn't migrate a better word here?

console/src/session/legacy/migrate.ts
```diff
@@ -0,0 +1,118 @@
...
+import { openReader, type Reader, readState } from "@/session/legacy/read";
+import { Project } from "@/session/project";
+import { Theme } from "@/session/theme";
+
+/**
+ * What the 0.56 blob held that this release still has a home for. Every branch is
+ * optional so a blob from any point in that release line parses, and anything that
+ * fails validation is left behind rather than failing the whole carry-over.
+ */
+const stateZ = z.object({
```
emilbon99: Should probably update this comment to explicitly state what versions this will migrate from correctly.

console/src/session/core/slice.ts
```diff
@@ -12,56 +12,67 @@ import { synnaxParamsZ } from "@synnaxlabs/client";
...
   name: "Demo",
   host: "demo.synnaxlabs.com",
   port: 9090,
   username: "synnax",
   password: "seldon",
   secure: true,
-};
+});
 
 export const sliceStateZ = z.object({
```
emilbon99, line 67: Why don't we instead shape this like an array instead? Instead of storing the host and port twice we only store it once. Then all lookups can be queried by host and port or by key? Also if the host and port change we don't have to do weird reconciliation aside from checking for duplicates?

emilbon99 (changes requested):
I'm a bit concerned with the core rekeying. Isn't one of the hard won lessons we've learned over time that we shouldn't be keying items by non-stable identities? I feel like every time we've made a decision to do that (racks, tasks, etc) it has completely backfired on us.

I'm also concerned about storing the session state by core host:port. The session state is designed to reflect document state, which lives on the core. The most reliable check of "did the cores document state change so radically that we need to reconcile session state" is whether its *key* changed, not its address. For example, we commonly restart cores in memory, and their key changes, so they no longer hold all of the same documents.

What yield to do you actually get from doing this that overcomes the complexities of needing to reconcile it all the time?

pjdotson:
> the hard won lessons we've learned over time that we shouldn't be keying items by non-stable identities? I feel like every time we've made a decision to do that (racks, tasks, etc) it has completely backfired on us.
> 
> I'm also concerned about storing the session state by core host:port. The session state is designed to reflect document state, which lives on the core. The most reliable check of "did the cores document state change so radically that we need to reconcile session state" is whether its _key_ changed, not

I changed this so that Cores have their own keys stored on

I didn't want to do this based off of cluster key because eventually when you have multi-node cluster support, you will want to store connection info for two Cores of the same cluster. So the Core key is now local and has cluster key as a field on it.

Agreed that going on host:port was a bad idea.

pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

emilbon99:
@greptile/rerun

emilbon99:
@greptile rerun

PR #2812, opened 2026-08-25
SY-4730: Support horizontal scale bars

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4730](https://linear.app/synnax/issue/SY-4730)

## Description

The scale symbol can now run horizontally. The grid rotate button toggles the bar between the two axes and swaps its dimensions along with the sides its ticks and readout sit on. The readout side is independent of the tick side, and a readout side that runs along the bar places the readout inside the bar beside the fill edge.

### Why extend the symbol instead of adding a horizontal one

A second symbol copies the whole scale config: telemetry, bounds, notation, precision, units, staleness, colors, and text level. Every field added later has to be added twice and the two drift. Rotation is also how the rest of the schematic works, so a scale rotates the same way a valve does.

### Why `orientation` instead of a new config field

`orientation` already exists on every symbol and is what the rotate button writes. A separate axis field would leave two fields describing one thing, with the button writing one and the renderer reading the other.

### Dimensions

The configured width and height are the bar's own. The symbol adds the 26px tick gutter on top, so a 0 width bar still occupies the gutter and the bar never mirrors onto the scale. `DEFAULT_DIMENSIONS` drops to 34x160 so a new scale keeps the footprint it had.

### Migration

Scales written before this stored `orientation: "left"` and sized their dimensions as the bar plus the gutter. `v8.ScaleMigration` restates both: the orientation becomes `"top"`, and the gutter comes off the stored width, so a 60x160 scale stays 60x160 on screen. A scale drawn without ticks reserved no gutter and keeps its width. `axis()` in the symbol reads any non-canonical orientation as vertical for the same reason the migration exists, since a 0.57 Console writing into a migrated cluster stores `"left"` again.

### `caretSide`

It defaults to `"right"` rather than falling back to `side`, so the control is a plain `Base.Field` and nothing is shown that is not stored. A stored scale with ticks on the left gets its readout on the right instead of following the ticks.

### `location.swapAxis`

Rotation carries each edge to the one facing the same way on the other axis. It lives in `x` because it is general geometry, and it is its own inverse, so rotating back restores the original sides.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR extends scale symbols to render horizontally while preserving existing scale footprints through bar-only dimensions and a one-time compatibility migration.
- Adds horizontal rendering, independent tick/readout sides, and axis-aware rotation and resizing.
- Normalizes stored and imported legacy scale geometry before the renderer adds its tick gutter.
- Adds migration, rendering, configuration, and geometry utility coverage.

The PR appears safe to merge.

No blocking failure remains; the migration and renderer now consistently convert legacy gutter-inclusive widths into bar-only dimensions and reconstruct the original footprint at render time.

| Filename | Overview |
|----------|----------|
| core/pkg/service/schematic/versions/v8/migrate.go | Adds an idempotent migration that normalizes legacy scale orientation and converts visible-tick scales to bar-only width. |
| core/pkg/service/schematic/versions/imex.go | Applies the same legacy scale normalization to imported schematics before creating a new resource. |
| pluto/src/schematic/node/general/scale/Symbol.tsx | Adds axis-aware rendering, rotation, and resize conversion while preserving the distinction between stored bar size and rendered outer size. |
| pluto/src/vis/scale/aether/scale.ts | Generalizes fill, tick, gutter, caret, and readout geometry across horizontal and vertical scales. |
| pluto/src/schematic/node/general/scale/config.ts | Defines bar-only defaults and documents the current orientation-to-axis compatibility contract. |
| x/ts/src/spatial/location/location.ts | Adds an involutive edge mapping used to preserve side semantics when rotating between axes. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  A[Legacy stored or imported scale] --> B[Normalize orientation to top]
  B --> C[Remove visible tick gutter from stored width]
  C --> D[Persist bar-only dimensions]
  D --> E[Scale renderer determines axis]
  E --> F[Add gutter to rendered outer dimensions]
  F --> G[Preserve original schematic footprint]
  G --> H[Rotation swaps bar dimensions and sides]
```

<sub>Reviews (2): Last reviewed commit: ["SY-4730: Migrate stored scales to bar-on..."](https://github.com/synnaxlabs/synnax/commit/19c41cc747901421659f72eeae59a83e11ad3d0f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56526321)</sub>


pluto/src/schematic/node/general/scale/Symbol.tsx
```diff
@@ -22,24 +26,56 @@ export const Symbol = ({
...
+  },
 }: NodeProps<Config>): ReactElement => {
+  const dir = axis(orientation);
+  // The configured dimensions are the bar's own. The ticks live beside it, so the
+  // symbol takes the gutter on top of them.
+  const gutter = BaseScale.gutter(indicator);
+  const outer: dimensions.Dimensions =
+    dir === "y"
+      ? { width: dims.width + gutter, height: dims.height }
+      : { width: dims.width, height: dims.height + gutter };
```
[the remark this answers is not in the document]
sy-nico: Fixed by `v8.ScaleMigration`. It takes the 26px gutter off the stored width when it restates the orientation, so a 60x160 scale migrates to a 34x160 bar and renders at 60x160, the footprint it had. A scale stored with `showScale: false` reserved no gutter, so its width is left alone. Covered in `versions/v8/migrate_test.go`.

sy-nico:
@greptile, rerun

core/pkg/service/schematic/versions/imex.go
```diff
@@ -71,6 +71,8 @@ func DecodeImExEnvelope(ctx context.Context, env imex.Envelope) (Schematic, erro
 	if err != nil {
 		return Schematic{}, err
 	}
+	// An envelope carries the same stale scale geometry the storage migration rewrites.
+	v8.NormalizeScales(sch)
```
pjdotson: shouldn't `v8.MigrateSchematic` apply the scale normalization?

core/pkg/service/schematic/versions/v8/migrate.go
```diff
@@ -29,3 +31,62 @@ func MigrateNode(ctx context.Context, old v7.Node) (Node, error) {
...
+	verticalScale    = "top"
+	horizontalScale  = "right"
+	scaleGutter      = 26
+	variantField     = "variant"
+	orientationField = "orientation"
+	dimensionsField  = "dimensions"
+	indicatorField   = "indicator"
+	showScaleField   = "showScale"
+	widthField       = "width"
+)
```
pjdotson, line 41: personally I don't think constants like `showScaleField` that are only used in one place are helpful, the variable name adds no meaning so it just makes the code harder to read.

pjdotson (changes requested):
1) should have options in the orientation selector for the scale orientation:

Image

core/pkg/service/schematic/versions/imex.go
sy-nico: I decided to remove this. It's only needed for the narrow case where you export a schematic from v0.57 and import that JSON into v0.58. I was hitting it because I was editing offline and uploading to the Runner Core.

Without it, those imported scales come in 26px wider, and resizing them fixes it. A normal upgrade is unaffected, since stored schematics are repaired by `ScaleMigration`.

core/pkg/service/schematic/versions/v8/migrate.go
sy-nico: Fixed. Dropped the field-name constants and kept only `verticalScale`, `horizontalScale`, and `scaleGutter`, where the name adds meaning.

sy-nico:
@pjdotson,
I started implementing the vertical scale tick labels, but every symbol already repaints in full on each animation frame, since canvas 2D has no retained layers. Turning a tick label adds a transform and an unwind per label on top of that, and at 100+ scales it roughly doubles the work on the tick path.

So we would need to cache the static parts of a symbol on their own canvas first, painting outlines and tick marks only when the config changes. That helps every symbol, not just scales, and makes the labels cost nothing per frame.

Filed [SY-4814](https://linear.app/synnax/issue/SY-4814/cache-static-symbol-drawing-then-support-vertical-scale-tick-labels) covering both.

PR #2727, opened 2026-08-16
SY-4607: Never let streamed telemetry claim historical read coverage

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4607](https://linear.app/synnax/issue/SY-4607)

## Description

The same sample carries a provisional leading alignment when streamed and a committed alignment when fetched, and the line renderer pairs x and y series by alignment overlap. Serving a span from streamed data while the paired channel held fetched data made the pair undrawable: missing backloads, permanent gaps, zero-floored auto bounds, and dead tooltips at the seam.

Reads now fetch every span not covered by fetched entries, converging both channels to committed alignments. A span may briefly return in both forms.

Also: the TPC simulator seeded every sensor with 0.0 one sample before the loop's first real write, which spiked plot auto bounds on task start.

## Basic Readiness

- [ ] I have performed a self-review of my code.
- [ ] I have added relevant, automated tests to cover the changes.
- [ ] I have updated documentation to reflect the changes.

<h3>Greptile Summary</h3>

The PR prevents streamed telemetry from claiming historical cache coverage so reads converge on committed alignments, while allowing both representations briefly.
- Separates fetched and streamed cache entries and derives gaps only from fetched data.
- Corrects Cesium auto-span sample boundaries and backward-read alignments.
- Seeds the TPC simulator with realistic initial sensor values.
- Adds cache, feed, iterator, alignment, rollover, and garbage-collection coverage.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge.

No blocking failure remains.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| client/ts/src/framer/cache/static.ts | Separates fetched and streamed entries, computes historical coverage exclusively from fetched data, and garbage-collects both lists. |
| client/ts/src/framer/cache/unary.ts | Preserves overlapping live-buffer data without allowing it to suppress historical gap fetching. |
| cesium/internal/unary/iterator.go | Corrects auto-span boundary selection, domain transitions, sample offsets, and backward alignment calculation. |
| client/py/examples/simulators/tpc.py | Reuses a complete initial-state map to seed sensors consistently with the simulation loop. |
| client/ts/src/framer/feed.spec.ts | Adds end-to-end coverage for alignment-compatible backloads and reads across leading-buffer rollovers. |

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Stream as Live stream
  participant Cache as Telemetry cache
  participant Reader as Historical reader
[...]

PR #2720, opened 2026-08-16
SY-4600: Fix permanent rolling plot gaps after a stream stops

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4600](https://linear.app/synnax/issue/SY-4600/rolling-plots-show-permanent-gaps-after-a-channels-stream-stops-and)

## Description

Rolling line plots showed permanent, double-digit-second holes between the historical fetch and the live stream whenever a channel had been streamed earlier in the session and then stopped. The cache's live leading buffer claims coverage of all time after its start, nothing ended that claim when streaming demand stopped, and `Unary.read` suppressed every gap behind it, so the reader never fetched the unstreamed span from the Core.

The `MultiplexedStreamer` now flushes a channel's leading buffer into the static cache whenever stream continuity breaks: the key leaves the stream on reconcile, the whole stream is released, the connection drops, the run loop dies, or a late frame arrives for a key with no remaining demand. The flush stamps the buffer's real data end (from the SY-4552 streamed-series stamping), so later reads see the true gap and refetch it.

One behavior change: a value consumer that re-subscribes to a channel no longer shows the last cached value instantly. It shows nothing until the first live sample arrives, instead of a potentially stale value.

Covered by a live-core regression test in `feed.spec.ts` (stream, unsubscribe, write, read misses nothing) plus unit tests for each flush path in `streamer.spec.ts` and `unary.spec.ts`.

## Basic Readiness

- [x] I have performed a self-review of my code.
- [x] I have added relevant, automated tests to cover the changes.
- [ ] I have updated documentation to reflect the changes.

<h3>Greptile Summary</h3>

The PR ends stale live-cache coverage when stream continuity stops and coordinates concurrent reconnect attempts.
- Flushes dynamic channel buffers into the static cache after demand removal, connection loss, run-loop termination, and late frames.
- Adds regression coverage for cache gaps and stream lifecycle paths.
- Uses one shared reconnect promise to prevent competing replacement streams.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge because no eligible blocking failure remains.

No blocking failure remains.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| client/ts/src/framer/cache/streamer.ts | Flushes leading buffers when stream continuity or demand ends. |
| client/ts/src/framer/cache/unary.ts | Moves a flushed dynamic series into the static cache. |
| client/ts/src/framer/cache/dynamic.ts | Exposes leading-buffer flush as a cache operation. |
[...]

PR #2821, opened 2026-08-25
SY-4742: Stop the live buffer from leaking into reads outside its data

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4742](https://linear.app/synnax/issue/SY-4742/stop-the-live-buffer-from-leaking-into-reads-outside-its-data)

## Description

While a channel streams, the client caches incoming samples in a live leading buffer whose recorded time range ends at `TimeStamp.MAX`, a placeholder for "still growing". `Unary.read` included the buffer in a read result whenever that recorded range overlapped the requested range, so the buffer "overlapped" every read that ended after its start, including ranges where it held no samples at all. In the v0.57.0 incident, epoch-stamped Arc samples made the buffer claim `[1970, forever)`: a read of the last 30 seconds correctly fetched nothing from the Core but returned the whole 1970 buffer anyway, exploding x auto-bounds and poisoning newly opened rolling plots through their backfill read.

The overlap test now judges the data the buffer actually holds. `Dynamic` exposes the buffer's real data range, ending at the last stamped write with the same wall-clock fallback the flush path uses, and `Unary.read` checks the requested range against it. The live edge is unaffected because a healthy live buffer's data end is essentially now, and a read that genuinely targets the buffer's era still gets it. This is the last layer of the coverage claim introduced in #2673: #2720 ended the claim when streaming stops and #2727 removed it from gap computation, both leaving the inclusion test on the provisional MAX end.

Two further defects in the same file came out of review, both of which put a wrong timestamp on a buffer boundary:

- A series that outgrew the buffer used to be split across two buffers, and both halves were stamped with the wall clock. The timestamp where a split falls is not in the data, and guessing it assumes samples arrive at an even rate, which a command channel never does. A series is now never split: the buffer grows to hold it whole, or it rotates, so every boundary sits between two series that carry real timestamps. The old split loop also allocated forever when a single variable-length sample exceeded the buffer's byte capacity.
- A frame trimmed of a re-sent prefix keeps the whole frame's start, so rotating on one gave the new buffer a start earlier than any sample it holds. The rotation now starts where the flushed buffer ends.

Auditing the series copy paths turned up a defect of its own: `convert`, `slice`, `sub` and `reAlign` all built their result without `alignmentMultiple`, so a decimated series silently reverted to one alignment step per sample, and `slice`/`sub` stepped the alignment by the sample count instead of the count times the multiple. Pluto's line renderer and its bounds computation both divide by that multiple. Every series-copying method now carries the full property set, pinned by a spec per method.

Covered by unit specs on `Dynamic.dataTimeRange`, rotation stamping, the `Unary.read` exclusion, the x series property preservation, plus a live-core regression in `feed.spec.ts` that streams epoch-stamped samples and asserts a read of the present comes back empty while a read of the buffer's own era still serves it. Every new discriminating spec fails against the code it replaces.

## Basic Readiness

### DIFF
--- a/client/ts/src/framer/cache/dynamic.spec.ts
+++ b/client/ts/src/framer/cache/dynamic.spec.ts
@@ -183,37 +183,61 @@ describe("DynamicCache", () => {
           );
         });
 
-        it("should fall back to the wall clock at a buffer split", () => {
+        it("should grow a buffer rather than split a series across two", () => {
           const cache = new Dynamic({ dynamicBufferSize: 2, now: () => WALL });
           const { flushed, allocated } = cache.write(
             new MultiSeries([
               stamped([1, 2, 3], TimeStamp.seconds(10).range(TimeStamp.seconds(13))),
             ]),
           );
-          // The split point's timestamp is unknowable from the data series, so the
-          // flushed buffer ends at the wall clock and the continuation starts there.
-          expect(flushed.series[0].timeRange).toEqual(
-            TimeStamp.seconds(10).range(WALL),
+          // Splitting would put the boundary at a timestamp the data does not carry.
+          expect(flushed).toHaveLength(0);
+          expect(allocated).toHaveLength(3);
+          expect(cache.dataTimeRange).toEqual(
+            TimeStamp.seconds(10).range(TimeStamp.seconds(13)),
           );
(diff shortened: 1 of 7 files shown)


emilbon99:
@greptile re-review

emilbon99:
@greptile re-review

PR #2823, opened 2026-08-25
SY-4738: Stamp literals from the cycle and stop stage re-entry rewrites

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4738](https://linear.app/synnax/issue/SY-4738/literal-values-in-arc-graph-blocks-stamp-timestamp-zero)
[SY-4739](https://linear.app/synnax/issue/SY-4739/re-entering-a-stage-rewrites-channel-data-it-already-wrote)

## Description

Two defects that put a bad timestamp into a channel index and kill the task. They land together because the fix for the second needs the first.

**SY-4738.** A number typed into a Console graph block becomes an inline param value, and the runtime backs it with a time series holding one zero sample. A node forwarding provenance could pick that side, so `5 > sensor` stamped its output at 1970. `ProvenanceIdx` now skips literal and reference inputs and picks the longest input that actually carries time; a node with no such input stamps the cycle instead. Reference inputs were already never candidates in spirit, since a `chan` param is a key rather than data.

**SY-4739.** Stage entry re-armed every input, so a sink whose upstream still held its last series wrote that series again. Two samples at 500s and 501s became `[500s, 501s, 500s, 501s]`, the index walked backwards, and Cesium rejected the overlap. `Reset` now re-arms literal inputs only, so `1 -> valve_cmd` still fires on every entry while an edge-fed input keeps its consumed position. This is expressible only because SY-4734 made freshness a revision counter rather than a timestamp comparison.

Both runtimes. `WriteTest.ResetRearmsInputsOnStageReentry` and its three siblings pinned the old re-fire and are renamed and flipped.

Stacked on #2822.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR changes both Arc runtimes to ignore literal and reference inputs when selecting timestamp provenance and to preserve consumed edge-fed inputs across stage re-entry.
- Adds literal-input tracking and cycle timestamp fallbacks in the Go and C++ runtimes.
- Changes reset behavior so only literal `rearmAlways` inputs are re-armed.
- Updates math, operator, channel, selector, and WASM tests for timestamp provenance and stage re-entry.

The PR should not merge until no-provenance operators preserve one timestamp per output sample instead of producing writes that channel sinks reject.

Multi-sample inputs without timestamps reach the new cycle-stamping fallback, which creates a single output timestamp and causes downstream channel writes to be dropped for mismatched sample counts.

**Files Needing Attention:** arc/go/stl/op/op.go, arc/go/stl/math/math.go, arc/cpp/stl/math/math.h

| Filename | Overview |
|----------|----------|
| arc/go/runtime/node/state.go | Adds literal-input classification, provenance selection, cycle stamping, and selective reset behavior; the one-sample stamping contract does not account for multi-sample inputs without timestamps. |
| arc/cpp/runtime/state/state.cpp | Implements the C++ provenance selector and one-sample cycle timestamp fallback in parity with Go. |
| arc/cpp/runtime/state/state.h | Stores literal classifications and changes reset semantics so consumed edge-fed inputs survive stage re-entry. |
| arc/go/stl/op/op.go | Uses a one-sample cycle fallback for unary and binary outputs, causing cardinality mismatches for multi-sample inputs without timestamps. |
| arc/go/stl/math/math.go | Applies provenance and cycle-stamping behavior to aggregate and derivative nodes; derivative outputs can inherit the same no-provenance cardinality issue. |
| arc/cpp/stl/math/math.h | Applies the corresponding provenance changes to C++ aggregate, derivative, binary, and unary math nodes. |
| arc/go/stl/op/op_test.go | Covers edge-plus-literal and all-literal scalar provenance but not multi-sample edge-fed data without timestamps. |
| arc/cpp/stl/math/math_test.cpp | Adds scalar literal provenance coverage and updates stage-reset expectations. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  A[Edge-fed multi-sample input] --> B{Timestamp series present?}
  B -- Yes --> C[Forward longest provenance series]
  B -- No --> D[Stamp one cycle timestamp]
  C --> E[Output data and timestamps]
  D --> E
  E --> F{Counts match?}
  F -- Yes --> G[Channel write]
  F -- No --> H[Report error and drop write]
```

<sub>Reviews (1): Last reviewed commit: ["SY-4738: Stamp the cycle in C++ when no ..."](https://github.com/synnaxlabs/synnax/commit/46d28b15be563f8bb6520611850cc479a2d1d727) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56628986)</sub>


emilbon99:
@sy-nico would also like to discuss this oen

arc/go/runtime/node/state.go
```diff
@@ -290,15 +297,23 @@ func (n *State) MarkFresh(paramIndex int) {
...
+// value it already emitted, which duplicates writes downstream.
 func (n *State) Reset(Context) {
 	for i := range n.accumulated {
 		switch n.rearm[i] {
 		case rearmOnFresh:
 		case rearmOnArrival:
 			n.absorbInput(i)
-		case rearmAlways, rearmOnReset:
+		case rearmAlways:
+			if !n.literal[i] {
```
sy-nico, line 311: Doesn't a node mixing a literal with an edge-fed input still re-fire on stage re-entry? `RefreshInputs` runs when any input is unconsumed, so re-arming the literal on `5 > sensor` makes the op re-emit the retained sensor samples with their retained timestamps, and the sink downstream sees a fresh rev and writes the same stamps again, which is the same overlap kill SY-4739 fixes on a direct edge.

arc/go/stl/math/math.go
```diff
@@ -379,16 +379,17 @@ func (r *avgNode) Next(ctx node.Context) {
...
 	}
 	r.sampleCount = r.process(inputData, r.sampleCount, r.Output(0))
-	if inputTime.Len() > 0 {
-		lastTimestamp := telem.ValueAt[telem.TimeStamp](inputTime, -1)
-		*r.OutputTime(0) = telem.NewSeriesV[telem.TimeStamp](lastTimestamp)
+	if timeIdx := r.ProvenanceIdx(); timeIdx >= 0 {
+		t := r.InputTime(timeIdx)
+		*r.OutputTime(0) = telem.NewSeriesV(telem.ValueAt[telem.TimeStamp](t, -1))
+	} else {
+		r.StampCycle(ctx, 0)
```
sy-nico, line 392: The avg and derivative halves of this change need tests for the Go and C++ runtimes. The new literal coverage only exercises the binary op. A literal-fed `avg` regressing back to the 1970 stamp would still go green today, so maybe worth one Go case mirroring the `op_test` ones?

sy-nico (changes requested):

arc/go/stl/math/math.go
```diff
@@ -492,6 +493,9 @@ func (d *derivativeNode) Next(ctx node.Context) {
 		&d.prevValue, &d.prevTimestamp, &d.hasPrev,
 		d.Output(0), d.OutputTime(0),
 	)
+	if d.ProvenanceIdx() < 0 {
+		d.StampCycle(ctx, 0)
+	}
 	d.Output(0).Alignment = inputData.Alignment
 	d.Output(0).TimeRange = inputData.TimeRange
 	d.OutputTime(0).Alignment = inputData.Alignment
```
sy-nico, line 500: This regresses `avg`, `min` and `max` when a `reset` edge is wired. `ProvenanceIdx` picks the longest non-literal input and the reset now counts, so a cycle where the reset batch outruns the data batch stamps the output with the reset channel's time instead of the data sample. A lagging reset channel then pushes the stamp backwards past the previous output, and Cesium rejects that write the same way it did in SY-4739. The old code always read `inputIdx` here, same in the C++ `Aggregator`, so I think the stamp should stay pinned to the data input in both.


At least needs a test to confirm/deny my suspicion.

sy-nico (changes requested):

arc/go/stl/math/math.go
emilbon99: *Emiliano + Claude: drafted in our review session.*

Confirmed with a test: a lagging reset stamped the output with the reset's time. The aggregator now stamps from its data input in both runtimes, and falls back to the cycle stamp only when that input is a literal. Regression tests pin this in both runtimes.

sy-nico (approved):

PR #2846, opened 2026-08-26
SY-4734: Add a stable per-cycle timestamp to the Arc runtime

emilbon99:
# Issue Pull Request

## Linear Issue

[SY-4734](https://linear.app/synnax/issue/SY-4734/add-a-stable-per-cycle-timestamp-to-the-arc-runtime)

## Description

Arc had 16 independent clocks. Every producer read its own `MonoClock` at its own moment in the scheduler pass, so samples written in one cycle carried several close-but-distinct timestamps, and `now()` inside a guest function disagreed with the node that fed it. Worse, two channels sharing an index always killed the task: each write appended its own index stamp, so the index ended up with more samples than any data channel and Cesium rejected the commit.

`node.Cycle` carries one stamp per scheduler pass. The runtime loop reads its clock once, builds the cycle, and hands it to `Scheduler.Next`; `Context.Now` exposes it and nothing downstream reads a clock of its own. Every node-side clock is deleted: `constant`, both variable nodes, the WASM node, the time module, `stable`, and the channel source, along with `stable.WithNow` and the C++ `NowFunc` parameter. `reset` now takes the Context, since stage activation happens inside a cycle. The `now()` binding returns the cycle stamp, so it is constant within a cycle, which `arc/docs/spec.md` now states.

Index stamping moved out of the write path and into the flush. `Flush(fr, now)` synthesizes one index series per group from the cycle stamp, one sample per data sample spaced 1ns apart, and returns its highest stamp so the loop advances its clock past it. An index a sink already filled with forwarded provenance is left alone. Channels sharing an index now produce one coherent index series per cycle instead of one per writer.

Input freshness also moves off timestamps. A node used to decide "is this input new?" by comparing its stamp against what it last consumed, which breaks as soon as every producer in a cycle stamps identically. `ProgramState.rev` is a monotonic counter bumped on every emit, and a node compares revisions instead. "The producer wrote this again" is now a fact independent of time.

Design in [RFC 0058](docs/tech/rfc/0058-arc-cycle-timestamp.md), included here. The three commits after the merge follow its phases: revision freshness, the cycle stamp, then index coalescing.

Recreates #2822, which could not be reopened after its release-branch base was deleted. The parent PR #2813 already landed on `main`.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR gives each Arc scheduler cycle one stable timestamp and replaces timestamp-based input freshness with write revisions. It also moves index generation to channel flush.
- Adds a cycle context shared by schedulers, nodes, and WASM time bindings.
- Coalesces generated timestamps for channels that share an index.
- Adds monotonic clock advancement after generated index ranges.

The shared-index length defect must be fixed before merge because valid per-cycle writes can produce an invalid output frame.

Index synthesis uses the first written channel's sample count and skips later channels in the same index group, even when their sample counts differ.

**Files Needing Attention:** arc/go/stl/channels/state.go and arc/cpp/stl/channels/state.cpp

| Filename | Overview |
|----------|----------|
| arc/go/stl/channels/state.go | Moves index stamping to flush, but derives a shared index length from only the first member written. |
| arc/cpp/stl/channels/state.cpp | Implements the same shared-index synthesis behavior and unequal-length failure as Go. |
| arc/go/runtime/node/state.go | Replaces timestamp freshness with program-wide output revisions and updates input consumption. |
| arc/cpp/runtime/state/state.cpp | Implements revision-based freshness and returns the highest synthesized timestamp from flush. |
| arc/cpp/runtime/runtime.h | Samples one timestamp per scheduler cycle and advances the runtime clock after flush. |
| x/go/telem/mono_clock.go | Adds clock-floor advancement for timestamps derived from a cycle stamp. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Clock[Runtime clock] --> Cycle[Cycle timestamp]
  Cycle --> Scheduler
  Scheduler --> Nodes[Arc nodes]
  Nodes --> Writes[Buffered channel writes]
  Cycle --> Flush[Channel flush]
  Writes --> Flush
  Flush --> Index[Shared index synthesis]
  Index --> Output[Output frame]
  Index --> Advance[Advance runtime clock]
```

<sub>Reviews (1): Last reviewed commit: ["SY-4734: Reflow the now() spec bullet to..."](https://github.com/synnaxlabs/synnax/commit/33bd129e834917260ec53d40d5c683b0e060c949) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57041544)</sub>


arc/go/stl/wasm/node.go
```diff
@@ -249,7 +248,7 @@ func (n *nodeImpl) Next(ctx node.Context) {
 		}
 		var ts uint64
 		if clockStamp {
-			ts = uint64(n.clock.Now())
+			ts = uint64(ctx.Now)
```
sy-nico: Regression:

Every sample in a batch now gets the identical stamp on the `clockStamp` path, where the old per-sample `clock.Now()` bumped 1ns each, so a `$sel` node fed a multi-sample channel batch writes N equal timestamps into the index through the sink and Cesium takes them without complaint. The channel source and `stampIndexes` both space their samples 1ns apart, so should this be `ctx.Now + i` too? Same loop in the C++ `wasm/node.h`.

sy-nico (changes requested):

arc/go/stl/channels/state.go
```diff
@@ -80,7 +82,34 @@ func (cs *ProgramState) Flush(
...
+
+// stampIndexes fills every empty index buffer whose channels wrote this cycle,
+// and returns the highest timestamp it wrote. A non-empty buffer already holds
+// the provenance a sink forwarded and is left alone. Every member of a group
+// starts at now, so one cycle's writes stay aligned across channels.
+func (cs *ProgramState) stampIndexes(now telem.TimeStamp) (highest telem.TimeStamp) {
+	for _, key := range cs.activeWriteKeys {
+		idx := cs.indexes[key]
+		if idx == 0 || len(cs.writes[idx].Data) > 0 {
+			continue
```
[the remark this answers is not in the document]
emilbon99: *Emiliano + Claude: drafted in our review session.*

Known and deliberate. Unequal or partial writes to one index group still error at Cesium's equal-length check, and the RFC (§6.2) records why we don't pad or hold writes in this PR: padding re-issues commands, and holding stalls on conditional writes. Before this change, any two channels on a shared index failed every time. Now only programs that actually write unequal counts fail. The policy question is tracked separately.

arc/go/stl/wasm/node.go
emilbon99: *Emiliano + Claude: drafted in our review session.*

Good catch. `ctx.Now + i` alone would let the next cycle land inside that range, because flush only advanced the clock past stamps it synthesized. The channel source already had the same gap for unindexed channels. Producers now reserve stamps from the cycle through the context, and the loop advances its clock past the highest one reserved. `$sel` samples are now 1 ns apart in both runtimes. Regression tests pin both fixes in both runtimes, and in the Core's Arc task and calculator loops.

sy-nico:
@greptile, rerun

sy-nico (approved):

PR #2838, opened 2026-08-26
SY-4760: Add cut, copy, and paste to the diagram context menus (Part 2)

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4760](https://linear.app/synnax/issue/SY-4760/schematic-context-menu-improvements)

## Description

Stacked on Part 1 (#2837).

Adds `Cut`, `Copy`, and `Paste` entries to the schematic and Arc graph context menus. The entries are hidden when the diagram is not editable and `Cut`/`Copy` are disabled without a selection, matching the code editor's menu.

Menu clicks reuse the native clipboard path instead of adding a second one: `useClipboard` now takes the diagram container ref and returns imperative triggers that focus the container and call `document.execCommand`, so the menu and the keyboard shortcuts share one serialization and one MIME type. Cut fires a native copy event and then removes what it wrote, because browsers only fire cut events in editable DOM. The new `Diagram.Menu.ClipboardItems` component follows `Menu.UndoRedoItems`.

Tests cover the trigger behavior and add a compile-time canary that fails the build if `execCommand` is ever dropped from the DOM standard.

## Basic Readiness

- [x] I have performed a self-review of my code.
- [x] I have added relevant, automated tests to cover the changes.

<h3>Greptile Summary</h3>

The PR adds Cut, Copy, and Paste actions to editable diagram context menus while sharing the existing native clipboard serialization path.
- Adds imperative clipboard triggers and fallback replay of the last copied diagram payload.
- Integrates clipboard menu items into schematic and Arc graph editors.
- Centralizes standard clipboard keyboard triggers and adds regression coverage for menu-trigger behavior.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge.

No blocking failure remains.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| pluto/src/vis/diagram/clipboard.ts | Adds menu-driven clipboard commands and now clears pending cut state in a `finally` block, resolving the previously reported exception path. |
| pluto/src/vis/diagram/clipboard.spec.ts | Covers copy, cut, paste, fallback replay, and the regression where a throwing cut must not affect a later copy. |
| pluto/src/vis/diagram/menu/ClipboardItems.tsx | Adds reusable Cut, Copy, and Paste menu entries with selection-based disabling. |
| pluto/src/arc/graph/Editor.tsx | Connects Arc graph context-menu actions to the shared clipboard hook. |
[...]

PR #2847, opened 2026-08-26
SY-4139: Schematic Symbol Groups

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4139](https://linear.app/synnax/issue/SY-4139)

## Description

Adds symbol grouping to the schematic, stacked on #2838.

### Grouping

`Ctrl+G` groups the selection, `Ctrl+U` ungroups, with matching context menu entries that appear only when they apply. The console suppresses the browser defaults on both shortcuts, which otherwise open view source and find next. A group is a hidden `groupBox` symbol: it renders behind its members, drags like any other symbol, and never appears in the symbols toolbar or search. Membership lives in the group's config as a `members` list, so there is no schema change and no migration. Groups nest: grouping resolves each selected symbol to its outermost group, and ungroup peels one layer, promoting members into the enclosing group.

### Movement

Members are position locked via `draggable: false`. Moving the group fans its delta out to members in the dispatch preprocess, so dragging, alignment, distribution, undo, and remote replay share one code path and coalesce into single undo steps.

### Selection and editing

Clicking any member selects the whole group, so copy, cut, and delete always cover the group and membership can never go stale. Double clicking a member drills into it for individual editing. A selected group shows the standard multi-symbol toolbar, so alignment, color, label, and rotation edits apply across the group. Paste rewrites a pasted group's `members` onto the pasted keys via `PasteResult.remap`, which supersedes the rc-only `newKeys` field (adapters derive the keys from the map).

### Sizing

The group box persists no dimensions. Its size derives at render time from member positions and measured sizes with a fixed padding, so it tracks member changes live and undo and remote edits stay correct with no refit machinery.

### Locking

A selected group shows a padlock chip that toggles `locked` in its config. A locked group cannot move, double clicking into its members is blocked, and bulk edits from a multi selection skip everything inside it.

### Membership integrity

Deleting a node splices its key out of every group's `members` in the `removeNode` reducer itself, on both the client and the Core. A raw action stream can no longer leave a dangling reference, undo restores the membership with the node, and the two reducers are pinned to identical behavior by a mirrored case matrix.

### Tests

Pure specs for the membership forest, selection closure and drill in, derived sizing, the lock chip, trigger cases, a mirrored reducer cascade matrix in TypeScript and Go, and hook, clipboard, and dispatch round trips against a live Core.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR adds nested schematic symbol grouping, group-aware movement and selection, derived group sizing, locking, clipboard remapping, and membership cleanup across the client and Core.
- Adds group and ungroup commands, context-menu actions, and keyboard triggers.
- Keeps group movement, selection, editing, locking, and nested membership consistent.
- Updates cut handling to remove incident edges and prevent dangling references.
- Adds mirrored TypeScript and Go reducer coverage plus UI, clipboard, query, and dispatch tests.

The PR appears safe to merge.

No blocking failure remains; the current cut adapter explicitly removes every incident edge for each cut node, resolving the previously reported dangling-edge issue.

| Filename | Overview |
|----------|----------|
| pluto/src/schematic/group/group.ts | Implements membership traversal, grouping, nested ungrouping, selection closure, and lock-aware behavior. |
| pluto/src/schematic/clipboard.ts | Remaps pasted group memberships and removes incident edges when cutting nodes, resolving the previously reported dangling-edge failure. |
| client/ts/src/schematic/actions.ts | Extends node removal to splice deleted keys from group membership while preserving undo targets and inverses. |
| core/pkg/service/schematic/actions.go | Mirrors client-side group membership cleanup in the persisted Core reducer. |
| pluto/src/schematic/node/groupBox/Symbol.tsx | Defines the hidden group symbol and its derived rendering behavior. |
| pluto/src/schematic/Schematic.tsx | Integrates grouping behavior into schematic rendering and interaction dispatch. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Selection[Selected symbols] --> Group[Create hidden groupBox]
  Group --> Members[Persist member keys]
  Members --> Movement[Propagate group movement]
  Members --> SelectionClosure[Expand selection to group]
  Members --> Clipboard[Remap membership on paste]
  Members --> Cleanup[Splice membership on node removal]
  Lock[Group lock] --> Movement
  Lock --> SelectionClosure
```

<sub>Reviews (7): Last reviewed commit: ["SY-4139: Cascade group membership remova..."](https://github.com/synnaxlabs/synnax/commit/3de11e3cec09c789f42f1ee208cb7cd1011b5717) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57286142)</sub>


sy-nico:
@greptile, rerun. Accidentally had branch targeting `rc` instead of [sy-4760-schematic-context-menu-improvements](https://github.com/synnaxlabs/synnax/tree/sy-4760-schematic-context-menu-improvements)

That just means the changes here are scoped to the Schematic grouping, making the change-set to review smaller.

pjdotson (changes requested):
Two major comments:

1) The UX for the grouping doesn't really feel there. Why do I have this pan thing in the top left corner:

Image

If I have grouped something, I can't edit, say, the color of anything in the group. So I have a group, I decide I want to change the color of one of the symbols, I now have to delete the group, change the color of a symbol, and remake a group
Shouldn't there be more options like color, etc to change the color of all of the symbols in the group?
- when you select a whole group should the individual symbols appear as selected as well? why do we need to have that and the general group selection highlight?
- If you select a group via clicking on it, vs dragging a select box across the whole group, that changes whether or not the background of the group is hilighted.
- Maybe you should have an option for whether you want to show the dotted border when not in edit mode? e.g. no border / dotted border / solid border

2) I am concerned about the data model chosen here. Why did you choose the model of just making a group the same as any other node? Now the data model means different nodes of the schematic can have relations. So, you could have an issue where a node gets deleted but its corresponding group node's `members` property doesn't get deleted. Why choose this instead of an explicitly nested structure a la what the mosaic tabs do?

Will probably be best to talk through these synchronously.

sy-nico:
@pjdotson , evaluation of the data models attached. This was also discussed thoroughly with Emiliano a few months ago on the right approach.

Very confident this is the right approach: [SCHEMATIC_GROUP_DATA_MODEL.md](https://github.com/user-attachments/files/31854047/SCHEMATIC_GROUP_DATA_MODEL.md)

sy-nico:
@greptile, rerun

PR #2848, opened 2026-08-27
SY-4769: Standardize schematic symbol heights

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4769](https://linear.app/synnax/issue/SY-4769)

## Description

Laying out a schematic that looks tidy has been an exercise in frustration. Drop a `Value` next to a `Setpoint` next to a `String display` and none of them line up, so you nudge each one by hand, zoom in, nudge again, and still end up with a row that reads as slightly broken. There was no size you could pick that made them agree, because no two symbols measured themselves the same way.

I deliberately kept schema changes and migrations out of this. The point was to keep the change scoped and easy to review, so every symbol keeps the config field it already stores, `level` for `Value` and `String display` and `size` for the control symbols, and nothing on disk changes shape. Where a new field was tempting, a size picker for `State indicator`, I left it alone rather than widen the diff.

Six symbols each sized themselves differently. `Value` derived its height from a typography formula, `String display` and `State indicator` from text line height plus padding, and `Input`, `Setpoint`, and `Select` from the shared component size scale. At their defaults they rendered at 26, 28, 28, 24, 24, and 24 pixels. They now share one ladder, taken from the `--pluto-height-*` scale every other control already uses:

| Size | Height |
| ---- | ------ |
| XS | 21 |
| S | 24 |
| M | 28 |
| L | 36 |
| XL | 48 |

`M` is the default everywhere, which is what `State indicator` was already fixed at.

### Size pickers

`Value` and `String display` store a text level, not a size. A new `Form.LevelSizeField` gives them the same `XS` through `XL` picker as the control symbols and translates through `SIZE_LEVELS`, so nothing stored changes shape. `String display` defaulted to `p`, a level its own picker never offered, so there was no way to select it back; it now defaults to `h4`. `SelectSize` gained `XS` and `XL`, and its labels were shifted, since what it called `M` was `small`, or 24 pixels.

### Field order

All six rows now read color, size, width, then anything else.

### User impact

Existing `Value` symbols resize, since the height is computed at render rather than stored: 26 to 24 at `S`, 29 to 28 at `M`, 36.5 to 36 at `L`, 45.5 to 48 at `XL`. Connected edges re-anchor. Control symbols do not move, because `size` is stored and only the default changed. Legacy `String display` configs keep their 28 pixel height through the `medium` fallback.

### Rendering cost

Nothing worth measuring. `levelSize` is a five element `find` per render, and `Value` dropped a `Theming` context subscription in exchange. No new canvas work, no new Aether state, no extra re-render triggers.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR standardizes schematic symbol heights around the shared component-size scale while preserving existing persisted configuration shapes.
- Adds shared mappings between component sizes, pixel heights, and text levels.
- Updates Value and String Display sizing controls and rendering.
- Aligns default sizes and form-field ordering across the affected symbols.
- Corrects the Level and Alignment picker key arrays so keyboard navigation follows the rendered options.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| pluto/src/select/text/Level.tsx | The navigation key array now exactly matches the five rendered buttons in display order, resolving the previous mismatch. |
| pluto/src/select/flex/Alignment.tsx | Restricts navigation keys to the three alignments actually rendered by this picker. |
| pluto/src/select/Button.spec.tsx | Adds assertions that each button picker's rendered item order matches its intended navigation order. |
| pluto/src/schematic/node/common/size.ts | Introduces the shared size-to-text-level mapping and a medium fallback for legacy unoffered levels. |
| pluto/src/schematic/node/general/value/Symbol.tsx | Replaces typography-derived Value height with the standardized shared height ladder. |
| pluto/src/schematic/node/general/stringDisplay/Primitive.tsx | Applies the standardized explicit height based on the configured text level. |

<sub>Reviews (2): Last reviewed commit: ["SY-4769: Align picker nav keys with rend..."](https://github.com/synnaxlabs/synnax/commit/9fcad2965e79150cb3995428284bfb35fedec98b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57369949)</sub>

### DIFF
--- a/pluto/src/schematic/node/common/form/Size.tsx
+++ b/pluto/src/schematic/node/common/form/Size.tsx
@@ -7,10 +7,12 @@
 // License, use of this software will be governed by the Apache License, Version 2.0,
 // included in the file licenses/APL.txt.
 
+import { type text } from "@synnaxlabs/x";
 import { type ReactElement } from "react";
 
 import { Component } from "@/component";
 import { Form } from "@/form";
+import { levelSize, SIZE_LEVELS } from "@/schematic/node/common/size";
 
 export const SizeField = (
   props: Partial<Form.FieldProps<Component.Size>>,
@@ -27,3 +29,23 @@ export const SizeField = (
 );
 
 const selectSize = Component.renderProp(Component.SelectSize);
+
+/** A size field for symbols that store a text level instead of a size. */
+export const LevelSizeField = (
+  props: Partial<Form.FieldProps<text.Level>>,
+): ReactElement => (
+  <Form.Field<text.Level>
+    path="level"
+    label="Size"
+    hideIfNull
+    padHelpText={false}
+    {...props}
+  >
+    {({ value, onChange }) => (
+      <Component.SelectSize
+        value={levelSize(value)}
+        onChange={(size: Component.Size) => onChange(SIZE_LEVELS[size])}
+      />
+    )}
+  </Form.Field>
+);
--- a/pluto/src/schematic/node/common/form/Size.tsx
+++ b/pluto/src/schematic/node/common/form/Size.tsx
@@ -12,7 +12,7 @@ import { type ReactElement } from "react";
 
 import { Component } from "@/component";
 import { Form } from "@/form";
-import { levelSize, SIZE_LEVELS } from "@/schematic/node/common/size";
+import { LEVEL_SIZES, SIZE_LEVELS } from "@/schematic/node/common/size";
 
 export const SizeField = (
   props: Partial<Form.FieldProps<Component.Size>>,
@@ -41,11 +41,21 @@ export const LevelSizeField = (
     padHelpText={false}
     {...props}
   >
-    {({ value, onChange }) => (
-      <Component.SelectSize
-        value={levelSize(value)}
-        onChange={(size: Component.Size) => onChange(SIZE_LEVELS[size])}
-      />
-    )}
+    {selectLevelSize}
   </Form.Field>
 );
+
+const SelectLevelSize = ({
+  value,
+  onChange,
+}: {
+  value: text.Level;
+  onChange: (level: text.Level) => void;
+}): ReactElement => (
+  <Component.SelectSize
+    value={LEVEL_SIZES[value]}
+    onChange={(size: Component.Size) => onChange(SIZE_LEVELS[size])}
+  />
+);
+
+const selectLevelSize = Component.renderProp(SelectLevelSize);
--- a/pluto/src/schematic/node/common/size.ts
+++ b/pluto/src/schematic/node/common/size.ts
@@ -9,7 +9,7 @@
 
 import { type text } from "@synnaxlabs/x";
 
-import { type Size, SIZES } from "@/component/size";
+import { type Size } from "@/component/size";
 
 /** The text level a symbol renders at for each size rung. */
 export const SIZE_LEVELS: Record<Size, text.Level> = {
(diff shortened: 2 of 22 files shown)


pluto/src/select/text/Level.tsx
```diff
@@ -17,20 +17,20 @@ export interface LevelProps extends Omit<Select.ButtonsProps<text.Level>, "keys"
...
       M
     </Button>
-    <Button itemKey="h5" square>
-      S
+    <Button itemKey="h3" square>
+      L
     </Button>
-    <Button itemKey="small" square>
-      XS
+    <Button itemKey="h2" square>
```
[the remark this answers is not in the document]
sy-nico: Confirmed, and fixed. `keys` is now the exact set the picker renders, in render order: `["small", "h5", "h4", "h3", "h2"]`.

`Select.Flex.Alignment` had the same defect independently of this PR. It passed `Flex.ALIGNMENTS` while rendering only three buttons, so arrow keys could land on `stretch` with nothing to select. It now passes an explicit array too.

Added a `picker button order` test in `select/Button.spec.tsx` covering all four button pickers, since the failure is keyboard only and invisible in review.

sy-nico:
@greptile, rerun.

pluto/src/schematic/node/common/form/Size.tsx
```diff
@@ -27,3 +29,23 @@ export const SizeField = (
...
+  props: Partial<Form.FieldProps<text.Level>>,
+): ReactElement => (
+  <Form.Field<text.Level>
+    path="level"
+    label="Size"
+    hideIfNull
+    padHelpText={false}
+    {...props}
+  >
+    {({ value, onChange }) => (
```
pjdotson: this should be a `Component.renderProp`

pluto/src/select/flex/Alignment.tsx
```diff
@@ -19,8 +19,11 @@ export interface AlignmentProps extends Omit<
   "keys"
 > {}
 
+/** The alignments the picker offers, in render order. */
+const KEYS: Flex.Alignment[] = ["start", "center", "end"];
+
 export const Alignment = ({ value, ...rest }: AlignmentProps): ReactElement => (
-  <Buttons {...rest} value={value} keys={Flex.ALIGNMENTS}>
+  <Buttons {...rest} value={value} keys={KEYS}>
```
pjdotson, line 26: why duplicate the `Flex.ALIGNMENTS`?

pluto/src/select/text/Level.tsx
```diff
@@ -7,30 +7,33 @@
...
 export interface LevelProps extends Omit<Select.ButtonsProps<text.Level>, "keys"> {}
 
+/** The levels the picker offers, in render order. */
+const KEYS: text.Level[] = ["small", "h5", "h4", "h3", "h2"];
+
 export const Level = (props: LevelProps): ReactElement => (
-  <Buttons {...props} keys={text.LEVELS}>
-    <Button itemKey="h2" square>
-      XL
+  <Buttons {...props} keys={KEYS}>
```
pjdotson, line 22: why duplicate `text.LEVELS`

pluto/src/schematic/node/common/size.ts
```diff
@@ -0,0 +1,25 @@
...
+export const SIZE_LEVELS: Record<Size, text.Level> = {
+  tiny: "small",
+  small: "h5",
+  medium: "h4",
+  large: "h3",
+  huge: "h2",
+};
+
+/** @returns the rung for a level, falling back to medium for unoffered levels. */
+export const levelSize = (level: text.Level): Size =>
```
pjdotson: why not just make this a `Record<text.Level, Size>`?

pjdotson (changes requested):
When I drop a new string display on there is no default size set:

Image

pluto/src/select/flex/Alignment.tsx
sy-nico: It's not a duplicate, it's the 3-of-4 subset the picker renders. Arrow keys walk `keys`, not the children, so passing `Flex.ALIGNMENTS` lets hover land on `stretch` and Enter select it with no button on screen.

pluto/src/select/text/Level.tsx
sy-nico: Same as `Alignment`. `keys` must be the rendered buttons in render order. `text.LEVELS` has 7 entries in descending order and the picker renders 5 ascending, so passing it ran arrow navigation backwards and let hover land on `h1` or `p` with no button. This was the Greptile P1 above.

sy-nico:
@pjdotson , Your screenshot is from a stale build
image

pjdotson (approved):

PR #2854, opened 2026-08-28
SY-4502: Upgrade Go dependencies

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4502](https://linear.app/synnax/issue/SY-4502)

## Description

Upgrades the Go toolchain to 1.27.0 and the module dependencies across all nine Go
modules. Also moves golangci-lint to v2.13.2 in every `.custom-gcl.yaml` and in the CI
workflow, and bumps `stringer` to v0.49.0.

## Basic Readiness

- [x] I have performed a self-review of my code.
- [x] I have added relevant, automated tests to cover the changes.
- [x] I have updated documentation to reflect the changes.

<h3>Greptile Summary</h3>

The PR upgrades the Go toolchain and dependencies across all nine Go modules.
- Raises module Go directives from 1.26.5 to 1.27.0.
- Updates module dependencies and checksums.
- Upgrades golangci-lint and stringer.
- Applies formatting changes required by the updated tooling.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge.

No blocking failure remains.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| .github/workflows/test.go.yaml | Updates golangci-lint and stringer versions while retaining module-derived Go toolchain setup. |
| .golangci.yaml | Migrates gofumpt configuration to the updated formatter option structure. |
| core/go.mod | Raises the core module to Go 1.27.0 and refreshes direct and indirect dependencies. |
| cesium/go.mod | Raises the Cesium module toolchain and updates dependency versions. |
| aspen/go.mod | Raises the Aspen module toolchain and updates dependency versions. |
| freighter/go/go.mod | Raises the Freighter Go module toolchain and refreshes transport dependencies. |
| oracle/go.mod | Raises the Oracle module toolchain and updates compiler and code-generation dependencies. |

<sub>Reviews (2): Last reviewed commit: ["Fix lint issues"](https://github.com/synnaxlabs/synnax/commit/02f45d725c9b9c3e56b73247363c387e34df967b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57848363)</sub>

PR #2856, opened 2026-08-28
SY-4645: Adopt Go 1.27 language features

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4645](https://linear.app/synnax/issue/SY-4645/upgrade-to-go-127)

## Description

The toolchain bump landed in #2854 and is already on `rc`. This PR adopts the three Go
1.27 language changes across all nine Go modules.

**Generic methods.** The tree declares 42 of them. Functions that took their subject as
the first argument became methods on it, across `validate`, `telem.Series`,
`plumber.Pipeline`, `imex.Envelope`, `status.Service`, `signals.Provider`, `cesium.DB`,
and `http.Router`. `status.Writer` and `kv.Subscriber` also stop being generic: no field
used the parameter, so it moved onto the methods that name it and callers now infer it.

**Promoted field keys.** A struct literal key may now name a field promoted through an
embedded struct — `Point3D{name: x}` compiles, `Point3D{Object.name: x}` does not. A
`go/packages` analyzer found the 1,007 eligible sites. Seventy-five hand-written ones are
rewritten here; the other 932 are `codec_gen_test.go` fixtures regenerated from an Oracle
marshal plugin change plus `oracle sync`. The plugin promotes a parent's fields into the
outer literal only when no sibling embed claims the name; a parent that collides keeps its
wrapper. Deeply nested task configs collapse from four levels to one flat list.

**Inheritance conflicts.** Oracle now rejects a field name two parents contribute, on
struct, union, and action `extends` clauses, diamonds included — the leftmost parent used
to win and the other field disappeared. Drop the name with `-name` or rename it. No schema
in the tree is affected.

**Function type inference.** Explicit type arguments that the target type now resolves are
dropped, on function values assigned into struct fields and on calls through the Arc
analyzer and compiler.

Also here: the `asynctimerchan` GODEBUG is gone permanently, so the drain-before-reset
dance in `arc/task` and `confluence/delta` is deleted, and
`docs/claude/toolchains/go.md` documented the old `validate` call form.

One trap worth recording. `oracle sync` shells out to whichever `clang-format` is first on
`PATH`, while CI pins version 22. A local clang-format 23 reindents the closing brace of a
generated static map, so three generated headers under `client/cpp` are restored to
version 22 output.

Verified per module: `go vet` clean, `custom-gcl run` clean on every changed package,
every changed package tested, `oracle check` passes all six gates, and
`scripts/check_clang_format.sh` passes for all six C++ directories.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR adopts Go 1.27 language features throughout the repository.
- Converts generic helper functions to generic methods and removes type arguments now resolved through target-type inference.
- Uses promoted embedded fields in generated and handwritten struct literals.
- Rejects ambiguous inherited fields during Oracle analysis and updates generated fixtures.
- Removes obsolete timer-channel draining now that Go 1.27 permanently uses synchronous timer semantics.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| oracle/plugin/resolver/inheritance.go | Updates inheritance handling so conflicting parent fields are rejected by analysis rather than silently resolved by parent order. |
| oracle/plugin/go/types/embed.go | Adds Go-specific embedding eligibility checks for promoted-name collisions. |
| oracle/plugin/go/marshal/test_gen.go | Generates promoted literal keys when unambiguous while retaining embedded wrappers for collisions. |
| core/pkg/service/arc/task/task.go | Converts pipeline helpers to generic methods and removes timer draining made obsolete by Go 1.27. |
| x/go/confluence/delta.go | Simplifies timer resets under the synchronous timer-channel semantics required by the module toolchain. |
| x/go/telem/series.go | Converts generic Series utilities to methods without changing their underlying behavior. |

<sub>Reviews (3): Last reviewed commit: ["remove type alias"](https://github.com/synnaxlabs/synnax/commit/8b2bc69f7626be12c63b783d56be65ebb62965a4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58050035)</sub>


pjdotson:
@greptile /rerun

oracle/plugin/go/marshal/test_gen.go
```diff
@@ -544,36 +545,80 @@ func (b *testValueBuilder) buildStructFieldExprs(
...
 	)
 	if err != nil {
 		return nil, err
 	}
+	taken := literalKeys(childFieldExprs)
+	var exprs []string
+	for _, e := range embeds {
+		// A promoted field may key the outer literal directly, so the embedded wrapper
+		// is only needed when the name is already spoken for.
+		if len(e.fields) > 0 && !anyKeyTaken(taken, e.fields) {
```
sy-nico, line 587: This check should also look at the other embeds' field names, not just keys flattened before it. If two parents share a field name, the first one flattens into a promoted key Go rejects as ambiguous, and `oracle sync` emits a fixture that doesn't compile. Oracle allows that schema shape (`resolution/type.go`, first wins on conflict).
Add a conflicting-parents case to the plugin tests.

sy-nico (changes requested):

pjdotson:
@greptile /rerun

oracle/plugin/go/marshal/test_gen.go
sy-nico: Re-opening because the field-name case is still broken: with `Meta { name }` and `Other { name }` both extended, `Meta` still flattens to a top-level `Name` key Go rejects as ambiguous, because `taken` only reserves embed names and keys already flattened. The new test covers the embed-name shadow, not two parents sharing a field, so the conflicting-parents case is still untested.

core/pkg/service/status/service.go
```diff
@@ -213,16 +207,19 @@ func (s *Service) SetByKeyOrName(
 	return key, multipleMatches, nil
 }
 
-func NewWriter[D any](s *Service, tx gorp.Tx) Writer[D] {
-	return Writer[D]{
+// NewWriter opens a Writer for statuses whose details are of type D. Pass a nil tx to
+// write directly against the service's DB.
+func (s *Service) NewWriter(tx gorp.Tx) Writer {
+	return Writer{
```
sy-nico: The doc comment still says "whose details are of type D", but `NewWriter` lost its type parameter in this PR, so there is no `D` in this signature anymore. Reads confusing next to `NewRetrieve[D]` right below, which really is generic.

arc/go/runtime/node/state.go
```diff
@@ -349,15 +349,15 @@ func (n *State) RefInput(paramIndex int) telem.Series {
...
 // else the configured value.
-func (n *State) StringInput(name string) string {
-	i, err := n.ResolveInput(name)
+func (s *State) StringInput(name string) string {
+	i, err := s.ResolveInput(name)
 	if err != nil {
 		return ""
 	}
-	if s := n.RefInput(i); s.Len() > 0 {
+	if s := s.RefInput(i); s.Len() > 0 {
```
sy-nico: Here and in `NumericInput` at line 374, `s := s.RefInput(i)` now shadows the `*State` receiver with a `telem.Series` after the receiver rename from `n` to `s`. Compiles fine, but it's easy to misread which `s` a line inside the block is talking about.

aspen/internal/kv/config.go
```diff
@@ -110,16 +110,16 @@ func (cfg Config) Override(other Config) Config {
...
-	validate.NotNil(v, "lease_transport_server", cfg.LeaseTransportClient)
-	validate.NotNil(v, "recovery_transport_client", cfg.RecoveryTransportClient)
-	validate.NotNil(v, "recovery_transport_server", cfg.RecoveryTransportServer)
-	validate.NotNil(v, "engine", cfg.Engine)
+	v.NotNil("cluster", cfg.Cluster)
+	v.NotNil("tx_transport_client", cfg.BatchTransportClient)
+	v.NotNil("tx_transport_server", cfg.BatchTransportServer)
+	v.NotNil("feedback_transport_client", cfg.FeedbackTransportClient)
+	v.NotNil("feedback_transport_server", cfg.FeedbackTransportServer)
+	v.NotNil("lease_transport_client", cfg.LeaseTransportServer)
```
sy-nico: Pre-existing, but the names here should swap, right?

`"lease_transport_server"` goes with `cfg.LeaseTransportServer` and `"lease_transport_client"` with `cfg.LeaseTransportClient`. 

As is a nil server reports as the client missing and sends whoever hits it to the wrong config field, and the report lines at 135 pair them correctly.

sy-nico (changes requested):

oracle/plugin/go/marshal/test_gen.go
pjdotson: I made these types where you inherit from two other types that have the same field a compile error now.

pjdotson:
@greptile /rerun

sy-nico (approved):

PR #2859, opened 2026-08-28
SY-4797: Fix stale Console assets and opaque read failures

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4797](https://linear.app/synnax/issue/SY-4797/debug-console-issues)

## Description

Two failures reported from a production deployment on 0.57.4, plus the
instrumentation that would have named the second one on its own.

**A browser kept running a pre-0.57.0 Console against a 0.57.4 Core.** The Core
served the Console with no `Cache-Control` at all, and every embedded file
carries a zero modification time, so a browser revalidating a stored copy is
answered 304 after any upgrade. Hashed assets now carry an immutable policy and
everything else is `no-store`. A request path that names a file the build no
longer emits answers 404 instead of the app shell: a module worker handed HTML
reports an error event with no message, which surfaced as
`[aether] worker error: unknown`.

**A permissions read failed with its reason thrown away.** `ConnectionGuard`
renders nothing until the subject's policies land, and flux wraps any failure as
`Failed to retrieve permissions`. The copied diagnostics never walked
`error.cause`, so the report named no reason. The crash screen and the copy text
now both carry the chain of causes, and a denial from the Core names the
subject, the action, and the object that stopped the request instead of being a
bare sentinel.

**A failed permissions read no longer settles on the first try.** Flux gains a
`retryUnreachable` option, built the same way `awaitCreation` is: three retries
at a widening wait when the Core could not be reached, and an immediate settle
for anything the Core actually answered. The permissions query opts in, so a
moment of dead connection at startup no longer looks the same as a lasting one.

Patch bumps for Pluto, the Console, and the Core so the fixes ship.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR prevents stale embedded Console assets, improves permission-denial diagnostics, and retries transiently unreachable permission reads.
- Applies immutable caching to hashed assets and `no-store` to stable filenames while returning 404 for obsolete asset paths.
- Adds subject, action, and object context to RBAC denial errors.
- Displays and copies nested error causes in the Console fallback.
- Adds bounded connection-error retries to Flux retrieval and enables them for permission queries.
- Bumps Console, Pluto, and Core patch versions.

The PR appears safe to merge, with no concrete blocking or independently actionable non-blocking defects identified.

The cache-policy, RBAC diagnostics, error-cause reporting, and bounded retry changes preserve their surrounding contracts and are covered by focused tests.

| Filename | Overview |
|----------|----------|
| core/pkg/console/console.go | Adds explicit cache policies and distinguishes missing asset requests from SPA shell routes. |
| core/pkg/service/access/rbac/service.go | Preserves denial matching while wrapping errors with the denied subject, action, and first uncovered object. |
| pluto/src/flux/retrieve.ts | Adds bounded retries for connection failures and propagates the option through all retrieval entry points. |
| pluto/src/access/queries.ts | Enables transient connection retries for the permissions query required during workspace startup. |
| pluto/src/errors/Fallback.tsx | Renders and copies a bounded chain of nested error causes. |

<sub>Reviews (1): Last reviewed commit: ["SY-4797: Fix stale Console assets and op..."](https://github.com/synnaxlabs/synnax/commit/b11ec4750300dce3161b90dd47edbd771ca49e60) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58093916)</sub>


pluto/src/access/queries.ts
sy-nico, line 74: A permission-gated button still reads as denied when its `granted` check hits a moment of dead connection, since this query settles the failure on the first try and `useGranted` folds it to `false` until something refetches. Was leaving `retryUnreachable` off here deliberate, given the permissions query right below opts in?

core/pkg/console/console.go
```diff
@@ -61,25 +64,63 @@ func New(cfgs ...Config) (*Console, error) {
 	return &Console{fs: cfg.FS}, nil
 }
 
+// Cache policies for the two kinds of file the Console build emits. A hashed asset name
+// changes with its content, so a browser may keep it forever. Every other name survives
+// a release, and embedded files all carry a zero modification time, so a revalidation
+// of a stored copy answers 304 even after an upgrade.
+const (
+	immutableCacheControl = "public, max-age=31536000, immutable"
+	noStoreCacheControl   = "no-store"
```
sy-nico, line 73: Just noting the customer's browser probably still needs a hard reload. Their stored copy has a zero `Last-Modified`, which makes it heuristically fresh for years, and a fresh copy never contacts the Core, so `no-store` never reaches it.

pluto/src/access/queries.ts
pjdotson: added `retryUnreachable` here

core/pkg/console/console.go
pjdotson: Noted but not anything we can do since it was already published code.

PR #2860, opened 2026-08-30
SY-4783: Reuse a call stack for Arc WASM guest calls

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4783](https://linear.app/synnax/issue/SY-4783)

## Description

The Arc runtime calls the WASM guest once per sample, and `api.Function.Call` allocates a fresh `[]uint64` result stack on every call. In an allocation profile of a `return a * b` calculated channel, this single call site produced 95% of all objects allocated in the calculation hot loop.

`nodeImpl` now owns one reusable stack, sized from the guest function definition to `max(params, results)`, and calls `CallWithStack`. The stack is kept separate from the existing `params` slice because `CallWithStack` overwrites it with the results, while `params` holds literal input values that must survive across calls.

Measured over a 10,000-sample frame through an affine node:

| | before | after |
|---|---|---|
| allocs/op | 10000 | 0 |
| B/op | 240005 | 0 |
| ns/sample | 107.2 | 94.95 |

`Create` now rejects a node whose declared input count disagrees with the guest signature. `Call` errored on a count mismatch; `CallWithStack` validates only the stack length, so the old guard had to be replaced.

This does not remove the guest call itself; batching it across a series is SY-4784.

The two existing benchmarks in `bench_test.go` show no change, because neither sets the driving input's data series, so `RefreshInputs` bails and the per-sample loop never runs. That is SY-4782 and is not addressed here.

`node.Factory.Create` also loses its unused `context.Context` parameter. No implementation ever read it: `Create` only wires an in-memory node, so it never blocks, does I/O, or honors cancellation. The `error` return is left alone, since `CompoundFactory` dispatches on `query.ErrNotFound` and several factories return real input-validation failures.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR reuses a preallocated WASM call stack for Arc guest invocations and validates guest parameter counts during node creation.
- Replaces allocating guest calls with `CallWithStack`.
- Keeps persistent parameters separate from result storage.
- Removes the unused context parameter from node factory creation.
- Updates tests and call sites for the revised factory API.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| arc/go/stl/wasm/node.go | Reuses a dedicated call stack while copying persistent parameters before each guest invocation. |
| arc/go/stl/wasm/wasm.go | Validates guest parameter counts and allocates the reusable stack from the function signature. |
| arc/go/stl/wasm/wasm_test.go | Multi-sample tests exercise repeated calls and demonstrate that literal parameters survive stack overwrites. |
| arc/go/runtime/node/factory.go | Removes the unused context argument from the node factory contract and compound dispatch. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Inputs[Input samples and literals] --> Params[Persistent parameter slice]
  Params -->|copy per sample| Stack[Reusable WASM call stack]
  Stack --> Guest[CallWithStack]
  Guest -->|overwrites stack| Results[Guest results]
  Results --> Outputs[Arc output series]
```

<sub>Reviews (3): Last reviewed commit: ["rf"](https://github.com/synnaxlabs/synnax/commit/ec0a26d6b1d84f510f6506bf37cb3460eaa247c2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58333945)</sub>


pjdotson:
@greptile /rerun

arc/go/stl/wasm/wasm.go
```diff
@@ -95,6 +94,11 @@ func (w *Module) Create(_ context.Context, cfg node.Config) (node.Node, error) {
 		stringOutputs[i] = out.Type.Kind == types.KindString
 	}
 
+	// The reused call stack is separate from params because CallWithStack overwrites it
+	// with the results, and params holds literal values that persist across calls.
+	def := fn.Definition()
+	stack := make([]uint64, max(len(def.ParamTypes()), len(def.ResultTypes())))
```
sy-nico, line 104: Should this also check that `len(params)` matches `len(def.ParamTypes())`? `Call` used to error loudly when the counts disagreed, but `CallWithStack` only validates stack length, so a compiler bug that miscounts inputs would now run the guest on truncated or zeroed args and quietly produce garbage instead of failing.

sy-nico (changes requested):
One nit/question. Approved after resolving

arc/go/stl/wasm/wasm.go
pjdotson: added the check in

pjdotson:
@greptile /rerun

PR #2861, opened 2026-08-31
SY-4787: Filter frames against a cached key set

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4787](https://linear.app/synnax/issue/SY-4787)

## Description

`Frame.KeepKeys` scanned the demanded key slice once per frame key, so the relay paid O(frame keys x subscriber keys) on every frame for every subscriber. A Console session that streams 50 channels out of a 500-channel frame did 25,000 comparisons per frame.

`KeepKeys` now takes a `set.Set`. Every caller filters many frames against stable keys, so each one caches the set and rebuilds it only when its keys change: the relay and Cesium streamers on each request, the relay free-write tap inside its atomic pointer, the codec per state update, and the calculation transforms at startup. The relay and Cesium streamers now hold that set as their only key state, with `SendOpenAck` and `ExcludeGroups` as explicit fields, so no stale copy of the open config survives.

No caller kept a slice, so the slice form is gone rather than left beside the set form. A future one-shot caller with a handful of keys must build a set, which costs more than the scan it replaces. `ExcludeKeys` still takes a slice and shares the same `filter` method: `filter` takes the membership source as a type parameter, so each call site keeps a static function value and adds no allocation.

Routing at the delta instead of broadcast-then-filter, which the issue also describes, is not part of this PR.

## Benchmarks

`x/go/telem/frame_bench_test.go`, `-count=10` through benchstat, Apple M3 Max. Shape is frame entries / demanded keys.

| Shape | Before (slice scan) | After (set) | Delta |
| --- | --- | --- | --- |
| 10 / 5 | 45.55 ns | 52.14 ns | +14.47% |
| 10 / 100 | 154.6 ns | 50.60 ns | -67.27% |
| 25 / 25 | 141.7 ns | 110.8 ns | -21.78% |
| 50 / 50 | 456.8 ns | 233.7 ns | -48.85% |
| 100 / 50 | 1355.5 ns | 666.6 ns | -50.82% |
| 100 / 500 | 7472.0 ns | 407.7 ns | -94.54% |
| 500 / 50 | 8.780 us | 4.812 us | -45.18% |
| 500 / 500 | 44.791 us | 8.711 us | -80.55% |
| geomean | 1.077 us | 408.7 ns | -62.07% |

Every delta has p=0.000, n=10, and allocations are identical in every shape. The gain grows with the length of the demanded key list, which is the relay's shape: one writer sends a few channels to a subscriber that reads many. The 10 / 5 shape is 6.6 ns slower, because a hash lookup costs more than a scan of five entries. That cost is not worth a special case.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR replaces repeated slice scans with cached key sets while preserving subscription-update behavior.
- Changes `KeepKeys` to accept a set and updates its filtering implementation.
- Caches key sets in relay, Cesium streamer, codec, free-write tap, iterator, and calculation paths.
- Adds set-based frame tests and filtering benchmarks.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| x/go/telem/frame.go | Changes `KeepKeys` to set membership and generalizes the shared filtering helper without altering filtering semantics. |
| core/pkg/distribution/framer/relay/streamer.go | Stores subscription keys as a cached set and preserves request-driven replacement and group exclusion behavior. |
| cesium/streamer.go | Replaces the retained channel slice with a set rebuilt whenever the subscription request changes. |
| core/pkg/distribution/framer/codec/codec.go | Adds an immutable key set to each codec state for frame filtering while retaining the sorted key slice for encoding. |
| core/pkg/distribution/framer/relay/tap.go | Atomically publishes newly constructed key sets for filtering free-channel writes. |
| core/pkg/service/framer/calculation/transform.go | Builds the calculation output key set once at startup and reuses it for frame filtering. |

<h3>Flowchart</h3>

```mermaid
flowchart LR
  K[Configured channel keys] --> S[Cached key set]
  S --> R[Relay and Cesium streamers]
  S --> C[Frame codec]
  S --> T[Calculation transforms]
  R --> F[Keep matching frame entries]
  C --> F
  T --> F
  F --> O[Filtered frame output]
```

<sub>Reviews (4): Last reviewed commit: ["address comment on duplicated state"](https://github.com/synnaxlabs/synnax/commit/6a44a6bc3e134db61a6d4524356ac4cc85fb845e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58767125)</sub>

- Knowledge Base — [Core storage and distribution](https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/core-storage-distribution.md)
- Knowledge Base — [Data infrastructure libraries](https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/data-infrastructure.md)
- Knowledge Base — [Cesium time-series storage](https://app.greptile.com/synnax/-/custom-context/knowledge-base/synnaxlabs/synnax/-/docs/cesium-time-series.md)
</details>


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

core/pkg/distribution/framer/relay/streamer.go
```diff
@@ -161,7 +167,7 @@ func (s *streamer) Flow(ctx signal.Context, opts ...confluence.Option) {
 					return nil
 				}
 				req.Keys = lo.Uniq(req.Keys)
-				s.cfg.Keys = req.Keys
+				s.keys = set.New(req.Keys...)
```
sy-nico, line 162: Now that the filter reads `s.keys`, `s.cfg.Keys` stops updating after the first request and quietly holds the original keys forever, same with `s.Channels` in the cesium streamer. They're still read once at startup so they can't go away, but maybe keep the one-line sync or mark them as initial values only so the next reader doesn't trust them.

sy-nico (changes requested):

pjdotson:
@greptile /rerun

sy-nico (approved):

PR #2866, opened 2026-09-01
SY-4761: Enable scale and rotation on the schematic switch symbol

sy-nico:
# Issue Pull Request

## Linear Issue

[SY-4761](https://linear.app/synnax/issue/SY-4761)

## Description

The `switch` symbol's Style tab showed a Scale control and the rotate button cycled its orientation, but neither changed the rendering. The switch is the only symbol with these controls that draws itself in HTML instead of through `Primitive.SVG`: its primitive dropped the `scale` prop, and the track and knob sizes were hard-coded in the shared input CSS.

- Scale: the shared `Input.Switch` geometry now multiplies by a `--pluto-switch-scale` variable (default 1, so switches elsewhere in the app are unchanged), and the symbol primitive sets it from `scale`. Symbol-scoped CSS zeros the button chrome so the node box hugs the track, and transitions only paint so drag-resize tracks instantly.
- Rotation: the primitive now carries the standard location class, and the symbol CSS swaps the layout box and rotates the indicator inside it, the HTML equivalent of the SVG symbols' box swap and rotation.

A switch that was drag-resized while the control was dead silently stored a `scale`. That value is now honored, a one-time visual jump that reflects what was dragged. Never-resized switches store `scale: 1` and render pixel-identical.

## Basic Readiness

<h3>Greptile Summary</h3>

The PR makes the HTML-based schematic switch honor its stored scale and orientation while preserving the default geometry of shared input switches.
- Scales the shared switch track, knob, offsets, and knob travel through `--pluto-switch-scale`.
- Adds orientation classes and symbol-specific layout-box rotation rules.
- Adds unit coverage for scale variables, orientation classes, color behavior, and enabled state.

The PR appears safe to merge, with no concrete changed-code failures identified.

The scaled geometry remains equivalent at scale 1, knob travel scales consistently, and the orientation-specific layout boxes agree with the switch’s handle placement.

| Filename | Overview |
|----------|----------|
| pluto/src/input/Switch.css | Replaces fixed switch geometry with scale-aware calculations while preserving equivalent geometry at the default scale. |
| pluto/src/schematic/node/general/switch/Primitive.tsx | Propagates scale and orientation to CSS through a custom property and standard location class. |
| pluto/src/schematic/node/general/switch/switch.css | Removes schematic button chrome and aligns rotated visual geometry with orientation-aware node bounds. |
| pluto/src/schematic/node/general/switch/Primitive.spec.tsx | Adds assertions for scale propagation and orientation classes alongside existing switch behavior coverage. |

<sub>Reviews (1): Last reviewed commit: ["SY-4761: Fix rotation on the schematic s..."](https://github.com/synnaxlabs/synnax/commit/72316f7a1165d1f3d7472a0c7613f166bdba3794) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58862815)</sub>

### DIFF
--- a/pluto/src/schematic/node/general/switch/Primitive.tsx
+++ b/pluto/src/schematic/node/general/switch/Primitive.tsx
@@ -9,6 +9,7 @@
 
 import "@/schematic/node/general/switch/switch.css";
 
+import { location } from "@synnaxlabs/x";
 import { type CSSProperties, type MouseEventHandler, type ReactElement } from "react";
 
 import { CSS } from "@/css";
@@ -18,9 +19,9 @@ import { Primitive } from "@/schematic/node/common/primitive";
 import { type Toggle } from "@/schematic/node/common/toggle";
 import { symbolColorVar } from "@/schematic/symbolColor";
 
-export interface Props
-  extends Omit<Toggle.ButtonProps, "onClick">, Primitive.SVGBasedProps {
+export interface Props extends Omit<Toggle.ButtonProps, "onClick"> {
   onClick?: MouseEventHandler<HTMLElement>;
+  scale?: number;
 }
 
 export const Switch = ({
@@ -41,7 +42,7 @@ export const Switch = ({
       className={CSS.cls(
         colorVar != null && CSS.B("symbol-colored"),
         colorVar != null && CSS.BM("switch-symbol", "colored"),
-        CSS.loc(orientation),
+        CSS.dir(location.direction(orientation)),
       )}
       style={style}
     >
--- a/pluto/src/schematic/node/general/switch/Primitive.tsx
+++ b/pluto/src/schematic/node/general/switch/Primitive.tsx
@@ -41,6 +41,7 @@ export const Switch = ({
       className={CSS.cls(
         colorVar != null && CSS.B("symbol-colored"),
         colorVar != null && CSS.BM("switch-symbol", "colored"),
+        CSS.loc(orientation),
       )}
       style={style}
     >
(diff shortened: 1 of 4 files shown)


pluto/src/schematic/node/general/switch/Primitive.tsx
```diff
@@ -18,7 +18,8 @@ import { Primitive } from "@/schematic/node/common/primitive";
 import { type Toggle } from "@/schematic/node/common/toggle";
 import { symbolColorVar } from "@/schematic/symbolColor";
 
-export interface Props extends Omit<Toggle.ButtonProps, "onClick"> {
+export interface Props
+  extends Omit<Toggle.ButtonProps, "onClick">, Primitive.SVGBasedProps {
```
pjdotson: the only prop we get from `Primitive.SVGBasedProps` is `scale`? doesn't make sense to use its props if we don't actually pass the props down

pluto/src/schematic/node/general/switch/switch.css
```diff
@@ -31,3 +31,47 @@
...
+            background 0.15s;
+    }
+}
+
+/*
+ * The shared svg rotation in primitive.css cannot rotate this HTML symbol, and
+ * rotating an HTML element leaves its layout box unrotated. So swap the box by hand
+ * and rotate the flex-centered indicator inside it.
+ */
+.pluto-symbol-primitive.pluto--location-top > .pluto-input__switch,
```
pjdotson: does it really make sense to be able to rotate a switch? IMO it only looks good the normal way and you shouldn't be allowed to rotate it.

pjdotson (commented):
When the switch is very big the gap between the knob and the edge is too big. There should probably be a maximum allowable gap between those:

Image

pluto/src/schematic/node/general/switch/switch.css
sy-nico: Talked offline and decided we would allow only horizontal and vertical in the positive directions

PR #2874, opened 2026-09-04
SY-4497: Upgrade TypeScript dependencies

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4497](https://linear.app/synnax/issue/SY-4497)

## Description

Upgrades the TypeScript toolchain and every workspace dependency, and moves the docs
site onto Astro's native Markdown pipeline.

| Dependency | From | To |
| --- | --- | --- |
| pnpm | 11.18.0 | 12.3.1 |
| zod | 4.4.3 | 4.5.4 |
| eslint | 10.8.0 | 10.9.1 |
| typescript-eslint | 8.66.0 | 8.69.0 |
| astro | 7.1.6 | 7.2.10 |
| @astrojs/mdx | 7.0.5 | 8.0.0 |
| @codingame/monaco-vscode-\* | 36.0.0 | 36.2.7 |
| proxy-memoize | 2.0.3 | 3.0.1 |
| @tauri-apps/plugin-updater | 2.10.1 | 2.11.0 |
| knip | 6.31.0 | 6.34.0 |

### pnpm and the catalog

Adopts the pnpm 12 settings — `shellEmulator`, `catalogMode: strict`, `catalogPrune`,
and `trustPolicy: no-downgrade` — and replaces `packageManager` with `devEngines`. Every
dependency now resolves through the catalog, workspace dependencies included, which pnpm
12.2 allows. `pluto/examples` is not a workspace member and keeps its own pins.

`chokidar` and `semver` predate provenance on their package, and `@vercel/routing-utils`
stopped attaching an attestation after 6.4.0 while still publishing from a trusted
publisher, so all three need trust exclusions.

The pnpm version is pinned exactly rather than as a range. `pnpm/setup` resolves a range
against the registry while pnpm resolves it against the lockfile, so `^12.1.0` installed
12.3.2 and then reported 12.1.0, which failed every CI job. Vercel needs its own pin: it
picks pnpm by project creation date, lands on 9, and cannot parse the two-document
lockfile pnpm 12 writes, so it was silently ignoring the lockfile and resolving fresh.

### Behavior changes

Zod 4.5 brings `zod/compile`, imported at every entry point that owns schemas so they
compile on first use. Libraries import it too, which forces the behavior on consumers.
`z.validate` replaces `safeParse().success` at all 92 call sites; it returns a type
predicate, so four unchecked assertions in the Client fall away.

proxy-memoize was pinned at 2.0.3 because Immer freezes reducer output in place, and
state read before the freeze and again after breaks proxy-compare's cached descriptor
decision. Freezing at store construction closes that window and lifts the pin.

The Monaco API is declared twice in Pluto, once under its own name and once aliased to
`vscode` for vscode-languageclient. The alias cannot use `catalog:` — pnpm writes a
dangling link rather than failing — so both carry the version and move together.

The Tauri updater moves on both the JS and the Rust side. The Windows fix is in the
crate: a failed installer spawn through `ShellExecuteW` now returns an error instead of
failing silently, which the existing handler reports as a status.

### Lint rules

ESLint 10.9 and typescript-eslint 8.69 fix the `prefer-template` and `return-await`
autofixes, both of which run as errors here. The bump also enables
`no-loss-of-precision`, `no-meaningless-void-operator`, `no-mixed-enums`,
`no-misused-promises` with the new `checksConditionals.flagUnions`, `unified-signatures`,
and `no-empty-object-type`.

`unified-signatures` found overloads that were already redundant: `channel.retrieve`
declared `Key | Name` above a `Key | string` that subsumed it, `device.create` declared
forms the schema-bearing overloads already covered, and `Scale.scale` carried an internal
widening overload nothing used. The `useField` overloads keep their order behind a
disable, because `ExtensionGetOptions` also accepts `optional: true` and hoisting it
above the optional overload would drop the null from the return type.

`no-empty-object-type` allows single-extends interfaces and empty object types. Without
those it reports 734 sites: the `interface X extends z.infer<typeof xZ> {}` idiom, 245 of
them in generated `types.gen.ts`, and deliberate uses of `{}` as a non-nullish constraint.

### Docs site

Astro 7's native pipeline (Sätteri) used to emit highlighted code fences as raw HTML that
bypassed the MDX `pre` override, so the site was pinned to `processor: unified()`.
`@astrojs/markdown-remark` 7.3.0 moves MDX compilation into both processors and removes
the blocker; `@astrojs/mdx` 8.0.0 is the same upstream change. Dropping the pin renders
every code block identically, verified against seven pages, so the explicit processor and
its dependency are gone.

`posthog-js` is removed. Nothing imported it — the site loads PostHog from the CDN
through the snippet in `PostHog.astro`, so the package shipped nothing.

The site link check no longer fails on two vendor hosts. Omron answers too slowly from a
datacenter IP to probe inside the budget, so it joins the ignore list. etherlab.org
refuses HEAD at the TLS layer but serves GET, so a GET now follows every failed HEAD
rather than only one that returned a status.

### Held back

`@types/node` moves to 24 to match the `node ^24.20.0` runtime, which is the current LTS.
Types for 26 would admit APIs that do not exist at runtime, so it moves with the runtime.

TypeScript 7 is already the default catalog. A `ts6` catalog pins madge,
typescript-eslint, `astro check`, and unplugin-dts, which need the JavaScript Compiler
API that TypeScript 7 no longer ships. typescript-eslint is the binding constraint, with
a `<6.1.0` peer ceiling.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR upgrades the TypeScript toolchain and workspace dependencies, adopts pnpm 12 catalog and setup conventions, and moves the documentation site to Astro’s native Markdown pipeline.
- Updates dependency catalogs, manifests, and lockfiles across TypeScript workspaces.
- Updates schema validation, store freezing, lint compatibility, and associated tests.
- Simplifies CI setup through `pnpm/setup` and updates the docs build and link checker.
- The previously reported benchmark-comment issue is resolved at the current head.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| client/ts/src/framer/adapter.bench.ts | Replaces the unnecessary `void` expression with a block-bodied async callback, resolving the prior feedback without explanatory syntax comments. |
| client/ts/src/framer/frame.bench.ts | Removes the unnecessary `void` expression from the frame-construction benchmark. |
| pnpm-workspace.yaml | Centralizes upgraded workspace dependency versions and pnpm 12 catalog settings. |
| drift/src/configureStore.ts | Freezes initial store state to keep proxy memoization behavior stable across Immer’s freezing lifecycle. |
| docs/site/astro.config.ts | Moves documentation rendering from the explicitly configured unified processor to Astro’s native Markdown pipeline. |

<sub>Reviews (2): Last reviewed commit: ["SY-4497: Stop the site link check failin..."](https://github.com/synnaxlabs/synnax/commit/157835532ada089c564ead3f5ed41c657d146ceb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=60260716)</sub>


pjdotson:
@greptile /rerun

sy-nico (approved):

PR #2886, opened 2026-09-10
SY-4796: Upgrade Go stdlib usage to 1.27

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4796](https://linear.app/synnax/issue/SY-4796/upgrade-go-stdlib-usage-to-127)

## Description

Moves every Go module onto the Go 1.27 standard library: `encoding/json/v2`, the stdlib
`uuid` package, `math/rand/v2`, `strings.CutLast`, ML-DSA keys, and a `go fix` pass.
`types.Nil` gives way to `struct{}` throughout.

**JSON.** `x/go/encoding/json` builds on `jsontext`, and Oracle emits `MarshalJSONTo` and
`UnmarshalJSONFrom`, so every `types.gen.go` regenerates. The codec takes the v2 defaults,
which changes the wire in visible ways: a nil slice or map encodes as `[]` or `{}`, a
retrieve response always carries its list, `omitzero` marks only schema-optional fields,
and a required field serializes as `null` instead of disappearing. Decoding is strict: a
duplicate name, invalid UTF-8, or a JSON `null` for a number-backed type such as
`telem.TimeStamp` or a resource key is an error. The Python JSON codec leaves `None`
fields off the wire to match. Optional fields carry `msgpack:",omitempty"`, the only
omission tag vmihailenco honors.

**UUID.** The stdlib `uuid.UUID` has no `MarshalBinary`, so `x/go/encoding/msgpack`
registers a codec that keeps the 16-byte form `google/uuid` wrote. The registration lives
in an `init()` because vmihailenco's registry is process-global.

**Cesium.** The meta file stores `is_index` at version 3. Older files open because the
meta codec folds case. A database opened by this build cannot be reopened by an older
release.

**Certificates.** `cert.KeyAlgorithm` adds opt-in ML-DSA keys. RSA stays the default and
the CLI does not expose the option. JWT has no ML-DSA method, so the token signing key is
split from the node TLS key and falls back to a persisted Ed25519 key when the TLS key
cannot sign.

**Breaker.** `breaker.Config` takes an injected clock. The synctest helper is gone.

**Oracle.** Generated fuzz tests compare round trips with `testutil.DeepEqual`, which
equates NaN in named float types and ignores a type's own `Equal` method; go-cmp is gone.
`oracle check --format=json` reports `elapsed_ns` as a string. The Prettier and ESLint
formatters run through `pnpm exec`.

## Basic Readiness

<h2><a href="https://app.greptile.com/api/retrigger?id=62714551"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge with no outstanding findings.

<h3>Summary</h3>

The PR moves the Go modules to Go 1.27 APIs and updates generated codecs, wire tags, certificate support, UUID handling, tests, and Python JSON encoding.
- Regenerates JSON v2 and MessagePack bindings.
- Preserves optional-field omission across Go and Python.
- Adds structural round-trip comparison with named-NaN support.
- Updates certificate and token-key handling.

<h3>Diagram</h3>

```mermaid
flowchart LR
    S[Oracle schemas] --> G[Go code generation]
    G --> J[JSON v2 codecs]
    G --> M[MessagePack codecs]
    J --> F[Freighter wire payloads]
    M --> F
    P[Python models] -->|omit None| F
    F --> C[Core services]
```

<sub>Reviews (4) · Last reviewed commit: ["reflow"](https://github.com/synnaxlabs/synnax/commit/0776301fbf2f8fb31435edaf5b74fbaa6231451d)</sub>


pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

pjdotson:
@greptile /rerun

x/go/encoding/json/json.go
```diff
@@ -25,143 +25,113 @@ import (
...
+func readDigits(dec *jsontext.Decoder) (string, error) {
+	tok, err := dec.ReadToken()
+	if err != nil {
+		return "", err
 	}
-	var str string
-	if err := json.Unmarshal(b, &str); err != nil {
-		return 0, err
+	if k := tok.Kind(); k != '0' && k != '"' {
+		return "", errors.Newf("cannot decode a number from JSON %s", k.String())
```
sy-nico, line 100: This breaks the released Python client (0.58.1) against a Core with this change. Its JSON codec still sends `null` for unset fields, so every iterator request carries `span: null` and `stamp: null` and task retrieve sends `rack: null`, and `TimeStamp`, `TimeSpan`, `rack.Key` and `task.Key` all decode through here. v1 read that null as zero, so after this lands `client.read` and `tasks.retrieve` fail on every Python install that has not upgraded, and the `exclude_none` change only helps clients built after it.

Nothing stored changes, so this is Core and client version skew rather than a migration. Null needs to keep decoding as zero here for as long as 0.58.x Python clients are supported.

sy-nico (changes requested):

x/go/encoding/json/json.go
pjdotson: This is fine. We don't compatibility guarantees between different minor-version client and Cores until we get to v1 major version.

sy-nico: Tested it. Python 0.58.1 against a Core built from this head and `client.read`/`tasks.retrieve` both fail, both pass on rc. Both Cores report 0.58.2, so the client sees the same minor and does not warn. The contract in both clients and RFC 0058 is same major.minor, and rc is still 0.58.2.

sy-nico (changes requested):
See thread, but from the claude testing:

Image

x/go/encoding/json/json.go
pjdotson: This code is not released yet. When we make a new release, we will have a Python client version v0.59 and a Core version v0.59. The Python v0.59 will be compatible with the Core v0.59. This is not going into a v0.58 Core release.

pjdotson:
@greptile /rerun

sy-nico (approved):

PR #2891, opened 2026-09-15
SY-4779: Fix flaky Cesium auto-index explicit timestamp spec

pjdotson:
# Issue Pull Request

## Linear Issue

[SY-4779](https://linear.app/synnax/issue/SY-4779)

## Description

`Writer AutoIndex` had a spec that failed on CI roughly once per many runs with
`commit timestamp ... must be strictly greater than the starting timestamp`.

The spec read the clock into `before`, opened the writer, and then wrote an explicit
index timestamp at `before + 10ms`. With `AutoIndex` set and no `Start`, `OpenWriter`
defaults `Start` to the clock at open. An open that took more than 10ms put `Start`
after the explicit timestamp, so Cesium correctly rejected the write. The CI run in the
issue took 12.1ms. The clock for the explicit timestamp is now read after the writer
opens, so the timestamp always falls inside the writer's domain no matter how long the
open takes.

The final assertion was also too weak to catch the behavior in the spec's name. It
checked that the auto-stamp is greater than the explicit timestamp, which holds even
when the caller's timestamp advances the auto-stamp clock, because the next stamp would
then be `explicit + 1`. The spec now records the clock once it passes the explicit
timestamp and asserts the auto-stamp is at or after that reading, which only holds when
the auto-stamp is a fresh clock read.

Test-only change; no version bump.

Injecting a 20ms sleep into `OpenWriter` reproduces the CI failure at the same line with
the old spec, and the new spec passes with that sleep in place.

## Basic Readiness

<h3>Greptile Summary</h3>

This PR makes the Cesium auto-index explicit timestamp test independent of writer-open latency. It also verifies that both writes are authorized and that the next automatic index uses a fresh clock reading.

The test-only change appears safe to merge.

The explicit timestamp is now created inside the writer domain, and the final assertion correctly distinguishes a fresh clock reading from an increment based on the explicit timestamp. No outstanding correctness or rule violations remain.

| Filename | Overview |
|----------|----------|
| cesium/writer_auto_index_test.go | Moves timestamp capture after writer creation and strengthens write and auto-index assertions. |

<sub>Reviews (2): Last reviewed commit: ["update assertions"](https://github.com/synnaxlabs/synnax/commit/33346dd1c9b9ed6ff71c41a851fc9fff99e02b22) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=64574215)</sub>


pjdotson:
@greptile /rerun

sy-nico (approved):

